PR #1, opened 2025-01-22
dev

MathisGD:
dev branch


src/interfaces/ITerms.sol
```diff
@@ -0,0 +1,33 @@
...
+}
+
+struct Offer {
+    bool buy;
+    address offering;
+    uint256 assets;
+    address loanToken;
+    Collateral[] collaterals;
+    uint256 maturity;
+    uint256 price;
```
gd-colin, line 24: For the verif, we could explore the following idea, `Offers` include an extra field `id`, their hash, to avoid recomputing it. The value of such hash is derived by computing it on the initial offer where the `id` is set to zero. Consider the following illustration.

``` ml
let newOffer(buy,…) =
  let rawOffer = newRawOffer(buy, …, id = 0) in
  let id' = keccak(encode(rawOffer)) in
  newRawOffer(buy,…,id = id')
```
This avoids recomputation and should help the prover.

gd-colin: To use it, we always `require(offer.id != 0)`.

gd-colin: FYI, this technique of pre-computing constructors is also known as [smart-constructors](https://wiki.haskell.org/index.php?title=Smart_constructors).

src/Terms.sol
```diff
@@ -0,0 +1,164 @@
...
+            // Relies on the fact that the collaterals are sorted.
+            // Note that we actually never check that.
+            // If they are not, the match could fail.
+            while (
+                bytes20(sellOffer.collaterals[i].token) < bytes20(buyOffer.collaterals[j].token)
+                    && j++ < buyOffer.collaterals.length
+            ) {}
+            require(sellOffer.collaterals[i].token == buyOffer.collaterals[j].token, "Collaterals tokens do not match");
+            require(sellOffer.collaterals[i].lltv <= buyOffer.collaterals[j].lltv, "LLTVs do not match");
+            require(sellOffer.collaterals[i].oracle == buyOffer.collaterals[j].oracle, "Oracles do not match");
```
QGarchery, line 244: There is something unexpected here where the list of the seller can be included in the list of the buyer but fail to match. Assume that the buyer is a lender that lists more oracles for the same collateral and LLTV. Then it can happen that the borrower would only list one oracle, and if it's not the first one in the list of the lender then it fails because of this line. The same goes for LLTVs, although it's unclear why you would have multiple LLTVs in the same list as a lender.

A fix would be to hash the whole (collateral, lltv, oracle) tuple, and compare hashes instead of comparing collaterals

MathisGD: The multi oracle thing isn't implemented yet (thought that we needed oracle lists). But actually you idea is great!

MathisGD: I got your comment now. We are not expecting to implement the "multi-oracle offer" feature by just adding some collaterals, because on the borrow side it doesn't behave like adding collaterals (you expect the lend collaterals to be a superset of the borrow collaterals)

PR #3, opened 2025-03-13
Liquidation poc

gd-colin:
This PR introduces a POC for the `liquidate` function.


src/Terms.sol
```diff
@@ -60,8 +60,17 @@ contract Terms is ITerms {
...
+            rest = buyerScaledPrice - sellerScaledPrice;
+        } else {
+            rest = 0;
+        }
+
         IERC20(buyOffer.loanToken).transferFrom(buyer, seller, sellerScaledPrice);
-        IERC20(buyOffer.loanToken).transferFrom(buyer, msg.sender, buyerScaledPrice - sellerScaledPrice);
+        if (rest > 0) {
+            IERC20(buyOffer.loanToken).transferFrom(buyer, msg.sender, rest);
+        }
```
MathisGD: this change isn't necessary because we have this
`require(buyOffer.price >= sellOffer.price, "Buy offer price is less than sell offer price");`

src/interfaces/ITerms.sol
```diff
@@ -30,4 +30,12 @@ struct Signature {
     bytes32 s;
 }
 
+struct Limit {
+    // Index in the collateral list of the term.
+    // TODO use something more robust than indexes.
+    uint256 collateralIndex;
```
MathisGD: why do we say that it's not robust? here the term is immutable

src/interfaces/ITerms.sol
```diff
@@ -30,4 +30,12 @@ struct Signature {
     bytes32 s;
 }
 
+struct Limit {
```
MathisGD: why "limit"?

src/Terms.sol
```diff
@@ -96,6 +105,88 @@ contract Terms is ITerms {
         IERC20(collateral).transfer(msg.sender, amount);
     }
 
+    /// @notice Liquidates the given collections `repaidAmounts` of debt asset
+    /// by colleral or seize the given `seizedAssets` of collateral on the given
+    /// `term` of the given `borrower`,
+    /// @dev Either `seizedAssets` or `repaidAmounts` should be empty.
```
MathisGD: you mean with amount=0 or really an empty list? Because then you require this `require(seizedAssets.length == repaidAmounts.length, "Incoherent limit arrays");`

src/Terms.sol
```diff
@@ -96,6 +105,88 @@ contract Terms is ITerms {
...
+    /// @return The list of amounts of assets seized by collateral index.
+    /// @return The list of amounts of assets repaid by collateral index.
+    function liquidate(Term memory term, Limit[] memory seizedAssets, Limit[] memory repaidAmounts, address borrower)
+        external
+        returns (Limit[] memory, Limit[] memory)
+    {
+        require(
+            seizedAssets.length <= term.collaterals.length, "Cannot seize more assets than the the supplied collaterals"
+        );
+        // TODO check that either the user is either seized or repaying amounts.
```
MathisGD: why couldn't one pass the seized as input for one collateral and repaid for an other?

src/interfaces/ITerms.sol
```diff
@@ -30,4 +30,12 @@ struct Signature {
...
+    // Index in the collateral list of the term.
+    // TODO use something more robust than indexes.
+    uint256 collateralIndex;
+    // Amount in either loan or collateral asset.
+    uint256 amount;
```
MathisGD: why not both, and then you require that one of the two is 0 (as discussed [here](https://github.com/morpho-org/v2/pull/3/files#r2000523350))
```suggestion
    uint256 seized;
    uint256 repaid;
```

src/Terms.sol
```diff
@@ -96,6 +105,88 @@ contract Terms is ITerms {
...
+    /// @param term The term of the bond.
+    /// @param borrower The debtor of the loan.
+    /// @param seizedAssets A collection of amounts of collateral to seize and the collateral index to seize..
+    /// @param repaidShares A collection of amounts of loan asset to repay and the collateral index to seize.
+    /// @param data Arbitrary data to pass to the `onMorphoLiquidate` callback. Pass empty data if not needed.
+    /// @return The list of amounts of assets seized by collateral index.
+    /// @return The list of amounts of assets repaid by collateral index.
+    function liquidate(Term memory term, Limit[] memory seizedAssets, Limit[] memory repaidAmounts, address borrower)
+        external
+        returns (Limit[] memory, Limit[] memory)
```
MathisGD: As discussed [here](https://github.com/morpho-org/v2/pull/3/files#r2000525498), I think that we could have one input.
```suggestion
    function liquidate(Term memory term, Limit[] memory collaterals, address borrower)
        external
        returns (Limit[] memory, Limit[] memory)
```

src/Terms.sol
```diff
@@ -96,6 +105,88 @@ contract Terms is ITerms {
...
+        uint256 totalRepaid;
+        for (uint256 i = 0; i < repaidAmounts.length; i++) {
+            totalRepaid += repaidAmounts[i].amount;
+        }
+
+        // Compute the repaid and seized amounts by collateral index.
+        if (totalRepaid > 0) {
+            for (uint256 i = 0; i < repaidAmounts.length; i++) {
+                Limit memory l;
+                uint256 collateralPrice = IOracle(term.collaterals[repaidAmounts[i].collateralIndex].oracle).price();
```
MathisGD: can be queried outside of the if

src/Terms.sol
```diff
@@ -96,6 +105,88 @@ contract Terms is ITerms {
...
+                l.collateralIndex = repaidAmounts[i].collateralIndex;
+                l.amount = (repaidAmounts[i].amount * liquidationIncentiveFactor / WAD) / collateralPrice;
```
MathisGD: the collateral price needs to be scaled (in morpho blue 1e36)
```suggestion
                l.amount = (repaidAmounts[i].amount * liquidationIncentiveFactor / WAD) * <some constant>  / collateralPrice;
```

src/Terms.sol
```diff
@@ -96,6 +105,88 @@ contract Terms is ITerms {
...
+                totalRepaid += l.amount;
+                repaidAmounts[i] = l;
+            }
+        }
+
+        debtOf[borrower][id] -= totalRepaid;
+        withdrawable[id] += totalRepaid;
+
+        // Transfer the repaid amount.
+        IERC20(term.loanToken).transferFrom(msg.sender, address(this), totalRepaid);
```
MathisGD: At some point we will probably want to have a callback (in the sense Morpho Blue) in here, so we can already have the repay after the withdraw collat

src/Terms.sol
```diff
@@ -96,6 +105,88 @@ contract Terms is ITerms {
...
+                uint256 collateralPrice = IOracle(term.collaterals[repaidAmounts[i].collateralIndex].oracle).price();
+                l.collateralIndex = repaidAmounts[i].collateralIndex;
+                l.amount = (repaidAmounts[i].amount * liquidationIncentiveFactor / WAD) / collateralPrice;
+                seizedAssets[i] = l;
+            }
+        } else {
+            for (uint256 i = 0; i < seizedAssets.length; i++) {
+                Limit memory l;
+                uint256 collateralPrice = IOracle(term.collaterals[seizedAssets[i].collateralIndex].oracle).price();
+                uint256 seizedAssetsQuoted = seizedAssets[i].amount * collateralPrice;
```
```suggestion
                uint256 seizedAssetsQuoted = seizedAssets[i].amount * collateralPrice / <some constant>;
```

src/Terms.sol
```diff
@@ -96,6 +105,88 @@ contract Terms is ITerms {
...
+                l.collateralIndex = repaidAmounts[i].collateralIndex;
+                l.amount = seizedAssetsQuoted * WAD / liquidationIncentiveFactor;
+                totalRepaid += l.amount;
```
MathisGD: ideally we unify this outside of the if else (if we implement my proposal)

src/Terms.sol
```diff
@@ -96,6 +105,88 @@ contract Terms is ITerms {
...
+                l.amount = (repaidAmounts[i].amount * liquidationIncentiveFactor / WAD) / collateralPrice;
+                seizedAssets[i] = l;
+            }
+        } else {
+            for (uint256 i = 0; i < seizedAssets.length; i++) {
+                Limit memory l;
+                uint256 collateralPrice = IOracle(term.collaterals[seizedAssets[i].collateralIndex].oracle).price();
+                uint256 seizedAssetsQuoted = seizedAssets[i].amount * collateralPrice;
+                l.collateralIndex = repaidAmounts[i].collateralIndex;
+                l.amount = seizedAssetsQuoted * WAD / liquidationIncentiveFactor;
```
MathisGD: the rounding goes in the wrong direction here. we should write a little math lib anyway

src/Terms.sol
```diff
@@ -96,6 +105,88 @@ contract Terms is ITerms {
...
+        // Transfer the seized collaterals.
+        for (uint256 i = 0; i < seizedAssets.length; i++) {
+            Limit memory l = seizedAssets[i];
+            address collateral = term.collaterals[l.collateralIndex].token;
+            collateralOf[borrower][_id(term)][collateral] -= l.amount;
+            IERC20(collateral).transfer(msg.sender, l.amount);
+        }
+
+        // Realize bad debt.
+        uint256 totalCollateralQuoted;
```
MathisGD: there is probably a way to combine this loop with the first one and avoid having to loop twice

src/Terms.sol
```diff
@@ -96,6 +105,88 @@ contract Terms is ITerms {
...
+        for (uint256 i = 0; i < term.collaterals.length; i++) {
+            uint256 price = IOracle(term.collaterals[i].oracle).price();
+            uint256 collateralQuoted = collateralOf[borrower][id][term.collaterals[i].token] * price / WAD;
+            totalCollateralQuoted += collateralQuoted;
+        }
+        if (totalCollateralQuoted == 0) {
+            uint256 badDebt = debtOf[borrower][id];
+            withdrawable[id] -= badDebt;
```
MathisGD: why do we remove from withdrawable?

src/Terms.sol
```diff
@@ -96,6 +105,88 @@ contract Terms is ITerms {
...
+            debtOf[borrower][id] = 0;
+        }
```
MathisGD: where is the bad debt realised exactly here? what prevents lenders to redeem at 1:1 their bonds

src/interfaces/ITerms.sol
gd-colin: Maybe it is, but I'm under the impression that this may be error prone.

src/interfaces/ITerms.sol
gd-colin: For a lack of better name, change in https://github.com/morpho-org/v2/pull/3/commits/3f694857478a9d1cba68bb2510670b265c1d7b1c.

src/Terms.sol
gd-colin: Changed in https://github.com/morpho-org/v2/pull/3/commits/3f694857478a9d1cba68bb2510670b265c1d7b1c.

src/interfaces/ITerms.sol
gd-colin: Implemented in https://github.com/morpho-org/v2/pull/3/commits/3f694857478a9d1cba68bb2510670b265c1d7b1c.

src/Terms.sol
gd-colin: Done in https://github.com/morpho-org/v2/pull/3/commits/3f694857478a9d1cba68bb2510670b265c1d7b1c.

src/Terms.sol
gd-colin: Added in https://github.com/morpho-org/v2/pull/3/commits/3f694857478a9d1cba68bb2510670b265c1d7b1c.

src/Terms.sol
gd-colin: Added in https://github.com/morpho-org/v2/pull/3/commits/3f694857478a9d1cba68bb2510670b265c1d7b1c.

src/Terms.sol
gd-colin: Do you think it's important to do this change at this point?

src/Terms.sol
gd-colin: Done in https://github.com/morpho-org/v2/pull/3/commits/3f694857478a9d1cba68bb2510670b265c1d7b1c.

src/Terms.sol
gd-colin: Done in https://github.com/morpho-org/v2/pull/3/commits/3f694857478a9d1cba68bb2510670b265c1d7b1c.

src/Terms.sol
gd-colin: Changed in https://github.com/morpho-org/v2/pull/3/commits/3f694857478a9d1cba68bb2510670b265c1d7b1c.

src/Terms.sol
gd-colin: Changed in https://github.com/morpho-org/v2/pull/3/commits/3f694857478a9d1cba68bb2510670b265c1d7b1c.

src/Terms.sol
gd-colin: I mean `amount == 0`, but this is handled differently since https://github.com/morpho-org/v2/pull/3/commits/3f694857478a9d1cba68bb2510670b265c1d7b1c.

src/Terms.sol
MathisGD: let's at least flag it with a comment

src/interfaces/ITerms.sol
MathisGD: I think that it's ok! And it suits well our current `term` type

src/Terms.sol
```diff
@@ -96,8 +107,112 @@ contract Terms is ITerms {
...
+            totalCollateralQuoted += collateralQuoted;
+            maxDebt += collateralQuoted * term.collaterals[i].lltv / WAD;
+        }
+
+        // Check that position not healthy.
+        require(debtOf[borrower][id] > maxDebt, "Healthy borrower");
+
+        // Compute the repaid and seized amounts by collateral index, remaining collateral and total repaid.
+        for (uint256 i = 0; i < seizures.length; i++) {
+            require(seizures[i].collateralIndex < term.collaterals.length, "INCONSISTENT_INPUT");
```
MathisGD: we should also check that the same collateral doesn't appear twice. the best way to do that might be to require that they are ordered (by index)

src/Terms.sol
```diff
@@ -96,8 +107,112 @@ contract Terms is ITerms {
...
+            (bool success, bytes memory returnData) = msg.sender.call(callbackData);
+            if (!success) lowLevelRevert(returnData);
+        }
+
+        IERC20(term.loanToken).transferFrom(msg.sender, address(this), totalRepaid);
+
+        return seizures;
+    }
+
+    function _seizeCollateral(Collateral memory c, Seizure memory s, uint256 lif, address borrower)
```
MathisGD: i would much prefer it to not be a standalone function. I find that more often than we think, these confuse more than they help (~leaky abstractions)

src/Terms.sol
gd-colin: Alternatively, we could have an array of booleans `seized` that are initialized to false (for instance while computing the `totalCollateralQuoted`), upon seizure we check if the index is marked to true, if it is the function reverts else the index is set as seized and the liquidation proceeds, this avoids any extra requirement or loopings.

I'm not sure why we should do it though, it's totally _possible_ to do that, it's pointless and expensive for the user, but semantically defined. Which makes me think that the original check on the length of seizures makes no sense.

src/Terms.sol
gd-colin: We should probably count the bad debt separately in a dedicated mapping (e.g. `badDebtOf`) and handle it upon bond withdrawal or minting. What do you think @MathisGD ?

MathisGD: yes, though it would be per bond (not per user)

src/Terms.sol
MathisGD: we can POC both, but I think that i prefer requiring that indexes are ordered. Alternatively we could ask to pass the full list of collaterals, with repaid=seized=0 for collaterals that the liquidator doesn't want

src/Terms.sol
MathisGD: and it should be relative (not absolute)

src/Terms.sol
gd-colin: Actually implemented here : https://github.com/morpho-org/v2/pull/3/commits/e406a6a388edec77de285133f804181cb2301944

src/Terms.sol
MathisGD: @colin-morpho

gd-colin: Actually, I don't remember introducing this change myself. Maybe I was on a bad/wrong commit.

src/Terms.sol
```diff
@@ -2,25 +2,32 @@
 pragma solidity 0.8.28;
 
 import "./libraries/Math.sol";
+import {MathLib, WAD} from "./libraries/MathLib.sol";
```
gd-colin: We should probably merge both math libs or move utility math functions (min,max) to a `UtilsLib`.

src/Terms.sol
```diff
@@ -2,25 +2,32 @@
...
+    uint256 public constant ORACLE_PRICE_SCALE = 1e36;
 
     /// STORAGE ///
 
     // Terms.
-    mapping(address => mapping(bytes32 => uint256)) public bondOf;
-    mapping(address => mapping(bytes32 => uint256)) public debtOf;
-    mapping(bytes32 => uint256) public withdrawable;
+    mapping(address => mapping(bytes32 => uint256)) public bondSharesOf;
+    mapping(address => mapping(bytes32 => uint256)) public debtSharesOf;
```
gd-colin: Alternatively, we could keep track of debt in assets and not shares. They both introduce a few conversions, so it may be a matter of preference. Please, comment if you have a preference or resolve otherwise.

src/Terms.sol
```diff
@@ -2,25 +2,32 @@
...
+    mapping(bytes32 => uint256) public withdrawableShares;
+    mapping(bytes32 => Market) public market;
```
gd-colin: We should probably pick a better suited name. Any ideas?

src/Terms.sol
```diff
@@ -96,8 +126,131 @@ contract Terms is ITerms {
         IERC20(collateral).transfer(msg.sender, amount);
     }
 
+    /// @notice Execute the given collection `seizures` on the given `term` of the given `borrower`.
```
gd-colin:
```suggestion
    /// @notice Execute the given collection of `seizures` on the given `term` of the given `borrower`.
```

src/Terms.sol
```diff
@@ -96,8 +126,131 @@ contract Terms is ITerms {
...
+    /// @dev On each seizure either `repaidAmounts` or `seizedAssets` should be equal to zero.
+    /// @param term The term of the bond.
+    /// @param seizures An array of amounts of debt to repay or assetd to seize with the index of the collateral in the term collaterals.
```
gd-colin:
```suggestion
    /// @param seizures An array of amounts of debt to repay or assets to seize with the index of the collateral in the term's collateral assets.
```

src/Terms.sol
```diff
@@ -96,8 +126,131 @@ contract Terms is ITerms {
...
+            (bool success, bytes memory returnData) = msg.sender.call(callbackData);
+            if (!success) lowLevelRevert(returnData);
+        }
+
+        IERC20(term.loanToken).transferFrom(msg.sender, address(this), totalRepaid);
+
+        return seizures;
+    }
+
+    function bondOf(address owner, bytes32 id) public view returns (uint256) {
```
gd-colin, line 206: Added this for convenience in tests, but they may not make sense in the final version.

src/Terms.sol
```diff
@@ -119,8 +272,8 @@ contract Terms is ITerms {
 
         require(buyOffer.offering != sellOffer.offering, "Same offering");
         require(buyOffer.loanToken == sellOffer.loanToken, "Loan tokens do not match");
+        uint256 j = 0;
```
gd-colin, line 233: Fixes a bug when verifying inclusion of collateral assets.

src/Terms.sol
```diff
@@ -131,6 +284,7 @@ contract Terms is ITerms {
             require(sellOffer.collaterals[i].token == buyOffer.collaterals[j].token, "Collaterals tokens do not match");
             require(sellOffer.collaterals[i].lltv <= buyOffer.collaterals[j].lltv, "LLTVs do not match");
             require(sellOffer.collaterals[i].oracle == buyOffer.collaterals[j].oracle, "Oracles do not match");
+            j++;
```
gd-colin, line 245: Here, we continue the search on where we left off. That is, on the next collateral asset from buyer.

src/interfaces/ITerms.sol
```diff
@@ -24,10 +24,25 @@ struct Offer {
     uint256 price;
 }
 
+struct Market {
```
gd-colin: Ditto on naming.

src/libraries/SharesMathLib.sol
```diff
@@ -0,0 +1,45 @@
...
+/// https://docs.openzeppelin.com/contracts/4.x/erc4626#inflation-attack.
+library SharesMathLib {
+    using MathLib for uint256;
+
+    /// @dev The number of virtual shares has been chosen low enough to prevent overflows, and high enough to ensure
+    /// high precision computations.
+    /// @dev Virtual shares can never be redeemed for the assets they are entitled to, but it is assumed the share price
+    /// stays low enough not to inflate these assets to a significant value.
+    /// @dev Warning: The assets to which virtual borrow shares are entitled behave like unrealizable bad debt.
+    uint256 internal constant VIRTUAL_SHARES = 1e6;
```
gd-colin, line 20: We should consider adjusting the precision of virtual shares, we may not need that much.

test/TermsTest.sol
```diff
@@ -1,37 +1,43 @@
 // SPDX-License-Identifier: UNLICENSED
 pragma solidity ^0.8.0;
 
-import {Test, console} from "../lib/forge-std/src/Test.sol";
-import "../src/Terms.sol";
-import {ERC20} from "./helpers/ERC20.sol";
+import "./BaseTest.sol";
+import {console} from "../lib/forge-std/src/Test.sol";
```
gd-colin:
```suggestion
```

src/Terms.sol
gd-colin: Implemented in https://github.com/morpho-org/v2/pull/3/commits/b7433df21bc0c2854f7a363a6320cb4ae1556dcb

src/Terms.sol
gd-colin: Fixed in https://github.com/morpho-org/v2/pull/3/commits/b7433df21bc0c2854f7a363a6320cb4ae1556dcb.

src/Terms.sol
gd-colin: Ok, interesting point. Although, I don't see the what is "leaky" here and the internal function is quite self-explanatory. It does exactly, what its name says. Noting less, nothing more. I actually find the code easier to read with this function. However, I will inline it later on once we settle on how we make sure that the list of seized collateral assets is coherent or sorted. It will make it easier to compare diffs.

src/Terms.sol
MathisGD: can you remove it please

src/Terms.sol
MathisGD: why would we account for the debt in shares? their value in underlying can't move

src/Terms.sol
MathisGD: why a struct? for the moment let's keep things simple.

also here we have a special case where we might not need the totalShares and totalAssets. Storing only an exchange rate might be fine, because we might not care that much about precision. And it's simpler

src/Terms.sol
```diff
@@ -2,25 +2,32 @@
...
 
     /// STORAGE ///
 
     // Terms.
-    mapping(address => mapping(bytes32 => uint256)) public bondOf;
-    mapping(address => mapping(bytes32 => uint256)) public debtOf;
-    mapping(bytes32 => uint256) public withdrawable;
+    mapping(address => mapping(bytes32 => uint256)) public bondSharesOf;
+    mapping(address => mapping(bytes32 => uint256)) public debtSharesOf;
+    mapping(bytes32 => uint256) public withdrawableShares;
```
MathisGD: why is this in shares? withdrawable token doesn't change when a bad debt is realised right?

src/libraries/MathLib.sol
MathisGD, line 1: you can remove things that we don't use from here

src/interfaces/ITerms.sol
```diff
@@ -24,10 +24,25 @@ struct Offer {
...
+    uint256 totalShares;
+}
+
 struct Signature {
     uint8 v;
     bytes32 r;
     bytes32 s;
 }
 
+struct Seizure {
```
MathisGD, line 33: I think that it would be simpler to require to pass the amounts for all collaterals, and not really more expensive

src/Terms.sol
```diff
@@ -96,8 +126,131 @@ contract Terms is ITerms {
...
+                collateralOf[borrower][id][term.collaterals[i].token].mulDivDown(price, ORACLE_PRICE_SCALE);
+            totalCollateralQuoted += collateralQuoted;
+            maxDebt += collateralQuoted.wMulDown(term.collaterals[i].lltv);
+        }
+
+        // Check that position is not healthy.
+        require(
+            debtSharesOf[borrower][id].toAssetsUp(market[id].totalAssets, market[id].totalShares) > maxDebt,
+            "Healthy borrower"
+        );
```
MathisGD: you can use isHealthy for now, we will do the optimizations later

src/Terms.sol
```diff
@@ -96,8 +126,131 @@ contract Terms is ITerms {
...
+        return debtSharesOf[owner][id].toAssetsDown(market[id].totalAssets, market[id].totalShares);
+    }
+
+    function withdrawable(bytes32 id) public view returns (uint256) {
+        return withdrawableShares[id].toAssetsDown(market[id].totalAssets, market[id].totalShares);
+    }
+
     /// INTERNAL ///
 
+    function _seizeCollateral(Collateral memory c, Seizure memory s, uint256 lif, address liquidator)
```
MathisGD: can you please inline it? or you see a good reason not to do it?

src/Terms.sol
MathisGD: nice catch

src/Terms.sol
MathisGD: nice catch

I don't think that the previous version wasn't working, but this was is more optimised indeed

src/libraries/SharesMathLib.sol
MathisGD: see [this](https://github.com/morpho-org/v2/pull/3#discussion_r2021767890)

src/Terms.sol
gd-colin: Looks like you missed this comment https://github.com/morpho-org/v2/pull/3#discussion_r2021740047

src/Terms.sol
gd-colin: As we discussed IRL, we need at least two values to account for bad debt, the index is not sufficient (we need to recompute total assets or total shares to update the index after a bad debt). Using total shares and total assets seems like a standard implementation. Resolving.

src/libraries/SharesMathLib.sol
gd-colin: Leaving unresolved to keep track of this.

src/Terms.sol
gd-colin: Done here https://github.com/morpho-org/v2/pull/3/commits/f967432bf9f2e5a3fe4f11484cb63f760107667e

src/Terms.sol
gd-colin: Done here https://github.com/morpho-org/v2/pull/3/commits/f967432bf9f2e5a3fe4f11484cb63f760107667e

src/interfaces/ITerms.sol
gd-colin: Implemented here https://github.com/morpho-org/v2/pull/3/commits/fe1115915f0bcdb3040b5bc76209923d03f623f3

src/libraries/MathLib.sol
gd-colin: Done here https://github.com/morpho-org/v2/pull/3/commits/f967432bf9f2e5a3fe4f11484cb63f760107667e

src/Terms.sol
gd-colin: Done here https://github.com/morpho-org/v2/pull/3/commits/f967432bf9f2e5a3fe4f11484cb63f760107667e

src/Terms.sol
gd-colin: @MathisGD you're right! Changed in https://github.com/morpho-org/v2/pull/3/commits/29d9e7d4faff0556760efd3856d8da7ffa7e8448

src/Terms.sol
gd-colin: I implemented your latest suggestion in https://github.com/morpho-org/v2/pull/3/commits/fe1115915f0bcdb3040b5bc76209923d03f623f3.

gd-colin: For reference, I added a comparison of gas [here](https://colab.research.google.com/drive/1L-pvtrh_EzYlWTicf_c-9oNz8hTHQ5p2?authuser=1).

The yellow and grey curves correspond to providing the full list of collateral assets whereas the green and red curves correspond to providing only the list of seized collateral assets.

We can observe that as the cardinal of collateral assets increase in the term, seizing fewer collateral suffers a considerable increase in gas. It's not dramatic, but we should keep this in mind since this may have an impact liquidation incentivization.

src/Terms.sol
gd-colin: You're actually right, I got mixed up. Fixed in https://github.com/morpho-org/v2/pull/3/commits/f81640181e223c3db37e86146116736f3d35181c.

src/Terms.sol
gd-colin: No, it wasn't working. Notably, it would fail with the collateral assets lists being equal and larger than one.

src/Terms.sol
```diff
@@ -68,9 +81,14 @@ contract Terms is ITerms {
     function withdrawBond(Term memory term, uint256 amount, address onBehalf) external {
         bytes32 id = _id(term);
 
-        bondOf[onBehalf][id] -= amount;
+        uint256 shares = amount.toSharesUp(totalAssets[id], totalShares[id]);
```
gd-colin: Maybe this should be rounded down.

MathisGD: no, it should be rounded up. You want to burn too much shares

src/Terms.sol
```diff
@@ -96,8 +114,100 @@ contract Terms is ITerms {
...
+    /// @param term The term of the bond.
+    /// @param seizures An array of amounts of debt to repay or assets to seize with the index of the collateral in the term's collateral assets.
+    /// @param borrower The debtor of the loan.
+    /// @param data Arbitrary data to pass to the callback. Pass empty data if not needed.
+    /// @return A collection of the actual amounts of debt repaid or asset seized with the collateral index.
+    function liquidate(Term memory term, Seizure[] memory seizures, address borrower, bytes calldata data)
+        external
+        returns (Seizure[] memory)
+    {
+        require(seizures.length == term.collaterals.length, "Cannot seize more assets than the supplied collaterals");
```
```suggestion
        require(seizures.length == term.collaterals.length, "Cannot seize more nor less assets than the bond's collaterals");
```

src/Terms.sol
```diff
@@ -47,13 +55,18 @@ contract Terms is ITerms {
         bytes32 id = _id(term);
 
         uint256 repaid = Math.min(debtOf[buyer][id], amount);
+        uint256 bought = amount - repaid;
         debtOf[buyer][id] -= repaid;
-        bondOf[buyer][id] += amount - repaid;
+        bondSharesOf[buyer][id] += bought.toSharesDown(totalAssets[id], totalShares[id]);
```
```suggestion
        bondSharesOf[buyer][id] += (amount - repaid).toSharesDown(totalAssets[id], totalShares[id]);
```

src/Terms.sol
```diff
@@ -96,8 +114,100 @@ contract Terms is ITerms {
...
+    {
+        require(seizures.length == term.collaterals.length, "Cannot seize more assets than the supplied collaterals");
+
+        bytes32 id = _id(term);
+
+        // Over approximation
+        uint256 liquidationIncentiveFactor = 1.15e18;
+
+        uint256 totalRepaid;
+        uint256 totalCollateralQuoted;
```
MathisGD: we only care about if this value is zero or not at the end. So we could have also boolean `allCollateralsAreZero` and update it accordingly. Btw, it fixes the problem of roundings when quoting. I think that we should go for that

src/Terms.sol
```diff
@@ -96,8 +114,100 @@ contract Terms is ITerms {
...
+        // Check that position is not healthy.
+        require(!_isHealthy(term, borrower), "Healthy borrower");
+
+        // Compute the repaid and seized amounts by collateral index, remaining collateral and total repaid.
+        for (uint256 i = 0; i < term.collaterals.length; i++) {
+            uint256 collateralPrice = IOracle(term.collaterals[i].oracle).price();
+            uint256 collateralQuoted =
+                collateralOf[borrower][id][term.collaterals[i].token].mulDivDown(collateralPrice, ORACLE_PRICE_SCALE);
+            if ((seizures[i].repaidAmount + seizures[i].seizedAssets) != 0) {
+                require(exactlyOneZero(seizures[i].seizedAssets, seizures[i].repaidAmount), "INCONSISTENT_INPUT");
```
MathisGD: with [this](https://github.com/morpho-org/v2/pull/3/files#r2027155294), we could have:
```suggestion
            require(minOneZero(seizures[i].repaidAmount, seizures[i].seizedAssets, "INCONSISTENT_INPUT");
```

src/Terms.sol
```diff
@@ -96,8 +114,100 @@ contract Terms is ITerms {
...
+            }
+        }
+
+        debtOf[borrower][id] -= totalRepaid;
+        withdrawable[id] += totalRepaid;
+
+        // Realize bad debt.
+        if (totalCollateralQuoted == 0) {
+            uint256 badDebt = debtOf[borrower][id];
+            totalAssets[id] -= badDebt;
```
```suggestion
            totalAssets[id] -= debtOf[borrower][id];
```

src/Terms.sol
```diff
@@ -96,8 +114,100 @@ contract Terms is ITerms {
...
+                totalCollateralQuoted += collateralQuoted - seizedAssetsQuoted;
+            } else {
+                totalCollateralQuoted += collateralQuoted;
+            }
+        }
+
+        debtOf[borrower][id] -= totalRepaid;
+        withdrawable[id] += totalRepaid;
+
+        // Realize bad debt.
```
MathisGD: you have to realise bad debt, right?

src/Terms.sol
```diff
@@ -96,8 +114,100 @@ contract Terms is ITerms {
...
+        // Compute the repaid and seized amounts by collateral index, remaining collateral and total repaid.
+        for (uint256 i = 0; i < term.collaterals.length; i++) {
+            uint256 collateralPrice = IOracle(term.collaterals[i].oracle).price();
+            uint256 collateralQuoted =
+                collateralOf[borrower][id][term.collaterals[i].token].mulDivDown(collateralPrice, ORACLE_PRICE_SCALE);
+            if ((seizures[i].repaidAmount + seizures[i].seizedAssets) != 0) {
+                require(exactlyOneZero(seizures[i].seizedAssets, seizures[i].repaidAmount), "INCONSISTENT_INPUT");
+
+                // Perform the seizure
+                uint256 seizedAssetsQuoted = seizures[i].seizedAssets.mulDivDown(collateralPrice, ORACLE_PRICE_SCALE);
```
MathisGD: this is only if the input is seizedAssets, right?

src/Terms.sol
```diff
@@ -96,8 +114,100 @@ contract Terms is ITerms {
...
+                if (seizures[i].seizedAssets > 0) {
+                    seizures[i].repaidAmount = seizedAssetsQuoted.wDivUp(liquidationIncentiveFactor);
+                } else {
+                    seizures[i].seizedAssets = seizures[i].repaidAmount.wMulDown(liquidationIncentiveFactor).mulDivUp(
+                        ORACLE_PRICE_SCALE, collateralPrice
+                    );
+                    seizedAssetsQuoted = seizures[i].seizedAssets.mulDivDown(collateralPrice, ORACLE_PRICE_SCALE);
+                }
```
```suggestion
                if (seizures[i].seizedAssets > 0) {
                    uint256 seizedAssetsQuoted = seizures[i].seizedAssets.mulDivDown(collateralPrice, ORACLE_PRICE_SCALE);
                    seizures[i].repaidAmount = seizedAssetsQuoted.wDivUp(liquidationIncentiveFactor);
                } else {
                    seizures[i].seizedAssets = seizures[i].repaidAmount.wMulDown(liquidationIncentiveFactor).mulDivUp(
                        ORACLE_PRICE_SCALE, collateralPrice
                    );
                }
```
based on [this](https://github.com/morpho-org/v2/pull/3/files#r2027155294) as well

src/Terms.sol
```diff
@@ -96,8 +114,100 @@ contract Terms is ITerms {
...
+                IERC20(term.collaterals[i].token).transfer(msg.sender, seizures[i].seizedAssets);
+
+                collateralOf[borrower][_id(term)][term.collaterals[i].token] -= seizures[i].seizedAssets;
+                totalRepaid += seizures[i].repaidAmount;
+                totalCollateralQuoted += collateralQuoted - seizedAssetsQuoted;
+            } else {
+                totalCollateralQuoted += collateralQuoted;
+            }
+        }
```
```suggestion
            if (collateralOf[borrower]... != 0) allCollateralIsZero = false;
        }
```

src/Terms.sol
```diff
@@ -96,8 +114,100 @@ contract Terms is ITerms {
...
+
+        // Realize bad debt.
+        if (totalCollateralQuoted == 0) {
+            uint256 badDebt = debtOf[borrower][id];
+            totalAssets[id] -= badDebt;
+            debtOf[borrower][id] = 0;
+        }
+
+        // Perform the callback.
+        // TODO: simplify with dedicated signature for callback
```
MathisGD: let's do it! you can basically take V1's interface

src/Terms.sol
```diff
@@ -96,8 +114,100 @@ contract Terms is ITerms {
...
+    }
+
+    function bondOf(address owner, bytes32 id) public view returns (uint256) {
+        return bondSharesOf[owner][id].toAssetsDown(totalAssets[id], totalShares[id]);
+    }
+
     /// INTERNAL ///
 
+    // TODO: move to a dedicated library
+    function exactlyOneZero(uint256 x, uint256 y) internal pure returns (bool z) {
```
MathisGD: I think that we can inline for the moment (no need for assembly)

test/LiquidationTest.sol
```diff
@@ -0,0 +1,236 @@
...
+        loanToken.transfer(liquidator, 1000);
+        Term memory t = liquidationTerms[n - 1];
+        Oracle(t.collaterals[0].oracle).setPrice(0.25e36);
+
+        vm.prank(liquidator);
+        uint256 gasBefore;
+        uint256 gasUsed;
+        if (n == 10) {
+            gasBefore = gasleft();
+            terms.liquidate(t, sK[k - 1], borrower, "0x0");
```
MathisGD: except if you were on purpose making an empty call
```suggestion
            terms.liquidate(t, sK[k - 1], borrower, hex"");
```

src/Terms.sol
gd-colin: The variable is actually used in several places and helps readability.

src/Terms.sol
gd-colin: What do you mean?

MathisGD: where is the bad debt realised? nothing happens for the moment. We just remove the debt of the borrower. Lenders don't loose anything

src/Terms.sol
```diff
@@ -96,8 +115,77 @@ contract Terms is ITerms {
...
+
+            allCollateralsAtZero = allCollateralsAtZero && collateralOf[borrower][id][term.collaterals[i].token] == 0;
+        }
+
+        debtOf[borrower][id] -= totalRepaid;
+        withdrawable[id] += totalRepaid;
+
+        // Realize bad debt.
+        if (allCollateralsAtZero) {
+            totalAssets[id] -= debtOf[borrower][id];
```
gd-colin: They do loose the bad debt, don't they @MathisGD

src/Terms.sol
gd-colin: See https://github.com/morpho-org/v2/pull/3/files#r2031146719

src/Terms.sol
gd-colin: Look carefully at the `-=`, which diminishes the value of the share

MathisGD: right, i misread

src/Terms.sol
MathisGD: yes, and remove toAssetsUp as we don't use it

src/Terms.sol
```diff
@@ -68,9 +82,14 @@ contract Terms is ITerms {
     function withdrawBond(Term memory term, uint256 amount, address onBehalf) external {
```
MathisGD: it's better to put the shares input (or both) to be able to withdraw all shares (here it might not be possible because of roundings)

src/Terms.sol
```diff
@@ -96,8 +115,77 @@ contract Terms is ITerms {
...
+    }
+
     /// INTERNAL ///
 
+    function lowLevelRevert(bytes memory returnData) internal pure {
+        assembly ("memory-safe") {
+            revert(add(32, returnData), mload(returnData))
+        }
+    }
+
```
```suggestion
```

src/Terms.sol
```diff
@@ -132,19 +132,19 @@ contract Terms is ITerms {
...
         uint256 maxDebt;
-        uint256 liquidatableDebt;
+        uint256 repayableDebt;
         for (uint256 i = 0; i < term.collaterals.length; i++) {
             uint256 price = IOracle(term.collaterals[i].oracle).price();
             uint256 collateralQuoted =
                 collateralOf[borrower][id][term.collaterals[i].token].mulDivDown(price, ORACLE_PRICE_SCALE);
             maxDebt += collateralQuoted.wMulDown(term.collaterals[i].lltv);
-            liquidatableDebt += collateralQuoted.wDivDown(liquidationIncentiveFactor);
+            repayableDebt += collateralQuoted.wDivUp(liquidationIncentiveFactor);
```
MathisGD, line 156: didn't we say it must be down?

gd-colin: The note said up, and rounding down made the test fail. But I'll investigate this thoroughly.

gd-colin: I double checked, here when we round up, it means that the bad debt will be rounded down, i.e. the liquidator will pay more and the borrowers will suffer less bad debt. We also need that the property totalRepaid <= repayableDebt holds.

src/Terms.sol
gd-colin: Done here https://github.com/morpho-org/v2/pull/3/commits/ab78e592b3bb8822f6e8eded960266806428a917

src/Terms.sol
gd-colin: Done here https://github.com/morpho-org/v2/pull/3/commits/ab78e592b3bb8822f6e8eded960266806428a917

src/libraries/SharesMathLib.sol
gd-colin: Tracked in follow-up issue #6

src/Terms.sol
```diff
@@ -1,26 +1,37 @@
...
-import "./libraries/Math.sol";
+import "./libraries/UtilsLib.sol";
+import {MathLib, WAD} from "./libraries/MathLib.sol";
+import {SharesMathLib} from "./libraries/SharesMathLib.sol";
 import "./interfaces/IERC20.sol";
 import "./interfaces/IOracle.sol";
 import "./interfaces/ITerms.sol";
+import "./interfaces/IMorphoLiquidationCallback.sol";
+
+import "../lib/forge-std/src/console.sol";
```
```suggestion
```

src/Terms.sol
```diff
@@ -46,14 +57,20 @@ contract Terms is ITerms {
...
+        uint256 bought = amount - repaid;
         debtOf[buyer][id] -= repaid;
-        bondOf[buyer][id] += amount - repaid;
+        bondSharesOf[buyer][id] += bought.toSharesDown(totalAssets[id], totalShares[id]);
 
-        uint256 withdrawn = Math.min(bondOf[seller][id], amount);
-        bondOf[seller][id] -= withdrawn;
+        uint256 withdrawn =
+            UtilsLib.min(bondSharesOf[seller][id].toAssetsDown(totalAssets[id], totalShares[id]), amount);
+        bondSharesOf[seller][id] -= withdrawn;
```
```suggestion
        bondSharesOf[seller][id] -= withdrawn.toSharesUp(totalAssets[id], totalShares[id]);
```
because `withdrawn` is an amount of underlying, not a quantity of shares

src/Terms.sol
```diff
@@ -46,14 +57,20 @@ contract Terms is ITerms {
...
         debtOf[seller][id] += amount - withdrawn;
 
+        uint256 boughtAmount = (bought - withdrawn);
```
QGarchery: This could very well underflow right ? We could imagine that `bought` is 0 notably, because the buyer had no previous debt, and that `withdrawn` is not 0 because the seller had bonds previously. This makes sense because it means that the overall quantity of bonds can decrease

src/Terms.sol
```diff
@@ -46,14 +57,20 @@ contract Terms is ITerms {
...
+        totalShares[id] += boughtAmount.toSharesDown(totalAssets[id], totalShares[id]);
```
QGarchery: I think this should be an exact computation, so that we have the invariant: "totalShares is the sum of all shares of all participants". To do that it should be modified to follow the above modifications of `bondSharesOf[buyer][id]` and `bondSharesOf[seller][id]`

src/Terms.sol
```diff
@@ -46,14 +57,20 @@ contract Terms is ITerms {
...
+        uint256 repaid = UtilsLib.min(debtOf[buyer][id], amount);
+        uint256 bought = amount - repaid;
         debtOf[buyer][id] -= repaid;
-        bondOf[buyer][id] += amount - repaid;
+        bondSharesOf[buyer][id] += bought.toSharesDown(totalAssets[id], totalShares[id]);
 
-        uint256 withdrawn = Math.min(bondOf[seller][id], amount);
-        bondOf[seller][id] -= withdrawn;
+        uint256 withdrawn =
+            UtilsLib.min(bondSharesOf[seller][id].toAssetsDown(totalAssets[id], totalShares[id]), amount);
```
QGarchery: When introducing shares here there is a tricky rounding consideration that does not exist in Morpho Blue for example. In Morpho Blue, every time a rounding occurs, it is at the expense of the user that is interacting with the protocol. It's different here because `MATCH` is applied also to the maker, and that user is not interacting directly with the protocol. 

QGarchery (commented):
Partial review for now: I only checked the shares mechanism, and I need to look into the `liquidate` function itself

src/Terms.sol
QGarchery: Here we have a discrepancy between how `totalShares[id]` changes and how `bondSharesOf[buyer][id]` plus `bondSharesOf[seller][id]` change. This means that this isn't an exact computation in the following sense: `totalShares` is not the sum of all shares of all participants

src/Terms.sol
gd-colin: `bought` is equal to zero if the `amount == debtOf[buyer]` not when the buyer had no debt, in this case it would underflow indeed

gd-colin: In such case, the bought amount should be equal to 0 as no new bonds are issued and the withdrawn bonds should be subtracted from the total shares and assets indeed.

src/Terms.sol
gd-colin: However, if we have the offer posted in shares, we would still have a similar issue, wouldn't we? The repaid amount in shares would be rounded (down most likely), such that bought bonds are rounded up.

src/Terms.sol
gd-colin: Implemented in https://github.com/morpho-org/v2/pull/3/commits/d9ac321ddccabb1a8c51e7863163b5bbf8b1e0df

src/Terms.sol
gd-colin: Fixed by https://github.com/morpho-org/v2/pull/3/commits/d9ac321ddccabb1a8c51e7863163b5bbf8b1e0df

src/Terms.sol
gd-colin: In https://github.com/morpho-org/v2/pull/3/commits/925feff66c7957b4a32dd0c1fe76b8b0bd419726 I pushed an attempt to limit unfavourable rounding on the seller. The idea is to use the exact seller shares of the seller when possible.

QGarchery: I don't think that it really solves the issue, because the amount could be chosen low enough so that the min returns `amount`. Then we have the same computation of `amount.toSharesUp(totalAssets[id], totalShares[id])` which is unfavorable for the `seller`

QGarchery: Also the issue could apply to other lines of code, and not just for the seller. It could be that the `seller` is initiating the transaction and then the buyer would lose because of the line `bondSharesOf[buyer][id] += bought.toSharesDown(totalAssets[id], totalShares[id])` that can be manipulated through choosing the `amount`

src/Terms.sol
QGarchery: Re-opening because I thought that we wanted to make the bad debt round up, so that we know that the bad debt cannot be realized multiple times. This ensures that the bad debt does not depend on the actual amounts liquidated (given in `seizures`), because it cannot be repeated. This prevents potential manipulations that would do a lot of small bad debt realizations. 
To me it's ok if bad debt is a bit over-estimated one time. The reasoning is that it it happens after the threshold of `repayableDebt`, a threshold that is higher than when the position becomes liquidatable

> it means that the bad debt will be rounded down, i.e. the liquidator will pay more and the borrowers will suffer less bad debt

More bad debt realized does not mean that liquidators pay more, the same ratio (LIF) is used independently of if there is bad debt or not. Also bad debt realization is only beneficial to borrowers

src/Terms.sol
```diff
@@ -96,6 +129,82 @@ contract Terms is ITerms {
...
+        for (uint256 i = 0; i < term.collaterals.length; i++) {
+            if (seizures[i].repaidAmount + seizures[i].seizedAssets > 0) {
+                require(
+                    UtilsLib.exactlyOneZero(seizures[i].repaidAmount, seizures[i].seizedAssets), "INCONSISTENT_INPUT"
+                );
+
+                uint256 collateralPrice = IOracle(term.collaterals[i].oracle).price();
+
+                if (seizures[i].seizedAssets > 0) {
+                    seizures[i].repaidAmount = seizures[i].seizedAssets.mulDivDown(collateralPrice, ORACLE_PRICE_SCALE)
```
QGarchery, line 171:
```suggestion
                    seizures[i].repaidAmount = seizures[i].seizedAssets.mulDivUp(collateralPrice, ORACLE_PRICE_SCALE)
```
because we round against the liquidator

src/Terms.sol
```diff
@@ -96,6 +129,82 @@ contract Terms is ITerms {
...
+                        .wDivUp(liquidationIncentiveFactor);
+                } else {
+                    seizures[i].seizedAssets = seizures[i].repaidAmount.wMulDown(liquidationIncentiveFactor).mulDivUp(
```
```suggestion
                    seizures[i].seizedAssets = seizures[i].repaidAmount.wMulDown(liquidationIncentiveFactor).mulDivDown(
```
same

src/Terms.sol
```diff
@@ -96,6 +129,82 @@ contract Terms is ITerms {
...
+                        ORACLE_PRICE_SCALE, collateralPrice
+                    );
+                }
+
+                totalRepaid += seizures[i].repaidAmount;
+                collateralOf[borrower][id][term.collaterals[i].token] -= seizures[i].seizedAssets;
+
+                IERC20(term.collaterals[i].token).transfer(msg.sender, seizures[i].seizedAssets);
```
QGarchery, line 182: It seems somewhat dangerous to do the transfers before updating the whole state. For example, if one token allows to call an arbitrary address (see ERC-777 notably), then it can execute arbitrary code and reenter at this point. Note that it is especially dangerous because the account is definitely liquidatable still: `debtOf[borrower][id]` is decreased only later, and `collateralOf[borrower][id]` is decreased before

src/Terms.sol
```diff
@@ -96,6 +129,82 @@ contract Terms is ITerms {
...
+        require(seizures.length == term.collaterals.length, "should have all collats");
+
+        bytes32 id = _id(term);
+        uint256 liquidationIncentiveFactor = 1.15e18;
+
+        uint256 maxDebt;
+        uint256 repayableDebt;
+
+        for (uint256 i = 0; i < term.collaterals.length; i++) {
+            uint256 price = IOracle(term.collaterals[i].oracle).price();
```
QGarchery, line 152: Not important for now, but we could look to cache this price for gas optimization, and also making sure that it's the same price used during this whole `liquidation` call. So it would mean to allocate a new memory array of size `seizure.length`

src/Terms.sol
```diff
@@ -96,6 +129,82 @@ contract Terms is ITerms {
...
+
+                totalRepaid += seizures[i].repaidAmount;
+                collateralOf[borrower][id][term.collaterals[i].token] -= seizures[i].seizedAssets;
+
+                IERC20(term.collaterals[i].token).transfer(msg.sender, seizures[i].seizedAssets);
+            }
+        }
+
+        debtOf[borrower][id] -= totalRepaid;
+        withdrawable[id] += totalRepaid;
```
QGarchery, line 197: Probably not a huge deal, but one downside I can see when asking the liquidator to come with the loan tokens instead of asking him to come with the bonds: if the liquidator has the bonds, the liquidity needs to be on the Term contract to begin with. This is because the liquidator needs to withdraw the bonds first, and at this step the repayment has not been made yet (it happens at the end of the function)

src/Terms.sol
gd-colin: Opening follow up issue #8  to track this as this challenges optimization, if we handle seized collateral transfers we need to add an extra loop after updating the state.

src/Terms.sol
gd-colin: Opening follow-up issue #9

src/Terms.sol
gd-colin: This is fixed in https://github.com/morpho-org/v2/pull/3/commits/f440d9180ff1ce2df1aad48c7990dfa6256f92a7. More about this in [slack comment](https://morpholabs.slack.com/archives/C02N7CZ088N/p1747755908900889).

src/Terms.sol
gd-colin: Fixed in https://github.com/morpho-org/v2/pull/3/commits/f440d9180ff1ce2df1aad48c7990dfa6256f92a7.

src/Terms.sol
gd-colin: Fixed in https://github.com/morpho-org/v2/pull/3/commits/f440d9180ff1ce2df1aad48c7990dfa6256f92a7.

src/Terms.sol
gd-colin: Opening followup issue #10. This problem is more general and goes beyond the scope of this PR. Let's handle it separately.

src/Terms.sol
gd-colin: If you have the bonds, you can always flash loan the loan asset, liquidate with the loan asset, withdraw the bonds and repay the flash loan, can't you?

QGarchery: Yes exactly ! So the advantage is only about not requiring some flashloan liquidity. Note that callbacks in morpho-blue are there only for that, as you could do the same things as callbacks but with flashloans (if there is enough flashloanable liquidity)

gd-colin: Opening follow up issue #11 to track this. Main observations are that this design is not optimized for the most profitable liquidation process (i.e. getting the liquidation bonus + withdrawing bonds) and it also requires liquidity of the loan asset.

MathisGD (approved):

PR #7, opened 2025-05-16
Feat/callback

peyha:
Implement callback in offers (and a test which withdraws from a morpho market to lend in a bond).
This PR was primarily made to have a rough estimate of the gas cost of matching an offer while executing a callback (I got 200k gas which seems ok)


MathisGD:
will re-implement later

PR #13, opened 2025-05-27
Feat liquidate with bonds

gd-colin:
This PR refactors liquidation to require the liquidator to use bonds instead of loan asset.


MathisGD:
closing, as discussed in the liquidation workshop. A bit argument was notably that it makes losses realisation weird, because it gets realised on less bonds

Issue #10, opened 2025-05-21
Grieffing attack when rounding seller/buyer offers


gd-colin:
When introducing shares here there is a tricky rounding consideration that does not exist in Morpho Blue for example. In Morpho Blue, every time a rounding occurs, it is at the expense of the user that is interacting with the protocol. It's different here because `MATCH` is applied also to the maker, and that user is not interacting directly with the protocol. 

With this specific line of code for example, it could be that the `amount` is chosen just so that an unfavorable rounding occurs for the `seller`. Done repeatedly this could be a serious griefing attack, depending on the particular decimals

_Originally posted by @QGarchery in https://github.com/morpho-org/v2/pull/3#discussion_r2046371893_

MathisGD:
see this comment https://github.com/morpho-org/v2/pull/16#discussion_r2175543517

MathisGD:
see this other comment: https://github.com/morpho-org/v2/pull/16/files#r2177725646

PR #16, opened 2025-06-30
new  interface

MathisGD:


src/Terms.sol
```diff
@@ -36,42 +33,29 @@ contract Terms is ITerms {
...
-        consumed[abi.encode(sellOffer)] += amount;
+        consumed[abi.encode(offer)] += amount;
 
-        Term memory term = Term(sellOffer.loanToken, sellOffer.collaterals, sellOffer.maturity);
         bytes32 id = _id(term);
 
         uint256 repaid = UtilsLib.min(debtOf[buyer][id], amount);
         uint256 bought = amount - repaid;
-        uint256 boughtShares = bought.toSharesDown(totalAssets[id], totalShares[id]);
+        uint256 boughtShares = bought.mulDivDown(totalShares[id] + 1, totalAssets[id] + 1);
```
gd-colin, line 54: Shouldn't rounding be done in function who takes and not who buys? Said otherwise, why a maker buyer will see *his* shares rounded down in lieux of the seller's? I'm not sure it makes any sense to round in "favor of protocol" and I don't see the justification to favor the seller.

test/TermsTest.sol
```diff
@@ -13,7 +13,7 @@ contract TermsTest is BaseTest {
     address private borrower;
     uint256 private lenderSK;
     address private lender;
-    address private liquidator;
+    address private liquidator = makeAddr("liquidator");
```
gd-colin, line 16: Why is `makeAddr` moved here *only* for this case?

MathisGD: for the other ones, we craft both an address an a PK

src/Terms.sol
```diff
@@ -36,42 +33,29 @@ contract Terms is ITerms {
...
+        uint256 withdrawn =
+            UtilsLib.min(bondSharesOf[seller][id].mulDivDown(totalAssets[id] + 1, totalShares[id] + 1), amount);
+        uint256 withdrawnShares = withdrawn.mulDivDown(totalShares[id] + 1, totalAssets[id] + 1);
```
```suggestion
        uint256 withdrawnShares = withdrawn.mulDivUp(totalShares[id] + 1, totalAssets[id] + 1);
```

src/Terms.sol
MathisGD: there was a mistake on the other line which is fixed one. Though I don't understand why rounding in favor of the protocol wouldn't be suited here, like in other protocols?

gd-colin: I think the point was (is?) that you can have a grieffing attack either way.

gd-colin: Related issue #10.

MathisGD: right, because it's at the expense of both users being matched, the taker and the maker. The solution to that is probably not to round the other way around anyway, because the maker could also well be the person making the transaction, and manipulating the protocol.

one great thing about bonds is that the share price can only go down, not up. Meaning that we will always have totalShares>=totalAssets (or sharePrice<=1), so roundings will never be greater than 1. We could simply rely on the fact that 1 asset is always smaller than the gas price required to do a transaction.

MathisGD: let's move this discussion to #10

src/Terms.sol
```diff
@@ -213,39 +194,23 @@ contract Terms is ITerms {
...
-            while (
-                bytes20(sellOffer.collaterals[i].token) < bytes20(buyOffer.collaterals[j].token)
-                    && j++ < buyOffer.collaterals.length
-            ) {}
-            require(sellOffer.collaterals[i].token == buyOffer.collaterals[j].token, "Collaterals tokens do not match");
-            require(sellOffer.collaterals[i].lltv <= buyOffer.collaterals[j].lltv, "LLTVs do not match");
-            require(sellOffer.collaterals[i].oracle == buyOffer.collaterals[j].oracle, "Oracles do not match");
+            for (; j < offer.collaterals.length; j++) {
+                if (offer.collaterals[j].token == term.collaterals[i].token || j == offer.collaterals.length) break;
+            }
```
```suggestion
            for (; offer.collaterals[j].token != term.collaterals[i].token; j++) {}
```
it should be equivalent except that if no match is found then it reverts with out of bounds (it will revert later with out of bounds anyway).
Note that `|| j == offer.collaterals.length` is dead in the code of this PR

src/Terms.sol
```diff
@@ -213,39 +194,23 @@ contract Terms is ITerms {
...
+            require(offer.collaterals[i].token == offer.collaterals[j].token, "Collaterals tokens do not match");
+            require(offer.collaterals[i].lltv <= offer.collaterals[j].lltv, "LLTVs do not match");
+            require(offer.collaterals[i].oracle == offer.collaterals[j].oracle, "Oracles do not match");
```
```suggestion
            require(term.collaterals[i].token == offer.collaterals[j].token, "Collaterals tokens do not match");
            require(term.collaterals[i].lltv <= offer.collaterals[j].lltv, "LLTVs do not match");
            require(term.collaterals[i].oracle == offer.collaterals[j].oracle, "Oracles do not match");
```

MathisGD: nice catch

>require(term.collaterals[i].lltv <= offer.collaterals[j].lltv, "LLTVs do not match");

though, actually this is not what we expect, will think about it

src/Terms.sol
```diff
@@ -36,42 +33,29 @@ contract Terms is ITerms {
...
-            buyOffer.assets - consumed[abi.encode(buyOffer)], sellOffer.assets - consumed[abi.encode(sellOffer)]
-        );
-        require(amount > 0, "No assets to match");
-        address buyer = buyOffer.offering;
-        address seller = sellOffer.offering;
+        (address buyer, address seller) = offer.buy ? (offer.offering, onBehalf) : (onBehalf, offer.offering);
 
-        consumed[abi.encode(buyOffer)] += amount;
-        consumed[abi.encode(sellOffer)] += amount;
+        consumed[abi.encode(offer)] += amount;
```
QGarchery, line 47:
```suggestion
        consumed[abi.encode(offer)] += amount;
        require(consumed[abi.encode(offer)] <= offer.assets);
```
otherwise an offer can be taken without limit

src/Terms.sol
QGarchery: Right, see #22

src/Terms.sol
```diff
@@ -213,39 +194,23 @@ contract Terms is ITerms {
...
-    function _checkOffers(
-        Offer memory buyOffer,
-        Signature memory buySig,
-        Offer memory sellOffer,
-        Signature memory sellSig
-    ) internal view {
-        // Check consistency.
-
-        require(buyOffer.buy && !sellOffer.buy, "Inconsistent lend flags");
-        require(buyOffer.maturity > block.timestamp, "Buy offer has expired");
```
QGarchery, line 225: Because it deletes this line, after this PR it is possible to trade after the term (see #23)

MathisGD: it's a mistake. I just wanted to move the check to the take function. I don't think that people should be able to take after the term

src/Terms.sol
```diff
@@ -36,42 +33,29 @@ contract Terms is ITerms {
...
-        Term memory term = Term(sellOffer.loanToken, sellOffer.collaterals, sellOffer.maturity);
         bytes32 id = _id(term);
 
         uint256 repaid = UtilsLib.min(debtOf[buyer][id], amount);
         uint256 bought = amount - repaid;
-        uint256 boughtShares = bought.toSharesDown(totalAssets[id], totalShares[id]);
+        uint256 boughtShares = bought.mulDivDown(totalShares[id] + 1, totalAssets[id] + 1);
+        uint256 withdrawn =
+            UtilsLib.min(bondSharesOf[seller][id].mulDivDown(totalAssets[id] + 1, totalShares[id] + 1), amount);
+        uint256 withdrawnShares = withdrawn.mulDivUp(totalShares[id] + 1, totalAssets[id] + 1);
```
QGarchery, line 57: Weirdly, this refactor could make so it's not possible to get to 0 shares exactly. It shouldn't be an issue though because 0 is not a magic value for `bondSharesOf` (contrary to `debtOf`, where a debt of 0 ensures that the position is healthy)

MathisGD: adding to the issue on roundings

src/Terms.sol
MathisGD: I don't think that I understand #22, let's discuss it

MathisGD: in the meanwhile i pushed a fix

src/Terms.sol
```diff
@@ -36,42 +33,31 @@ contract Terms is ITerms {
 
     /// ENTRY-POINTS ///
 
-    /// @dev This function is used for both primary and secondary markets.
-    function MATCH(Offer memory buyOffer, Signature memory buySig, Offer memory sellOffer, Signature memory sellSig)
+    /// @dev Same function used to buy and sell.
+    /// @dev If one wants to make to offers without taking a position, they can batch take them and not have a position at the end.
+    function take(Term memory term, uint256 amount, address onBehalf, Offer memory offer, Signature memory sig)
```
QGarchery, line 38: There is an authorization issue, because how do we know that `onBehalf` is ok to `take` here ? Before there was a signature that was required for both parties, but now, for example, `onBehalf` could be the buyer and the sender could be the seller that crafts a worthless bond

src/Terms.sol
```diff
@@ -213,39 +196,23 @@ contract Terms is ITerms {
...
-            // If they are not, the match could fail.
-            while (
-                bytes20(sellOffer.collaterals[i].token) < bytes20(buyOffer.collaterals[j].token)
-                    && j++ < buyOffer.collaterals.length
-            ) {}
-            require(sellOffer.collaterals[i].token == buyOffer.collaterals[j].token, "Collaterals tokens do not match");
-            require(sellOffer.collaterals[i].lltv <= buyOffer.collaterals[j].lltv, "LLTVs do not match");
-            require(sellOffer.collaterals[i].oracle == buyOffer.collaterals[j].oracle, "Oracles do not match");
+            // If they are not, the matching could fail.
+            for (; superset[j].token != subset[i].token; j++) {}
```
```suggestion
            while (superset[j].token != subset[i].token) j++;
```
(optional) maybe slightly more natural

QGarchery (changes requested):

src/Terms.sol
MathisGD: yes authorizations are not implemented yet

QGarchery (approved):

PR #20, opened 2025-07-01
nonce

### LINKED ISSUE #17: nonces

MathisGD:
fixes #17

### DIFF
--- a/src/Terms.sol
+++ b/src/Terms.sol
@@ -28,8 +28,10 @@ contract Terms is ITerms {
     mapping(bytes32 => uint256) public totalAssets;
     mapping(bytes32 => uint256) public totalShares;
     mapping(address => mapping(bytes32 => mapping(address => uint256))) public collateralOf;
+
     // Offers.
-    mapping(bytes => uint256) public consumed;
+    // users -> nonce -> consumed. Multiple offers can have the same nonce.
+    mapping(address => mapping(uint256 => uint256)) public consumed;
 
     /// ENTRY-POINTS ///
 
@@ -43,7 +45,8 @@ contract Terms is ITerms {
 
         (address buyer, address seller) = offer.buy ? (offer.offering, onBehalf) : (onBehalf, offer.offering);
 
-        consumed[abi.encode(offer)] += amount;
+        require(amount <= offer.assets - consumed[offer.offering][offer.nonce], "consumed");
+        consumed[offer.offering][offer.nonce] += amount;
 
         bytes32 id = _id(term);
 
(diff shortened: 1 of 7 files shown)


src/interfaces/ITerms.sol
```diff
@@ -22,6 +22,7 @@ struct Offer {
     Collateral[] collaterals;
     uint256 maturity;
     uint256 price;
+    uint256 nonce;
```
MathisGD, line 26: if you remove the nonce from the struct, the sanity rules in certora pass..

MathisGD: i ended up simplifying the sanity check. don't know what was happening

QGarchery: > if you remove the nonce from the struct, the sanity rules in certora pass..

Why do you feel like it shouldn't pass, the nonce didn't appear in the previous spec

MathisGD: it's more: just adding a field in the struct (and all things equal) breaks certora

QGarchery: Will look into it in #31

QGarchery (approved):

src/interfaces/ITerms.sol
```diff
@@ -23,6 +23,7 @@ struct Offer {
...
```
adhusson, line 26: Missing corresponding OFFER_TYPEHASH update.

src/Terms.sol
```diff
@@ -23,31 +23,33 @@ contract Terms is ITerms {
...
+    mapping(address user => mapping(bytes32 termId => uint256)) public debtOf;
+    mapping(bytes32 termId => uint256) public withdrawable;
+    mapping(bytes32 termId => uint256) public totalAssets;
+    mapping(bytes32 termId => uint256) public totalShares;
+    mapping(address user => mapping(bytes32 termId => mapping(address collateralToken => uint256))) public collateralOf;
+
+    /// @dev Multiple offers can have the same nonce. This allows to implement easy and efficient batch-cancelling and
+    /// OCO (One-Cancels-the-Other) orders. Note that OCO orders work better if all offers have the same amount,
+    /// otherwise one might not be takable anymore while an other one at the same nonce is still takeable.
+    mapping(address user => mapping(uint256 nonce => uint256)) public consumed;
```
adhusson, line 36: Missing "cancel all" function. Separately, there could be a single-cancel function, as mentioned [here](https://www.notion.so/morpho-labs/V2-cancellations-nonces-223d69939e6d80bcb940c367611189aa?source=copy_link#223d69939e6d80f797cfc8bf23db64e5).

Issue #19, opened 2025-07-01
Implement liquidation after maturity


QGarchery:
See https://morpholabs.slack.com/archives/C02N7CZ088N/p1741604099248589?thread_ts=1739378276.566829&cid=C02N7CZ088N

PR #21, opened 2025-07-01
Remove timestamp check from health check

QGarchery:
Fixes checking deadline to authorize users to withdraw collateral. There could well be some borrowers with some collateral after maturity, and we should not prevent them from withdrawing it.
Note that there is a timestamp check in `_checkOffers` already


src/Terms.sol
```diff
@@ -258,11 +258,11 @@ contract Terms is ITerms {
     }
 
     function _isHealthy(Term memory term, address borrower) internal view returns (bool) {
-        if (term.maturity < block.timestamp) {
```
MathisGD, line 228: we need liquidations to be always possible after the term. So the function should return false if the maturity is passed (but true if there is no debt, as you mentioned)

QGarchery: I didn't include it because those liquidations would be performed differently. See #19

gd-colin: Right, but what about offsetting positions? You don't want to account those positions when the term is met, do you?

QGarchery: Not sure what you mean @colin-morpho, do you mean that users should not be able to withdrawCollateral after the term ?

gd-colin: Is it ok to sell an overdue bond in secondary?

QGarchery: It is prohibited at the moment, because there is a timestamp check in `_checkOffers`. This PR doesn't change that

QGarchery: Created a new issue #23 to track this open question

MathisGD (approved):

gd-colin (approved):

PR #27, opened 2025-07-03
assets as interface

### LINKED ISSUE #26: interface is inconvenient
whatever the decision on the quoting model (price or rate), how much you lend/borrow when you take assets=x is not x, it's f(x, y), where y is the price or rate input. This can be annoying, especially if f depends on the time (which is the case if the quoting is done by rate).

MathisGD:
fixes #26


MathisGD:
let's go directly with #28

PR #28, opened 2025-07-03
quote with rate

### LINKED ISSUE #26: interface is inconvenient

MathisGD:
nb: with apr, it's much simpler to take P=1 rather than F=1

### DIFF
--- a/src/Terms.sol
+++ b/src/Terms.sol
@@ -36,7 +36,7 @@ contract Terms is ITerms {
 
     /// @dev Same function used to buy and sell.
     /// @dev If one wants to make to offers without taking a position, they can batch take them and not have a position at the end.
-    function take(Term memory term, uint256 amount, address onBehalf, Offer memory offer, Signature memory sig)
+    function take(Term memory term, uint256 assets, address onBehalf, Offer memory offer, Signature memory sig)
         public
     {
         require(term.maturity >= block.timestamp, "maturity");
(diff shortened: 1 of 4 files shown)


src/Terms.sol
```diff
@@ -36,81 +36,82 @@ contract Terms is ITerms {
...
+    function take(Term memory term, uint256 assets, address onBehalf, Offer memory offer, Signature memory sig)
         public
     {
         require(term.maturity >= block.timestamp, "maturity");
         _checkSignature(offer, sig);
         _checkOffer(term, offer);
 
-        (address buyer, address seller) = offer.buy ? (offer.offering, onBehalf) : (onBehalf, offer.offering);
+        uint256 timeToMaturity = term.maturity - block.timestamp;
+        uint256 bonds = assets * (1e18 + timeToMaturity * offer.rate) / 1e18;
```
QGarchery, line 50: One side effect: price of a bond is enforced to be <= 1 (seems ok)

src/interfaces/ITerms.sol
```diff
@@ -22,7 +22,7 @@ struct Offer {
     address loanToken;
     Collateral[] collaterals;
     uint256 maturity;
-    uint256 price;
+    uint256 rate;
```
QGarchery, line 27:
```suggestion
    // The rate is expressed in percentage per second and is scaled by WAD, so `0.01e18 / uint256(365 days)` represents 1% APR.
    uint256 rate;
```

QGarchery (approved):

PR #30, opened 2025-07-04
[Certora] some invariants

### LINKED ISSUE #132: `sharePriceBelow1` invariant
of PR #30 should pass. why doesn't it passes? I get why the share price is not decreasing (roundings), but it shouldn't go above 1 right?

MathisGD:


src/Terms.sol
```diff
@@ -43,12 +43,13 @@ contract Terms is ITerms {
         external
     {
         require(term.maturity >= block.timestamp, "maturity");
-        _checkSignature(offer, sig);
+        // _checkSignature(offer, sig);
```
MathisGD: this makes the invariants vacuous apparently

MathisGD:
let's see later

src/Terms.sol
QGarchery: See [this thread](https://morpholabs.slack.com/archives/C09EAFEJUMV/p1759331567109389) for an explanation

certora/MorphoV2.spec
```diff
@@ -2,10 +2,59 @@
...
+strong invariant totalUnitsEqualsSumDebtPlusWithdrawable(bytes32 id)
+    totalUnits(id) == sumDebtOf[id] + withdrawable(id);
+
+strong invariant totalSharesEqualsSumSharesOf(bytes32 id)
+    totalShares(id) == sumSharesOf[id];
+
+// this is not true because of the roundings in shares to/from assets conversions
+// strong invariant sharePriceBelow1(bytes32 id)
+//     totalShares(id) >= totalUnits(id);
+
```
MathisGD: how can this not be True

certora/MorphoV2.conf
```diff
@@ -3,14 +3,16 @@
...
+  ],
+  "solc": "solc-0.8.28",
+  "solc_via_ir": true,
   "optimistic_loop": true,
+  "loop_iter": "2",
   "optimistic_hashing": true,
-  "solc_via_ir": true,
-  "hashing_length_bound": "320",
-  "msg": "MorphoV2" 
+  "hashing_length_bound": "640",
```
MathisGD, line 15: let's add a comment on this

certora/MorphoV2.conf
```diff
@@ -3,14 +3,16 @@
...
+  "server": "production",
```
MathisGD: it's the default right?

certora/MorphoV2.spec
```diff
@@ -1,11 +1,39 @@
...
 
-/// METHODS ///
+methods {
+    function withdrawable(bytes32 id) external returns uint256 envfree;
+    function totalUnits(bytes32 id) external returns (uint256) envfree;
+    function totalShares(bytes32 id) external returns (uint256) envfree;
 
-methods {}
+    function _.transfer(address, uint256) external => DISPATCHER(true);
+    function _.transferFrom(address, address, uint256) external => DISPATCHER(true);
```
MathisGD: probably not needed
```suggestion
```

certora/MorphoV2.spec
QGarchery: moved to #157

certora/MorphoV2.conf
QGarchery: removed

certora/MorphoV2.spec
QGarchery: removed

certora/MorphoV2.conf
QGarchery: Actually something is still unclear to me in [this thread](https://morpholabs.slack.com/archives/C09EAFEJUMV/p1759397695869969?thread_ts=1759331567.109389&cid=C09EAFEJUMV). In particular, the [corresponding run](https://prover.certora.com/output/7508195/f1cffcc838144079bd414ca86a3d86fd/?) was run against an old version of the code where the Offer struct was much smaller (8 words, structs in memory are not packed), so it's unclear to me why   the 320 limit isn't enough

QGarchery: As a side note, JSON doesn't support comments (and conf files are actually json files)

MathisGD: are we confortable moving forward anyway?

MathisGD (commented):
LGTM

certora/MorphoV2.conf
QGarchery: Yes, let's merge this PR and we can have the full explanation / comments later

QGarchery (approved):

PR #32, opened 2025-07-04
separate matching

MathisGD:
- allows future matching / offer structure that live concurrently or replace to the one in this pr, eg contract offers


src/Terms.sol
```diff
@@ -8,50 +8,39 @@ import "./interfaces/IERC20.sol";
...
+    mapping(address user => mapping(address matching => bool)) public isMatching;
 
     /// ENTRY-POINTS ///
 
     /// @dev Same function used to buy and sell.
     /// @dev If one wants to match two offers without taking a position, they can batch take them and not have a
     /// position at the end.
-    function take(Term memory term, uint256 assets, address onBehalf, Offer memory offer, Signature memory sig)
-        public
+    function take(Term memory term, uint256 assets, address onBehalf, bool buy, address matching, bytes calldata data)
```
QGarchery: The boolean `buy` is not validated, which means that the taker can switch who is the seller and who is the buyer

src/Terms.sol
```diff
@@ -196,40 +185,16 @@ contract Terms is ITerms {
         return bondSharesOf[owner][id].mulDivDown(totalBonds[id] + 1, totalShares[id] + 1);
     }
 
+    function setMatching(address matching, bool _isMatching) external {
+        isMatching[msg.sender][matching] = _isMatching;
+    }
```
QGarchery, line 219: Other options, to avoid having to do an interaction with the terms contract:
- have the matching contract be the entrypoint. It's not good because it forces the taker to be the caller, and also it creates many entrypoints potentially
- have a signature-based authorization of the matching contract

src/Matching.sol
```diff
@@ -0,0 +1,80 @@
...
+    // 1% APR.
+    uint256 rate;
+    uint256 nonce;
+}
+
+struct Signature {
+    uint8 v;
+    bytes32 r;
+    bytes32 s;
+}
```
QGarchery: We could move those definitions to `IMatching.sol`

QGarchery (commented):
Looks great !

src/Terms.sol
MathisGD: note that you would do it once and for all, so it might not be such a big deal. Though it's great if it can be batched yes, so we could have a signature based one

src/Terms.sol
MathisGD: fixed. we don't need it in the interface now

MathisGD: actually, should we have it anyway and validate? right now a bad matching could trick a taker into doing something they don't want (whereas they are not supposed to trust the matching)

src/Terms.sol
```diff
@@ -8,50 +8,38 @@ import "./interfaces/IERC20.sol";
...
-    {
+    function take(Term memory term, uint256 assets, address onBehalf, address matching, bytes calldata data) external {
         require(term.maturity >= block.timestamp, "maturity");
-        _checkSignature(offer, sig);
-        _checkOffer(term, offer);
-
-        uint256 bonds = assets * (1e18 + (term.maturity - block.timestamp) * offer.rate) / 1e18;
 
-        require((consumed[offer.offering][offer.nonce] += assets) <= offer.assets, "consumed");
+        (bool buy, address counterparty, uint256 bonds) = IMatching(matching).take(term, assets, data);
```
MathisGD: there should be the onBehalf, it helps take time compliance

MathisGD: should there be a limit price, to avoid having to trust the matching contract

src/Terms.sol
QGarchery: Tracked in #48

src/Terms.sol
QGarchery: Tracked in #48

QGarchery (approved):

src/Terms.sol
```diff
@@ -44,20 +37,21 @@ contract Terms is ITerms {
         Term memory term,
         uint256 assets,
         address onBehalf,
```
QGarchery, line 39:
```suggestion
        address taker,
```
I find that easier to understand. See the next 2 suggestions as well

src/Terms.sol
```diff
@@ -44,20 +37,21 @@ contract Terms is ITerms {
...
-        Offer memory offer,
-        Signature memory sig,
+        address matching,
+        bytes calldata data,
         address callbackAddress,
         bytes memory callbackData
```
QGarchery, line 43:
```suggestion
        address takerCallbackAddress,
        bytes memory takerCallbackData
```

src/Terms.sol
```diff
@@ -44,20 +37,21 @@ contract Terms is ITerms {
...
         require(term.maturity >= block.timestamp, "bond maturity");
-        _checkSignature(offer, sig);
-        _checkOffer(term, offer);
-
-        uint256 bonds = assets * (1e18 + (term.maturity - block.timestamp) * offer.rate) / 1e18;
 
-        require((consumed[offer.offering][offer.nonce] += assets) <= offer.assets, "consumed");
+        (
+            bool buyOffer,
+            address counterparty,
```
QGarchery, line 49:
```suggestion
            address maker,
```

QGarchery (approved):
LGTM

src/Matching.sol
```diff
@@ -0,0 +1,92 @@
...
+        external
+        returns (
+            bool buy,
+            address counterparty,
+            uint256 bonds,
+            address makerCallbackAddress,
+            bytes memory makerCallbackData
+        )
+    {
+        (Offer memory offer, Signature memory sig) = abi.decode(data, (Offer, Signature));
```
adhusson, line 40: Since the data is always decoded and always to the same type, why not include it as arguments?

src/Matching.sol
```diff
@@ -0,0 +1,92 @@
...
+
+    function _checkCanUseOffer(Offer memory offer, Signature memory sig) internal view {
+        if (sig.v == 0) {
+            require(ratified[offer.offering][abi.encode(offer)], "offer not ratified");
+        } else {
+            bytes32 hashStruct = keccak256(abi.encode(OFFER_TYPEHASH, offer));
+            bytes32 domainSeparator = keccak256(abi.encode(DOMAIN_TYPEHASH, block.chainid, address(this)));
+            bytes32 digest = keccak256(bytes.concat("\x19\x01", domainSeparator, hashStruct));
+            address signatory = ecrecover(digest, sig.v, sig.r, sig.s);
+            require(signatory != address(0) && offer.offering == signatory, "Invalid sig");
```
adhusson, line 84:
```suggestion
            require(offer.offering != address(0) && offer.offering == ecrecover(digest, sig.v, sig.r, sig.s), "Invalid sig");
```

src/Matching.sol
QGarchery: Maybe I misunderstood your question but the idea of a separate matching is that you could have a different type for `Offer` (for example adding startPrice and expiryPrice)

src/Matching.sol
QGarchery: Not that much better for me, because it's less clear that it checks that the returned value of ecrecover is not address(0) (which is a special ouput for this function)

MathisGD: agree with quentin

MathisGD:
replaced by #102

PR #37, opened 2025-07-04
offer groups: opco + oco

adhusson:
Keeps these features:

- partial fill
- parallelization
- OPCO: use same nonce, same index
- group cancel: make a 0 amount offer in same nonce, take it

Adds these features:

- OCO: use same nonce, different index
- nonce cancel: set nonce to a value higher than all nonces you want to cancel


src/Terms.sol
MathisGD, line 1: > - partial fill
> - parallelization
> - OPCO: use same nonce, same index
> - group cancel: make a 0 amount offer in same nonce, take it

we technically already have all this with the current mechanism.

>OCO: use same nonce, different index

I'm not sure about this. others makers could attack you by taking one wei 

>nonce cancel: set nonce to a value higher than all nonces you want to cancel

This is indeed an added feature, but i'm not sure it's so useful.

MathisGD: >I'm not sure about this. others makers could attack you by taking one wei

edit: it's not as bad as i first thought, the offer that got taken is still takeable (but still the attack prevents to take the other ones)

MathisGD, line 1: pseudo spec:
- offers specify a nonce and an index inside this nonce
- the nonce can only be increased by the user, manually
- "consumed" is accounted per nonce (like the current code)
- when an offer is taken, the nonce index is updated to the one of the offer
- inside a nonce, if the index is non zero, only offers that have this index can be taken

adhusson: yes the 2 added features are nonce cancel and oco, will make it clearer

> the attack prevents to take the other ones

This seems inherent to OCO + partial fill. But creating offers is free while taking for 1 wei costs gas.

adhusson: - to be taken, an offer's nonce must be at ~most~ least its account's current nonce

adhusson: > This is indeed an added feature, but i'm not sure it's so useful.

If the user takes the current timestamp in ms as the nonce, he can cancel all his past offers in one tx. So he can quickly remove all his exposure to new borrows/lends if his internal evaluation of the correct prices change. Without it either
- the cancel cost is linear in the number of outstanding offers.
- all his offers use a callback that reads a "cancel?" oracle.

Less useful than OCO which helps with differing amounts / tokens but still interesting. 

edit: due to the [mass invalidation problem](https://www.notion.so/morpho-labs/Handling-reverts-in-maker-callbacks-1bcd69939e6d80848d33f166e9f15efa?source=copy_link#1bcd69939e6d80ae9f97f119de758981), being able to cancel too many offers with a constant amount of gas could be an attack, unless routers only keep a small amount of offers per account

src/Terms.sol
```diff
@@ -196,6 +201,17 @@ contract Terms is ITerms {
         return bondSharesOf[owner][id].mulDivDown(totalBonds[id] + 1, totalShares[id] + 1);
     }
 
+    /// @dev Invalidate all offers with a nonce < newNonce.
+    function setNonce(uint256 newNonce) external {
+        require(newNonce <= type(uint248).max, "max");
```
QGarchery: We probably want to do the same as Permit2 and only allow increase of 2^16 at a time, so that we have the property that an account can always be used with all the features of nonces.
Also here some nonces are not cancellable, but it doesn't cost much to also have that property

src/Terms.sol
```diff
@@ -196,6 +201,17 @@ contract Terms is ITerms {
         return bondSharesOf[owner][id].mulDivDown(totalBonds[id] + 1, totalShares[id] + 1);
     }
 
+    /// @dev Invalidate all offers with a nonce < newNonce.
+    function setNonce(uint256 newNonce) external {
```
QGarchery, line 229: Weirdly with this implementation the order of the nonces has an importance. Making so nonces are completely independent is not completely trivial actually, but it's probably ok to not do it

src/Terms.sol
QGarchery: I think it's:
> to be taken, an offer's nonce must be at **least** its account's current nonce

src/Terms.sol
adhusson: Is there a noncancellable nonce except max nonce? (actually it is cancellable but using the group cancel approach only)

adhusson: > only allow increase of 2^16 at a time

not sure about this because 
* it breaks "current ms timestamp as nonce" trick (2^16 ms is ~1mn), but maybe it's not a good trick to begin with
* if a UI / script bug maxes the nonce of an account the worst case is that the account cannot make new offers, they can just use a new account

QGarchery: > Is there a noncancellable nonce except max nonce? (actually it is cancellable but using the group cancel approach only)

All nonces greater than `type(uint248).max` are not cancellable I think

adhusson: Indeed 248 was a mistake it's uint256 now.

adhusson:
might close because
- onchain mempool makes OCO risky against griefing (take 1 wei forces recreating the offers), but this can be fixed by a callback that forces a minmum fill. Also as @MathisGD [mentions](https://github.com/morpho-org/morpho-v2/pull/37#discussion_r2188655612) the taken offer can still be fully taken. So it just forces a choice.
- risk of mass canceling if many offers can be removed in constant gas

PR #39, opened 2025-07-07
start and expiry

### LINKED ISSUE #33: start and expiry

MathisGD:
fixes #33


src/interfaces/ITerms.sol
```diff
@@ -21,7 +21,9 @@ struct Offer {
     uint256 assets;
     address loanToken;
     Collateral[] collaterals;
-    uint256 maturity;
+    uint256 bondMaturity;
```
MathisGD: not sure about this renaming. it creates a discrepancy with bonds, but clarifies offerExpiry

adhusson (approved):

src/interfaces/ITerms.sol
QGarchery: what about renaming `maturity` into `bondMaturity` in `Term` ?

QGarchery (approved):

src/interfaces/ITerms.sol
MathisGD: at the end I rolled back to maturity

QGarchery (approved):

PR #40, opened 2025-07-07
safe transfer

### LINKED ISSUE #35: safe transfer

MathisGD:
fixes #35 

I first thought that we wouldn't need the code check anymore, because lender and borrowers are "matched" so it would be "the problem of the borrower if the token doesn't exist yet", but actually it could happen with collaterals too. We could just do the check for collateral, but it's probably overkill, it's cheap in gas anyway


adhusson (approved):

test/SafeTransferLib.sol
```diff
@@ -0,0 +1,103 @@
...
+
+/// @dev Token returning false.
+contract ERC20WithBooleanAlwaysFalse {
+    function transfer(address to, uint256 value) external returns (bool res) {}
+    function transferFrom(address from, address to, uint256 value) external returns (bool res) {}
+    function approve(address, uint256) external pure returns (bool res) {}
+}
+
+/// @dev Normal token.
+contract ERC20Normal {
```
QGarchery: Not so normal ^^

test/SafeTransferLib.sol
QGarchery, line 1: There are some "common implementation" of tokens in `certora/dispatch`. Maybe worth reusing that, and put them in `test/helpers` instead ?

src/Terms.sol
```diff
@@ -3,6 +3,7 @@
 pragma solidity 0.8.28;
 
 import "./libraries/UtilsLib.sol";
+import "./libraries/SafeTransferLib.sol";
 import "./libraries/MathLib.sol";
 import "./interfaces/IERC20.sol";
```
```suggestion
```

test/SafeTransferLib.sol
QGarchery, line 1: Rename this file into `SafeTransferLibTest`

QGarchery (approved):

test/SafeTransferLib.sol
MathisGD: I like the fact that the behavior is very clear like that. it's not so much LOC, and it will not be reused elsewhere anyway. wdyt?

QGarchery (approved):

PR #45, opened 2025-07-09
`bool asBorrower`: include withdraw vs. borrow or repay vs. lend information in offer

### LINKED ISSUE #42: issue with offer structure
buy=false makes you withdraw and borrow at the same time, as the taker chooses the bond you can end up borrowing while you wanted to withdraw on a bond in which you are already in. Even worse for repay

adhusson:
fixes #42 

A buy offer could be intended as a -debt offer (repay) but end up in a different term where there is no debt, and it's taken as a +bond (lend) offer. 

Or: a sell offer could be intended as a -bond (withdraw) offer but end up in a different term where there is are bonds, and it's taken a +debt (borrow) offer. 

With `asBorrower`, on a buy offer you say you strictly want to repay if `true`, or strictly lend if `false`. 

On a sell offer you say you strictly want to borrow if `true`, or strictly withdraw if `false`.

### DIFF
--- a/src/Matching.sol
+++ b/src/Matching.sol
@@ -23,13 +23,18 @@ contract Matching is IMatching {
 
     function take(Term memory term, uint256 assets, bytes calldata data)
         external
-        returns (bool buy, address counterparty, uint256 bonds)
+        returns (bool buy, bool asBorrower, address counterparty, uint256 bonds)
     {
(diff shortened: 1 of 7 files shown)


MathisGD:
pending the reconsideration of offers

MathisGD:
replaced by #131

PR #49, opened 2025-07-09
Refactor collateral inclusion

QGarchery:
To avoid having to copy arrays, it should not change the behavior


adhusson (commented):
Assignment from memory references to memory references doesn't copy data, see [this part](https://docs.soliditylang.org/en/v0.8.28/types.html#reference-types:~:text=Assignments%20from%20memory%20to%20memory%20only%20create%20references.%20This%20means%20that%20changes%20to%20one%20memory%20variable%20are%20also%20visible%20in%20all%20other%20memory%20variables%20that%20refer%20to%20the%20same%20data.) of the Solidity docs.

QGarchery:
Awesome, closing this PR then !

PR #41, opened 2025-07-07
callbacks

MathisGD:
fixes #14

- when there is a buy callback, should the funds be pulled from the offering of from the callback?

PR #56, opened 2025-07-09
matching+callback refactoring

### LINKED ISSUE #48: Remove need for taker to trust matching
# Have a `buy` argument on `take`

actually, should we have it anyway and validate? right now a bad matching could trick a taker into doing something they don't want (whereas they are not supposed to trust the matching)

_Originally posted by @MathisGD in https://github.com/morpho-org/v2/pull/32#discussion_r2190573130_

# Have a limit price argument on `take`

should there be a limit price, to avoid having to trust the matching contract

_Originally posted by @MathisGD in https://github.com/morpho-org/v2/pull/32#discussion_r2192724371_

### LINKED ISSUE #50: everything should be input
bonds, counterparty should be an input of take, and matching should validate them, but not return anything

MathisGD:
does multiple things:
- merge matching and callbacks
- change the interface so that it's symmetric and simpler (in the Terms code)
- fixes #48 
- fixes #50 
- replaces #41 (#14)

todo:
- redo tests of #41


src/Matching.sol
```diff
@@ -21,15 +40,17 @@ contract Matching is IMatching {
...
+        require(offer.buy == buy, "not a buy");
+        require(offer.offering == user, "not offering");
+        consumed[user][offer.nonce] += assets;
+        require(consumed[user][offer.nonce] <= offer.assets, "consumed");
+        require(bonds == assets * (1e18 + (term.maturity - block.timestamp) * offer.rate) / 1e18, "bonds");
         _checkSignature(offer, sig);
         _checkOffer(term, offer);
-        require((consumed[offer.offering][offer.nonce] += assets) <= offer.assets, "consumed");
-        return (offer.buy, offer.offering, assets * (1e18 + (term.maturity - block.timestamp) * offer.rate) / 1e18);
     }
```
MathisGD, line 58: the function will have to call callbacks here. We will want to have the same interface as the hook, which could be misleading..

src/Matching.sol
```diff
@@ -21,15 +40,17 @@ contract Matching is IMatching {
...
+    function hook(Term memory term, uint256 assets, uint256 bonds, bool buy, address user, bytes calldata data)
         external
-        returns (bool buy, address counterparty, uint256 bonds)
     {
         (Offer memory offer, Signature memory sig) = abi.decode(data, (Offer, Signature));
+        require(offer.buy == buy, "not a buy");
+        require(offer.offering == user, "not offering");
+        consumed[user][offer.nonce] += assets;
+        require(consumed[user][offer.nonce] <= offer.assets, "consumed");
+        require(bonds == assets * (1e18 + (term.maturity - block.timestamp) * offer.rate) / 1e18, "bonds");
```
MathisGD, line 55: for buy
```suggestion
        require(bonds >= assets * (1e18 + (term.maturity - block.timestamp) * offer.rate) / 1e18, "bonds");
```
and for sell 
```suggestion
        require(bonds <= assets * (1e18 + (term.maturity - block.timestamp) * offer.rate) / 1e18, "bonds");
```
this would allow MATCH of crossed offers. unclear though how MEV would profit

src/Terms.sol
```diff
@@ -26,44 +26,51 @@ contract Terms is ITerms {
...
 
-    /// @dev Same function used to buy and sell.
     /// @dev If one wants to match two offers without taking a position, they can batch take them and not have a
     /// position at the end.
-    function take(Term memory term, uint256 assets, address onBehalf, address matching, bytes calldata data) external {
+    /// @dev The hook of the maker is validated.
+    function take(
+        Term memory term,
+        uint256 assets,
+        uint256 bonds,
```
MathisGD, line 39: if your transaction doesn't get included in the next block, bonds will not be what you expect and it will revert. Should it be a limit price?

gd-colin: wait why would the bond not be what's expected? because the next block is closer to the term, is that it?

MathisGD: exactly

MathisGD:
see #32

PR #57, opened 2025-07-10
include hook in offer

adhusson:
prevents an offer from being used with an unintended hook

### DIFF
--- a/src/Matching.sol
+++ b/src/Matching.sol
@@ -50,7 +50,6 @@ contract Matching is IHook {
         consumed[user][offer.nonce] += assets;
         require(consumed[user][offer.nonce] <= offer.assets, "consumed");
         require(bonds == assets * (1e18 + (term.maturity - block.timestamp) * offer.rate) / 1e18, "bonds");
-        require(offer.hook == address(this), "hook");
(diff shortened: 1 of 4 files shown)


src/Matching.sol
```diff
@@ -15,6 +15,7 @@ struct Offer {
     // 1% APR.
     uint256 rate;
     uint256 nonce;
+    address hook;
```
MathisGD, line 18: the naming will clash with the actual hook we will want to add, see my comment in #56

adhusson: hook/callback?

src/Matching.sol
```diff
@@ -49,6 +50,7 @@ contract Matching is IHook {
         consumed[user][offer.nonce] += assets;
         require(consumed[user][offer.nonce] <= offer.assets, "consumed");
         require(bonds == assets * (1e18 + (term.maturity - block.timestamp) * offer.rate) / 1e18, "bonds");
+        require(offer.hook == address(this), "hook");
```
adhusson: not needed, the content of the offer is already checked by the signature

adhusson:
```suggestion
```

PR #59, opened 2025-07-11
multiple hooks

adhusson:
+ extra callback to be able to test multiple hooks


adhusson:
closing because we are going in the direction of treating matching & hook separately (which may also remove the need for a hooks array)

PR #60, opened 2025-07-12
simplify liquidate

adhusson:
easier to read, gas optimisations, single for loop

### DIFF
--- a/src/Terms.sol
+++ b/src/Terms.sol
@@ -117,11 +117,6 @@ contract Terms is ITerms {
         IERC20(collateral).transfer(msg.sender, assets);
     }
 
-    struct Vars {
-        uint256 maxDebt;
-        uint256 repayableDebt;
-    }
-
     /// @notice Execute the given collection of `seizures` on the given `term` of the given `borrower`.
     /// @dev On each seizure either `repaidAmounts` or `seizedAssets` should be equal to zero.
     /// @param term The term of the bond.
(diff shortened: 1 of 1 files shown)


src/Terms.sol
```diff
@@ -136,55 +131,55 @@ contract Terms is ITerms {
     {
         require(seizures.length == term.collaterals.length, "should have all collats");
 
-        Vars memory vars;
```
gd-colin, line 139: I think this was intentional for testing to disable compilation with IR

gd-colin:
The PR is a good idea however, we should consider other optimization points.
For instance, the reason why there original code has two for loops is because an alternative design is for the liquidator to pass an array that targets only the assets he needs, to avoid long loops. We should compare which implementation is actually more efficient.

src/Terms.sol
adhusson: other changes like [this one](https://github.com/morpho-org/morpho-v2/pull/60/files#diff-2f530276825e96ff84d75b14bb8f85575e2ba72ee56a29129525a672faada763R145-R150) maintain compatibility with the non-ir pipeline

src/Terms.sol
```diff
@@ -117,11 +117,6 @@ contract Terms is ITerms {
         IERC20(collateral).transfer(msg.sender, assets);
     }
 
-    struct Vars {
-        uint256 maxDebt;
-        uint256 repayableDebt;
-    }
-
     /// @notice Execute the given collection of `seizures` on the given `term` of the given `borrower`.
     /// @dev On each seizure either `repaidAmounts` or `seizedAssets` should be equal to zero.
```
peyha, line 121:
```suggestion
    /// @dev On each seizure either `repaidBonds` or `seizedAssets` should be equal to zero.
```

peyha (commented):
I've mixed feelings about this PR
On the one hand, using a simple loop simplifies the logic but on the other hand as @colin-morpho mentionned, it's probably not gas efficient when liquidating a small subset of the bonds collateral (up to some point because it still saves an oracle call).
Note that if we end up with a dutch auction liquidation, we will have to keep two loops (because either the close factor or the liquidation incentive will be computed with `maxDebt` and `debtOf[borrower][id]`)

src/Terms.sol
gd-colin: For example, have a look at an older gas measurement in https://colab.research.google.com/drive/1L-pvtrh_EzYlWTicf_c-9oNz8hTHQ5p2?authuser=1, for each case, the more gas efficient curves were measured with an implementation that has two separate loops (i.e. one for collat pricing and one for seizures array), my point is, if seizing a single assey we might want to look into this optimization option.

src/Terms.sol
```diff
@@ -136,55 +131,55 @@ contract Terms is ITerms {
     {
         require(seizures.length == term.collaterals.length, "should have all collats");
```
MathisGD, line 132: why did we end up checking this already @colin-morpho @QGarchery? With the current state of the code it doesn't really make sense.

adhusson: Ok I understand better, if we loop once through collaterals and once through seizures it’s probably better

gd-colin: I think it's related to this comment https://github.com/morpho-org/morpho-v2/pull/3#discussion_r2003670326.

gd-colin: At that time the context was simplicity first and it was a tradeoff to ensure that seizures are unique. This is actually important as it can add a complexity overhead, even if you require that the assets of seizures is sorted, then it's still O(|seizures|) to check it's sorted and that seizures are unique.

src/Terms.sol
gd-colin: I think this is the https://github.com/morpho-org/morpho-v2/commit/fe1115915f0bcdb3040b5bc76209923d03f623f3#diff-2f530276825e96ff84d75b14bb8f85575e2ba72ee56a29129525a672faada763L136 where implementations changed

adhusson:
I thought it would be better to loop only once, and skip the seizure parts for a very low cost. But I do see +6% difference (+20k gas over 300k for 10 collaterals / 1 seizure) between a) looping once, `|seizures| == |collaterals|` vs.  b) looping twice, `|seizures| == 1`.

I don't understand why, but it seems to be a sum of small differences. For instance checking 9 times `seizure.repaidBonds * seizure.seizedAssets == 0` costs 7k gas. Sidenote: it only costs 4k gas to check 9 times `seizure.repaidBonds == 0 || seizure.seizedAssets == 0`.

It would take some digging to understand better so I'm closing this for PR for now and pushing another one (https://github.com/morpho-org/morpho-v2/pull/62) that does reduce the size of `seizures` + other changes. 

Note that if we keep `|seizures| == |collaterals|`, gas difference between 1 loop and 2 loops are very small, so it makes sense to have a small seizures array.

Issue #8, opened 2025-05-21
Perform transfers after updating the state.


gd-colin:
It seems somewhat dangerous to do the transfers before updating the whole state. For example, if one token allows to call an arbitrary address (see ERC-777 notably), then it can execute arbitrary code and reenter at this point. Note that it is especially dangerous because the account is definitely liquidatable still: `debtOf[borrower][id]` is decreased only later, and `collateralOf[borrower][id]` is decreased before

_Originally posted by @QGarchery in https://github.com/morpho-org/v2/pull/3#discussion_r2054013855_

MathisGD:
one solution would be to implement: a liquidatable account is liquidatable during all the transaction (kept in a transient storage slot) to implement the liquidation of multiple collaterals, instead of having a loop in the protocol like that. It would increase the gas cost of liquidating multiple collaterals though

MathisGD:
actually, maybe not if we keep everything in transient storage

MathisGD:
done in #62

PR #62, opened 2025-07-21
save gas & simplify in `liquidate`

adhusson:
- small seizures array, saves about 20k gas with 10 collaterals and 1 seizure
- transfer collaterals after updating internal accounting to be reentrancy safe
- single debt update
- ~fix totalRepaid > debt~ there is still the issue that due to roundings, it may not be possible to both 1) completely repay a debt and 2) seize all collateral

### DIFF
--- a/src/Terms.sol
+++ b/src/Terms.sol
@@ -139,11 +139,12 @@ contract Terms is ITerms {
     {
         Vars memory vars;
         bytes32 id = _id(term);
+        uint256[] memory prices = new uint256[](term.collaterals.length);
 
         for (uint256 i = 0; i < term.collaterals.length; i++) {
-            uint256 price = IOracle(term.collaterals[i].oracle).price();
+            prices[i] = IOracle(term.collaterals[i].oracle).price();
             uint256 collateralQuoted =
-                collateralOf[borrower][id][term.collaterals[i].token].mulDivDown(price, ORACLE_PRICE_SCALE);
+                collateralOf[borrower][id][term.collaterals[i].token].mulDivDown(prices[i], ORACLE_PRICE_SCALE);
             vars.maxDebt += collateralQuoted.mulDivDown(term.collaterals[i].lltv, 1e18);
             vars.repayableDebt += collateralQuoted.mulDivUp(1e18, LIQUIDATION_INCENTIVE_FACTOR);
         }
@@ -158,15 +159,12 @@ contract Terms is ITerms {
             collateralIndex = seizure.collateralIndex;
             require(UtilsLib.exactlyOneZero(seizure.repaidBonds, seizure.seizedAssets), "INCONSISTENT_INPUT");
 
-            uint256 collateralPrice = IOracle(term.collaterals[seizure.collateralIndex].oracle).price();
-
             if (seizure.seizedAssets > 0) {
-                seizure.repaidBonds = seizure.seizedAssets.mulDivUp(collateralPrice, ORACLE_PRICE_SCALE).mulDivUp(
-                    1e18, LIQUIDATION_INCENTIVE_FACTOR
-                );
+                seizure.repaidBonds = seizure.seizedAssets.mulDivUp(prices[seizure.collateralIndex], ORACLE_PRICE_SCALE)
+                    .mulDivUp(1e18, LIQUIDATION_INCENTIVE_FACTOR);
             } else {
                 seizure.seizedAssets = seizure.repaidBonds.mulDivDown(LIQUIDATION_INCENTIVE_FACTOR, 1e18).mulDivDown(
-                    ORACLE_PRICE_SCALE, collateralPrice
+                    ORACLE_PRICE_SCALE, prices[seizure.collateralIndex]
                 );
             }
 
--- a/src/Terms.sol
+++ b/src/Terms.sol
@@ -207,7 +207,6 @@ contract Terms is ITerms {
         // Realize bad debt
         uint256 badDebt;
 
-        if (totalRepaid > originalDebt) totalRepaid = originalDebt;
         if (repayableDebt < originalDebt) {
             // Because roundings are not aligned the effective bad debt is either the remaining debt or the original
             // debt minus the theoretical repayable debt.
--- a/src/Terms.sol
+++ b/src/Terms.sol
@@ -119,12 +119,6 @@ contract Terms is ITerms {
         SafeTransferLib.safeTransfer(collateral, msg.sender, assets);
     }
 
-    struct Vars {
-        uint256 maxDebt;
-        uint256 repayableDebt;
-        uint256 totalRepaid;
-    }
-
     /// @notice Execute the given collection of `seizures` on the given `term` of the given `borrower`.
     /// @dev On each seizure either `repaidBonds` or `seizedAssets` should be equal to zero.
     /// @param term The term of the bond.
@@ -137,20 +131,22 @@ contract Terms is ITerms {
         external
         returns (Seizure[] memory)
     {
-        Vars memory vars;
+        uint256 repayableDebt;
+        uint256 maxDebt;
         bytes32 id = _id(term);
         uint256[] memory prices = new uint256[](term.collaterals.length);
 
         for (uint256 i = 0; i < term.collaterals.length; i++) {
             prices[i] = IOracle(term.collaterals[i].oracle).price();
-            uint256 collateralQuoted =
-                collateralOf[borrower][id][term.collaterals[i].token].mulDivDown(prices[i], ORACLE_PRICE_SCALE);
-            vars.maxDebt += collateralQuoted.mulDivDown(term.collaterals[i].lltv, 1e18);
-            vars.repayableDebt += collateralQuoted.mulDivUp(1e18, LIQUIDATION_INCENTIVE_FACTOR);
+            {
+                address token = term.collaterals[i].token;
+                uint256 collateralQuoted = collateralOf[borrower][id][token].mulDivDown(prices[i], ORACLE_PRICE_SCALE);
+                maxDebt += collateralQuoted.mulDivDown(term.collaterals[i].lltv, 1e18);
+                repayableDebt += collateralQuoted.mulDivUp(1e18, LIQUIDATION_INCENTIVE_FACTOR);
+            }
         }
-        uint256 originalDebt = debtOf[borrower][id];
-        require(originalDebt > vars.maxDebt, "position is healthy");
 
+        uint256 totalRepaid;
         uint256 collateralIndex = type(uint256).max;
         Seizure memory seizure;
         for (uint256 i = seizures.length; i > 0; i--) {
@@ -168,31 +164,39 @@ contract Terms is ITerms {
                 );
             }
 
-            vars.totalRepaid += seizure.repaidBonds;
-            collateralOf[borrower][id][term.collaterals[seizure.collateralIndex].token] -= seizure.seizedAssets;
+            totalRepaid += seizure.repaidBonds;
+            address token = term.collaterals[seizure.collateralIndex].token;
+            collateralOf[borrower][id][token] -= seizure.seizedAssets;
         }
 
         // Realize bad debt
-        uint256 badDebt;
-        if (vars.repayableDebt < originalDebt) {
-            // Because roundings are not aligned the effective bad debt is either the remaining debt or the original
-            // debt minus the theoretical repayable debt.
-            badDebt = UtilsLib.min(originalDebt - vars.totalRepaid, originalDebt - vars.repayableDebt);
-            totalBonds[id] -= badDebt;
+        {
+            uint256 originalDebt = debtOf[borrower][id];
+            require(originalDebt > maxDebt, "position is healthy");
+
+            uint256 badDebt;
+
+            if (repayableDebt < originalDebt) {
+                // Because roundings are not aligned the effective bad debt is either the remaining debt or the original
+                // debt minus the theoretical repayable debt.
+                badDebt = UtilsLib.min(originalDebt - totalRepaid, originalDebt - repayableDebt);
+                totalBonds[id] -= badDebt;
+            }
+
+            withdrawable[id] += totalRepaid;
+            debtOf[borrower][id] = originalDebt - totalRepaid - badDebt;
         }
 
-        withdrawable[id] += vars.totalRepaid;
-        debtOf[borrower][id] = originalDebt - vars.totalRepaid - badDebt;
-
         for (uint256 i = 0; i < seizures.length; i++) {
+            seizure = seizures[i];
             SafeTransferLib.safeTransfer(
-                term.collaterals[seizures[i].collateralIndex].token, msg.sender, seizures[i].seizedAssets
+                term.collaterals[seizure.collateralIndex].token, msg.sender, seizure.seizedAssets
             );
         }
 
         if (data.length > 0) IMorphoLiquidationCallback(msg.sender).onLiquidate(seizures, borrower, msg.sender, data);
 
-        SafeTransferLib.safeTransferFrom(term.loanToken, msg.sender, address(this), vars.totalRepaid);
+        SafeTransferLib.safeTransferFrom(term.loanToken, msg.sender, address(this), totalRepaid);
 
         return seizures;
     }
--- a/src/Terms.sol
+++ b/src/Terms.sol
@@ -122,13 +122,14 @@ contract Terms is ITerms {
     struct Vars {
         uint256 maxDebt;
         uint256 repayableDebt;
+        uint256 totalRepaid;
     }
 
     /// @notice Execute the given collection of `seizures` on the given `term` of the given `borrower`.
-    /// @dev On each seizure either `repaidAmounts` or `seizedAssets` should be equal to zero.
+    /// @dev On each seizure either `repaidBonds` or `seizedAssets` should be equal to zero.
     /// @param term The term of the bond.
     /// @param seizures An array of amounts of debt to repay or assets to seize with the index of the collateral in the
-    /// term's collateral assets.
+    /// term's collateral assets. Must be sorted by collateral index in increasing order.
     /// @param borrower The debtor of the loan.
     /// @param data Arbitrary data to pass to the callback. Pass empty data if not needed.
     /// @return A collection of the actual amounts of debt repaid or asset seized with the collateral index.
@@ -136,8 +137,6 @@ contract Terms is ITerms {
         external
         returns (Seizure[] memory)
     {
-        require(seizures.length == term.collaterals.length, "should have all collats");
-
         Vars memory vars;
         bytes32 id = _id(term);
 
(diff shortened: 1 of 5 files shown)


gd-colin:
Fixes #8, #9

src/Terms.sol
```diff
@@ -148,50 +147,54 @@ contract Terms is ITerms {
...
-                        .mulDivUp(1e18, LIQUIDATION_INCENTIVE_FACTOR);
-                } else {
-                    seizures[i].seizedAssets = seizures[i].repaidBonds.mulDivDown(LIQUIDATION_INCENTIVE_FACTOR, 1e18)
-                        .mulDivDown(ORACLE_PRICE_SCALE, collateralPrice);
-                }
+        uint256 collateralIndex = type(uint256).max;
+        Seizure memory seizure;
+        for (uint256 i = seizures.length; i > 0; i--) {
+            seizure = seizures[i - 1];
+            require(collateralIndex > seizure.collateralIndex, "not in order");
```
gd-colin: This checks that the seizures is sorted and that seized assets are seized at most once right? But where is it checked that the collateral array is sorted and that collaterals are unique?

gd-colin:
Also this may be a good opportunity to tackle #9, what do you think?

src/Terms.sol
gd-colin: Which is needed for this check to make sense.

src/Terms.sol
```diff
@@ -148,50 +147,54 @@ contract Terms is ITerms {
             vars.maxDebt += collateralQuoted.mulDivDown(term.collaterals[i].lltv, 1e18);
             vars.repayableDebt += collateralQuoted.mulDivUp(1e18, LIQUIDATION_INCENTIVE_FACTOR);
```
gd-colin: Let's check here that collaterals array is sorted and that collats are unique as well.

peyha: is it necessary as it's supposed to be done during take ? from [this comment](https://github.com/morpho-org/morpho-v2/blob/09c2ae49a81b9132fbb81b1ec5820bafd0917de8/src/Terms.sol#L219), it should be done in the matching contract so the liquidation would either revert or not liquidate anything because debt would be 0

gd-colin: Right good point @peyha, maybe we should add a similar comment in the liquidation function as well to clarify this instead!

src/Terms.sol
```diff
@@ -148,50 +147,54 @@ contract Terms is ITerms {
...
-                if (seizures[i].seizedAssets > 0) {
-                    seizures[i].repaidBonds = seizures[i].seizedAssets.mulDivUp(collateralPrice, ORACLE_PRICE_SCALE)
-                        .mulDivUp(1e18, LIQUIDATION_INCENTIVE_FACTOR);
-                } else {
-                    seizures[i].seizedAssets = seizures[i].repaidBonds.mulDivDown(LIQUIDATION_INCENTIVE_FACTOR, 1e18)
-                        .mulDivDown(ORACLE_PRICE_SCALE, collateralPrice);
-                }
+        uint256 collateralIndex = type(uint256).max;
+        Seizure memory seizure;
+        for (uint256 i = seizures.length; i > 0; i--) {
```
peyha: why are you iterating in this order ?

src/Terms.sol
adhusson: Since the term is checked by hashing (`id = _id(term)`) we know its collateral list has not been altered. 

If the check is added, I think it should be in _checkOffer (which currently has a [comment](https://github.com/morpho-org/morpho-v2/blob/09c2ae49a81b9132fbb81b1ec5820bafd0917de8/src/Terms.sol#L218) about not checking), otherwise it would be possible to create unliquidatable terms.

src/Terms.sol
```diff
@@ -148,50 +147,54 @@ contract Terms is ITerms {
...
             // debt minus the theoretical repayable debt.
-            uint256 badDebt = UtilsLib.min(debtOf[borrower][id], originalDebt - vars.repayableDebt);
-            debtOf[borrower][id] -= badDebt;
+            badDebt = UtilsLib.min(originalDebt - vars.totalRepaid, originalDebt - vars.repayableDebt);
             totalBonds[id] -= badDebt;
         }
 
-        withdrawable[id] += totalRepaid;
+        withdrawable[id] += vars.totalRepaid;
+        debtOf[borrower][id] = originalDebt - vars.totalRepaid - badDebt;
```
peyha: Isn't it more gas-optimized to only consider badDebt when realizing bad debt  as before? I would guess bad debt only happens in 1% of liquidations so it would save a tiny bit of gas 99% of the time

src/Terms.sol
```diff
@@ -122,22 +122,21 @@ contract Terms is ITerms {
     struct Vars {
```
peyha:
```suggestion
    struct LiquidationVars {
```

adhusson (commented):
updated to query oracles once per collateral. it increases the gas cost by ~300/collateral.

src/Terms.sol
peyha: note that it is more gas optimized to remove liquidationVars and to use `maxDebt`, `repayableDebt` and `totalRepaid` as local variable on the stack in `liquidate`  although the struct improves clarity

gd-colin (approved):

src/Terms.sol
```diff
@@ -122,22 +122,21 @@ contract Terms is ITerms {
...
+        uint256 totalRepaid;
     }
 
     /// @notice Execute the given collection of `seizures` on the given `term` of the given `borrower`.
-    /// @dev On each seizure either `repaidAmounts` or `seizedAssets` should be equal to zero.
+    /// @dev On each seizure either `repaidBonds` or `seizedAssets` should be equal to zero.
     /// @param term The term of the bond.
     /// @param seizures An array of amounts of debt to repay or assets to seize with the index of the collateral in the
-    /// term's collateral assets.
+    /// term's collateral assets. Must be sorted by collateral index in increasing order.
```
peyha: just wondering, why do we want to keep these indexes sorted exactly ? the liquidator can still liquidate in any order by liquidating one asset at a time consecutively even though bad debt is not realized the same way

peyha: also, removing the collateral index ordering check is a (small) gas improvement

src/Terms.sol
adhusson: You can't start at `collateralIndex = 0` unless you use an int and starts at -1. Assuming we want to stay with uints, we must start from the top. So we can either require that seizures are sorted by decreasing collateral index, or iterate from the end of the array. 

It's simple to iterate from the end of the array and means the caller doesn't have to think about reversing the order, which is slightly less natural than sorting by increasing collateral index.

src/Terms.sol
adhusson: In a scenario without bad debt, it costs ~100 more to do it the previous way because `debtOf` is read one more time.

src/Terms.sol
adhusson: agree, but the struct was here for compatibility with the classic (no viar ir) compilation pipeline, which is used by `forge coverage`.

But now I found a way to remove it, at the cost of checking the health after going through the seizures instead of before (0a211688700efc441dbcb3b5e1a7ebce05ef94f9).

adhusson: I'm not sure it's better actually, it costs slightly more gas to use a struct but the code is cleaner.

We could also give up on forge coverage.

src/Terms.sol
adhusson: If indices are not sorted then the same asset could be seized multiple times, which means the oracle is called multiple times.

gd-colin: Well, not anymore actually now that the prices are checked once 🤔

adhusson: done here 78d2c106fbdda60caeb97d7cba3425d6bba30fb4

peyha (approved):
LGTM

src/Terms.sol
adhusson: I'm not fond of it after all, it simplifies reasoning to be able to say that every seizure uniquely maps a collateral and that there is no "seizure slicing". For instance you don't have to reason about how slicing affects roundings (although it seems like it's not a problem here) inside of a liquidation.

gd-colin: I actually agree with you Adrien. I'm not sure that the gas improvement justifies dismissing this requirement.

peyha: I feel like rounding is not an issue if we round in the correct direction. The current version is quite clean so I prefer it but it's not a strong preference

peyha: Also the liquidator can liquidate the same asset consecutively (even if it's gas costly because oracles will be called multiple times) so it's not a complete fix

gd-colin: But beyond being clean, do we have a good reason to allow, say, an asset to be seized in multiple steps in a single liquidation?

Case we allow it:
  pros -> it's cheaper in gas, this is already possible in multiple transactions so why forbid it here;
  cons -> seizures can be un-predictable, which may open a vulnerability vector difficult to anticipate, may lead to anti-patterns for liquidations (makes it harder to reason for the user), does it affect recovery close factor?, it can be tricky...
  
Case we forbid it:
  pros -> we can predict how seizures will go through (i.e. they are predictable), simple from user perspective,
  cons -> its a little bit more expensive gas-wise.

gd-colin: >  (even if it's gas costly because oracles will be called multiple times)


no they won't oracle calls are memoized

peyha: > do we have a good reason to allow say an asset

my argument is that you can still achieve a similar trick using several liquidation calls even with ordered indexes
Suppose I have a liquidatable position collateralized by 10 ETH and 1 WBTC. Then I can do two liquidation calls consecutively
- Liquidate 1 ETH
- Liquidate 1 ETH (and possibly some WBTC)
Instead of liquidating 2 ETH and some WBTC at once
It's not exactly equivalent in terms of rounding because `collateralOf`, `debtOf` are updated but it could still lead to issues

peyha: I think I'm starting to change my mind, sorting indices would possibly fix many possible vulnerabilities and it's almost zero cost (my example is not as equivalent to using a single call with duplicate collaterals as I expected)

gd-colin: But I wonder, if there's a vulnerability here, why is there none when we seize a single asset in multiple liquidation transactions rather that in a single liquidation? 🤔

peyha: I'm not sure there is none but more checks are done. The health check is done during each consecutive calls, the bad debt and the collateral/debt accounting is updated while it's not the case with duplicate indices

src/Terms.sol
```diff
@@ -136,58 +131,66 @@ contract Terms is ITerms {
...
-        uint256 originalDebt = debtOf[borrower][id];
-        debtOf[borrower][id] -= totalRepaid;
+            totalRepaid += seizure.repaidBonds;
+            address collateralToken = term.collaterals[seizure.collateralIndex].token;
+            collateralOf[borrower][id][collateralToken] -= seizure.seizedAssets;
+        }
 
         // Realize bad debt
-        if (vars.repayableDebt < originalDebt) {
+        uint256 badDebt;
```
MathisGD, line 208: would be great to realize at the beginning, after the health check!

src/Terms.sol
```diff
@@ -136,58 +131,66 @@ contract Terms is ITerms {
...
     {
-        require(seizures.length == term.collaterals.length, "should have all collats");
-
-        Vars memory vars;
+        uint256 repayableDebt;
+        uint256 maxDebt;
         bytes32 id = _id(term);
+        uint256[] memory prices = new uint256[](term.collaterals.length);
 
         for (uint256 i = 0; i < term.collaterals.length; i++) {
```
MathisGD, line 173: impossible to re-use the health check? (maybe yes as we need the max repayable debt and the collateral prices)

src/Terms.sol
```diff
@@ -136,58 +131,66 @@ contract Terms is ITerms {
...
+            totalRepaid += seizure.repaidBonds;
+            address collateralToken = term.collaterals[seizure.collateralIndex].token;
+            collateralOf[borrower][id][collateralToken] -= seizure.seizedAssets;
+        }
 
         // Realize bad debt
-        if (vars.repayableDebt < originalDebt) {
+        uint256 badDebt;
+
+        if (totalRepaid > originalDebt) totalRepaid = originalDebt;
```
MathisGD: why is that possible?

adhusson: see this https://github.com/morpho-org/morpho-v2/pull/91/files

src/Terms.sol
adhusson: > would be great to realize at the beginning, after the health check!

realizing before reducing by repayment can be nice but having 2 state updates separated would be worse for readability/understanding no?

what do you think about https://github.com/morpho-org/morpho-v2/pull/95

src/Terms.sol
```diff
@@ -136,58 +131,66 @@ contract Terms is ITerms {
...
 
         withdrawable[id] += totalRepaid;
+        debtOf[borrower][id] = originalDebt - totalRepaid - badDebt;
+
+        for (uint256 i = 0; i < seizures.length; i++) {
+            seizure = seizures[i];
+            SafeTransferLib.safeTransfer(
+                term.collaterals[seizure.collateralIndex].token, msg.sender, seizure.seizedAssets
+            );
+        }
```
adhusson, line 225: the cost of being reentrancy safe by doing this is about 300 gas per seizure

src/Terms.sol
adhusson: > impossible to re-use the health check? (maybe yes as we need the max repayable debt and the collateral prices)

Best I could do is https://github.com/morpho-org/morpho-v2/pull/97

- oracle prices no longer reused
- "`isHealthy`" does extra work now

src/Terms.sol
```diff
@@ -160,58 +155,66 @@ contract Terms is ITerms {
...
 
-        for (uint256 i = 0; i < term.collaterals.length; i++) {
-            if (seizures[i].repaidBonds + seizures[i].seizedAssets > 0) {
-                require(
-                    UtilsLib.exactlyOneZero(seizures[i].repaidBonds, seizures[i].seizedAssets), "INCONSISTENT_INPUT"
+        uint256 totalRepaid;
+        Seizure memory seizure;
+
+        for (uint256 i = 0; i < seizures.length; i++) {
+            seizure = seizures[i];
```
QGarchery: Why do this ? It's not really an optimization right ?

src/Terms.sol
```diff
@@ -160,58 +155,66 @@ contract Terms is ITerms {
...
-        if (vars.repayableDebt < originalDebt) {
+        uint256 badDebt;
+
+        if (totalRepaid > originalDebt) totalRepaid = originalDebt;
+        if (repayableDebt < originalDebt) {
             // Because roundings are not aligned the effective bad debt is either the remaining debt or the original
             // debt minus the theoretical repayable debt.
-            uint256 badDebt = UtilsLib.min(debtOf[borrower][id], originalDebt - vars.repayableDebt);
-            debtOf[borrower][id] -= badDebt;
+            badDebt = UtilsLib.min(originalDebt - totalRepaid, originalDebt - repayableDebt);
```
QGarchery, line 213: I found the previous code to be easier to understand here: it was clear that the min was here to cap the bad debt to the current debt of the user (`debtOf[borrower][id]`), so I would prefer if the update was done twice `debtOf[borrower][id] -= totalRepaid` and `debtOf[borrower][id] -= badDebt` even if it is slightly less optimized

src/Terms.sol
QGarchery: In the test it seems like it's just borrowing max, but I don't really understand what's happening and why it creates a totalRepaid greater than originalDebt. Is it a rounding issue ?

adhusson: yes, it's due to roundings

src/Terms.sol
adhusson: It is, almost 10k less gas for 10 seizures (~1k/seizure).

src/Terms.sol
adhusson: Interesting, updating storage variables in fewer places seemed much better to me clarity-wise.

Similarly the min now shows that what matters is comparing `totalRepaid` to `repayableDebt`. Before, the min used to be of the form `badDebt = min(updatedDebt, oldDebt - repayableDebt)`. So you had to check that the updated was itself of the form `oldDebt - totalRepaid`, and then you could start to think about those 2 values.

src/Terms.sol
QGarchery: In favor also to not allow multiple seizures to have the same collateral token, see #107

src/Terms.sol
QGarchery: > the min now shows that what matters is comparing `totalRepaid` to `repayableDebt`.

It is what it's doing in effect, but not what is intuitively understandable. For example, would the following refactor make sense to you (as it's comparing directly `totalRepaid` to `repayableDebt`) ?
```suggestion
            badDebt = originalDebt - UtilsLib.max(totalRepaid, repayableDebt);
```

src/Terms.sol
QGarchery: I mean just declaring `seizure` instead of calling `seizures[i]` everywhere. Is it what saves gas ? If so, I don't understand why since it's fetching the data in memory in both cases

src/Terms.sol
adhusson: Yes, even better! (Although note that now I even prefer something like [this](https://github.com/morpho-org/morpho-v2/pull/95/files#diff-2f530276825e96ff84d75b14bb8f85575e2ba72ee56a29129525a672faada763R174-R175) where you can the total repaid to the repayable debt.)

src/Terms.sol
adhusson: Yes it's the variable declaration that saves gas. I think it's just due to the fact that there are a lot of uses of this variable : I count 14, so for a 1k gas diff per seizure you only need to waste ~70 gas per use. The memory access function for seizures is (commented with approximate gas numbers):

```
function memory_array_index_access_struct_Seizure_dyn(baseRef, index) -> addr // 8+3+3 = 14
{
    if iszero(lt(index, array_length_array_struct_Seizure_dyn(baseRef))) { panic_error_0x32() } // 3+(3+3+3) + (14+expansion) = 26+expansion
    let offset := mul(index, 32) // 5+3+3 = 11
    offset := add(offset, 32) // 3+3+3 = 9
    addr := add(baseRef, offset) // 3+3+3 = 9
} // total 14+26+11+9+9+expansion = 69+expansion
            
function array_length_array_struct_Seizure_dyn(value) -> length { // ~ 8+3 = 11
    length := mload(value) // 3+expansion
} // total 14+expansion
```

adhusson: To clarify, the compiler does not reuse the memory offset computation required to access `seizures[i]`, unless we make a variable, it recomputes it every time.

src/Terms.sol
```diff
@@ -175,10 +175,9 @@ contract Terms is ITerms {
         require(originalDebt > maxDebt, "position is healthy");
 
         uint256 totalRepaid;
-        Seizure memory seizure;
 
         for (uint256 i = 0; i < seizures.length; i++) {
-            seizure = seizures[i];
+            Seizure memory seizure = seizures[i];
```
adhusson, line 190: Note that this does not suffer from the 'solidity never deallocates memory' problem commonly found in loops as this is only a reference assignment and does not allocate memory.

src/Terms.sol
adhusson: replaced by #110

src/Terms.sol
QGarchery: Thanks for the explanation, indeed we use it a lot !

src/Terms.sol
QGarchery: This point becomes moot after #95 or #110 anyway

QGarchery (approved):

peyha (approved):

peyha (approved):

src/Terms.sol
```diff
@@ -170,58 +165,64 @@ contract Terms is ITerms {
...
 
-                totalRepaid += seizures[i].repaidBonds;
-                collateralOf[borrower][id][term.collaterals[i].token] -= seizures[i].seizedAssets;
+        for (uint256 i = 0; i < seizures.length; i++) {
+            Seizure memory seizure = seizures[i];
+            require(UtilsLib.exactlyOneZero(seizure.repaidBonds, seizure.seizedAssets), "INCONSISTENT_INPUT");
 
-                SafeTransferLib.safeTransfer(term.collaterals[i].token, msg.sender, seizures[i].seizedAssets);
+            if (seizure.seizedAssets > 0) {
+                seizure.repaidBonds = seizure.seizedAssets.mulDivUp(prices[seizure.collateralIndex], ORACLE_PRICE_SCALE)
```
MathisGD, line 194: we could make a stack variable of collateralIndex

src/Terms.sol
```diff
@@ -170,58 +165,64 @@ contract Terms is ITerms {
...
+                    .mulDivUp(1e18, LIQUIDATION_INCENTIVE_FACTOR);
+            } else {
+                seizure.seizedAssets = seizure.repaidBonds.mulDivDown(LIQUIDATION_INCENTIVE_FACTOR, 1e18).mulDivDown(
+                    ORACLE_PRICE_SCALE, prices[seizure.collateralIndex]
```
MathisGD, line 198: unsure why the order here is different. opening an issue

src/Terms.sol
```diff
@@ -170,58 +165,64 @@ contract Terms is ITerms {
...
-                    seizures[i].seizedAssets = seizures[i].repaidBonds.mulDivDown(LIQUIDATION_INCENTIVE_FACTOR, 1e18)
-                        .mulDivDown(ORACLE_PRICE_SCALE, collateralPrice);
-                }
+        uint256 totalRepaid;
 
-                totalRepaid += seizures[i].repaidBonds;
-                collateralOf[borrower][id][term.collaterals[i].token] -= seizures[i].seizedAssets;
+        for (uint256 i = 0; i < seizures.length; i++) {
+            Seizure memory seizure = seizures[i];
+            require(UtilsLib.exactlyOneZero(seizure.repaidBonds, seizure.seizedAssets), "INCONSISTENT_INPUT");
```
MathisGD, line 191: this prevents sending twice the same collat, but it's not very straightforward

MathisGD (approved):

PR #64, opened 2025-07-23
Recovery close factor POC

peyha:
Very naive way to implement the recovery close factor


peyha:
can be closed safely as we chose another liquidation mechanism

PR #66, opened 2025-07-26
merge storage mappings to reduce keccak computations

adhusson:
* simpler code
* save gas (eg 7k gas /1.2% in `testMatch`) note that the screenshots compares in the wrong direction so it looks like a gas loss

image


MathisGD:
not convinced for the moment, I would wait that the contract stabilizes a bit

MathisGD:
let's open an issue, this is getting very old

PR #67, opened 2025-07-28
no shares

adhusson:
simplifies code a little bit

con: no bond minting after a bad debt (would be a net loss for the lender)

### DIFF
--- a/src/Terms.sol
+++ b/src/Terms.sol
@@ -23,11 +23,11 @@ contract Terms is ITerms {
 
     /// STORAGE ///
 
-    mapping(address => mapping(bytes32 => uint256)) public bondSharesOf;
+    mapping(address => mapping(bytes32 => uint256)) public bondsOf;
     mapping(address => mapping(bytes32 => uint256)) public debtOf;
     mapping(bytes32 => uint256) public withdrawable;
     mapping(bytes32 => uint256) public totalBonds;
-    mapping(bytes32 => uint256) public totalShares;
+    mapping(bytes32 => uint256) public totalBadDebt;
     mapping(address => mapping(bytes32 => mapping(address => uint256))) public collateralOf;
 
     /// @dev Multiple offers can have the same nonce. This allows to implement easy and efficient batch-cancelling and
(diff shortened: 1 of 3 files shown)


src/Terms.sol
```diff
@@ -59,18 +59,16 @@ contract Terms is ITerms {
...
 
             debtOf[buyer][id] -= repaid;
-            bondSharesOf[buyer][id] += boughtShares;
-            bondSharesOf[seller][id] -= withdrawnShares;
-            debtOf[seller][id] += bonds - withdrawn;
+            bondsOf[buyer][id] += bought;
+            bondsOf[seller][id] -= withdrawn;
+            debtOf[seller][id] += borrowed;
+
+            require(repaid <= borrowed, "Cannot reduce total debt");
```
adhusson, line 70: maybe only force this when there is > 0 bad debt

adhusson:
closing because not useful in the end (having shares is fine)

src/Terms.sol
```diff
@@ -81,20 +79,16 @@ contract Terms is ITerms {
...
-        require(UtilsLib.exactlyOneZero(bonds, shares), "INCONSISTENT_INPUT");
+    function withdrawBond(Term memory term, uint256 bonds, address onBehalf) external {
         bytes32 id = _id(term);
 
-        if (bonds > 0) shares = bonds.mulDivUp(totalShares[id] + 1, totalBonds[id] + 1);
-        else bonds = shares.mulDivDown(totalBonds[id] + 1, totalShares[id] + 1);
-
-        bondSharesOf[onBehalf][id] -= shares;
-        withdrawable[id] -= bonds;
+        uint256 scaledBonds = bonds.mulDivDown(totalBonds[id] - totalBadDebt[id], totalBonds[id]);
```
adhusson, line 85: must scale bad debt down here

src/Terms.sol
```diff
@@ -23,11 +23,11 @@ contract Terms is ITerms {
 
     /// STORAGE ///
 
-    mapping(address => mapping(bytes32 => uint256)) public bondSharesOf;
+    mapping(address => mapping(bytes32 => uint256)) public bondsOf;
     mapping(address => mapping(bytes32 => uint256)) public debtOf;
     mapping(bytes32 => uint256) public withdrawable;
     mapping(bytes32 => uint256) public totalBonds;
-    mapping(bytes32 => uint256) public totalShares;
+    mapping(bytes32 => uint256) public totalBadDebt;
```
MathisGD, line 30: as it doesn't allow to remove these variables, not so worth it imo

PR #69, opened 2025-08-22
Recovery close factor for liquidation

peyha:
Not very optimized for now, we could remove an oracle call

### DIFF
--- a/src/Terms.sol
+++ b/src/Terms.sol
@@ -151,10 +151,12 @@ contract Terms is ITerms {
             vars.maxDebt += collateralQuoted.mulDivDown(term.collaterals[i].lltv, 1e18);
             vars.repayableDebt += collateralQuoted.mulDivUp(1e18, LIQUIDATION_INCENTIVE_FACTOR);
         }
+        uint256 thresholdProportion = 0.2e18;
+
         require(debtOf[borrower][id] > vars.maxDebt, "position is healthy");
 
         uint256 totalRepaid;
-
+        uint256 newMaxDebt;
         for (uint256 i = 0; i < term.collaterals.length; i++) {
             if (seizures[i].repaidBonds + seizures[i].seizedAssets > 0) {
                 require(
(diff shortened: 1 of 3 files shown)


src/Terms.sol
```diff
@@ -175,9 +175,20 @@ contract Terms is ITerms {
             }
         }
 
+        // Recovery close factor check
+        uint256 postLiquidationMaxDebt = 0;
+        for (uint256 i = 0; i < term.collaterals.length; i++) {
+            uint256 price = IOracle(term.collaterals[i].oracle).price();
+            uint256 collateralQuoted =
+                collateralOf[borrower][id][term.collaterals[i].token].mulDivDown(price, ORACLE_PRICE_SCALE);
+            postLiquidationMaxDebt += collateralQuoted.mulDivDown(term.collaterals[i].lltv, 1e18);
```
gd-colin: So here we round down as it would force to leave "healthier" positions, right? If we round the other way around it's less conservative but how bad can it be? Also, it could balance out the super conservative bad debt realization. I think it's worth convincing ourselves that it is the right rounding.

src/Terms.sol
```diff
@@ -175,9 +175,20 @@ contract Terms is ITerms {
...
+        }
+
         uint256 originalDebt = debtOf[borrower][id];
         debtOf[borrower][id] -= totalRepaid;
 
+        require(debtOf[borrower][id] > postLiquidationMaxDebt, "liquidation above recovery close factor");
```
gd-colin: Maybe greater or equal, if it's just equal the liquidation should go through.

test/LiquidationTest.sol
```diff
@@ -142,4 +142,19 @@ contract LiquidationTest is BaseTest {
         recordedLiquidator = liquidator;
         recordedData = data;
     }
+
+    function testLiquidateAboveRecoveryCloseFactor(uint256 bonds) public {
```
gd-colin, line 147: Maybe assert that the position js healthy after the liquidation in the test?

adhusson (approved):

src/Terms.sol
peyha: Kept a strict inequality to be consistent with `_isHealthy` (that I ended up re-implementing)

src/Terms.sol
peyha: I used the usual health rounding without thinking too deeply. If we want the liquidator to liquidate as much as possible with the recovery close factor, we should indeed round down as it will make the `postLiquidationMaxDebt` lower meaning that the liquidator can repay more debt

PR #75, opened 2025-09-01
feat: full liquidation backup

peyha:
draft


peyha:
can be closed safely as we chose another liquidation mechanism

PR #76, opened 2025-09-01
flash accounting

adhusson:
Works by calling `interact(callbackData)` and then doing actions in the callback.

Pros
* When withdrawing multiple collaterals with > 0 debt, single health check at the end instead of one per collat.
* No need for flashloans when rolling.
* Saves useless transfers on rolling.
* Saves useless transfers when moving collateral from one id to the other.

Cons
* Complex


adhusson:
Closing because too complex despite the benefits.

PR #77, opened 2025-09-02
no secondary trading with a different rate

adhusson:
see [this thread](https://morpholabs.slack.com/archives/C08A40KKN9W/p1756722331759709)


adhusson:
Likely not needed since
- the 2nd issue in the thread is not real
- the first issue is not worth fragmenting secondary that much

PR #79, opened 2025-09-05
Post maturity liquidation

### LINKED ISSUE #19: Implement liquidation after maturity

peyha:
- [x] More tests
- [ ] ~Refactor the common liquidation logic between the two functions~
- [ ] Justify roundings
- [x] Make the price decay a function of the incentive
- [ ] Precise the cliff and the duration
- [x] Improve math logic (exp, log)


src/Terms.sol
```diff
@@ -167,22 +169,68 @@ contract Terms is ITerms {
...
+
+            SafeTransferLib.safeTransfer(term.collaterals[i].token, msg.sender, seizures[i].seizedAssets);
+        }
 
         uint256 originalDebt = debtOf[borrower][id];
         debtOf[borrower][id] -= totalRepaid;
 
         // Realize bad debt
-        if (vars.repayableDebt < originalDebt) {
+        if (totalRepaid > originalDebt) {
```
adhusson: I don't understand this

src/Terms.sol
```diff
@@ -167,22 +169,68 @@ contract Terms is ITerms {
...
             // Because roundings are not aligned the effective bad debt is either the remaining debt or the original
             // debt minus the theoretical repayable debt.
-            uint256 badDebt = UtilsLib.min(debtOf[borrower][id], originalDebt - vars.repayableDebt);
+            uint256 badDebt = UtilsLib.min(debtOf[borrower][id], originalDebt - totalRepaid);
```
adhusson: `debtOf[borrower][id]` is already equal to `originalDebt - totalRepaid`

src/Terms.sol
```diff
@@ -167,22 +169,68 @@ contract Terms is ITerms {
...
+        require(seizures.length == term.collaterals.length, "should have all collats");
+
+        bytes32 id = _id(term);
+
+        int256 logCliff = -1.60943791243e18; // ln(20%)
+        uint256 auctionDuration = 1 hours;
 
-                totalRepaid += seizures[i].repaidBonds;
-                collateralOf[borrower][id][term.collaterals[i].token] -= seizures[i].seizedAssets;
+        uint256 totalRepaid = 0;
```
adhusson: not used (also not used in `liquidate`)

src/Terms.sol
adhusson, line 1: I think that for now it could be easier to not factorize with `liquidateInternal`, and we can factorize later, wdyt?

peyha: Yes! I didn't expect`liquidate` to be re-worked when I factorized it

src/Terms.sol
peyha: fixed it, wdyt ?

src/libraries/LogLib.sol
```diff
@@ -10,8 +10,16 @@ library LogLib {
 
     /// @dev Returns an approximation of ln.
     function wLn(int256 x) internal pure returns (int256 y) {
-        unchecked {
-            y = x - x * x / (2 * WAD_INT) + x * x / WAD_INT * x / (3 * WAD_INT);
+        uint256 N_ROUNDS = 10;
```
adhusson: Previous version was not precise enough in the expected input range? The 10 rounds of looping are scary.

peyha: The 10 rounds are arbitrary, not sure what would be the best precision (3 seems fine) but using the Mclaurin series is probably not the best way to have an accurate result

adhusson: solady has an [lnWad](https://github.com/Vectorized/solady/blob/main/src/utils/FixedPointMathLib.sol#L277), could be interesting to compare gas cost & accuracy of all 3.

peyha: 10 rounds only guarantees ~10decimals while solady's lnWad is ~15 decimals so it seems like solady is way better

peyha: Solady's lnWad only consumes 614 gas as stated [here](https://xn--2-umb.com/22/exp-ln/) I don't think we can do much better

src/libraries/MathLib.sol
```diff
@@ -13,6 +13,16 @@ library MathLib {
...
+    /// @dev Returns (`x` * `y`) / `d` rounded down.
+    function mulDivDown(int256 x, int256 y, int256 d) internal pure returns (int256) {
+        return (x * y) / d;
+    }
+
+    /// @dev Returns (`x` * `y`) / `d` rounded up.
+    function mulDivUp(int256 x, int256 y, int256 d) internal pure returns (int256) {
+        return (x * y + (d - 1)) / d;
+    }
+
```
MathisGD: to remove right?

MathisGD: oh no it's for int

MathisGD: actually there are no int anymore right?

src/MorphoV2.sol
```diff
@@ -12,6 +12,7 @@ import "./interfaces/ICallbacks.sol";
 
 contract MorphoV2 is IMorphoV2 {
     using MathLib for uint256;
+    using MathLib for int256;
```
```suggestion
```

src/MorphoV2.sol
```diff
@@ -274,13 +224,80 @@ contract MorphoV2 is IMorphoV2 {
...
+
+        for (uint256 i = 0; i < obligation.collaterals.length; i++) {
+            uint256 price = IOracle(obligation.collaterals[i].oracle).price();
+
+            uint256 discountFactor = 1e18
+                - (1e18 - 1e36 / LIQUIDATION_INCENTIVE_FACTOR).mulDivDown(
+                    UtilsLib.min(1e18 * (block.timestamp - obligation.maturity) / auctionTimeToTarget, 1e18), 1e18
+                );
+
+            uint256 auctionPrice = price.mulDivDown(discountFactor, 1e18);
```
MathisGD, line 331: idk if the abstraction should be discount, or just LIF

MathisGD: also, it's really tempting to factorize with the other liquidation

src/MorphoV2.sol
MathisGD, line 1: some thinking about separating vs merging the functions:
- if we separate them, arguably the post maturity liquidation shouldn't realize the bad debt, so there is no computation of the health at the beginning at all.
- will it make liquidator's life easier or harder

MathisGD, line 1: there are some conflicts as well

src/libraries/LogLib.sol
peyha: removed

src/MorphoV2.sol
peyha: good point, I think separating is better
- if post maturity liquidation does not realize the bad debt, then there is no need to check all oracles (except the ones corresponding to the asset that gets liquidated) so liquidating one collateral on an obligation with multiple collaterals can get quite cheap
- it will not make liquidator's life much harder as the interface is the same and only the function name differs (and I guess liquidators will run two distinct logic to handle normal liquidation and post-maturity liquidation

peyha: now I'm not sure that post maturity liquidation should not realize the bad debt (how would it be realized if not there)

src/MorphoV2.sol
```diff
@@ -335,9 +403,8 @@ contract MorphoV2 is IMorphoV2 {
                 address currentCollateralToken = obligation.collaterals[i].token;
                 require(currentCollateralToken > previousCollateralToken, "collaterals not sorted");
                 uint256 price = IOracle(obligation.collaterals[i].oracle).price();
-                maxDebt += collateralOf[borrower][id][currentCollateralToken].mulDivDown(price, ORACLE_PRICE_SCALE)
-                    .mulDivDown(obligation.collaterals[i].lltv, 1e18);
-                previousCollateralToken = currentCollateralToken;
```
MathisGD, line 340: wrong rebasing I think

MathisGD:
replaced by #188

PR #85, opened 2025-09-11
loan asset as collateral

adhusson:
WIP

See issue #87.

* loan asset immediately withdrawable at maturity.
* loan asset cannot be liquidated.
* loan asset treated as collateral (but not included in list).
* debt does not ignore covered amount

I find it less clean.


adhusson:
not clean, there is no reason to reuse the `collateralOf` mapping in general, and especially if the cover does not behave as collateral at all

PR #86, opened 2025-09-12
loan asset as collateral (can be liquidated)

adhusson:
See issue #87.

* loan asset immediately withdrawable at maturity.
* loan asset can be liquidated.
* loan asset treated as collateral (but not included in list).
* debt does not ignore covered amount


adhusson:
not clean, will try to find a better version of this

Issue #87, opened 2025-09-13
Should the loan asset be usable as collateral


adhusson:
Early liquidations & early "repayments" would just cover the debt instead of irreversibly repaying. The borrower could later buy a bond to offset his debt and save the difference.

As a good side effect, since early liquidations would no longer repay but just swap collateral, there is no 0% loan / MEV extractable from early liquidations ([slack thread](https://morpholabs.slack.com/archives/C08A40KKN9W/p1756722331759709)).

Not clear if:
- The collateral is immediately available for withdraw after maturity, or if there must be an explicit repayment / post-maturity liquidation.
- The loan asset collateral can be liquidated, or if debt covered by loan assets is ignored in liquidations.
- The loan asset should be treated separately from collaterals, or treated as one, or forced to be included in collaterals.

Related PRs: 
- #84 makes loan asset collat immediately available for withdraw. It cannot be liquidated, and covered debt is not counted.
  - Variation: #89 reuses the supply/withdraw collateral functions, but I don't think it's worth it. Clearer to have dedicated functions for this.
-  In #85 reuses the collateral mappings to store covered amounts, and `debtOf` does not ignore the covered amount. I don't like it, it just makes the code more complex.
-  In #86 and #88 it can be liquidated, always at 1:1, but I don't think it's useful and is also more complex than #84. In #86  the health/liquidation loops explicitly consider the cover, in #88 the loan asset is force included in the collateral list.

Notes:
* From [this suggestion](https://docs.google.com/document/d/1xsTro-S1CE_MxYD6oPqKct8G0tS7vyvExW_wHRazRVk/edit?usp=sharing)

MathisGD:
[latest update](https://morpholabs.slack.com/archives/C02N7CZ088N/p1761821926095609)

MathisGD:
[latest update](https://morpholabs.slack.com/archives/C02N7CZ088N/p1761901481239409)

PR #88, opened 2025-09-13
loan asset forced to be collateral

adhusson:
See issue #87 

- loan asset not immediately withdrawable at maturity (must post-maturity liquidate first)
- loan asset can be liquidated (always at price 1)
- loan asset force included as first collateral
- debt does not ignore covered amount


adhusson:
not clean, will try to find a better version of this

PR #89, opened 2025-09-14
debt can be covered with loan asset (reuse collateral functions)

adhusson:
See issue #87 

* loan asset immediately withdrawable at maturity
* loan asset cannot be liquidated
* loan asset treated separately from collaterals, but the `supplyCollateral` and `withdrawCollateral` functions are used
* debt ignores covered amount


MathisGD:
I don't think it's the right design, #176 is better

PR #90, opened 2025-09-16
separate matching with take and make structs

adhusson:
- Matching only considers loan token / collaterals / pricing
- Terms manages nonces and authorizations
- Offers are ratified before take, by direct call, signature, or callback.
  - What is the purpose of a callback here vs. asking a contract to call Terms directly? Not clear. But going without a callback requires having a standard for calling contracts to get them to ratify their offers.
  - Unlike with ratify-when-taking, a ratified offer stays ratified. It should probably be limited to the current tx.

Note on https://github.com/morpho-org/morpho-v2/pull/90/commits/57f399180f8caa8a7343b028956f0ed9a82353b5 that requires a magic value in hooks: due to 7702 users will have contract code & will also approve protocol contracts. If a user has 7702 code with a fallback that does nothing instead of reverting, he will be exploitable (by setting sig.v == 1) unless we ask for a magic return value.


adhusson:
replaced by #138

PR #91, opened 2025-09-16
fix: cap total repaid to debt

adhusson:
It is possible to be in a state of bad debt such that for one collateral,
* if you try to repay all bonds, not all the collateral is seized.
* if you try to seize all the collateral, the repaid bonds are overestimated and `originalDebt - totalRepaid` underflows.


MathisGD:
not relevant after #119

PR #93, opened 2025-09-17
feat: liquidation per collateral, transient health

MathisGD:


src/Terms.sol
```diff
@@ -143,81 +144,68 @@ contract Terms is ITerms {
         SafeTransferLib.safeTransfer(collateral, msg.sender, assets);
     }
 
-    struct Vars {
-        uint256 maxDebt;
-        uint256 repayableDebt;
-    }
-
     /// @notice Execute the given collection of `seizures` on the given `term` of the given `borrower`.
```
adhusson:
```suggestion
    /// @notice Execute a seizure on the given `term` of the given `borrower`.
```

src/Terms.sol
```diff
@@ -143,81 +144,68 @@ contract Terms is ITerms {
...
     /// @dev On each seizure either `repaidAmounts` or `seizedAssets` should be equal to zero.
```
adhusson:
```suggestion
```

src/Terms.sol
```diff
@@ -143,81 +144,68 @@ contract Terms is ITerms {
...
+                repayableDebt += collateralQuoted.mulDivUp(1e18, LIQUIDATION_INCENTIVE_FACTOR);
+            }
+            require(debtOf[borrower][id] > maxDebt, "position is healthy");
+            TransientLib.setMapping(borrower, true);
+
+            // Realize bad debt
+            if (repayableDebt < debtOf[borrower][id]) {
+                // Because roundings are not aligned the effective bad debt is either the remaining debt or the original
+                // debt minus the theoretical repayable debt.
+                uint256 badDebt = UtilsLib.min(debtOf[borrower][id], debtOf[borrower][id] - repayableDebt);
```
adhusson: in the previous version at this point `debtOf` had been decrement by the repaid amount, and the second argument of min was `originalDebt - repayableDebt`. Here `min(debt,debt-x)` will be `debt-x`.

```suggestion
                uint256 badDebt = debtOf[borrower][id] - repayableDebt;
```

src/Terms.sol
```diff
@@ -143,81 +144,68 @@ contract Terms is ITerms {
...
+                debtOf[borrower][id] -= badDebt;
+                totalBonds[id] -= badDebt;
             }
         }
 
-        uint256 originalDebt = debtOf[borrower][id];
-        debtOf[borrower][id] -= totalRepaid;
+        uint256 collateralPrice = IOracle(collateral).price();
```
adhusson: In the previous version a collateral price insie a tx change could affect separate liquidations in the tx, but they all recomputed max debt. In this version you can have set max debt with a price and liquidate at another price.

src/Terms.sol
```diff
@@ -143,81 +144,68 @@ contract Terms is ITerms {
...
+            );
         }
 
-        withdrawable[id] += totalRepaid;
-
-        if (data.length > 0) ICallbacks(msg.sender).onLiquidate(seizures, borrower, msg.sender, data);
-
-        SafeTransferLib.safeTransferFrom(term.loanToken, msg.sender, address(this), totalRepaid);
+        collateralOf[borrower][id][collateral] -= seizedAssets;
+        debtOf[borrower][id] -= repaidBonds;
```
adhusson, line 198: could need this or zeroFloorSub, because the round up may go over the debt (not sure in the single collateral case), see https://github.com/morpho-org/morpho-v2/pull/91

```suggestion
        if (repaidBonds > debtOf[borrower][id]) repaidBonds = debtOf[borrower][id];
        debtOf[borrower][id] -= repaidBonds;
```

src/Terms.sol
```diff
@@ -143,81 +144,68 @@ contract Terms is ITerms {
...
     /// @param term The term of the bond.
-    /// @param seizures An array of amounts of debt to repay or assets to seize with the index of the collateral in the
-    /// term's collateral assets.
+    /// @param collateral The collateral to seize.
+    /// @param repaidBonds The amount of debt to repay.
+    /// @param seizedAssets The amount of assets to seize.
     /// @param borrower The debtor of the loan.
     /// @param data Arbitrary data to pass to the callback. Pass empty data if not needed.
-    /// @return A collection of the actual amounts of debt repaid or asset seized with the collateral index.
-    function liquidate(Term memory term, Seizure[] memory seizures, address borrower, bytes calldata data)
```
adhusson, line 159: Not completely convinced because the final code is not so much simpler that it's a strong reason to switch. 

On the other hand I need to think more about what happens about the possible effect of arbitrary interactions with a term between 2 liquidations.

src/Terms.sol
```diff
@@ -147,26 +147,26 @@ contract Terms is ITerms {
...
         uint256 repaidBonds,
         uint256 seizedAssets,
         address borrower,
         bytes calldata data
     ) external returns (uint256, uint256) {
-        require(UtilsLib.exactlyOneZero(repaidBonds, seizedAssets), "INCONSISTENT_INPUT");
-
-        uint256 maxDebt;
-        uint256 repayableDebt;
+        require(UtilsLib.atLeastOneZero(repaidBonds, seizedAssets), "INCONSISTENT_INPUT");
```
adhusson, line 163: why not exactly? to allow a liquidate call that just enables the liquidation but does nothing?

MathisGD: to allow to realize bad debt without liquidating anything

src/Terms.sol
```diff
@@ -143,81 +144,64 @@ contract Terms is ITerms {
...
+            if (repayableDebt < debtOf[borrower][id]) {
+                uint256 badDebt = debtOf[borrower][id] - repayableDebt;
+                debtOf[borrower][id] -= badDebt;
+                totalBonds[id] -= badDebt;
             }
         }
 
-        uint256 originalDebt = debtOf[borrower][id];
-        debtOf[borrower][id] -= totalRepaid;
+        uint256 collateralPrice = IOracle(term.collaterals[collateralIndex].oracle).price();
```
MathisGD, line 186: not ideal

MathisGD:
closing, because it makes thinking about the behavior too complex. also it doesn't allow to not call multiple times the oracles

PR #95, opened 2025-09-18
feat: realize bad debt first

adhusson:
a difference with #62 : in 62, if rounding makes you repay more than `repayableDebt`, it will reduce your bad debt. Here it won't, the total repaid amount is at most `repayableDebt`.

### DIFF
--- a/src/Terms.sol
+++ b/src/Terms.sol
@@ -171,19 +171,17 @@ contract Terms is ITerms {
             collateralOf[borrower][id][collateralToken] -= seizure.seizedAssets;
         }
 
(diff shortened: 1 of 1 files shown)


peyha (commented):
The readability of this PR is a bit difficult to me, reasoning with `min` values is less natural than with ifs

adhusson:
Closing in favor of #110

adhusson:
Note that there is an issue in this PR as noted here https://github.com/morpho-org/morpho-v2/pull/110#discussion_r2382355607

PR #96, opened 2025-09-18
test: add assumption for user address in supply collateral test

adhusson:


QGarchery (approved):
Completely fine assumption to make, but why does it matter / why is it needed here ?

adhusson:
The check "balance of user" fails when `user == terms`.

MathisGD (approved):

PR #97, opened 2025-09-18
refactor: reuse health computation in liquidation

adhusson:
oracle prices are no longer reused

### DIFF
--- a/src/Terms.sol
+++ b/src/Terms.sol
@@ -75,7 +75,7 @@ contract Terms is ITerms {
             totalBonds[id] += bought;
             totalBonds[id] -= withdrawn;
 
-            (uint maxDebt, ) = health(term,seller);
+            (uint256 maxDebt,) = health(term, seller);
             require(debtOf[seller][id] <= maxDebt, "Seller is unhealthy");
         }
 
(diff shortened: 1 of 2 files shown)


src/Terms.sol
```diff
@@ -257,21 +249,18 @@ contract Terms is ITerms {
         require(signatory != address(0) && offer.offering == signatory, "Invalid signature");
     }
 
-    function _isHealthy(Term memory term, address borrower) internal view returns (bool) {
+    ///@dev Returns the max debt and the repayable debt.
+    function health(Term memory term, address borrower) public view returns (uint256, uint256) {
```
peyha, line 253: `health` could also return an array with collateral prices to prevent multiple oracle price calls

peyha (commented):
I personally prefer this version of `health` as it can be re-used more easily (especially in the post-maturity dutch auction)

QGarchery (changes requested):
To me it's more important that we reuse oracle price values instead of calling them twice

MathisGD:
agree with quentin. Health could also return the list of prices, but then it becomes heavy. Closing for the moment

PR #99, opened 2025-09-19
force loan asset as collateral

adhusson:
compare to #100

in this one, all borrow positions must be postmaturity liquidated before lenders can withdraw

### DIFF
--- a/src/Terms.sol
+++ b/src/Terms.sol
@@ -152,8 +152,6 @@ contract Terms is ITerms {
         }
         require(debtOf[borrower][id] > vars.maxDebt, "position is healthy");
 
-        uint256 totalRepaid;
-
         for (uint256 i = 0; i < term.collaterals.length; i++) {
             if (seizures[i].repaidBonds + seizures[i].seizedAssets > 0) {
                 require(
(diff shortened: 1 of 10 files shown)


peyha (commented):
There seems to be an issue with this approach. A borrower can collateralize an amount x of debt with an amount x of the same loan asset and never repay his loan. In this situation, lenders will never be able to withdraw from the protocol.
A solution would be to make withdrawable liquidity increase when users supplyCollateral the loan asset but it might create other issues because the fund of the user might not be seizable

update: just saw that the solution was implemented in #100

src/Terms.sol
```diff
@@ -49,6 +49,10 @@ contract Terms is ITerms {
         require(block.timestamp >= offer.offerStart, "offer not started");
         require(block.timestamp <= offer.offerExpiry, "offer expired");
         require(term.maturity >= block.timestamp, "bond maturity");
+        require(
```
peyha, line 52: could also require `term.collaterals[0].lltv == WAD`

adhusson:
> lenders will never be able to withdraw from the protocol.

post-maturity, borrowers are liquidated and the withdrawable amount does increase, so there is no issue in this PR either

peyha:
Ah yes I see, worst case is that borrowers with debt & coveredDebt are not liquidated by liquidators so lenders can just liquidate to increase withdrawable liquidity

MathisGD:
same as #89

PR #100, opened 2025-09-19
force loan asset as collateral + immediate withdraw

adhusson:
All borrow positions' loan asset collateral are immediately withdrawable at maturity, no need to liquidate each

Since it's possible that a position is never repaid, the debt could stay > 0 forever even if it was effectively repaid through loan asset collateralization. 

The solution is to redefine debt as `full debt - loan asset collateralization`. But `debtOf(borrower,id)` does not contain the loan token address, so the effective debt can't be read in the `collateralOf` mapping. To fix this the loan asset collateral is stored at a constant key (`address(0)`).

Should be able to show:
```
  withdrawable - (sum_user collateralOf[user][loanToken])
= (sum_interaction repaid in postmaturity liq) - (sum_interaction withdrawn) 
= totalBonds - (sum_user fullDebtOf[user])
```

In the end it should behave the same as https://github.com/morpho-org/morpho-v2/pull/84 (but it's more complex)

### DIFF
--- a/src/Terms.sol
+++ b/src/Terms.sol
@@ -6,6 +6,7 @@ import "./libraries/UtilsLib.sol";
 import "./libraries/SafeTransferLib.sol";
 import "./libraries/MathLib.sol";
 import "./libraries/ExpLib.sol";
+import "./libraries/LogLib.sol";
 import "./interfaces/IOracle.sol";
 import "./interfaces/ITerms.sol";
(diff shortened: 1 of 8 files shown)


adhusson:
closing because having a postmaturity liquidation step for all borrowers is too complex, versions where cover is immediately withdrawable at maturity are better (see #87 for references)

PR #102, opened 2025-09-23
feat: linear price

MathisGD:


src/Terms.sol
```diff
@@ -49,13 +49,16 @@ contract Terms is ITerms {
...
+        require(block.timestamp >= offer.start, "offer not started");
+        require(block.timestamp <= offer.end, "offer expired");
         require(term.maturity >= block.timestamp, "bond maturity");
         _checkSignature(offer, sig);
         _checkOffer(term, offer);
 
-        uint256 bonds = assets * (1e18 + (term.maturity - block.timestamp) * offer.rate) / 1e18;
+        uint256 price = offer.startPrice
+            + (offer.endPrice - offer.startPrice) * (block.timestamp - offer.start) / (offer.end - offer.start);
+        require(price <= 1e18, "price too high");
```
peyha: can't `price > 1e18` make sense if for instance there are rewards ?

test/BaseTest.sol
```diff
@@ -94,9 +94,10 @@ abstract contract BaseTest is Test {
             loanToken: term.loanToken,
             collaterals: term.collaterals,
             maturity: block.timestamp + 100,
-            offerStart: block.timestamp,
-            offerExpiry: block.timestamp + 200,
-            rate: 0,
+            start: block.timestamp,
+            end: block.timestamp + 1,
```
adhusson: why +200 -> +1

test/TakeTest.sol
```diff
@@ -40,8 +40,10 @@ contract TakeTest is BaseTest {
         lendOffer.assets = 100;
         lendOffer.loanToken = address(loanToken);
         lendOffer.maturity = block.timestamp + 100;
-        lendOffer.offerExpiry = block.timestamp + 200;
-        lendOffer.rate = 0.01e18 / 100;
+        lendOffer.start = block.timestamp;
+        lendOffer.end = block.timestamp + 200;
+        lendOffer.startPrice = 0.99 ether;
+        lendOffer.endPrice = 0.99 ether;
```
adhusson: rate has become 0

test/TakeTest.sol
```diff
@@ -53,8 +55,9 @@ contract TakeTest is BaseTest {
         borrowOffer.assets = 100;
         borrowOffer.loanToken = address(loanToken);
         borrowOffer.maturity = block.timestamp + 100;
-        borrowOffer.offerExpiry = block.timestamp + 200;
-        borrowOffer.rate = 0.01e18 / 100;
+        borrowOffer.end = block.timestamp + 200;
+        borrowOffer.startPrice = 0.99 ether;
+        borrowOffer.endPrice = 0.99 ether;
```
adhusson: same here, rate is now 0

src/interfaces/ITerms.sol
```diff
@@ -22,11 +22,10 @@ struct Offer {
     address loanToken;
     Collateral[] collaterals;
     uint256 maturity;
-    uint256 offerStart;
-    uint256 offerExpiry;
-    // The rate is expressed in percentage per second and is scaled by WAD, so `0.01e18 / uint256(365 days)` represents
-    // 1% APR.
-    uint256 rate;
+    uint256 start;
+    uint256 end;
```
adhusson:
```suggestion
    uint256 validFrom;
    uint256 validUntil;
```


```suggestion
    uint256 validityStart;
    uint256 validityEnd;
```


```suggestion
    uint256 start;
    uint256 expiry;
```

src/Terms.sol
MathisGD: opening an issue

test/BaseTest.sol
MathisGD: actually + 0 is fine

test/TakeTest.sol
MathisGD: hmm no it's not zero (interest is 0.01 ether / wad)

src/Terms.sol
```diff
@@ -43,19 +43,30 @@ contract Terms is ITerms {
...
         require(term.maturity >= block.timestamp, "bond maturity");
-        _checkSignature(offer, sig);
-        _checkOffer(term, offer);
+        require(offer.loanToken == term.loanToken, "Loan tokens do not match");
+        require(offer.maturity == term.maturity, "Maturities do not match");
+        require(signatureIsValid(offer, sig), "Invalid signature");
+        _checkCollateralInclusion(term, offer);
 
-        uint256 bonds = assets * (1e18 + (term.maturity - block.timestamp) * offer.rate) / 1e18;
+        uint256 denominator = offer.expiry - offer.start;
```
adhusson:
```suggestion
        uint256 offerDuration = offer.expiry - offer.start;
```

src/Terms.sol
```diff
@@ -43,19 +43,30 @@ contract Terms is ITerms {
...
+        uint256 price = denominator > 0
+            ? offer.startPrice + (offer.expiryPrice - offer.startPrice) * (block.timestamp - offer.start) / denominator
+            : offer.startPrice;
+
+        // todo check rounding
```
QGarchery: I don't think there is any "good" rounding direction here:
- if you choose to underestimate bonds (which is what is done here), then the buyer is better off
- if you choose to overestimage bonds, then the seller is better off
but the caller could be either the seller of the buyer (or acting on either one behalf)

src/Terms.sol
```diff
@@ -245,13 +253,12 @@ contract Terms is ITerms {
         }
     }
 
-    function _checkSignature(Offer memory offer, Signature memory signature) internal view {
+    function signatureIsValid(Offer memory offer, Signature memory signature) internal view returns (bool) {
```
QGarchery, line 255: Curious why did you do this change

src/Terms.sol
```diff
@@ -226,10 +237,7 @@ contract Terms is ITerms {
         return keccak256(abi.encode(term));
     }
 
-    function _checkOffer(Term memory term, Offer memory offer) internal pure {
-        require(offer.loanToken == term.loanToken, "Loan tokens do not match");
-        require(offer.maturity == term.maturity, "Maturities do not match");
-
+    function _checkCollateralInclusion(Term memory term, Offer memory offer) internal pure {
```
QGarchery, line 239: same

src/Terms.sol
```diff
@@ -43,19 +43,30 @@ contract Terms is ITerms {
...
+        uint256 bonds,
         address onBehalf,
         Offer memory offer,
         Signature memory sig,
         address callbackAddress,
         bytes memory callbackData
     ) public {
-        require(block.timestamp >= offer.offerStart, "offer not started");
-        require(block.timestamp <= offer.offerExpiry, "offer expired");
+        require(assets * bonds == 0, "inconsistent input");
```
```suggestion
        require(assets == 0 || bonds == 0, "inconsistent input");
```
[optional] clearer (to me) and doesn't introduce a potential overflow

src/Terms.sol
MathisGD: yes, no right direction but I choose them randomly, maybe there is still a "better" direction. maybe advantaging the maker?

src/Terms.sol
MathisGD: the require is clearer, wdyt

src/Terms.sol
MathisGD: checks are less hidden, just the collateral inclusion is externalized

src/Terms.sol
```diff
@@ -43,19 +43,30 @@ contract Terms is ITerms {
...
 
-        uint256 bonds = assets * (1e18 + (term.maturity - block.timestamp) * offer.rate) / 1e18;
+        uint256 denominator = offer.expiry - offer.start;
+        uint256 price = denominator > 0
+            ? offer.startPrice + (offer.expiryPrice - offer.startPrice) * (block.timestamp - offer.start) / denominator
+            : offer.startPrice;
+
+        // todo check rounding
+        if (assets > 0) bonds = assets.mulDivDown(1e18, price);
+        else assets = bonds.mulDivUp(price, 1e18);
```
adhusson: Not sure what the principle behind choosing a rounding must be here.

The rounding is controlled by the taker so we could do



```suggestion
        if (assets > 0) bonds = taker == buyer ? assets.mulDivDown(1e18,price) : assets.mulDivUp(1e18,price);
        else assets = taker == buyer ? bonds.mulDivUp(price,1e18) : bonds.mulDivDown(price,1e18);
```

src/Terms.sol
```diff
@@ -43,19 +43,30 @@ contract Terms is ITerms {
     function take(
         Term memory term,
         uint256 assets,
+        uint256 bonds,
         address onBehalf,
```
adhusson:
```suggestion
        address taker,
```

repeated from https://github.com/morpho-org/morpho-v2/pull/32#discussion_r2379179565

src/Terms.sol
```diff
@@ -43,19 +43,30 @@ contract Terms is ITerms {
...
         Offer memory offer,
         Signature memory sig,
         address callbackAddress,
         bytes memory callbackData
```
adhusson:
```suggestion
        address takerCallbackAddress,
        bytes memory takerCallbackData
```

repeated from https://github.com/morpho-org/morpho-v2/pull/32#discussion_r2379180397

src/Terms.sol
```diff
@@ -245,13 +253,12 @@ contract Terms is ITerms {
...
 
-    function _checkSignature(Offer memory offer, Signature memory signature) internal view {
+    function signatureIsValid(Offer memory offer, Signature memory signature) internal view returns (bool) {
         bytes32 hashStruct = keccak256(abi.encode(OFFER_TYPEHASH, offer));
         bytes32 domainSeparator = keccak256(abi.encode(DOMAIN_TYPEHASH, block.chainid, address(this)));
         bytes32 digest = keccak256(bytes.concat("\x19\x01", domainSeparator, hashStruct));
         address signatory = ecrecover(digest, signature.v, signature.r, signature.s);
-
-        require(signatory != address(0) && offer.offering == signatory, "Invalid signature");
+        return signatory != address(0) && offer.offering == signatory;
```
adhusson, line 260:
```suggestion
        require(offer.offering != address(0) && offer.offering == ecrecover(digest, signature.v, signature.r, signature.s), "Invalid sig");
```

repeated from https://github.com/morpho-org/morpho-v2/pull/32#discussion_r2379472021

src/Terms.sol
adhusson: shouldn't also at least one of them be nonzero? what would be the purpose of taking for 0?

MathisGD: depend if we allow no-ops or not. we can open an issue for that

src/Terms.sol
QGarchery: No strong opinion on this

src/Terms.sol
QGarchery: Good point, that would be the best. But we don't do it everywhere for now, see [this PR](https://github.com/morpho-org/morpho-v2/pull/24) (old, and not ready for review at all)

MathisGD: I went for that

QGarchery: Because we don't do it everywhere I don't think it makes sense to do it just here (it adds code complexity for no guarantee)

MathisGD: fair, I'll rollback

adhusson (approved):

peyha (approved):

QGarchery (approved):

PR #106, opened 2025-09-25
trading fee

MathisGD:
"primitive":
- trading fee = (bonds-buyerAssets) * fee
- makerAssets = bonds * price

the computations should be such that
- feePct = 0 => fee = 0
- buyerAssets >= sellerAssets


src/MorphoV2.sol
```diff
@@ -35,22 +35,37 @@ contract Terms is ITerms {
     /// otherwise one might not be takable anymore while an other one at the same nonce is still takeable.
     mapping(address user => mapping(uint256 nonce => uint256)) public consumed;
 
+    /// @dev Cut on interest at each trade.
+    mapping(address loanToken => uint256) public tradingFeePct;
+    mapping(address loanToken => address) public tradingFeeRecipient;
```
peyha: is this the right granularity ? 
On Morpho-Blue, the fee is per market (so per loan token, collat token, lltv, oracle and irm). On uniswap-V3 it's per pool (which is very annoying because the DAO needs to set the fee for each pool).
The best of both world could be to have a global trading fee but to still be able to set fee per loan asset (or per term) as suggested by banteg for univ3 [here](https://gov.uniswap.org/t/temperature-check-activate-uniswap-protocol-governance/22936/39)

src/Terms.sol
```diff
@@ -35,22 +35,37 @@ contract Terms is ITerms {
...
         uint256 bonds,
         address taker,
         Offer memory offer,
         Signature memory sig,
         address takerCallbackAddress,
         bytes memory takerCallbackData
     ) public {
-        require(assets == 0 || bonds == 0, "inconsistent input");
+        uint256 sum = buyerAssets + sellerAssets + bonds;
+        require(sum == buyerAssets || sum == sellerAssets || sum == bonds, "inconsistent input");
```
peyha: is it ok if all three are 0 ?

src/Terms.sol
```diff
@@ -35,22 +35,37 @@ contract Terms is ITerms {
...
+        uint256 sellerAssets,
         uint256 bonds,
         address taker,
         Offer memory offer,
         Signature memory sig,
         address takerCallbackAddress,
         bytes memory takerCallbackData
     ) public {
-        require(assets == 0 || bonds == 0, "inconsistent input");
+        uint256 sum = buyerAssets + sellerAssets + bonds;
```
peyha:
```suggestion
        uint256 zeroValueCount = (buyerAssets == 0) + (sellerAssets == 0) + (bonds == 0);
```
might be clearer

MathisGD: you cannot sum bools in solidity

src/Terms.sol
MathisGD: see #108

src/MorphoV2.sol
MathisGD: from our latest [discussion](https://www.notion.so/morpho-labs/Bootcamp-V2-fees-granularity-229d69939e6d806c95e1d42c7cdba3ee) it's per loan, but it could be re-challenged (notably for like yearly obligations, a thinner granularity could make sense).

src/Terms.sol
peyha:
```suggestion
        uint256 zeroValueCount;
        assembly {
            zeroValueCount := add(iszero(buyerAssets), add(iszero(sellerAssets), iszero(bonds)))
         }
```
would work but using yul for this is probably not worth it

src/MorphoV2.sol
peyha: there should not be so much yearly obligation no ? 
Then a loan token settable trading fee + the possibility of setting it per bond id could also work

src/Terms.sol
MathisGD:
```suggestion
        uint256 zeroValueCount = buyerAssets == 0 ? 1 : 0 + sellerAssets == 0 ? 1 : 0 + bonds == 0 ? 1 : 0;
```
but isn't it heavier than the current code?

src/MorphoV2.sol
MathisGD: and both are added?

peyha: It could be the bond id trading fee if it has been set and otherwise the loan token trading fee but I haven't thought too much about it

src/Terms.sol
peyha: I think it's quite close in term of clarity but it potential overflows on `sum` and the gas cost is probably equivalent (but at this point of development gas is not relevant)

src/MorphoV2.sol
MathisGD: opened an issue #118 to discuss separately

src/MorphoV2.sol
```diff
@@ -35,22 +35,38 @@ contract Terms is ITerms {
...
+
     /// ENTRY-POINTS ///
 
     /// @dev Same function used to buy and sell.
     /// @dev If one wants to match two offers without taking a position, they can batch take them and not have a
     /// position at the end.
     function take(
         Term memory term,
-        uint256 assets,
+        uint256 buyerAssets,
```
adhusson, line 67: why are both arguments needed?

src/Terms.sol
```diff
@@ -35,22 +35,38 @@ contract Terms is ITerms {
     /// otherwise one might not be takable anymore while an other one at the same nonce is still takeable.
     mapping(address user => mapping(uint256 nonce => uint256)) public consumed;
 
+    /// @dev Cut on interest at each trade.
+    mapping(address loanToken => uint256) public tradingFeePct;
```
adhusson: 'Pct' not accurate here

src/Terms.sol
```diff
@@ -74,11 +90,25 @@ contract Terms is ITerms {
...
             ? offer.startPrice + (offer.expiryPrice - offer.startPrice) * (block.timestamp - offer.start) / offerDuration
             : offer.startPrice;
+        require(price <= 1e18, "price too high");
+
+        if (buyerAssets > 0) {
+            bonds = buyerAssets.mulDivDown(1e18, price);
+            sellerAssets = buyerAssets - (bonds - buyerAssets).mulDivDown(tradingFeePct[term.loanToken], 1e18);
+        } else if (sellerAssets > 0) {
+            buyerAssets = sellerAssets.mulDivDown(
+                1e18, 1e18 + tradingFeePct[term.loanToken] - tradingFeePct[term.loanToken].mulDivDown(1e18, price)
```
adhusson: ok with this

src/MorphoV2.sol
MathisGD: if you are the buyer you want buyerAssets, and if you are the seller you want sellerAssets. what you have in mind is that the fee is ~constant so asynchrony doesn't matter, people can compute offchain? I'm not sure this addition is not worth it. People are so used to be able to just say the amount they want.

src/Terms.sol
MathisGD: why?

src/Terms.sol
MathisGD: you mean all the computations?

src/MorphoV2.sol
adhusson: given the amount of offchain work (you go through a router to get offers) maybe we should expect more sophistication from tx senders, not sure. Code would be simpler at least, at the most extreme you can remove shares/units inputs + here seller/buyer assets.

src/Terms.sol
adhusson: it's in per-1e18. For instance in VaultV2 we just named fees `performanceFee` and `managementFee`, and they were in the same unit.

MathisGD: oh ok i see what you mean

src/MorphoV2.sol
adhusson: One thing convincing me even more: people are used to inputting the amount they want on etherscan directly. But it's not a function that anybody will use through etherscan. Instead we can assume there is some frontend code that can do the computation before calling the function. 

What do you think ? To me it looks like putting offchain helpers onchain.

src/MorphoV2.sol
QGarchery, line 1: The computations look good to me given:
1. the formula `trading fee = (bonds-buyerAssets) * fee` in the description of the PR
2. the formula `buyerAssets = bonds * price`

For 2, it means that it will make it easier for a buyer to know the interest. For example, a price of 0.8 for a 1 year obligation means an APR of 25%. 
But I'm wondering if we don't want to make it easier for sellers as well. Take the same example, but also assume of 50% fee. Then a price of 0.8 doesn't mean that the seller will pay 25% interest if he is borrowing, but actually it will be 1 / (0.8 - 0.2 * 0.5) - 1 ~= 43%. So the idea would be to change 2 into `sellerAssets = bonds * price` when the offer is a sell offer

And it makes me wonder about 1 as well, what would happen if we swap buyerAssets with sellerAssets in 1

MathisGD: indeed I forgot 2 in the description of the PR (added now)

very good point. I think that offers should be able to specify both.

>And it makes me wonder about 1 as well, what would happen if we swap buyerAssets with sellerAssets in 1

this is arbitrary right, we could return it but it's fundamentally asymmetrical

QGarchery: > this is arbitrary right, we could return it but it's fundamentally asymmetrical

Yes good point. And we can go from one to another, so we can always choose the fee so that it's the same amount => this means we can choose the one that we want. Proof:
Let's write:
- $$tF1$$ the trading fee taken in scenario 1, meaning in case where it is defined as `trading fee = (bonds-buyerAssets) * fee`
- $$tF2$$ the trading fee taken in scenario 2, meaning in case where it is defined as `trading fee = (bonds-sellerAssets) * fee`
- $$bA$$ the `buyerAssets`. Note that $$bA$$ can be supposed the same in both scenario, it just means that the same price is taken in both scenarios (because of formula 2)
- $$sA1$$ and $$sA2$$ the `sellerAssets` in scenario 1 and in scenario 2

We have
(a) $$tF1 = (bonds - bA) * fee$$
(b) $$tF2 = (bonds - sA2) * fee$$
(c) $$bA - sA2 = tF2$$

So with (b) & (c) we get $$tF2 = (bonds - (bA - tF2)) * fee$$, and $$tF2 * (1 - fee) = (bonds - bA) * fee$$. That gives us $$tF2 = (bonds - bA) * fee / (1 - fee)$$. Looking at (a), that means that the effective fee in scenario 2 is $$fee / (1 - fee)$$, and $$f := x \rightarrow x / (1 - x)$$ is inversible on $$[0, 0.5] \rightarrow [0,1]$$, so we can go from one to another

peyha: Trading fees are usually paid by takers so it could make sense to apply it to the taker side (which could be both buyer and seller but it should usually be the seller)
Another idea could be to split the fee between the buyer and the seller by using
`trading fee = (bonds-buyerAssets) * fee * lambda + (bonds-sellerAssets) * fee * (1-lambda)` 
These ideas are more complex though so it would make sense to go for a simpler solution

src/MorphoV2.sol
peyha: Can't we use `takerAssets`/`makerAsssets` instead of buyer and seller assets? Takers will probably always specify their own side assets
But at the same time I wonder if there can be an use-case for a taker targeting a specific maker asset amount. If there is one, using buyer and seller asset parameters seems necessary because the relation between one and the other depends on the price (which depend on the timestamp of the block to inclusion so it cannot be computed beforehand)

peyha:
If we end up using two arguments `buyerAssets` and `sellerAssets`, it's probably worth returning both at the end of `take`, it could simplify integration

src/MorphoV2.sol
```diff
@@ -35,22 +35,57 @@ contract MorphoV2 is IMorphoV2 {
     /// otherwise one might not be takable anymore while an other one at the same nonce is still takeable.
     mapping(address user => mapping(uint256 nonce => uint256)) public consumed;
 
+    /// @dev Cut on interest at each trade.
+    mapping(address loanToken => uint256) public tradingFee;
+    mapping(address loanToken => address) public tradingFeeRecipient;
+
+    /// @dev Contract owner for administrative functions.
+    address public owner;
```
peyha, line 35: Would it make sense to have a `feeCurator` settable by the owner ?
If the owner is DAO-owned and it requires a DAO vote to change a fee, it might be difficult to handle new tokens/new chains efficiently (for a similar reason as why the curator sets fee parameters in vault V2 when it was the owner in vault V1). 
It can especially be interesting if we end up implementing more granularity on fees (e.g., higher fees on longer term obligations) because we'll have to change the fee on potential many bonds.

src/MorphoV2.sol
```diff
@@ -35,22 +35,57 @@ contract MorphoV2 is IMorphoV2 {
...
+    /// ADMIN FUNCTIONS ///
+
+    function setOwner(address newOwner) external {
+        require(msg.sender == owner, "Only owner");
+        owner = newOwner;
+    }
+
+    function setTradingFee(address loanToken, uint256 fee) external {
+        require(msg.sender == owner, "Only owner");
+        require(fee <= 1e18, "Fee too high");
```
peyha, line 52: should we implement a maximum fee ?
for instance, it was 25% on blue, 50% on vault-v2 (max perf fee)
(an issue can also be opened if we want to decide later)

src/MorphoV2.sol
```diff
@@ -75,11 +110,26 @@ contract MorphoV2 is IMorphoV2 {
...
+
+        if (buyerAssets > 0) {
+            obligationUnits = buyerAssets.mulDivDown(1e18, price);
+            sellerAssets =
+                buyerAssets - (obligationUnits - buyerAssets).mulDivDown(tradingFee[obligation.loanToken], 1e18);
+        } else if (sellerAssets > 0) {
+            buyerAssets = sellerAssets.mulDivDown(
+                1e18, 1e18 + tradingFee[obligation.loanToken] - tradingFee[obligation.loanToken].mulDivDown(1e18, price)
+            );
+            obligationUnits = buyerAssets.mulDivDown(1e18, price);
```
```suggestion
            obligationUnits = sellerAssets.mulDivDown(
                1e18, price.mulDivDown(1e18 + tradingFee[obligation.loanToken])-tradingFee[obligation.loanToken]
            );
            buyerAssets = obligationUnits.mulDivDown(price, 1e18);
```
looks a bit simpler

src/MorphoV2.sol
```diff
@@ -101,13 +151,16 @@ contract MorphoV2 is IMorphoV2 {
...
 
         if (buyerCallbackAddress != address(0)) {
-            ICallbacks(buyerCallbackAddress).onTake(obligation, buyer, assets, buyerCallbackData);
+            ICallbacks(buyerCallbackAddress).onTake(obligation, buyer, buyerAssets, buyerCallbackData);
         }
 
-        SafeTransferLib.safeTransferFrom(offer.loanToken, buyer, seller, assets);
+        SafeTransferLib.safeTransferFrom(
+            offer.loanToken, buyer, tradingFeeRecipient[offer.loanToken], buyerAssets - sellerAssets
+        );
```
peyha, line 151: Thinking about the best order to execute these transfers, seems pretty equivalent but placing it after the `sellerCallbackExecution` saves some gas if the callback reverts

src/MorphoV2.sol
MathisGD: we decided to keep both inputs, at least for now

src/MorphoV2.sol
MathisGD: we decided to take the price of the maker.
and the fees on the seller assets

peyha (commented):
confirming that I have the same formula for `sellerAssets`,`buyerAssets` and `obligationUnits`

peyha (commented):
The PR is correct for me but I prefer #137's shorter formula

src/MorphoV2.sol
MathisGD: I don't think that we care much about revert gas cost. we can open an issue if you feel that it's worth discussing

adhusson (approved):

peyha:
The "primitive" in the PR message should be
- trading fee = (bonds - sellerAssets) * fee
- makerAssets = bonds * price 
settled during [this discussion](https://morpholabs.slack.com/archives/C08A40KKN9W/p1759421741683889?thread_ts=1759421057.665429&cid=C08A40KKN9W)

peyha:
The justification of the current formula is that we have 
- `units * buyerPrice = buyerAssets`
- `units * sellerPrice = sellerAssets`
- `buyerPrice = sellerPrice (1 - fee) + fee`
- `sellerPrice = (buyerPrice - fee) / (1 - fee)`

cf [this](https://github.com/morpho-org/morpho-v2/pull/146#discussion_r2410993715)

and that we use `buyerAssets = sellerAssets * buyerPrice / sellerPrice` and `sellerAssets = buyerAssets * sellerPrice / buyerPrice` to mitigate rounding issues (all multiplications rounded down)

src/MorphoV2.sol
peyha, line 159: we could
```suggestion
        return (buyerAssets, sellerAssets, obligationUnits) 
    }
```
(but we can also raise an issue and do it later)

peyha (approved):

QGarchery (approved):
LGTM. Verified informally the properties in the description of this PR, and also `buyerPrice >= sellerPrice` (added in the verif wish list)

src/MorphoV2.sol
MathisGD: yes, will open an other PR!

PR #107, opened 2025-09-25
Prevent multiple seizure of the same collateral

QGarchery:
Rationale:
- there is no point in allowing multiple seizure of the same collateral, it can only make things more complicated to understand or allow some weird rounding manipulation
- there is no loss in UX with this version, because it does not require the list to be sorted

### DIFF
--- a/src/Terms.sol
+++ b/src/Terms.sol
@@ -11,6 +11,7 @@ import "./interfaces/ICallbacks.sol";
 
 contract Terms is ITerms {
     using MathLib for uint256;
+    using UtilsLib for uint256;
 
     /// CONSTANTS ///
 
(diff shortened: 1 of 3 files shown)


src/interfaces/ITerms.sol
```diff
@@ -38,6 +38,11 @@ struct Signature {
     bytes32 s;
 }
 
+struct Price {
+    uint248 price;
+    bool seen;
```
QGarchery: Sadly packing is needed because memory is not optimized for structs that have sizes that are not multiple of 32 bytes. See [here](https://github.com/morpho-org/morpho-sandbox/pull/43) for a proof

adhusson: Not clear to me, the memory layout pads everything to 32 bytes anyway so the packing will not happen in memory. [Example](https://docs.soliditylang.org/en/latest/internals/layout_in_memory.html#example-for-difference-in-struct-layout). (But I didn't understand your linked repo)

src/Terms.sol
```diff
@@ -158,14 +159,14 @@ contract Terms is ITerms {
         uint256 repayableDebt;
         uint256 maxDebt;
         bytes32 id = _id(term);
-        uint256[] memory prices = new uint256[](term.collaterals.length);
+        Price[] memory prices = new Price[](term.collaterals.length);
```
adhusson, line 161: Isn't it easier to force seizure collaterals to be sorted? As done [here](https://github.com/morpho-org/morpho-v2/blob/0a211688700efc441dbcb3b5e1a7ebce05ef94f9/src/Terms.sol#L154-L155) for instance (note the loop from the end so that collaterals can be sorted by increasing index, which is more natural than decreasing).

src/interfaces/ITerms.sol
QGarchery: I just realized that, my bad. Updated the corresponding PR to show that (previous proof was not wrong but it was not showing what you say)

src/Terms.sol
QGarchery: I mostly agree, it's simpler in the code, and lighter on the memory. But for the liquidator it's a bit more work (should be ok)

QGarchery:
See comments

PR #110, opened 2025-09-26
realize bad debt first 2

QGarchery:
Rationale is that the bad debt is really done first, and every computation after that is optional (by passing 0 length seizures) or is a no-op (assuming that 0 transfers are no-ops). I find it slightly easier to read

It should be functionally the same as #95:
- for the bad debt, it is the same computation, just moved up
- for the total repaid
  - if `originalDebt > repayableDebt` then there is some bad debt so the computation `totalRepaid = UtilsLib.min(totalRepaid, debtOf[borrower][id])` is the same as doing `totalRepaid = UtilsLib.min(totalRepaid, repayableDebt)`, and in the other PR `repayableDebt` stays the same even after the min in this case
  - otherwise, there is no bad debt, so the computation `totalRepaid = UtilsLib.min(totalRepaid, debtOf[borrower][id])` is the same as doing `totalRepaid = UtilsLib.min(totalRepaid, originalDebt)`, and in the other PR `repayableDebt` is equal to `originalDebt`

### DIFF
--- a/src/Terms.sol
+++ b/src/Terms.sol
@@ -171,19 +171,17 @@ contract Terms is ITerms {
             collateralOf[borrower][id][collateralToken] -= seizure.seizedAssets;
         }
 
-        // Realize bad debt
-        uint256 badDebt;
+        repayableDebt = UtilsLib.min(repayableDebt, originalDebt);
+        totalRepaid = UtilsLib.min(totalRepaid, repayableDebt);
 
-        if (totalRepaid > originalDebt) totalRepaid = originalDebt;
-        if (repayableDebt < originalDebt) {
-            // Because roundings are not aligned the effective bad debt is either the remaining debt or the original
-            // debt minus the theoretical repayable debt.
-            badDebt = UtilsLib.min(originalDebt - totalRepaid, originalDebt - repayableDebt);
-            totalBonds[id] -= badDebt;
+        // Realize bad debt.
+        if (originalDebt > repayableDebt) {
+            debtOf[borrower][id] = repayableDebt;
+            totalBonds[id] -= originalDebt - repayableDebt;
         }
 
         withdrawable[id] += totalRepaid;
-        debtOf[borrower][id] = originalDebt - totalRepaid - badDebt;
+        debtOf[borrower][id] -= totalRepaid;
 
         for (uint256 i = 0; i < seizures.length; i++) {
             seizure = seizures[i];
--- a/src/Terms.sol
+++ b/src/Terms.sol
@@ -150,6 +150,13 @@ contract Terms is ITerms {
         uint256 originalDebt = debtOf[borrower][id];
         require(originalDebt > maxDebt, "position is healthy");
 
+        // Realize bad debt.
+        if (originalDebt > repayableDebt) {
+            uint256 badDebt = originalDebt - repayableDebt;
+            debtOf[borrower][id] -= badDebt;
+            totalBonds[id] -= badDebt;
+        }
+
         uint256 totalRepaid;
         Seizure memory seizure;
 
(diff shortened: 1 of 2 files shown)


adhusson (approved):
Better than #95. Same note as there: a difference with https://github.com/morpho-org/morpho-v2/pull/62 (and main) : in 62, if rounding makes you repay more than repayableDebt, it will reduce your bad debt. Here it won't, the total repaid amount is at most repayableDebt. But I think it's fine.

Also ok with the 2 distant places where the debt is updated.

src/Terms.sol
```diff
@@ -195,19 +202,9 @@ contract Terms is ITerms {
...
-
-        if (totalRepaid > originalDebt) totalRepaid = originalDebt;
-        if (repayableDebt < originalDebt) {
-            // Because roundings are not aligned the effective bad debt is either the remaining debt or the original
-            // debt minus the theoretical repayable debt.
-            badDebt = UtilsLib.min(originalDebt - totalRepaid, originalDebt - repayableDebt);
-            totalBonds[id] -= badDebt;
-        }
-
+        totalRepaid = UtilsLib.min(totalRepaid, debtOf[borrower][id]);
```
peyha: isn't there an issue with this line ?
If a position is liquidatable without bad debt, I can seize 100% of all collaterals of the user and the equivalent `totalRepaid` will be `debtOf[borrower][id]` and the liquidation won't revert

peyha (commented):
Realizing the bad debt just after the health check is much cleaner imo

src/Terms.sol
QGarchery: Indeed, it can sort of bypass the liquidation incentive, by closing the position entirely. In fact it should probably revert if totalRepaid is greater than the remaining debt of the borrower

QGarchery: Note that this issue is there in #62 and #95

QGarchery: Just removing the line fixes the issue actually :D

adhusson: fixed in #62 [here](https://github.com/morpho-org/morpho-v2/pull/62/commits/34590c007bddf75db791e3c8155b75024c2e8e54)

adhusson: (pushed a merge)

* In this PR with the offending line removed `testTotalRepaidTooHigh` fails both when trying to seize all collateral and also when trying to repay all debt. 
* In #62 with this offending line removed, that is not the case: `testTotalRepaidTooHigh` fails when trying to seize all collateral (because the repaid debt ends up > the current debt, which creates an underflow), but not when trying to repay all debt.

The difference is caused #62 using excess the repaid amount to reduce the bad debt, whereas this PR computes the bad debt first. So not all the debt can be repaid.

I think at least one of the approaches should work when a position has some bad debt. What do you think about capping `totalRepaid` to the debt, but only when the bad debt is > 0 ?

Or same result require that `totalRepaid <= originalDebt`. With this, when there is bad debt, you can liquidate all the bonds (but not necessarily all the collateral, due to roundings).

src/Terms.sol
```diff
@@ -184,6 +184,13 @@ contract Terms is ITerms {
         uint256 originalDebt = debtOf[borrower][id];
         require(originalDebt > maxDebt, "position is healthy");
 
+        // Realize bad debt.
+        if (originalDebt > repayableDebt) {
+            uint256 badDebt = originalDebt - repayableDebt;
```
MathisGD, line 189: the change of behavior is basically that if there is a rounding with totalRepaid, the protocol realizes "too much" bad debt (but only 1 Wei). This seems fine, and more predictable actually (repayable debt and original debt are well defined).

MathisGD: >for the bad debt, it is the same computation, just moved up

@QGarchery this is not true right?

src/Terms.sol
MathisGD, line 1: >for the total repaid

this is not true anymore right?

src/Terms.sol
```diff
@@ -204,18 +211,8 @@ contract Terms is ITerms {
             collateralOf[borrower][id][collateralToken] -= seizure.seizedAssets;
         }
 
-        // Realize bad debt
-        uint256 badDebt;
-
-        if (repayableDebt < originalDebt) {
-            // Because roundings are not aligned the effective bad debt is either the remaining debt or the original
-            // debt minus the theoretical repayable debt.
-            badDebt = UtilsLib.min(originalDebt - totalRepaid, originalDebt - repayableDebt);
```
MathisGD, line 213: this had the good effect that you can always seize all the collat if there is some bad debt created (what we expect, high level). maybe a reason not to merge this?

MathisGD: basically what @adhusson is saying [here](https://github.com/morpho-org/morpho-v2/pull/110#discussion_r2382733256)

QGarchery: Note that after #119, you can always seize the whole collateral if you pass the collateral of the borrower, because in that case we have the property `totalRepaid <= repayableDebt`. This property is always true (even when passing `repaidBonds` as input) after #120, so this PR would not change the behavior actually

QGarchery:
closing in favor of #119

PR #117, opened 2025-09-29
prevent crossing and refactor computations

MathisGD:


peyha (commented):
~~A lender can't repay a loan and lend in the same transaction anymore ? (similarly a borrower cannot withdraw and borrow)~~(it's said as a dev comment)
in term of clarity the previous version was slightly difficult to understand at first, the new one seems easier to understand (but I'm not a huge fan of nested conditions)

src/Terms.sol
```diff
@@ -92,26 +92,22 @@ contract Terms is ITerms {
...
-                debtOf[buyer][id] -= bondsUnits;
-                bondSharesOf[seller][id] -= bondsShares;
-                totalShares[id] -= bondsShares;
-                totalBonds[id] -= bondsUnits;
-            }
+            debtOf[buyer][id] -= bondsUnits;
+            bondSharesOf[seller][id] -= bondsShares;
+            totalShares[id] -= bondsShares;
+            totalBonds[id] -= bondsUnits;
         }
```
adhusson, line 111: worth these comments or similar

```suggestion
        // buyer gains shares, seller gains debt
        if (debtOf[buyer][id] == 0 && bondSharesOf[seller][id] == 0) {
            bondSharesOf[buyer][id] += bondsShares;
            debtOf[seller][id] += bondsUnits;
            totalShares[id] += bondsShares;
            totalBonds[id] += bondsUnits;
        // buyer gains shares, seller loses shares
        } else if (debtOf[buyer][id] == 0 && bondSharesOf[seller][id] > 0) {
            bondSharesOf[buyer][id] += bondsShares;
            bondSharesOf[seller][id] -= bondsShares;
        // buyer loses debt, seller gains debt
        } else if (debtOf[buyer][id] > 0 && bondSharesOf[seller][id] == 0) {
            debtOf[buyer][id] -= bondsUnits;
            debtOf[seller][id] += bondsUnits;
        // buyer loses debt, seller loses shares
        } else {
            debtOf[buyer][id] -= bondsUnits;
            bondSharesOf[seller][id] -= bondsShares;
            totalShares[id] -= bondsShares;
            totalBonds[id] -= bondsUnits;
}
```

peyha:
I'm starting to get convinced that crossing should be prevented. 
In terms of functionality, we do not lose anything because users can still split a crossing transaction into a repaying & lend transaction (resp. withdraw & borrow transaction) but it makes accounting easier (especially for the interest cut fee when we will have to compute the cashflow of lenders). Maybe we should compare an implementation of the interest cut fee with and without crossing to have a clear view on the size of the simplification

PR #119, opened 2025-09-29
Roundings in liquidate

QGarchery:
What this PR does:
- round up consistently in the computation of `repayableDebt`. The rationale is that rounding up or down doesn't give more guarantees that it will not realize multiple dust bad debts. But rounding up ensures that the computed `totalRepaid`, when passing `seizedAssets` as input, will be lower than `repayableDebt`
- move up the computation of the bad debt. This makes so, if there is bad debt, the `totalRepaid` cannot be more than the `repayableDebt`. This is because of the computation `debtOf[borrower][id] -= totalRepaid` and when there is bad debt `debtOf[borrower][id]` is equal to `repayableDebt`

### DIFF
--- a/src/Terms.sol
+++ b/src/Terms.sol
@@ -171,19 +171,17 @@ contract Terms is ITerms {
             collateralOf[borrower][id][collateralToken] -= seizure.seizedAssets;
         }
 
-        // Realize bad debt
-        uint256 badDebt;
+        repayableDebt = UtilsLib.min(repayableDebt, originalDebt);
+        totalRepaid = UtilsLib.min(totalRepaid, repayableDebt);
 
-        if (totalRepaid > originalDebt) totalRepaid = originalDebt;
-        if (repayableDebt < originalDebt) {
-            // Because roundings are not aligned the effective bad debt is either the remaining debt or the original
-            // debt minus the theoretical repayable debt.
-            badDebt = UtilsLib.min(originalDebt - totalRepaid, originalDebt - repayableDebt);
-            totalBonds[id] -= badDebt;
+        // Realize bad debt.
+        if (originalDebt > repayableDebt) {
+            debtOf[borrower][id] = repayableDebt;
+            totalBonds[id] -= originalDebt - repayableDebt;
         }
 
         withdrawable[id] += totalRepaid;
-        debtOf[borrower][id] = originalDebt - totalRepaid - badDebt;
+        debtOf[borrower][id] -= totalRepaid;
 
         for (uint256 i = 0; i < seizures.length; i++) {
             seizure = seizures[i];
--- a/src/Terms.sol
+++ b/src/Terms.sol
@@ -174,10 +174,13 @@ contract Terms is ITerms {
             prices[i] = IOracle(term.collaterals[i].oracle).price();
             {
                 address collateralToken = term.collaterals[i].token;
-                uint256 collateralQuoted =
-                    collateralOf[borrower][id][collateralToken].mulDivDown(prices[i], ORACLE_PRICE_SCALE);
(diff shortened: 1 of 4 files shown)


src/Terms.sol
```diff
@@ -174,10 +174,13 @@ contract Terms is ITerms {
...
                 address collateralToken = term.collaterals[i].token;
-                uint256 collateralQuoted =
-                    collateralOf[borrower][id][collateralToken].mulDivDown(prices[i], ORACLE_PRICE_SCALE);
-                maxDebt += collateralQuoted.mulDivDown(term.collaterals[i].lltv, 1e18);
-                repayableDebt += collateralQuoted.mulDivUp(1e18, LIQUIDATION_INCENTIVE_FACTOR);
+                uint256 collateralAmount = collateralOf[borrower][id][collateralToken];
+                maxDebt += collateralAmount.mulDivDown(prices[i], ORACLE_PRICE_SCALE).mulDivDown(
+                    term.collaterals[i].lltv, 1e18
+                );
+                repayableDebt += collateralAmount.mulDivUp(prices[i], ORACLE_PRICE_SCALE).mulDivDown(
```
```suggestion
                repayableDebt += collateralAmount.mulDivUp(prices[i], ORACLE_PRICE_SCALE).mulDivUp(
```

test/LiquidationTest.sol
```diff
@@ -122,7 +122,8 @@ contract LiquidationTest is BaseTest {
         assertEq(recordedData, data, "data");
     }
 
-    // Check that it is not always possible to seize all assets due to roundings.
+    // Check that due to roundings, even if there is bad debt it is not always possible to seize all assets or repay all
+    // debt.
```
adhusson: Shouldn't the test check that when there is bad debt, it is always possible to seize all asset?

src/Terms.sol
```diff
@@ -174,16 +174,26 @@ contract Terms is ITerms {
             prices[i] = IOracle(term.collaterals[i].oracle).price();
             {
                 address collateralToken = term.collaterals[i].token;
-                uint256 collateralQuoted =
-                    collateralOf[borrower][id][collateralToken].mulDivDown(prices[i], ORACLE_PRICE_SCALE);
-                maxDebt += collateralQuoted.mulDivDown(term.collaterals[i].lltv, 1e18);
-                repayableDebt += collateralQuoted.mulDivUp(1e18, LIQUIDATION_INCENTIVE_FACTOR);
+                uint256 collateralAmount = collateralOf[borrower][id][collateralToken];
```
```suggestion
                uint256 collateralAmount = collateralOf[borrower][id][term.collaterals[i].token];
```

src/Terms.sol
```diff
@@ -174,16 +174,26 @@ contract Terms is ITerms {
...
 
         uint256 originalDebt = debtOf[borrower][id];
         require(originalDebt > maxDebt, "position is healthy");
 
+        // Realize bad debt.
+        if (originalDebt > repayableDebt) {
+            uint256 badDebt = originalDebt - repayableDebt;
+            debtOf[borrower][id] -= badDebt;
+            totalBonds[id] -= badDebt;
+        }
```
MathisGD: we will need the variable later for the event
```suggestion
        uint256 badDebt = originalDebt > repayableDebt ? originalDebt - repayableDebt : 0;
        if (badDebt > 0) {
            debtOf[borrower][id] -= badDebt;
            totalBonds[id] -= badDebt;
        }
```

MathisGD (commented):
agree with the rationale

MathisGD (approved):

adhusson (approved):

peyha (approved):
nice, being able to seize 100% of the collateral if `originalDebt >repayableDebt` is a good feature 

under this condition, the liquidator is incentivized to seize all the collaterals to maximize the profit

PR #120, opened 2025-09-29
Cap current repaid bonds to repayable debt

QGarchery:
Important point: we get the property that, for a given seizure with `repaidBonds` as input, the corresponding amount repaid will be less than the contribution of that collateral to the `repayableDebt`. In other words, `repayableDebt` starts to be aptly named


QGarchery:
Closing as it's probably not needed to be that precise, we can instead cap `totalRepaid` directly to be less than `repayableDebt`

PR #121, opened 2025-09-29
[WIP] seize all or repay all

QGarchery:
It's a refactor of #120 where we now assume that the liquidator only passes inputs for the assets to seize, or only passes the inputs for the assets to repay. The point is code/behavior simplicity, and this assumption seems fine to make because the liquidator is mostly interested in either:
- knowing easily `totalRepaid` in advance, in which case passing only repaid amounts makes sense
- knowing easily each collateral amount in advance, in which case passing only seized amounts makes sense

Note:
- this PR allows to easily cap current repayable debts, without having to recompute it in a second loop
- this PR abandons the idea of passing an array that has smaller length than the collaterals of the bond


QGarchery:
Closing as it doesn't simplify the code that much

PR #122, opened 2025-09-29
chore: license update

MathisGD:
[context](https://www.notion.so/morpho-labs/Morpho-V2-License-Strategy-208d69939e6d80479689e761824baa9f)

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -15,11 +15,8 @@ Licensed Work:        Morpho V2 Core
 Additional Use Grant: Any uses listed and defined at
                       morpho-v2-core-license-grants.morpho.eth
 
-Change Date:          The earlier of (i) 2028-01-01, or (ii) a date specified
-                      at morpho-v2-core-license-date.morpho.eth, or (iii) upon
-                      the activation of the setFee function of the Licensed 
-                      Work’s applicable protocol smart contracts deployed for
-                      production use.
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -15,11 +15,8 @@ Licensed Work:        Morpho V2 Core
 Additional Use Grant: Any uses listed and defined at
                       morpho-v2-core-license-grants.morpho.eth
 
-Change Date:          The earlier of (i) 2028-01-01, or (ii) a date specified
-                      at morpho-v2-core-license-date.morpho.eth, or (iii) upon
-                      the activation of the setFee function of the Licensed 
-                      Work’s applicable protocol smart contracts deployed for
-                      production use.
+Change Date:          The earlier of 2029-10-01, or a date specified at 
+                      morpho-v2-core-license-date.morpho.eth.
```
MathisGD, line 19: we could update the date later as the rule is "the fourth anniversary of the first publicly available distribution". though it's not important

qdrn (approved):

LICENSE
MattLsb: couldn't we put 'the fourth anniversary of the first publicly available distribution' directly instead of a precise date? so we don't have to update if there is some delay

MathisGD: according to chat you can't do that

MattLsb (approved):

PR #124, opened 2025-09-30
refactor: minor changes

MathisGD:


test/TakeTest.sol
peyha, line 1: should we also implement a non-reverting test with lendOffer.expiry = lendOffer.start ?

src/Terms.sol
```diff
@@ -56,23 +56,20 @@ contract Terms is ITerms {
...
-
-        uint256 offerDuration = offer.expiry - offer.start;
-        uint256 price = offerDuration > 0
-            ? offer.startPrice + (offer.expiryPrice - offer.startPrice) * (block.timestamp - offer.start) / offerDuration
+        address buyer = offer.buy ? offer.offering : taker;
+        address buyerCallbackAddress = offer.buy ? offer.callbackAddress : takerCallbackAddress;
+        bytes memory buyerCallbackData = offer.buy ? offer.callbackData : takerCallbackData;
+        address seller = offer.buy ? taker : offer.offering;
+        address sellerCallbackAddress = offer.buy ? takerCallbackAddress : offer.callbackAddress;
+        bytes memory sellerCallbackData = offer.buy ? takerCallbackData : offer.callbackData;
```
QGarchery: I found this bit clearer before, because it lead the reader into thinking about the disjonction once and checking each variable after that "Ok if offer.buy, then the buyer is in the offer, same as for the callback. But in the other case, the buyer is the taker, same as the callback"

QGarchery (approved):

src/Terms.sol
MathisGD: curious about what the others think

peyha: I usually find sextuple hard to read but in this case it's ok because it factorizes some code and the order is clear so I would agree with Quentin

adhusson: For me it is less clear. The condition occurs only once in the tuple and you can see what's happening. With the conditions I have to check every one to make sure there isn't one different from the others.

MathisGD: ok, rolling back

QGarchery (approved):

adhusson (approved):

PR #128, opened 2025-09-30
chore: abandon legacy pipeline

MathisGD:
it's painful maintenance and now via-ir is quite fast.


peyha (approved):
semi duplicate with #124 but no big deal

QGarchery (approved):

adhusson (approved):

PR #129, opened 2025-09-30
feat: swap price/lif mul order

peyha:
Before, all debt computation were doing `collateral*price/LIF` (resp `collateral*price*lltv` for max debt). The new order is `collateral/LIF*price` (resp `collateral*lltv*price`).


QGarchery (approved):
It looks cleaner because then LIF and LLTV are always applied to an amount of collateral directly

peyha:
it's also consistent with Blue's implementation

src/MorphoV2.sol
MathisGD, line 1: why exactly is it better? one could also argue the opposite, that it's weird to apply the LIF/LLTV before converting prices

QGarchery: Note that it was not consitent before:
```solidity
            if (seizure.seizedAssets > 0) {
                seizure.repaidBonds = seizure.seizedAssets.mulDivUp(prices[seizure.collateralIndex], ORACLE_PRICE_SCALE)
                    .mulDivUp(1e18, LIQUIDATION_INCENTIVE_FACTOR);
            } else {
                seizure.seizedAssets = seizure.repaidBonds.mulDivDown(LIQUIDATION_INCENTIVE_FACTOR, 1e18).mulDivDown(
                    ORACLE_PRICE_SCALE, prices[seizure.collateralIndex]
                );
```
So the biggest change in this PR is to choose one option consistently. Then to choose between the 2, I find it more natural to say that LLTV is as if the collateral was multiplied by that factor to consider the collateralization, rather than to say that the collateral *value* is multiplied by that factor, because it introduces the oracle in that definition for no reason (btw what is *value* in this case ? Is it rounded up or down ?)

peyha: It was consistent before no ? The LIF is always multiplied against a bond amount

QGarchery: You're right, it was consistent before. So it's just about choosing to apply it to bonds or to collateral amount

QGarchery: I think I still prefer collateral amount: it's as if LLTV reduces your collateral when it comes to collateralization, and as if LIF increases the collateral when you are thinking about the seized part in the liquidation

peyha: Also slightly prefer collateral amount, collaterals usually have more decimals than loan token (e.g., stablecoins with 6 decimals) so the multiplication should be slightly more accurate with the collateral

MathisGD (approved):
convinced by the fact that "intermediate variables" makes more sense (see the thread)

adhusson (commented):
For lltv it makes less sense: the ltv is a loan to collateral value ratio. You can write it as `ltv = debt / (value of collateral)`. So `(value of collateral) * lltv = max debt` makes more sense than `value of (collateral * lltv) = max debt`.

For lif I don't know if it makes more sense one way or the other.

peyha:
For the LIF, I would say it depends if we consider this bonus to be an additional proportion of collateral or an additional proportion of corresponding debt, I think both version would be fine but I'm leaning for the collateral because it seems cleaner to me

peyha:
For the LLTV, I'm starting to get convinced by Adrien's point

QGarchery:
> You can write it as ltv = debt / (value of collateral). So (value of collateral) * lltv = max debt makes more sense than value of (collateral * lltv) = max debt.

Being the devil's advocate but you could also define `ltv = value of debt / collateral`, in which case you could write `collateral * lltv = max debt value`. Although I agree that we didn't write max debt this way, and that the price oracle is in the other direction.

Ok with both in the end, even though I still prefer merging this PR because, as @MathisGD said, in the computation `seizure.seizedAssets = seizure.repaidBonds.mulDivDown(LIQUIDATION_INCENTIVE_FACTOR, 1e18).mulDivDown(  ORACLE_PRICE_SCALE, prices[seizure.collateralIndex]);` the intermediate computation of `seizure.repaidBonds.mulDivDown(LIQUIDATION_INCENTIVE_FACTOR, 1e18)` make less sense to me. Whereas if we did `seizure.seizedAssets = seizure.repaidBonds.mulDivDown(ORACLE_PRICE_SCALE, prices[seizure.collateralIndex]).mulDivDown(LIQUIDATION_INCENTIVE_FACTOR, 1e18);` it's clear to me that the repaid amount is converted to a collateral to seize, and some extra asset can be seized

MathisGD:
what about making the change for the LIF but not for the LLTV?

MathisGD (approved):

adhusson (approved):

QGarchery (approved):

PR #131, opened 2025-10-01
refactor: oco first and merkle tree

MathisGD:
The current offer structure is not fully satisfying for these reasons:
- there is an overlap between them and OCO. And in particular OCO can do everything the offers can do.
- forcing a lender to be ok with any subset of collateral / a borrower to be ok with any superset (same for LLTV etc.) can be detrimental.
- we need things such as #45 for the secondary.

So we switch to "OCO only". To avoid the maker to make a lot of signatures, we allow to sign a merkle root. The mempool will use some form of compression to avoid the maker to publish a ton of data.

### DIFF
--- a/src/MorphoV2.sol
+++ b/src/MorphoV2.sol
@@ -15,9 +15,7 @@ contract MorphoV2 is IMorphoV2 {
     /// CONSTANTS ///
 
     bytes32 public constant DOMAIN_TYPEHASH = keccak256("EIP712Domain(uint256 chainId,address verifyingContract)");
-    bytes32 public constant OFFER_TYPEHASH = keccak256(
-        "Offer(bool lend,address offering,uint256 assets,address loanToken,Collateral[] collaterals,uint256 maturity,uint256 start,uint256 expiry,uint256 startPrice,uint256 expiryPrice,uint256 nonce)"
-    );
+    bytes32 public constant ROOT_TYPEHASH = keccak256("bytes32 root");
     uint256 public constant ORACLE_PRICE_SCALE = 1e36;
     uint256 public constant LIQUIDATION_INCENTIVE_FACTOR = 1.15e18;
 
@@ -41,24 +39,23 @@ contract MorphoV2 is IMorphoV2 {
     /// @dev If one wants to match two offers without taking a position, they can batch take them and not have a
     /// position at the end.
     function take(
-        Obligation memory obligation,
         uint256 assets,
         uint256 obligationUnits,
         address taker,
         Offer memory offer,
         Signature memory sig,
+        bytes32 root,
+        bytes32[] memory proof,
         address takerCallbackAddress,
         bytes memory takerCallbackData
     ) public {
         require(assets == 0 || obligationUnits == 0, "inconsistent input");
         require(block.timestamp >= offer.start, "offer not started");
         require(block.timestamp <= offer.expiry, "offer expired");
-        require(obligation.maturity >= block.timestamp, "maturity");
-        require(offer.loanToken == obligation.loanToken, "Loan tokens do not match");
-        require(offer.maturity == obligation.maturity, "Maturities do not match");
+        require(offer.obligation.maturity >= block.timestamp, "maturity");
         require(offer.start < offer.expiry || offer.expiryPrice == offer.startPrice, "inconsistent prices");
-        require(signatureIsValid(offer, sig), "Invalid signature");
-        _checkCollateralInclusion(obligation, offer);
+        require(signer(root, sig) == offer.offering, "invalid signature");
+        require(MathLib.isLeaf(root, keccak256(abi.encode(offer)), proof), "invalid proof");
 
         (
             address buyer,
@@ -81,7 +78,7 @@ contract MorphoV2 is IMorphoV2 {
 
         require((consumed[offer.offering][offer.nonce] += assets) <= offer.assets, "consumed");
 
-        bytes32 id = _id(obligation);
+        bytes32 id = _id(offer.obligation);
 
         uint256 repaid = UtilsLib.min(debtOf[buyer][id], obligationUnits);
         uint256 bought = obligationUnits - repaid;
@@ -101,16 +98,16 @@ contract MorphoV2 is IMorphoV2 {
         totalUnits[id] -= withdrawn;
 
         if (buyerCallbackAddress != address(0)) {
-            ICallbacks(buyerCallbackAddress).onTake(obligation, buyer, assets, buyerCallbackData);
+            ICallbacks(buyerCallbackAddress).onTake(offer.obligation, buyer, assets, buyerCallbackData);
         }
 
-        SafeTransferLib.safeTransferFrom(offer.loanToken, buyer, seller, assets);
+        SafeTransferLib.safeTransferFrom(offer.obligation.loanToken, buyer, seller, assets);
 
         if (sellerCallbackAddress != address(0)) {
-            ICallbacks(sellerCallbackAddress).onTake(obligation, seller, assets, sellerCallbackData);
+            ICallbacks(sellerCallbackAddress).onTake(offer.obligation, seller, assets, sellerCallbackData);
         }
 
-        require(_isHealthy(obligation, seller), "Seller is unhealthy");
+        require(_isHealthy(offer.obligation, seller), "Seller is unhealthy");
     }
 
     /// @dev Will revert if there is no withdrawable funds.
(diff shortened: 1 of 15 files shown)


src/MorphoV2.sol
```diff
@@ -15,9 +15,7 @@ contract MorphoV2 is IMorphoV2 {
     /// CONSTANTS ///
 
     bytes32 public constant DOMAIN_TYPEHASH = keccak256("EIP712Domain(uint256 chainId,address verifyingContract)");
-    bytes32 public constant OFFER_TYPEHASH = keccak256(
-        "Offer(bool lend,address maker,uint256 assets,address loanToken,Collateral[] collaterals,uint256 maturity,uint256 start,uint256 expiry,uint256 startPrice,uint256 expiryPrice,uint256 nonce)"
-    );
+    bytes32 public constant ROOT_TYPEHASH = keccak256("bytes32 root");
```
QGarchery: EIP 712 defines structured data as data that is contained in a struct type.

> `encode(domainSeparator : 𝔹²⁵⁶, message : 𝕊) = "\x19\x01" ‖ domainSeparator ‖ hashStruct(message)`

> `hashStruct(s : 𝕊) = keccak256(typeHash ‖ encodeData(s))`

>  The `typeHash` is a constant for a given struct type

Here the signed data is unstructured, so I think we should use the second option:

> `encode(message : 𝔹⁸ⁿ) = "\x19Ethereum Signed Message:\n" ‖ len(message) ‖ message`

Another option is to create a wrapper struct, and keep using the same signing method

QGarchery: That also makes me think that one nice property that we lose is that signing is clear: before this PR the wallet could show you what you are signing

src/MorphoV2.sol
```diff
@@ -238,28 +235,11 @@ contract MorphoV2 is IMorphoV2 {
...
-
-    function signatureIsValid(Offer memory offer, Signature memory signature) internal view returns (bool) {
-        bytes32 hashStruct = keccak256(abi.encode(OFFER_TYPEHASH, offer));
+    function signer(bytes32 root, Signature memory signature) internal view returns (address) {
+        bytes32 hashStruct = keccak256(abi.encode(ROOT_TYPEHASH, root));
         bytes32 domainSeparator = keccak256(abi.encode(DOMAIN_TYPEHASH, block.chainid, address(this)));
         bytes32 digest = keccak256(bytes.concat("\x19\x01", domainSeparator, hashStruct));
-        address signatory = ecrecover(digest, signature.v, signature.r, signature.s);
-        return signatory != address(0) && offer.maker == signatory;
+        return ecrecover(digest, signature.v, signature.r, signature.s);
```
QGarchery: I think it would be cleaner to require that the return value isn't address(0) because:
- it prevents any signature that is going through without knowing the private key associated
- even though in only this context it's probably fine (because it only allows to manipulate address(0) as the maker), if this function is used in other contexts then it can be misused

src/libraries/MathLib.sol
```diff
@@ -19,4 +19,19 @@ library MathLib {
             z := mul(gt(x, y), sub(x, y))
         }
     }
+
+    /// @dev Returns true if hash(hash(leafHash, proof[0]), proof[1])... = root (hash sorts the inputs
+    /// lexicographically).
+    function isLeaf(bytes32 root, bytes32 leafHash, bytes32[] memory proof) internal pure returns (bool) {
```
QGarchery, line 25: Showing that it's equivalent to the OZ implementation in this PR: #141

src/MorphoV2.sol
MathisGD: see slack threads:
- https://morpholabs.slack.com/archives/C02N7CZ088N/p1759745201337129
- https://morpholabs.slack.com/archives/C02N7CZ088N/p1759745295533129

src/MorphoV2.sol
```diff
@@ -35,24 +33,23 @@ contract MorphoV2 is IMorphoV2 {
...
         require(block.timestamp >= offer.start, "offer not started");
         require(block.timestamp <= offer.expiry, "offer expired");
-        require(obligation.maturity >= block.timestamp, "maturity");
-        require(offer.loanToken == obligation.loanToken, "Loan tokens do not match");
-        require(offer.maturity == obligation.maturity, "Maturities do not match");
+        require(offer.obligation.maturity >= block.timestamp, "maturity");
         require(offer.start < offer.expiry || offer.expiryPrice == offer.startPrice, "inconsistent prices");
-        require(signatureIsValid(offer, sig), "Invalid signature");
-        _checkCollateralInclusion(obligation, offer);
+        require(signer(root, sig) == offer.maker, "invalid signature");
```
peyha, line 52: how would this work with a smart contract such as a vault v2? the vault would authorize an EOA signer that would sign the merkle tree root on its behalf ?

MathisGD: I would keep this discussion in the issue for vaultV2 connexion (which is unsolved). Nothing fundamental changes in the PR regarding this.

src/libraries/MathLib.sol
```diff
@@ -19,4 +19,19 @@ library MathLib {
...
+        bytes32 currentHash = leafHash;
+        for (uint256 i = 0; i < proof.length; i++) {
+            currentHash = keccak256(sort(currentHash, proof[i]));
```
QGarchery, line 28: Probably ok to not write this, but nice to have in mind: you should not hash values of 64 bytes. It's ok here because offers are way more than 64 bytes.
This requirement is written [here](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/765b288a6a273f8c267adcd409094212dbd5e3cb/contracts/utils/cryptography/MerkleProof.sol#L16-L21), and more details in [this page](https://www.notion.so/morpho-labs/Merkle-tree-hash-collision-1548e78093d044418c3e13e5fd627243#1548e78093d044418c3e13e5fd627243). Note how the problem is not only interpreting a node as leaf value but also interpreting a leaf value as a node (it's actually a bigger issue)

src/MorphoV2.sol
```diff
@@ -95,16 +92,16 @@ contract MorphoV2 is IMorphoV2 {
         totalUnits[id] -= withdrawn;
 
         if (buyerCallbackAddress != address(0)) {
-            ICallbacks(buyerCallbackAddress).onTake(obligation, buyer, assets, buyerCallbackData);
+            ICallbacks(buyerCallbackAddress).onTake(offer.obligation, buyer, assets, buyerCallbackData);
```
peyha, line 96: just wondering, wouldn't it be worth passing either `price` or `obligationUnits` to callbacks ? might be useful for complex DeFi interactions

test/BaseTest.sol
```diff
@@ -49,8 +49,31 @@ abstract contract BaseTest is Test {
...
+    }
+
+    // assumes the offer is the first one!
+    function proof(Offer[2] memory offers) internal pure returns (bytes32[] memory) {
+        bytes32[] memory proof = new bytes32[](1);
+        proof[0] = keccak256(abi.encode(offers[1]));
+        return proof;
+    }
+
+    function sortHash(bytes32 h1, bytes32 h2) internal pure returns (bytes32) {
```
peyha: why not re-use MathLib.sort ?

peyha (commented):
lgtm, will approve once my comments are resolved

src/MorphoV2.sol
peyha: all right, was wondering to what extent this feature introduced change but it's the same

src/libraries/MathLib.sol
MathisGD: very interesting

src/MorphoV2.sol
MathisGD: see #144

peyha (approved):

src/interfaces/IMorphoV2.sol
```diff
@@ -3,6 +3,7 @@
 pragma solidity ^0.8.0;
 
 struct Obligation {
+    uint256 chainId;
```
peyha, line 6: We could use `uint64` for chainId to pack it with `loanToken` but we have no guarantee on the bound of `chainId` as per [EIP-155](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md) even though execution clients usually store it as `u64`

MathisGD: calldata is never packed anyway

QGarchery (approved):

src/libraries/MathLib.sol
```diff
@@ -19,4 +19,19 @@ library MathLib {
             z := mul(gt(x, y), sub(x, y))
         }
     }
+
+    /// @dev Returns true if hash(hash(leafHash, proof[0]), proof[1])... = root (hash sorts the inputs
```
adhusson:
```suggestion
    /// @dev Returns true if hash( ... hash(hash(leafHash, proof[0]), proof[1])..., ...) = root (hash sorts the inputs
```

adhusson (approved):

PR #116, opened 2025-09-29
feat: shares as input of take

MathisGD:
fixes #114

PR #133, opened 2025-10-01
[Certora] lender or borrower

### LINKED ISSUE #73: Invariant debt = 0 or bonds = 0 not maintained
If a user self-lends he will increase both debt and bonds, it might be better to enforce the invariant.

MathisGD:
fixes #73


adhusson:
> fixes #73 The rule is vacuous for take(), but I don't get why

I don't understand either. fyi like this it is also vacuous for take (I think it's because all executions of take revert):

```
rule notBorrowerAndLenderRule(bytes32 id, address user, env e, method f, calldataarg args)
{
    require sharesOf(user, id) == 0 || debtOf(user, id) == 0;
    f(e, args);
    assert false;
}
```

src/MorphoV2.sol
```diff
@@ -125,10 +126,11 @@ contract MorphoV2 is IMorphoV2 {
...
         bytes32 id = _id(offer.obligation);
 
-        uint256 sellerSharesDecrease =
-            UtilsLib.min(obligationUnits.mulDivDown(totalShares[id] + 1, totalUnits[id] + 1), sharesOf[seller][id]);
+        uint256 obligationShares = obligationUnits.mulDivUp(totalShares[id] + 1, totalUnits[id] + 1);
+
+        uint256 sellerSharesDecrease = UtilsLib.min(obligationShares, sharesOf[seller][id]);
         uint256 sellerDebtIncrease =
-            obligationUnits - sellerSharesDecrease.mulDivDown(totalUnits[id] + 1, totalShares[id] + 1);
+            (obligationShares - sellerSharesDecrease).mulDivUp(totalUnits[id] + 1, totalShares[id] + 1);
```
MathisGD: is it ok to do that though? see #116 (let's merge it first)

certora/MorphoV2.spec
```diff
@@ -1,17 +1,54 @@
...
+persistent ghost mapping(bytes32 => mathint) sumDebtOf {
+    init_state axiom (forall bytes32 id. sumDebtOf[id] == 0);
+}
+
+hook Sstore debtOf[KEY address owner][KEY bytes32 id] uint256 newDebt (uint256 oldDebt) {
+    sumDebtOf[id] = sumDebtOf[id] - oldDebt + newDebt;
 }
 
-/// SANITY ///
+rule takeInputs(env e, uint256 buyerAssets, uint256 sellerAssets, uint256 obligationUnits, uint256 obligationShares, address taker, MorphoV2.Offer offer, MorphoV2.Signature signature, bytes32 root, bytes32[] proof, address takerCallbackAddress, bytes takerCallbackData) {
```
adhusson: take outputs?

adhusson (approved):

certora/MorphoV2.spec
MathisGD:
```suggestion
rule takeInputOutputConsistency(env e, uint256 buyerAssets, uint256 sellerAssets, uint256 obligationUnits, uint256 obligationShares, address taker, MorphoV2.Offer offer, MorphoV2.Signature signature, bytes32 root, bytes32[] proof, address takerCallbackAddress, bytes takerCallbackData) {
```

MathisGD:
changes are no longer needed after #159

QGarchery (approved):

Issue #94, opened 2025-09-18
Gasless takes


adhusson:
Should takers be able to gaslessly take? It would require a `takeSignature` argument so that takers can sign a take and someone else can fill. This is always possible in symmetric implementations of `take`, meaning where take has 2 `Offer` arguments. For instance [here](https://github.com/morpho-org/morpho-v2/blob/feat/no-callback-nesting/src/Terms.sol#L37). 

But we can separately consider gasless takes even if the take function is asymmetric.

If yes then
- The `Term` should be inside the take/order struct.
- The take/order struct should specify if it's buying or selling
- It should also have nonce/start/expiry

So it will be like an offer, except no matching and its assets/bonds field means something else.

MathisGD:
https://morpholabs.slack.com/archives/C08A40KKN9W/p1772388909510559

PR #138, opened 2025-10-04
feat: authorizations

### LINKED ISSUE #25: authorization system
There is an authorization issue, because how do we know that `onBehalf` is ok to `take` here ? Before there was a signature that was required for both parties, but now, for example, `onBehalf` could be the buyer and the sender could be the seller that crafts a worthless bond

_Originally posted by @QGarchery in https://github.com/morpho-org/v2/pull/16#discussion_r2179343913_

adhusson:


src/interfaces/IMorphoV2.sol
```diff
@@ -37,6 +37,14 @@ struct Signature {
     bytes32 s;
 }
 
+struct Authorization {
+    address authorizer;
+    address authorizee;
+    bool authorized;
```
adhusson, line 42: Compared to Blue:

`authorized` -> `authorizee`
`isAuthorized` -> `authorized`

src/MorphoV2.sol
```diff
@@ -57,7 +63,13 @@ contract MorphoV2 is IMorphoV2 {
         require(offer.loanToken == obligation.loanToken, "Loan tokens do not match");
         require(offer.maturity == obligation.maturity, "Maturities do not match");
         require(offer.start < offer.expiry || offer.expiryPrice == offer.startPrice, "inconsistent prices");
-        require(signatureIsValid(offer, sig), "Invalid signature");
+        require(msg.sender == taker || authorized[taker][msg.sender], "order not authorized");
+        require(
+            msg.sender == offer.maker
+                || (sig.v != 0 && _signatureIsValid(abi.encode(OFFER_TYPEHASH, offer), sig, offer.maker))
```
adhusson: `sig.v != 0` shortcircuits to avoid wasting gas

src/interfaces/IMorphoV2.sol
peyha: agree that `authorizee` is better than `authorized` however I prefer `isAuthorized` because it makes the boolean type obvious and disambiguates the conflict with Blue's `authorized` variable

src/MorphoV2.sol
```diff
@@ -223,13 +233,42 @@ contract MorphoV2 is IMorphoV2 {
...
+    function setRatified(Offer memory offer, bool newRatified) external {
+        require(msg.sender == offer.maker || authorized[offer.maker][msg.sender], "ratification not authorized");
+        _ratified[abi.encode(offer)] = newRatified;
+    }
+
+    function setAuthorized(address authorizee, bool newAuthorized) external {
+        authorized[msg.sender][authorizee] = newAuthorized;
+    }
+
+    function setAuthorizedWithSig(Authorization memory authorization, Signature calldata signature) external {
```
MathisGD, line 251: here an EOA, is allowing a smart-contract to post offers on their behalf. is that really something though? why not signing offers yourself directly

src/MorphoV2.sol
```diff
@@ -223,13 +233,42 @@ contract MorphoV2 is IMorphoV2 {
         return seizures;
     }
 
+    function ratified(Offer memory offer) external view returns (bool) {
```
MathisGD, line 236: would it be an offer or a root?

src/MorphoV2.sol
```diff
@@ -49,8 +52,15 @@ contract MorphoV2 is IMorphoV2 {
         require(offer.obligation.maturity >= block.timestamp, "maturity");
         require(offer.obligation.chainId == block.chainid, "chain id mismatch");
         require(offer.start < offer.expiry || offer.expiryPrice == offer.startPrice, "inconsistent prices");
-        require(signer(root, sig) == offer.maker, "invalid signature");
-        require(MathLib.isLeaf(root, keccak256(abi.encode(offer)), proof), "invalid proof");
+        require(msg.sender == taker || authorized[taker][msg.sender], "order not authorized");
+        require(
+            msg.sender == offer.maker
```
MathisGD, line 57: what's the use-case exactly for that? it means that you are authorized for the taker btw (see #73)

MathisGD: the question applies to authorized[offer.maker][msg.sender] as well

src/MorphoV2.sol
```diff
@@ -21,6 +21,9 @@ contract MorphoV2 is IMorphoV2 {
     mapping(bytes32 => uint256) public totalUnits;
     mapping(bytes32 => uint256) public totalShares;
     mapping(address => mapping(bytes32 => mapping(address => uint256))) public collateralOf;
+    mapping(address => mapping(address => bool)) public authorized;
```
MathisGD, line 24: let's add it to withdraw, withdrawCollat etc

src/MorphoV2.sol
adhusson: It's an EOA that allows another account (EOA or smart contract) to post on their behalf. The difference with the other function is that its gasless for the EOA.

src/MorphoV2.sol
adhusson: > what's the use-case exactly for that?

For `msg.sender == offer.maker` not sure, maybe I had gasless takes (see #94) in mind. Will remove.

For `authorized[offer.maker][msg.sender]` it's for the general case where 2 users completely trust a third-party, he can match them directly, used in conjunction with `authorized[taker][msg.sender]`.

src/MorphoV2.sol
adhusson: Since ratification is onchain you probably only ratify what is about to be taken. But it's more uniform to only sign/ratify one object type (roots). What do you think?

src/interfaces/IMorphoV2.sol
adhusson: Will change it but not sure, it makes sense that they have the same name since it refers to the same value: `authorized[authorizer][authorizee] = authorization.authorized`.

peyha: then I would suggest also changing the mapping name to "isAuthorized" (just like it was done on [Blue](https://github.com/morpho-org/morpho-blue/blob/9686ec502c8583a76d30f31acd7f70a38aa30818/src/Morpho.sol#L66) ) even though the `authorized` naming is a bit leaner

adhusson:
replaced by #187

PR #139, opened 2025-10-04
feat: ratification callback

adhusson:
Add a ratification callback to #138.

If callback addresses need to be whitelisted by the router, ratifier addresses will need to be whitelisted as well. 

In #138 , the router has to (somehow) know that there is a whitelisted ratifier contract authorized by the offer. The taker also has to (somehow) know about this ratifier, and then the taker must do this multicall:
1. Call `ratifier.<someMethod>` to get the offer ratified.
2. Call `morpoV2.take`.

In this PR the router can check if `offer.ratifier` is whitelisted, and the taker can just call `take`. It also costs less gas (you don't write to the `_ratified` mapping).

(Why not merge the callback hook with the ratification hook? You would need to whitelist any combination of `(callback,ratifier)` and create a single contract that does the combination of both.)

### DIFF
--- a/src/MorphoV2.sol
+++ b/src/MorphoV2.sol
@@ -249,7 +249,6 @@ contract MorphoV2 is IMorphoV2 {
         return _ratified[abi.encode(offer)];
     }
 
-    /// @dev No ratification by callback, use multicall.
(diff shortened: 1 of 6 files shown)


adhusson:
replaced by #187

PR #141, opened 2025-10-06
[Certora] Show equivalent hashing

QGarchery:


certora/helpers/Hasher.sol
```diff
@@ -0,0 +1,25 @@
...
+contract Hasher {
+    function usedKeccak256(bytes32 x, bytes32 y) external pure returns (bytes32) {
+        return keccak256(MathLib.sort(x, y));
+    }
+
+    function commutativeKeccak256(bytes32 a, bytes32 b) public pure returns (bytes32) {
+        return a < b ? efficientKeccak256(a, b) : efficientKeccak256(b, a);
+    }
+
+    function efficientKeccak256(bytes32 a, bytes32 b) public pure returns (bytes32 value) {
```
MathisGD, line 18: makes me think that we should try the gas improvement

certora/Hasher.spec
```diff
@@ -0,0 +1,20 @@
...
+
+rule noRevertsUsedKeccak256(bytes32 x, bytes32 y) {
+    usedKeccak256@withrevert(x, y);
+    assert !lastReverted;
+}
+
+rule noRevertsCommutativeKeccak256(bytes32 x, bytes32 y) {
+    commutativeKeccak256@withrevert(x, y);
+    assert !lastReverted;
+}
```
MathisGD, line 20: we could also have checked the equivalence of `verify` and `isLeaf`

QGarchery: Yes, I didn't include it for now because `processProof` is very similar code, and it requires to define a `loop_iter` bound

certora/helpers/Hasher.sol
adhusson: Tried it yesterday if you mean [EfficientHashLib](https://github.com/Vectorized/solady/blob/main/src/utils/EfficientHashLib.sol#L796-L801) and it cost more gas (don't know why).

QGarchery: Maybe it's coming from bytes being copied ?

For this function `efficientKeccak256`, I get that it's much more efficient than the solidity `keccak256`, which seems a bit off: 505 gas improvement
Screenshot from 2025-10-06 14-52-35

certora/Hasher.spec
MathisGD: ok, let's forget about this

certora/helpers/Hasher.sol
MathisGD: opened an issue about it

MathisGD (approved):

certora/helpers/Hasher.sol
adhusson: I was trying in the context of hashing arbitrary bytes but here the value are bytes32, so solidity probably does inefficient things in that case.

MathisGD (approved):

PR #144, opened 2025-10-06
feat: non struct sign

MathisGD:
https://morpholabs.slack.com/archives/C02N7CZ088N/p1759745201337129

### DIFF
--- a/src/MorphoV2.sol
+++ b/src/MorphoV2.sol
@@ -47,6 +47,7 @@ contract MorphoV2 is IMorphoV2 {
         require(block.timestamp >= offer.start, "offer not started");
         require(block.timestamp <= offer.expiry, "offer expired");
         require(offer.obligation.maturity >= block.timestamp, "maturity");
+        require(offer.obligation.chainId == block.chainid, "chain id mismatch");
         require(offer.start < offer.expiry || offer.expiryPrice == offer.startPrice, "inconsistent prices");
         require(signer(root, sig) == offer.maker, "invalid signature");
         require(MathLib.isLeaf(root, keccak256(abi.encode(offer)), proof), "invalid proof");
@@ -230,10 +231,8 @@ contract MorphoV2 is IMorphoV2 {
     }
 
     function signer(bytes32 root, Signature memory signature) internal view returns (address) {
-        bytes32 hashStruct = keccak256(abi.encode(ROOT_TYPEHASH, root));
-        bytes32 domainSeparator = keccak256(abi.encode(DOMAIN_TYPEHASH, block.chainid, address(this)));
-        bytes32 digest = keccak256(bytes.concat("\x19\x01", domainSeparator, hashStruct));
-        address tentativeSigner = ecrecover(digest, signature.v, signature.r, signature.s);
+        bytes32 messageHash = keccak256(bytes.concat("\x19\x01\x45thereum signed message:\n32", root));
+        address tentativeSigner = ecrecover(messageHash, signature.v, signature.r, signature.s);
         require(tentativeSigner != address(0), "invalid signature");
         return tentativeSigner;
     }
--- a/src/MorphoV2.sol
+++ b/src/MorphoV2.sol
@@ -231,7 +231,7 @@ contract MorphoV2 is IMorphoV2 {
     }
 
     function signer(bytes32 root, Signature memory signature) internal view returns (address) {
-        bytes32 messageHash = keccak256(bytes.concat("\x19\x01\x45thereum signed message:\n32", root));
+        bytes32 messageHash = keccak256(bytes.concat("\x19\x45thereum Signed Message:\n32", root));
         address tentativeSigner = ecrecover(messageHash, signature.v, signature.r, signature.s);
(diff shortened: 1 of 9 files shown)


src/MorphoV2.sol
```diff
@@ -230,10 +231,8 @@ contract MorphoV2 is IMorphoV2 {
     }
 
     function signer(bytes32 root, Signature memory signature) internal view returns (address) {
-        bytes32 hashStruct = keccak256(abi.encode(ROOT_TYPEHASH, root));
-        bytes32 domainSeparator = keccak256(abi.encode(DOMAIN_TYPEHASH, block.chainid, address(this)));
-        bytes32 digest = keccak256(bytes.concat("\x19\x01", domainSeparator, hashStruct));
-        address tentativeSigner = ecrecover(digest, signature.v, signature.r, signature.s);
+        bytes32 messageHash = keccak256(bytes.concat("\x19\x01\x45thereum signed message:\n32", root));
```
```suggestion
        bytes32 messageHash = keccak256(bytes.concat("\x19\x45thereum signed message:\n32", root));
```

```suggestion
        bytes32 messageHash = keccak256(bytes.concat("\x19\x45thereum signed message:\n32", root));
```
it would be good to have a test for this, to match a standard implementation. One other thing I'm unsure about is the "32" string, maybe it's encoded as `0x20`

adhusson: or

```suggestion
        bytes32 messageHash = keccak256(bytes.concat("\x19Ethereum signed message:\n32", root));
```

test/BaseTest.sol
```diff
@@ -69,12 +69,9 @@ abstract contract BaseTest is Test {
     }
 
     function sig(bytes32 _root, uint256 sk) internal view returns (Signature memory) {
-        bytes32 hashStruct = keccak256(abi.encode(ROOT_TYPEHASH, _root));
-        bytes32 domainSeparator = keccak256(abi.encode(DOMAIN_TYPEHASH, block.chainid, address(morphoV2)));
-        bytes32 digest = keccak256(bytes.concat("\x19\x01", domainSeparator, hashStruct));
-
+        bytes32 messageHash = keccak256(bytes.concat("\x19\x01\x45thereum signed message:\n32", _root));
```
adhusson:
```suggestion
        bytes32 messageHash = keccak256(bytes.concat("\x19\x45thereum signed message:\n32", _root));
```

src/MorphoV2.sol
adhusson: Just looked up implementations & asked chatgpt, apparently it is the ascii-encoding of the length. Not clear to me from the EIP.

adhusson: original commit that introduced the pattern, it is the ascii representation : https://github.com/ethereum/go-ethereum/pull/2940/files#diff-c426ecd2f7d247753b9ea8c1cc003f21fa412661c1f967d203d4edf8163da344R308

test/SignatureTest.sol
```diff
@@ -0,0 +1,31 @@
...
+
+contract SignatureTest is Test, MorphoV2 {
+    function testFormat(bytes32 root) public {
+        bytes32 messageHash = keccak256(bytes.concat("\x19\x45thereum Signed Message:\n32", root));
+        assertEq(toEthSignedMessageHash(root), messageHash);
+    }
+
+    function testSigner(bytes32 root, uint256 sk) public {
+        vm.assume(sk != 0);
+        vm.assume(sk < SECP256K1);
```
adhusson:
```suggestion
        sk = boundPrivateKey(sk)
```

test/SignatureTest.sol
```diff
@@ -0,0 +1,31 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Copyright (c) 2025 Morpho Association
+pragma solidity ^0.8.0;
+
+import "./BaseTest.sol";
+
+uint256 constant SECP256K1 = 115792089237316195423570985008687907852837564279074904382605163141518161494337;
```
adhusson:
```suggestion
```

src/MorphoV2.sol
MathisGD: >Just looked up implementations & asked chatgpt, apparently it is the ascii-encoding of the length. Not clear to me from the EIP.

agree. neither the rationale. though it's sure that it's like that because it's OZ's implementation

adhusson (approved):

src/MorphoV2.sol
```suggestion
        bytes32 messageHash = keccak256(bytes.concat("\x19\x45thereum signed message:\n32", root));
```
it would be good to have a test for this, to match a standard implementation. One other thing I'm unsure about is the "32" string, maybe it's encoded as `0x20`

QGarchery (approved):

src/MorphoV2.sol
QGarchery: Tested in SignatureTest.sol against the reference implementation found [here](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/cryptography/MessageHashUtils.sol)

PR #146, opened 2025-10-07
feat: trading fee 2

MathisGD:


src/MorphoV2.sol
QGarchery, line 1: we have:
- `(units - sellerAssets) * fee = trading fee = buyerAssets - sellerAssets`
- `units * buyerPrice = buyerAssets`
- `units * sellerPrice = sellerAssets`

so that gives us: `(1 - sellerPrice) * fee = buyerPrice - sellerPrice`
and so the formula used in this PR are:
`buyerPrice = sellerPrice (1 - fee) + fee`
`sellerPrice = (buyerPrice - fee) / (1 - fee)`

QGarchery (approved):

src/MorphoV2.sol
peyha: What rounding direction should we use for each multiplication in these formulas ?

Issue #143, opened 2025-10-06
callback interface


MathisGD:
just wondering, wouldn't it be worth passing either `price` or `obligationUnits` to callbacks ? might be useful for complex DeFi interactions

_Originally posted by @peyha in https://github.com/morpho-org/morpho-v2/pull/131#discussion_r2406462629_

peyha:
We should pass 
- all amounts
- counterpart
- offer hash

We should also use different endpoints `onBuy`/`onSell`

peyha:
done by #168

PR #148, opened 2025-10-07
feat: add price as arg of take callback

peyha:
We probably don't put enough information as callback data
Should we use `price` or `obligationUnits` ?
Should we send `feeAssets` ?


MathisGD:
let's brainstorm first in #143

MathisGD:
replaced by #168

PR #149, opened 2025-10-07
chore: remove warnings & unify internal function naming

adhusson:


src/MorphoV2.sol
```diff
@@ -229,7 +229,7 @@ contract MorphoV2 is IMorphoV2 {
         return keccak256(abi.encode(obligation));
     }
 
-    function signer(bytes32 root, Signature memory signature) internal view returns (address) {
+    function _signer(bytes32 root, Signature memory signature) internal pure returns (address) {
```
MathisGD, line 232: didn't we say it should be the opposite now?

test/BaseTest.sol
```diff
@@ -57,18 +57,18 @@ abstract contract BaseTest is Test {
...
     }
 
     // assumes the offer is the first one!
     function proof(Offer[2] memory offers) internal pure returns (bytes32[] memory) {
-        bytes32[] memory proof = new bytes32[](1);
-        proof[0] = keccak256(abi.encode(offers[1]));
-        return proof;
+        bytes32[] memory _proof = new bytes32[](1);
+        _proof[0] = keccak256(abi.encode(offers[1]));
+        return _proof;
```
MathisGD:
```suggestion
        bytes32[] memory res = new bytes32[](1);
        res[0] = keccak256(abi.encode(offers[1]));
        return res;
```

peyha:
Can't the CI catch warnings ?

src/MorphoV2.sol
adhusson: yes, it's just to harmonize we can clean later

adhusson:
> Can't the CI catch warnings ?

Good point, see https://github.com/morpho-org/morpho-v2/pull/152

MathisGD (approved):

Issue #70, opened 2025-08-23
only check health if seller increased debt


adhusson:
in take, health of the seller does not need to be checked if he only reduced his bonds

MathisGD:
note that there is this:
```
if (debt == 0) {
            return true;
```
you think that it would be better to do it externally of isHealthy?

adhusson:
no this is good enough actually

PR #150, opened 2025-10-07
refactor: minor change in isHealthy

MathisGD:
trying #70 anyway


src/MorphoV2.sol
QGarchery, line 1: If we do this then the max debt computation is not skipped on `withdrawCollateral` which is too bad because I think it's a common use case (when rolling notably)

PR #154, opened 2025-10-08
refactor: computations in take

MathisGD:
I find it clearer like that (notably 'bought' is very misleading today)

### DIFF
--- a/src/MorphoV2.sol
+++ b/src/MorphoV2.sol
@@ -75,22 +75,20 @@ contract MorphoV2 is IMorphoV2 {
 
         bytes32 id = _id(offer.obligation);
 
-        uint256 repaid = UtilsLib.min(debtOf[buyer][id], obligationUnits);
-        uint256 bought = obligationUnits - repaid;
-        uint256 boughtShares = bought.mulDivDown(totalShares[id] + 1, totalUnits[id] + 1);
-        uint256 withdrawn =
-            UtilsLib.min(sharesOf[seller][id].mulDivDown(totalUnits[id] + 1, totalShares[id] + 1), obligationUnits);
-        uint256 withdrawnShares = withdrawn.mulDivUp(totalShares[id] + 1, totalUnits[id] + 1);
+        uint256 sellerSharesDecrease =
+            UtilsLib.min(obligationUnits.mulDivDown(totalShares[id] + 1, totalUnits[id] + 1), sharesOf[seller][id]);
+        uint256 sellerDebtIncrease =
+            obligationUnits - sellerSharesDecrease.mulDivDown(totalUnits[id] + 1, totalShares[id] + 1);
+        uint256 buyerDebtDecrease = UtilsLib.min(obligationUnits, debtOf[buyer][id]);
+        uint256 buyerSharesIncrease =
+            (obligationUnits - buyerDebtDecrease).mulDivDown(totalShares[id] + 1, totalUnits[id] + 1);
 
-        debtOf[buyer][id] -= repaid;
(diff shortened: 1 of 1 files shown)


src/MorphoV2.sol
```diff
@@ -75,22 +75,20 @@ contract MorphoV2 is IMorphoV2 {
...
-        totalUnits[id] -= withdrawn;
+        uint256 sellerSharesDecrease =
+            UtilsLib.min(obligationUnits.mulDivDown(totalShares[id] + 1, totalUnits[id] + 1), sharesOf[seller][id]);
+        uint256 sellerDebtIncrease =
+            obligationUnits - sellerSharesDecrease.mulDivDown(totalUnits[id] + 1, totalShares[id] + 1);
+        uint256 buyerDebtDecrease = UtilsLib.min(obligationUnits, debtOf[buyer][id]);
+        uint256 buyerSharesIncrease =
+            (obligationUnits - buyerDebtDecrease).mulDivDown(totalShares[id] + 1, totalUnits[id] + 1);
+
+        debtOf[buyer][id] -= buyerDebtDecrease;
```
MathisGD, line 86: there is some loss of gas to not do that. I think that we can do it (slightly ugly but it's ok)
```suggestion
        if (buyerDebtDecrease > 0) debtOf[buyer][id] -= buyerDebtDecrease;
```

adhusson: for all?

adhusson (approved):

src/MorphoV2.sol
MathisGD: yes

QGarchery (approved):

src/MorphoV2.sol
QGarchery, line 1: Note that all subtractions could be `unchecked` (but in theory there could be overflows)

MathisGD: you mean that in theory there couldn't be overflows right?

src/MorphoV2.sol
```diff
@@ -75,22 +75,20 @@ contract MorphoV2 is IMorphoV2 {
...
+        sharesOf[buyer][id] += buyerSharesIncrease;
+        sharesOf[seller][id] -= sellerSharesDecrease;
+        debtOf[seller][id] += sellerDebtIncrease;
+        totalShares[id] = totalShares[id] + buyerSharesIncrease - sellerSharesDecrease;
+        totalUnits[id] = totalUnits[id] + sellerDebtIncrease - buyerDebtDecrease;
```
MathisGD, line 91:
```suggestion
        if (buyerDebtDecrease > 0) debtOf[buyer][id] -= buyerDebtDecrease;
        if (buyerSharesIncrease > 0) sharesOf[buyer][id] += buyerSharesIncrease;
        if (sellerSharesDecrease > 0) sharesOf[seller][id] -= sellerSharesDecrease;
        if (sellerDebtIncrease > 0) debtOf[seller][id] += sellerDebtIncrease;
        if (buyerSharesIncrease != sellerSharesDecrease) {
            totalShares[id] = totalShares[id] + buyerSharesIncrease - sellerSharesDecrease;
        }
        if (sellerDebtIncrease != buyerDebtDecrease) {
            totalUnits[id] = totalUnits[id] + sellerDebtIncrease - buyerDebtDecrease;
        }
```

src/MorphoV2.sol
QGarchery: In theory there could be overflows, for broken tokens with total supply > 2**256 for example. Also if the rate is high it could create even more obligation units than the tokens

PR #156, opened 2025-10-09
feat: return amounts

### LINKED ISSUE #155: return shares, assets...

MathisGD:
will test later, when we refactor the tests

fixes #155

### DIFF
--- a/certora/MorphoV2.conf
+++ b/certora/MorphoV2.conf
@@ -11,6 +11,6 @@
   "optimistic_loop": true,
   "optimistic_hashing": true,
   "solc_via_ir": true,
-  "hashing_length_bound": "320",
+  "hashing_length_bound": "640",
   "msg": "MorphoV2" 
 }
--- a/certora/MorphoV2.spec
+++ b/certora/MorphoV2.spec
@@ -9,3 +9,17 @@ methods {}
 rule sanity() {
     assert true;
 }
+
+rule takeInputs(env e, uint256 buyerAssets, uint256 sellerAssets, uint256 obligationUnits, uint256 obligationShares, address taker, MorphoV2.Offer offer, MorphoV2.Signature signature, bytes32 root, bytes32[] proof, address takerCallbackAddress, bytes takerCallbackData) {
+    uint256 buyerAssetsOutput;
+    uint256 sellerAssetsOutput;
+    uint256 obligationUnitsOutput;
+    uint256 obligationSharesOutput;
+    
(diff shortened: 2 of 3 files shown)


peyha (approved):
did not check the certora files

certora/MorphoV2.conf
```diff
@@ -11,6 +11,6 @@
   "optimistic_loop": true,
   "optimistic_hashing": true,
   "solc_via_ir": true,
-  "hashing_length_bound": "320",
+  "hashing_length_bound": "640",
```
adhusson, line 14: Why increase it?

QGarchery: The general idea is that you have an optimistic hashing (meaning that it inserts something like `require length <= 320` for `length` the length of the data to hash). So it could create some vacuity if the data is always more than 320 bytes. 

The details are not completely clear to me, see this question [here](https://morpholabs.slack.com/archives/C09EAFEJUMV/p1760015932730959?thread_ts=1759331567.109389&cid=C09EAFEJUMV). What's not clear is that the Offer struct was, at that time, 8 words (so < 320 bytes) in [this run](https://prover.certora.com/output/7508195/b99a2a3d3d79480899f68d0b00164fa0/?anonymousKey=48cd8b7d1fe66379d9fb8d20b95d68937b3e845c), so I'm not sure why it was a vacuity error

QGarchery (approved):

certora/MorphoV2.conf
adhusson: Offers included collaterals before so their size was not fixed right? But anyway my understanding was that with optimistic hashing, increasing the length bound can only create new violations.

adhusson (approved):

certora/MorphoV2.conf
QGarchery: > Offers included collaterals before so their size was not fixed right?

Yes but hashing of a struct that contains a list first hashes the list and uses that hash as the field element to hash the whole struct. So in effect the `collaterals` list in the `Offer` struct was taking 32 bytes

QGarchery: > Yes but hashing of a struct that contains a list first hashes the list and uses that hash as the field element to hash the whole struct

Actually let me double check that

adhusson: Since the call is `keccak256(abi.encode(offer))`, the hash operates on raw bytes so I think it doesn't have access to enough structure to do this nested hashing.

QGarchery: Yes you're right, I will investigate further

PR #159, opened 2025-10-10
feat: prevent crossing

MathisGD:
replaces #117

### DIFF
--- a/src/MorphoV2.sol
+++ b/src/MorphoV2.sol
@@ -64,6 +64,7 @@ contract MorphoV2 is IMorphoV2 {
     /// @dev Same function used to buy and sell.
     /// @dev If one wants to match two offers without taking a position, they can batch take them and not have a
     /// position at the end.
+    /// @dev Neither the taker nor the maker can pass from having obligation shares to having debt in one take.
     function take(
         uint256 buyerAssets,
         uint256 sellerAssets,
--- a/src/MorphoV2.sol
+++ b/src/MorphoV2.sol
@@ -133,19 +133,23 @@ contract MorphoV2 is IMorphoV2 {
             (consumed[offer.maker][offer.nonce] += (offer.buy ? buyerAssets : sellerAssets)) <= offer.assets, "consumed"
         );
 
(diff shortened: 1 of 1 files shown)


src/MorphoV2.sol
```diff
@@ -133,19 +133,23 @@ contract MorphoV2 is IMorphoV2 {
             (consumed[offer.maker][offer.nonce] += (offer.buy ? buyerAssets : sellerAssets)) <= offer.assets, "consumed"
         );
```
peyha, line 136:
```suggestion
     bool buyerHasDebt = (debtOf[buyer][id] > 0);
     bool sellerHasObligation = (sharesOf[seller][id] > 0);
```
small prop which is slightly more readable (don't think it saves gas if the solidity compiler is correctly optimized) but current version is already ok

src/MorphoV2.sol
peyha, line 66: worth adding a dev comment (like in the other PR)

peyha (approved):

adhusson (approved):

QGarchery (approved):

src/MorphoV2.sol
MathisGD: I don't find it that much clearer

src/MorphoV2.sol
MathisGD: done

peyha (approved):

src/MorphoV2.sol
QGarchery, line 1: Weirdly I think this breaks the invariant sharePriceLessThanOne (see [this PR](https://github.com/morpho-org/morpho-v2/pull/157), and we have a complete run [here](https://prover.certora.com/output/17512/bb7b483e9ed34663b445b1ce4814deb6/?anonymousKey=a35e736571631dbbc0acad4d6837facacf367e90)). The reason is that we fix the conversion direction `obligationUnit` <-> `obligationShares` when those are defined, but the total shares / units can go up or down

QGarchery: See #166 for a fix

MathisGD: actually it seems to [pass](https://github.com/morpho-org/morpho-v2/pull/157)

PR #161, opened 2025-10-13
feat: offer inputs

### LINKED ISSUE #123: obligationUnit/shares as input of offers?
seems important. how would it work with the consumed mapping? we could ignore that, the same way you can mix loan assets on the same nonce

MathisGD:
fixes #123 

allows to, for example:
- get a precise amount of debt
- sell all your shares

I made some verification but not tests, is it fine for the moment?


src/MorphoV2.sol
```diff
@@ -23,8 +23,10 @@ contract MorphoV2 is IMorphoV2 {
     mapping(address => mapping(bytes32 => mapping(address => uint256))) public collateralOf;
 
     /// @dev Multiple offers can have the same nonce. This allows to implement easy and efficient batch-cancelling and
-    /// OCO (One-Cancels-the-Other) orders. Note that OCO orders work better if all offers have the same amount,
-    /// otherwise one might not be takable anymore while an other one at the same nonce is still takeable.
+    /// OCO (One-Cancels-the-Other) offers.
```
adhusson: They should also all be buy or all sell?

adhusson (approved):

src/MorphoV2.sol
MathisGD: hmmm it seems yes. but for other reasons: it's more that it doesn't make economical sense

peyha (approved):
some test will need to be implemented for this #165

certora/MorphoV2.spec
QGarchery, line 1: I don't get the use cases:

 > get a precise amount of debt

There is no guarantee that you will be taken fully, so you cannot guarantee that anyway right ?

> sell all your shares

Same, you cannot guarantee that all your shares will be sold. And if you want to make sure that the `assets` parameter doesn't limit the takers' ability to take all your shares, you can always put `type(uint256).max`.

Say that I'm a borrower and that I want to make an offer to roll to the next obligation. In my offer I put no limit on the debt, and in the callback it would withdraw all collateral (to do something else with it), and so the callback would fail if not all the debt was repaid

adhusson: You could say it makes the life of the taker easier they just need to copy the offer obligation amount instead of having to check assets too big -> query maker obligations -> change input, but still I agree with you.

adhusson: Without this, it goes from annoying to impossible without a custom callback for a lender to guarantee they never become a borrower.

QGarchery: Hm good point, though the user could also not put any collateral in this particular obligation. But the symmetric is more important: I'm a borrower and I want to exit all, but I don't want to become a lender

QGarchery (approved):
See [this discussion](https://github.com/morpho-org/morpho-v2/pull/161/files#r2430311680) for motivation

src/MorphoV2.sol
```diff
@@ -25,7 +25,7 @@ contract MorphoV2 is IMorphoV2 {
     mapping(address => mapping(bytes32 => mapping(address => uint256))) public collateralOf;
 
     /// @dev Groups are useful to have a global offered amount shared accross multiple offers ("OCO").
-    /// @dev To work as expected, all offers in a same group should have the same amount and the same loan asset.
+    /// @dev To work as expected, all offers in a same group should have the same innput, amount and loan token.
```
```suggestion
    /// @dev To work as expected, all offers in a same group should have the same input, amount and loan token.
```

peyha: thinking about it, "input" is not really clear,
```suggestion
    /// @dev To work as expected, all offers in a same group should have the same assets, obligationUnits, obligationShares and loan token.
```

src/MorphoV2.sol
```diff
@@ -157,9 +161,16 @@ contract MorphoV2 is IMorphoV2 {
             sellerAssets = obligationUnits.mulDivDown(sellerPrice, WAD);
         }
 
-        require(
-            (consumed[offer.maker][offer.group] += (offer.buy ? buyerAssets : sellerAssets)) <= offer.assets, "consumed"
-        );
+        if (offer.assets > 0) {
+            require(
+                (consumed[offer.maker][offer.group] += offer.buy ? buyerAssets : sellerAssets) <= offer.assets,
```
peyha, line 167: I think it was already mentioned somewhere but I would prefer having one `consumed` array per asset type (or a custom Consumed struct) to prevent weird behaviors. The gas cost is theoretically the same but it's slightly heavier in the code so it's not a strong opinion

peyha (commented):
did not verify the certora part

src/MorphoV2.sol
MathisGD: you would update only one of the 3? if not, the gas cost is very different

peyha: yes only one update storage update in total

MathisGD:
@codex please review

src/MorphoV2.sol
MathisGD: isn't it misleading that only one of the consumed is updated?

src/MorphoV2.sol
```diff
@@ -25,7 +25,8 @@ contract MorphoV2 is IMorphoV2 {
     mapping(address => mapping(bytes32 => mapping(address => uint256))) public collateralOf;
 
     /// @dev Groups are useful to have a global offered amount shared accross multiple offers ("OCO").
-    /// @dev To work as expected, all offers in a same group should have the same amount and the same loan asset.
+    /// @dev To work as expected, all offers in a same group should have the same assets, obligationUnits,
+    /// obligationShares and loan token.
     mapping(address user => mapping(bytes32 group => uint256)) public consumed;
```
MathisGD, line 30: an other possibility could be to store in the slot the unit type to avoid issues, but sounds heavy for not much

peyha (approved):

PR #164, opened 2025-10-13
feat: recover address from signature

### LINKED ISSUE #29: nick's method
should we allow nick's method? (=> should offering be a field of offers)

MathisGD:
fixes #29

pros:
- smaller offer
- anything else?

cons:
- offchain debugging
- less explicit errors if signature problems

### DIFF
--- a/src/MorphoV2.sol
+++ b/src/MorphoV2.sol
@@ -87,8 +87,9 @@ contract MorphoV2 is IMorphoV2 {
         require(offer.obligation.maturity >= block.timestamp, "maturity");
         require(offer.obligation.chainId == block.chainid, "chain id mismatch");
         require(offer.start < offer.expiry || offer.expiryPrice == offer.startPrice, "inconsistent prices");
-        require(_signer(root, sig) == offer.maker, "invalid signature");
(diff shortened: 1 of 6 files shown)


src/MorphoV2.sol
MathisGD, line 1: not fully sure tbh

adhusson:
fyi not really "nick's method" which is about extracting an address from a clearly random signature so that you know you don't have the private key and can't sign other things. Here it's just recovering the address from the signature.

src/interfaces/IMorphoV2.sol
```diff
@@ -19,7 +19,6 @@ struct Collateral {
 struct Offer {
     Obligation obligation;
     bool buy;
-    address maker;
```
adhusson, line 22: Will we do EIP-1271? It requires a signer field.

peyha:
How would it work with smart contract offers ?

MathisGD:
oh right! I thought that they would just not be able to do that, but then we can't remove `maker` in the offer struct...

src/MorphoV2.sol
adhusson: I think it's easier to have it in the struct overall, but not sure either. But calldata gas cost difference is ~368.

PR #157, opened 2025-10-09
[Certora] Share price less than one

QGarchery:
~~See [this counter-example](https://prover.certora.com/output/66603/f860f458816e465896046f41b96acfdc) without the code modification of this PR~~

edit: not needed since #159

PR #166, opened 2025-10-14
Primary rounding

QGarchery:


src/MorphoV2.sol
```diff
@@ -111,20 +111,22 @@ contract MorphoV2 is IMorphoV2 {
         uint256 buyerPrice = offer.buy ? offerPrice : offerPrice.mulDivDown(WAD - _tradingFee, WAD) + _tradingFee;
         uint256 sellerPrice = offer.buy ? (offerPrice - _tradingFee).mulDivDown(WAD, WAD - _tradingFee) : offerPrice;
 
+        bool primary = debtOf[buyer][id] == 0 && sharesOf[seller][id] == 0;
```
QGarchery, line 114: We can probably find something that looks better than this, reusing it in the `if (debtOf[buyer][id] == 0 && sharesOf[seller][id] == 0)` below for example

QGarchery:
Useless in the end, see #157

PR #168, opened 2025-10-15
callback interface

peyha:
New callback interface as discussed in #143

### DIFF
--- a/src/MorphoV2.sol
+++ b/src/MorphoV2.sol
@@ -166,7 +166,6 @@ contract MorphoV2 is IMorphoV2 {
             ICallbacks(buyerCallbackAddress)
                 .onBuy(
                     offer.obligation,
-                    (buyer == offer.maker),
                     buyer,
                     seller,
                     buyerAssets,
@@ -187,7 +186,6 @@ contract MorphoV2 is IMorphoV2 {
             ICallbacks(sellerCallbackAddress)
                 .onSell(
                     offer.obligation,
-                    (seller == offer.maker),
                     buyer,
                     seller,
                     buyerAssets,
--- a/src/MorphoV2.sol
+++ b/src/MorphoV2.sol
@@ -97,7 +97,6 @@ contract MorphoV2 is IMorphoV2 {
         );
         require(block.timestamp >= offer.start, "offer not started");
         require(block.timestamp <= offer.expiry, "offer expired");
-        require(offer.obligation.maturity >= block.timestamp, "maturity");
         require(offer.obligation.chainId == block.chainid, "chain id mismatch");
         require(offer.start < offer.expiry || offer.expiryPrice == offer.startPrice, "inconsistent prices");
         require(offer.maker != taker, "buyer and seller cannot be the same");
--- a/src/MorphoV2.sol
+++ b/src/MorphoV2.sol
@@ -2,13 +2,13 @@
 // Copyright (c) 2025 Morpho Association
 pragma solidity 0.8.28;
 
-import "./libraries/UtilsLib.sol";
-import "./libraries/SafeTransferLib.sol";
-import "./libraries/ConstantsLib.sol";
-import "./libraries/MathLib.sol";
-import "./interfaces/IOracle.sol";
-import "./interfaces/IMorphoV2.sol";
-import "./interfaces/ICallbacks.sol";
+import {UtilsLib} from "./libraries/UtilsLib.sol";
+import {SafeTransferLib} from "./libraries/SafeTransferLib.sol";
+import {WAD, ORACLE_PRICE_SCALE, LIQUIDATION_INCENTIVE_FACTOR} from "./libraries/ConstantsLib.sol";
+import {MathLib} from "./libraries/MathLib.sol";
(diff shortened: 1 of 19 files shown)


src/MorphoV2.sol
```diff
@@ -171,7 +183,19 @@ contract MorphoV2 is IMorphoV2 {
         SafeTransferLib.safeTransferFrom(offer.obligation.loanToken, buyer, seller, sellerAssets);
 
         if (sellerCallbackAddress != address(0)) {
-            ICallbacks(sellerCallbackAddress).onTake(offer.obligation, seller, sellerAssets, sellerCallbackData);
+            ICallbacks(sellerCallbackAddress)
+                .onSell(
+                    offer.obligation,
+                    (seller == offer.maker),
```
MathisGD: as discussed, not needed

src/MorphoV2.sol
```diff
@@ -171,7 +183,19 @@ contract MorphoV2 is IMorphoV2 {
...
+                    buyer,
+                    seller,
+                    buyerAssets,
+                    sellerAssets,
+                    obligationUnits,
+                    obligationShares,
+                    id,
```
MathisGD: this is the obligation hash, not the offer hash right?

src/MorphoV2.sol
peyha: yes, it could be easily passed in the callback data

src/interfaces/ICallbacks.sol
```diff
@@ -5,6 +5,27 @@ pragma solidity 0.8.28;
 import {Seizure, Obligation} from "./IMorphoV2.sol";
 
 interface ICallbacks {
-    function onTake(Obligation memory obligation, address borrower, uint256 assets, bytes memory data) external;
+    function onBuy(
+        Obligation memory obligation,
```
MathisGD, line 9: imo not needed, as you can put it in the data

src/interfaces/ICallbacks.sol
```diff
@@ -5,6 +5,27 @@ pragma solidity 0.8.28;
...
+        uint256 sellerAssets,
+        uint256 obligationUnits,
+        uint256 obligationShares,
+        bytes32 offerHash,
+        bytes memory data
+    ) external;
+    function onSell(
+        Obligation memory obligation,
+        address buyer,
+        address seller,
```
MathisGD, line 19: the seller is you no?

MathisGD, line 19: should there be msg.sender?

src/interfaces/ICallbacks.sol
peyha: if we limit the arguments to only what's necessary and what cannot be passed through `callbackData` then we should keep `buyerAssets`, `sellerAssets`, `obligationUnits`, `obligationShares`
`offerHash` and counterpart address would make sense for the maker callback (because it cannot be easily put into the callback data)
I don't think it's worth splitting these callbacks into `onTakeBuy`/`onMakeBuy`/`onTakeSell`/`onMakeSell` because it would make implementations more complex

src/interfaces/ICallbacks.sol
peyha: indeed, if `msg.sender` is not the sender (which would be the case with a bundler or with integrations) it could be useful

src/interfaces/ICallbacks.sol
peyha: an issue is that we are pushing some of the complexity to product to craft callbacks, if we cover the vast majority of callback usecases by passing parameters such as an obligation, it could simplify the router's work

peyha:
Settled during the bootcamp
we should pass
- obligation
- all assets
- buyer if onBuy and seller if onSell
- maybe counterpart depending on gate workshop conclusion
- data

It does not seem like `offerHash` and `msg.sender` cover relevant use-cases

src/interfaces/ICallbacks.sol
```diff
@@ -5,6 +5,25 @@ pragma solidity 0.8.28;
 import {Seizure, Obligation} from "./IMorphoV2.sol";
 
 interface ICallbacks {
-    function onTake(Obligation memory obligation, address borrower, uint256 assets, bytes memory data) external;
+    function onBuy(
+        Obligation memory obligation,
+        address buyer,
+        address seller,
```
MathisGD: pending gate discussion, let's remove for now
```suggestion
        address buyer,
```

src/interfaces/ICallbacks.sol
```diff
@@ -5,6 +5,25 @@ pragma solidity 0.8.28;
...
+        uint256 buyerAssets,
+        uint256 sellerAssets,
+        uint256 obligationUnits,
+        uint256 obligationShares,
+        bytes memory data
+    ) external;
+    function onSell(
+        Obligation memory obligation,
+        address buyer,
```
MathisGD:
```suggestion
```

MathisGD (approved):

adhusson (approved):

QGarchery (approved):

PR #169, opened 2025-10-15
fix: require collats to be sorted

MathisGD:
right now isHealthy double counts duplicated collat. One could say it's the responsability of the lender to avoid it, but it's cheap to check

### DIFF
--- a/src/MorphoV2.sol
+++ b/src/MorphoV2.sol
@@ -319,14 +319,12 @@ contract MorphoV2 is IMorphoV2 {
         } else {
             uint256 maxDebt;
             for (uint256 i = 0; i < obligation.collaterals.length; i++) {
-                require(
-                    bytes20(obligation.collaterals[i].token) > bytes20(previousCollateralToken),
-                    "collaterals not sorted"
-                );
+                address currentCollateralToken = obligation.collaterals[i].token;
+                require(currentCollateralToken > previousCollateralToken, "collaterals not sorted");
                 uint256 price = IOracle(obligation.collaterals[i].oracle).price();
-                maxDebt += collateralOf[borrower][id][obligation.collaterals[i].token]
-                    .mulDivDown(price, ORACLE_PRICE_SCALE).mulDivDown(obligation.collaterals[i].lltv, 1e18);
-                previousCollateralToken = obligation.collaterals[i].token;
+                maxDebt += collateralOf[borrower][id][currentCollateralToken].mulDivDown(price, ORACLE_PRICE_SCALE)
+                    .mulDivDown(obligation.collaterals[i].lltv, 1e18);
+                previousCollateralToken = currentCollateralToken;
             }
 
             return debt <= maxDebt;
--- a/src/MorphoV2.sol
+++ b/src/MorphoV2.sol
@@ -313,11 +313,11 @@ contract MorphoV2 is IMorphoV2 {
     function _isHealthy(Obligation memory obligation, address borrower) internal view returns (bool) {
         bytes32 id = _id(obligation);
         uint256 debt = debtOf[borrower][id];
-        address previousCollateralToken; // used to enforce that collateral are sorted.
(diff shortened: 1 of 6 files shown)


src/MorphoV2.sol
```diff
@@ -311,14 +313,20 @@ contract MorphoV2 is IMorphoV2 {
...
+        address previousCollateralToken; // used to enforce that collateral are sorted.
         if (debt == 0) {
             return true;
         } else {
             uint256 maxDebt;
             for (uint256 i = 0; i < obligation.collaterals.length; i++) {
+                require(
+                    bytes20(obligation.collaterals[i].token) > bytes20(previousCollateralToken),
+                    "collaterals not sorted"
+                );
```
MathisGD: ideally we wouldn't check at each health check, like if the obligation already exists it means that it has already been checked. But is it worth the overhead?

adhusson: * The gas diff i get for the check is ~2k gas for 2 collaterals, not negligible imo. 
* With an intermediary variable, it goes down to 1k gas diff. 
* The `bytes20(...)` is costly (and not used?). You can compare addresses directly and it brings the difference down to 500 gas.
* Finally assembly brings the diff down to ~250, negligible at this point.

src/MorphoV2.sol
```diff
@@ -311,14 +313,20 @@ contract MorphoV2 is IMorphoV2 {
...
                 uint256 price = IOracle(obligation.collaterals[i].oracle).price();
                 maxDebt += collateralOf[borrower][id][obligation.collaterals[i].token]
                     .mulDivDown(price, ORACLE_PRICE_SCALE).mulDivDown(obligation.collaterals[i].lltv, 1e18);
+                previousCollateralToken = obligation.collaterals[i].token;
```
```suggestion
                address currentCollateralToken = obligation.collaterals[i].token;
                require(currentCollateralToken > previousCollateralToken, "collaterals not sorted");
                uint256 price = IOracle(obligation.collaterals[i].oracle).price();
                maxDebt += collateralOf[borrower][id][currentCollateralToken].mulDivDown(price, ORACLE_PRICE_SCALE)
                    .mulDivDown(obligation.collaterals[i].lltv, 1e18);
                previousCollateralToken = currentCollateralToken;
```

src/MorphoV2.sol
MathisGD: ideally we avoid assembly. should we skip the check anyway if the obligation already exists?

adhusson: yes re: assembly it's not worth it. Surprising to have such a diff on `bool x; assembly { x := gt(a,b) } require(x,"err");` vs. `require(a>b,"err");`.

Checking if the obligation exists without touching more storage would be checking debt > 0 (which can create false negatives)? But you're adding a jump on every check then, which reduces the overall saving, which makes the additional lines less worth it.

peyha:
~~Would it make sense to use the same collateral with different oracle/LLTV ? 
Suppose I want to collateralize my obligation by wstETH with the exchange rate oracle at 94.5% and wstETH with an UniV4 oracle at 91.5% ?  
Not yet sure we want to limit this kind of usecase~~
It does not work in the current version because `supplyCollateral` takes a collateral token as input and not a triplet (collateral, oracle, lltv)

src/MorphoV2.sol
```diff
@@ -311,14 +313,18 @@ contract MorphoV2 is IMorphoV2 {
     function _isHealthy(Obligation memory obligation, address borrower) internal view returns (bool) {
         bytes32 id = _id(obligation);
         uint256 debt = debtOf[borrower][id];
+        address previousCollateralToken; // used to enforce that collateral are sorted.
```
QGarchery: (optional) we can put this declaration just after the declaration of `maxDebt`

QGarchery (approved):

adhusson (approved):

peyha (approved):

PR #84, opened 2025-09-09
debt can be covered with loan asset

adhusson:
See issue #87 

* loan asset immediately withdrawable at maturity
* loan asset cannot be liquidated
* loan asset treated separately from collaterals
* debt ignores covered amount

These invariants should hold:
```
  withdrawable - (sum_user coverOf[user])
= (sum_interaction repaid in postmaturity liq) - (sum_interaction withdrawn) 
= totalBonds - (sum_user fullDebtOf[user])
```

PR #170, opened 2025-10-16
[pre-repay] cover

### LINKED ISSUE #87: Should the loan asset be usable as collateral

MathisGD:
new version of #84. fixes #87


certora/MorphoV2.spec
```diff
@@ -47,8 +47,8 @@ rule takeInputOutputConsistency(env e, uint256 buyerAssets, uint256 sellerAssets
 strong invariant notBorrowerAndLender(bytes32 id, address user)
     sharesOf(user, id) == 0 || debtOf(user, id) == 0;
 
-strong invariant totalUnitsEqualsSumDebtPlusWithdrawable(bytes32 id)
-    totalUnits(id) == sumDebtOf[id] + withdrawable(id);
+strong invariant totalUnitsEqualsSumDebtPlusPreRepaid(bytes32 id)
+    totalUnits(id) == sumDebtOf[id] + totalPreRepaid(id);
```
MathisGD: we actually loose this property, I think.

src/MorphoV2.sol
```diff
@@ -179,32 +180,36 @@ contract MorphoV2 is IMorphoV2 {
...
 
         SafeTransferLib.safeTransfer(obligation.loanToken, msg.sender, obligationUnits);
     }
 
-    function repay(Obligation memory obligation, uint256 obligationUnits, address onBehalf) external {
-        bytes32 id = _id(obligation);
-
-        debtOf[onBehalf][id] -= obligationUnits;
-        withdrawable[id] += obligationUnits;
+    function repay(Obligation memory obligation, uint256 assets, address onBehalf) external {
```
MathisGD: so actually people never really repay:
- either they buy back an obligation to "burn" their debt
- or they pre repay. in this case their debt is never decreased. isn't it bad?

src/MorphoV2.sol
```diff
@@ -179,32 +180,36 @@ contract MorphoV2 is IMorphoV2 {
...
+        preRepaid[onBehalf][_id(obligation)] += assets;
+        totalPreRepaid[_id(obligation)] += assets;
```
adhusson: It's useless to rehash the obligation.

```suggestion
        bytes32 id = _id(term);
        preRepaid[onBehalf][id] += assets;
        totalPreRepaid[id] += assets;
```

src/MorphoV2.sol
```diff
@@ -179,32 +180,36 @@ contract MorphoV2 is IMorphoV2 {
...
+        SafeTransferLib.safeTransferFrom(obligation.loanToken, msg.sender, address(this), assets);
+    }
 
-        SafeTransferLib.safeTransferFrom(obligation.loanToken, msg.sender, address(this), obligationUnits);
+    function withdrawRepaid(Obligation memory obligation, uint256 assets, address onBehalf) external {
+        preRepaid[onBehalf][_id(obligation)] -= assets;
+        totalPreRepaid[_id(obligation)] -= assets;
```
```suggestion
        bytes32 id = _id(term);
        preRepaid[onBehalf][id] -= assets;
        totalPreRepaid[id] -= assets;
```

src/MorphoV2.sol
MathisGD, line 1: an other possible abstraction to achieve the same thing: you repay (and liquidate) normally ("full"), but it's not withdrawable before maturity except by you (if you happen to be a lender in the future). Then if you buy some obligations in the future you can make an instant profit.

adhusson: The pre-repaid amount is not deducted from the debt, so a borrower with 100 pre-repaid, 100 debt, and < 100/lltv collat will be liquidated.

src/MorphoV2.sol
```diff
@@ -314,7 +319,7 @@ contract MorphoV2 is IMorphoV2 {
         if (debt == 0) {
             return true;
         } else {
-            uint256 maxDebt;
+            uint256 maxDebt = totalPreRepaid[id];
```
```suggestion
            uint256 maxDebt = preRepaid[borrower][id];
```

src/MorphoV2.sol
adhusson, line 390: Same here, badDebt will would be underestimated.

src/MorphoV2.sol
```diff
@@ -17,10 +17,11 @@ contract MorphoV2 is IMorphoV2 {
 
     mapping(address => mapping(bytes32 => uint256)) public sharesOf;
     mapping(address => mapping(bytes32 => uint256)) public debtOf;
```
adhusson, line 30: A usability difference with #84 is that the debt getter returns the "raw debt". So we would need to explain to all integrators that they should always compute `effectiveDebt = debtOf[borrower][id] - preRepaid[borrower][id]`.

I think it could avoid a good amount of future errors related to this if we had a standard getter that does the subtraction (also some lines below already use `debtOf` instead of `debtOf - preRepaid`).

src/MorphoV2.sol
```diff
@@ -17,10 +17,11 @@ contract MorphoV2 is IMorphoV2 {
...
-    mapping(bytes32 => uint256) public withdrawable;
+    mapping(address => mapping(bytes32 => uint256)) public preRepaid;
```
adhusson: `preRepaidOf` to follow the `sharesOf`/`debtOf`/`collateralOf` style? nitpick.

MathisGD: `preRepaid` and `debtOf` are never reset to zero, even once the obligation is "finished" (all lenders are gone)

src/MorphoV2.sol
MathisGD: indeed, I forgot that it was not using `isHealthy`

src/MorphoV2.sol
MathisGD: there are two possible abstractions here: either the preRepaid is seen as collateral, either it's seen as a debt reduction (same in liquidate)

src/MorphoV2.sol
MathisGD: but doesn't it make the "buying" later weirder. You buy so your debt becomes negative..

src/MorphoV2.sol
MathisGD: implemented in #172

src/MorphoV2.sol
adhusson: it stays at 0 if it's zeroFloorSub

src/MorphoV2.sol
```diff
@@ -179,32 +180,38 @@ contract MorphoV2 is IMorphoV2 {
...
+        totalPreRepaid[id] -= obligationUnits;
 
         totalShares[id] -= shares;
         totalUnits[id] -= obligationUnits;
 
         SafeTransferLib.safeTransfer(obligation.loanToken, msg.sender, obligationUnits);
     }
 
-    function repay(Obligation memory obligation, uint256 obligationUnits, address onBehalf) external {
+    function repay(Obligation memory obligation, uint256 assets, address onBehalf) external {
```
adhusson: Currently it is not possible to fully repay-by-trade (buy secondary) until you early-withdrew, because `debtOf` goes down when you early repay. 

So it's more complex to get out of an early repayment. You can't just make an offer for a good price and then withdraw your assets later. You can't fix it with a callback in your offer either because the callback happens after the debt is modified. So the only solution is to be a taker, flashloan collateral, and remove cover so you can buy.

adhusson: due to this i rebased (redid really) #84

adhusson: this comment is wrong, it should apply to #172

src/MorphoV2.sol
```diff
@@ -231,32 +232,38 @@ contract MorphoV2 is IMorphoV2 {
         return (buyerAssets, sellerAssets, obligationUnits, obligationShares);
     }
 
-    /// @dev Will revert if there is no withdrawable funds.
     function withdraw(Obligation memory obligation, uint256 obligationUnits, uint256 shares, address onBehalf)
         external
     {
+        require(obligation.maturity > block.timestamp, "maturity not reached");
```
```suggestion
        require(obligation.maturity < block.timestamp, "maturity not reached");
```

```suggestion
        require(block.timestamp > obligation.maturity, "maturity not reached");
```

src/MorphoV2.sol
MathisGD: see #176  as well

src/MorphoV2.sol
```diff
@@ -388,7 +397,7 @@ contract MorphoV2 is IMorphoV2 {
         if (debt == 0) {
```
adhusson, line 468:
```suggestion
        if (debt <= preRepaidOf[borrower][id]) {
```

MathisGD:
replaced by recovery close factor

PR #177, opened 2025-10-21
nonce=root (nonce option 1)

MathisGD:


src/MorphoV2.sol
```diff
@@ -24,10 +24,10 @@ contract MorphoV2 is IMorphoV2 {
     mapping(bytes32 => uint256) public totalShares;
     mapping(address => mapping(bytes32 => mapping(address => uint256))) public collateralOf;
 
-    /// @dev Multiple offers can have the same nonce. This allows to implement easy and efficient batch-cancelling and
+    /// @dev Multiple offers can have the same root. This allows to implement easy and efficient batch-cancelling and
     /// OCO (One-Cancels-the-Other) orders. Note that OCO orders work better if all offers have the same amount,
-    /// otherwise one might not be takable anymore while an other one at the same nonce is still takeable.
-    mapping(address user => mapping(uint256 nonce => uint256)) public consumed;
+    /// otherwise one might not be takable anymore while an other one at the same root is still takeable.
+    mapping(address user => mapping(bytes32 root => uint256)) public consumed;
```
adhusson, line 30: need to add smart contract offers to see how it works together

MathisGD:
we decided to go for https://github.com/morpho-org/morpho-v2/pull/179 because this one merges the concept of OCO and merkle tree, which we are not sure is good. Plus it makes it harder to cancel after a mistake (you need to know the root)

PR #178, opened 2025-10-21
feat: nonce option 2

MathisGD:


MathisGD:
we prefer #179 or #177

PR #179, opened 2025-10-21
global nonce (nonce option 3)

MathisGD:
add a global nonce concept, to allow cancelling everything quickly/easily (just generate a random number, no other input).

renamed nonce => group
and globalNonce => nonce

### DIFF
--- a/src/MorphoV2.sol
+++ b/src/MorphoV2.sol
@@ -299,8 +299,8 @@ contract MorphoV2 is IMorphoV2 {
         return seizures;
     }
 
-    function shuffleGlobalNonce() external {
-        globalNonce[msg.sender] = uint256(keccak256(abi.encode(globalNonce[msg.sender], block.prevrandao)));
+    function shuffleGlobalNonce(uint256 salt) external {
+        globalNonce[msg.sender] = uint256(keccak256(abi.encode(globalNonce[msg.sender], salt)));
     }
 
     /// INTERNAL ///
--- a/src/MorphoV2.sol
+++ b/src/MorphoV2.sol
@@ -305,6 +305,7 @@ contract MorphoV2 is IMorphoV2 {
         consumed[msg.sender][group] += amount;
     }
 
+    /// @dev TODO: is it safe enough?
     function shuffleNonce() external {
         nonce[msg.sender] = uint256(keccak256(abi.encode(nonce[msg.sender], blockhash(0))));
     }
--- a/src/MorphoV2.sol
+++ b/src/MorphoV2.sol
@@ -28,6 +28,7 @@ contract MorphoV2 is IMorphoV2 {
     /// OCO (One-Cancels-the-Other) orders. Note that OCO orders work better if all offers have the same amount,
     /// otherwise one might not be takable anymore while an other one at the same nonce is still takeable.
     mapping(address user => mapping(uint256 nonce => uint256)) public consumed;
+    mapping(address user => uint256) public globalNonce;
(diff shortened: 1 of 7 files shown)


adhusson (commented):
the best one imo

For normal group offer canceling (offers with the same nonce), the user can make an offer with the same nonce & consume it, and it keeps other nonces active. For emergency global nonce avoids having to move to a new account.

src/MorphoV2.sol
```diff
@@ -297,6 +301,14 @@ contract MorphoV2 is IMorphoV2 {
         return seizures;
     }
 
+    function consume(uint256 group, uint256 amount) external {
+        consumed[msg.sender][group] += amount;
+    }
+
+    function shuffleNonce(uint256 salt) external {
```
adhusson: need authorization mgmt (made an issue)

adhusson (approved):

src/MorphoV2.sol
```diff
@@ -297,6 +301,14 @@ contract MorphoV2 is IMorphoV2 {
...
+        nonce[msg.sender] = uint256(keccak256(abi.encode(nonce[msg.sender], salt)));
+    }
```
QGarchery: (remark) those should be doable by an account that `msg.sender` authorized

MathisGD: yes see #182

src/MorphoV2.sol
```diff
@@ -297,6 +301,15 @@ contract MorphoV2 is IMorphoV2 {
         return seizures;
     }
 
+    function consume(uint256 group, uint256 amount) external {
+        consumed[msg.sender][group] += amount;
+    }
+
+    /// @dev TODO: is it safe enough?
+    function shuffleNonce() external {
+        nonce[msg.sender] = uint256(keccak256(abi.encode(nonce[msg.sender], blockhash(0))));
```
```suggestion
        nonce[msg.sender] = uint256(keccak256(abi.encode(nonce[msg.sender], blockhash(block.number - 1))));
```

adhusson (approved):

src/MorphoV2.sol
```diff
@@ -24,10 +24,13 @@ contract MorphoV2 is IMorphoV2 {
...
-    /// OCO (One-Cancels-the-Other) orders. Note that OCO orders work better if all offers have the same amount,
-    /// otherwise one might not be takable anymore while an other one at the same nonce is still takeable.
-    mapping(address user => mapping(uint256 nonce => uint256)) public consumed;
+    /// @dev Groups are useful to have a global offered amount shared accross multiple offers ("OCO").
+    /// @dev To work as expected, all offers in a same group should have the same amount and the same loan asset.
+    mapping(address user => mapping(uint256 group => uint256)) public consumed;
+
+    /// @dev Offers should have this exact nonce to be valid.
+    /// @dev The nonce can be shuffled by the user to cancel everything easily/efficiently.
+    mapping(address user => uint256) public nonce;
```
```suggestion
    mapping(address user => bytes32) public nonce;
```
as the nonce is now a hash, it could be clearer to use `bytes32`

QGarchery (approved):

peyha (commented):
[Took a look](https://chatgpt.com/c/68f763d3-4088-832c-bca8-397b6635659f) at potential blockhash behaviors on other EVM chains than Ethereum, the opcode is quite consistant across different chains (only exception being some ZK-rollup such as Scroll return a pseudo-hash but it seems to still work well)

adhusson (approved):

PR #180, opened 2025-10-21
feat: consume nonce

MathisGD:
avoids having to craft an offer and consume it to cancel a nonce


MathisGD (commented):
nb: useless if we go for #177

src/MorphoV2.sol
```diff
@@ -297,6 +297,10 @@ contract MorphoV2 is IMorphoV2 {
         return seizures;
     }
 
+    function consumeNonce(uint256 nonce, uint256 amount) external {
+        consumed[msg.sender][nonce] += amount;
```
adhusson, line 303: todo add authorizations

PR #183, opened 2025-10-22
feat: post maturity trading

### LINKED ISSUE #23: Should it be possible to trade after the term ?
Right, but what about offsetting positions? You don't want to account those positions when the term is met, do you?

_Originally posted by @colin-morpho in https://github.com/morpho-org/v2/pull/21#discussion_r2177505506_

MathisGD:
https://morpholabs.slack.com/archives/C02N7CZ088N/p1761117921639909

### DIFF
--- a/test/TakeTest.sol
+++ b/test/TakeTest.sol
@@ -57,16 +57,6 @@ contract TakeTest is BaseTest {
         morphoV2.supplyCollateral(obligation, obligation.collaterals[0].token, 135, borrower);
     }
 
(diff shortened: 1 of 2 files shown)


test/TakeTest.sol
peyha, line 27: should there be a test with `maturity < block.timestamp` ?

peyha (approved):

adhusson (approved):

QGarchery (approved):

PR #187, opened 2025-10-24
[ratification] ratify callback, arbitrary data

adhusson:


src/MorphoV2.sol
```diff
@@ -85,9 +94,7 @@ contract MorphoV2 is IMorphoV2 {
         uint256 obligationShares,
         address taker,
         Offer memory offer,
-        Signature memory sig,
-        bytes32 root,
-        bytes32[] memory proof,
+        bytes memory ratificationData,
```
adhusson: arbitrary ratification data because 
- signature can change in the future
- ratification contract may want more data

src/MorphoV2.sol
```diff
@@ -101,9 +108,27 @@ contract MorphoV2 is IMorphoV2 {
         require(offer.obligation.chainId == block.chainid, "chain id mismatch");
         require(offer.start < offer.expiry || offer.expiryPrice == offer.startPrice, "inconsistent prices");
         require(offer.maker != taker, "buyer and seller cannot be the same");
-        require(_signer(root, sig) == offer.maker, "invalid signature");
-        require(MathLib.isLeaf(root, keccak256(abi.encode(offer)), proof), "invalid proof");
         require(offer.nonce == nonce[offer.maker], "invalid nonce");
+        require(msg.sender == taker || authorized[taker][msg.sender], "order not authorized");
+
+        if (offer.ratifier != address(0)) {
```
adhusson: offer is OK if

1. an authorized contract ratifies it, or
2. a correct merkle root is signed, or
3. a correct merkle root was ratified, or
4. the sender is authorized

rationale:

1. Is for contracts like vaults that ratify in the take. They need to be whitelisted by the router. 
2. Is the normal EOA case.
3. Is for contracts that need to ratify in advance. They don't need to be whitelisted.
4. Is when users allow a third party to trade for them (they are doing a match of 2 offers using 2 takes).

PR #188, opened 2025-10-24
feat: post-maturity liq (merged)

### LINKED ISSUE #19: Implement liquidation after maturity

MathisGD:
alternative for #79 

fixes #19

some notes:
- we use the fact that LIF<=market LIF. This allows us to compute the maxRepayableDebt always in the same way.
- the require "not liquidatable" makes it clear what can make you liquidatable
- thinking in terms of LIF instead of discount allows to basically factorize everything

### DIFF
--- a/src/MorphoV2.sol
+++ b/src/MorphoV2.sol
@@ -229,8 +229,9 @@ contract MorphoV2 is IMorphoV2 {
         SafeTransferLib.safeTransfer(collateral, msg.sender, assets);
     }
 
-    /// @notice Execute the given collection of `seizures` on the given `obligation` of the given `borrower`.
     /// @dev On each seizure at least one of `repaid` or `seized` should be equal to zero.
+    /// @dev Accounts are liquidatable if they are unhealthy or of the maturity is reached.
+    /// @dev If an account is healthy, the LIF grows linearly from 1 at maturity to MAX_LIF at maturity + 15 minutes.
     /// @param obligation The obligation.
     /// @param seizures An array of amounts of debt to repay or assets to seize with the index of the collateral in the
     /// obligation's collateral assets.
--- a/src/MorphoV2.sol
+++ b/src/MorphoV2.sol
@@ -249,7 +249,8 @@ contract MorphoV2 is IMorphoV2 {
 
     /// @dev On each seizure at least one of `repaid` or `seized` should be equal to zero.
     /// @dev Accounts are liquidatable if they are unhealthy or if the maturity is reached.
-    /// @dev If an account is healthy, the LIF grows linearly from 1 at maturity to MAX_LIF at maturity + AUCTION_DURATION.
+    /// @dev If an account is healthy, the LIF grows linearly from 1 at maturity to MAX_LIF at maturity +
+    /// AUCTION_DURATION.
     /// @param obligation The obligation.
     /// @param seizures An array of amounts of debt to repay or assets to seize with the index of the collateral in the
     /// obligation's collateral assets.
--- a/src/MorphoV2.sol
+++ b/src/MorphoV2.sol
@@ -248,8 +248,8 @@ contract MorphoV2 is IMorphoV2 {
     }
 
     /// @dev On each seizure at least one of `repaid` or `seized` should be equal to zero.
-    /// @dev Accounts are liquidatable if they are unhealthy or of the maturity is reached.
-    /// @dev If an account is healthy, the LIF grows linearly from 1 at maturity to MAX_LIF at maturity + 15 minutes.
+    /// @dev Accounts are liquidatable if they are unhealthy or if the maturity is reached.
+    /// @dev If an account is healthy, the LIF grows linearly from 1 at maturity to MAX_LIF at maturity + AUCTION_DURATION.
     /// @param obligation The obligation.
     /// @param seizures An array of amounts of debt to repay or assets to seize with the index of the collateral in the
     /// obligation's collateral assets.
--- a/src/MorphoV2.sol
+++ b/src/MorphoV2.sol
@@ -4,7 +4,7 @@ pragma solidity 0.8.28;
 
 import {UtilsLib} from "./libraries/UtilsLib.sol";
 import {SafeTransferLib} from "./libraries/SafeTransferLib.sol";
-import {WAD, ORACLE_PRICE_SCALE, MAX_LIF, DUTCH_SPEED} from "./libraries/ConstantsLib.sol";
+import {WAD, ORACLE_PRICE_SCALE, MAX_LIF, AUCTION_DURATION} from "./libraries/ConstantsLib.sol";
 import {MathLib} from "./libraries/MathLib.sol";
 import {IOracle} from "./interfaces/IOracle.sol";
 import {IMorphoV2, Obligation, Offer, Signature, Seizure} from "./interfaces/IMorphoV2.sol";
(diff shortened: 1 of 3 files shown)


src/libraries/ConstantsLib.sol
```diff
@@ -4,4 +4,5 @@ pragma solidity ^0.8.0;
 
 uint256 constant WAD = 1e18;
 uint256 constant ORACLE_PRICE_SCALE = 1e36;
-uint256 constant LIQUIDATION_INCENTIVE_FACTOR = 1.15e18;
+uint256 constant MAX_LIF = 1.15e18;
+uint256 constant DUTCH_SPEED = (MAX_LIF - WAD) / 15 minutes;
```
peyha:
```suggestion
uint256 constant AUCTION_DURATION = 15 minutes;
uint256 constant DUTCH_SPEED = (MAX_LIF - WAD) / AUCTION_DURATION
```
prob a bit cleaner if people want to tweak the duration

peyha (commented):
Merging both liquidation endpoints is quite clean indeed
One of the main difference with #79 is that the bad debt check is computed with `MAX_LIF` and the oracle price here while it was computed with the discounted oracle price before. Theoretically, it does not change anything because in the current setting (dutch auction ending with same bonus as liquidation) there can only be bad debt at the discounted oracle price if there is already bad debt at `MAX_LIF` and oracle price

src/libraries/ConstantsLib.sol
MathisGD: idk as the auction is technically not finished at 15 min, just it stops decreasing the price

src/libraries/ConstantsLib.sol
```diff
@@ -4,4 +4,5 @@ pragma solidity ^0.8.0;
 
 uint256 constant WAD = 1e18;
 uint256 constant ORACLE_PRICE_SCALE = 1e36;
-uint256 constant LIQUIDATION_INCENTIVE_FACTOR = 1.15e18;
+uint256 constant MAX_LIF = 1.15e18;
```
QGarchery: We could add a comment explaining what `LIF` means

src/libraries/ConstantsLib.sol
peyha: then `DECREASE_DURATION` instead of `AUCTION_DURATION` ?

QGarchery (approved):
General idea seems good so I'm ok to merge, not sure if we want those exact params though

Those tests would be nice to have:
- check that the LIF formula can't go over MAX_LIF
- check when positions are liquidatable

src/libraries/ConstantsLib.sol
MathisGD: actually the current version is creating a rounding, I'll change it

test/LiquidationTest.sol
```diff
@@ -150,4 +174,48 @@ contract LiquidationTest is BaseTest {
...
         recordedData = data;
     }
+
+    // post maturity liquidation
+
+    function testLiquidatePostMaturityFullLIF(uint256 delay) public {
+        delay = bound(delay, 0, 100 weeks);
+
+        setupObligation(obligation, 100);
+        vm.warp(obligation.maturity + 15 minutes + delay);
```
peyha:
```suggestion
        vm.warp(obligation.maturity + AUCTION_DURATION + delay);
```

src/MorphoV2.sol
```diff
@@ -230,7 +230,7 @@ contract MorphoV2 is IMorphoV2 {
     }
 
     /// @notice Execute the given collection of `seizures` on the given `obligation` of the given `borrower`.
```
peyha: We could add a dev comment to explain the behavior between maturity and maturity + AUCTION_DURATION

QGarchery (approved):

src/MorphoV2.sol
```diff
@@ -247,8 +247,9 @@ contract MorphoV2 is IMorphoV2 {
         SafeTransferLib.safeTransfer(collateral, msg.sender, assets);
     }
 
-    /// @notice Execute the given collection of `seizures` on the given `obligation` of the given `borrower`.
-    /// @dev On each seizure either `repaid` or `seized` should be equal to zero.
+    /// @dev On each seizure at least one of `repaid` or `seized` should be equal to zero.
+    /// @dev Accounts are liquidatable if they are unhealthy or of the maturity is reached.
```
```suggestion
    /// @dev Accounts are liquidatable if they are unhealthy or if the maturity is reached.
```

src/MorphoV2.sol
```diff
@@ -247,8 +247,9 @@ contract MorphoV2 is IMorphoV2 {
...
+    /// @dev If an account is healthy, the LIF grows linearly from 1 at maturity to MAX_LIF at maturity + 15 minutes.
```
```suggestion
    /// @dev If an account is healthy, the LIF grows linearly from 1 at maturity to MAX_LIF at maturity + AUCTION_DURATION.
```

src/MorphoV2.sol
```diff
@@ -269,12 +270,15 @@ contract MorphoV2 is IMorphoV2 {
...
-            repayableDebt += collateralAmount.mulDivUp(WAD, LIQUIDATION_INCENTIVE_FACTOR)
-                .mulDivUp(prices[i], ORACLE_PRICE_SCALE);
+            repayableDebt += collateralAmount.mulDivUp(WAD, MAX_LIF).mulDivUp(prices[i], ORACLE_PRICE_SCALE);
         }
 
         uint256 originalDebt = debtOf[borrower][id];
-        require(originalDebt > maxDebt, "position is healthy");
+        require(block.timestamp > obligation.maturity || originalDebt > maxDebt, "position is not liquidatable");
+
+        uint256 lif = originalDebt > maxDebt
```
peyha, line 280:
```suggestion
        uint256 lif = originalDebt > maxDebt || block.timestamp > obligation.maturity + AUCTION_DURATION
```
looks clearer (and removes the min on the auction lif computation)

MathisGD: @QGarchery thoughts?

MathisGD: one can argue that right now is clearer. you have the path for unhealthy and the path for healthy (but post maturity)

QGarchery: No strong opinion. I like the the current code slightly better, one nice thing is that it doesn't duplicate `AUCTION_DURATION`

test/LiquidationTest.sol
```diff
@@ -150,4 +174,48 @@ contract LiquidationTest is BaseTest {
         recordedLiquidator = liquidator;
```
peyha, line 150: we might be missing a test with bad debt after maturity (though it's supposed to be the same computation as before maturity)

peyha (approved):

src/libraries/ConstantsLib.sol
```diff
@@ -4,4 +4,5 @@ pragma solidity ^0.8.0;
 
 uint256 constant WAD = 1e18;
 uint256 constant ORACLE_PRICE_SCALE = 1e36;
-uint256 constant LIQUIDATION_INCENTIVE_FACTOR = 1.15e18;
+uint256 constant MAX_LIF = 1.15e18; // Liquidation Incentive Factor
+uint256 constant AUCTION_DURATION = 15 minutes;
```
adhusson, line 8: I agree with [this comment](https://github.com/morpho-org/morpho-v2/pull/188#discussion_r2460116326) that the name is misleading.

Issue #134, opened 2025-10-02
max fees


MathisGD:
should we implement a maximum fee ?
for instance, it was 25% on blue, 50% on vault-v2 (max perf fee)
(an issue can also be opened if we want to decide later)

_Originally posted by @peyha in https://github.com/morpho-org/morpho-v2/pull/106#discussion_r2399157908_

MathisGD:
see #398

PR #191, opened 2025-10-24
feat: trading fee with max (version 2)

MathisGD:
variant of #186


test/TradingFeeTest.sol
```diff
@@ -176,6 +254,73 @@ contract TradingFeeTest is BaseTest {
...
+        assertApproxEqAbs(
+            loanToken.balanceOf(feeRecipient), feeRecipientBalanceBefore + expectedFee, 100, "fee recipient balance"
+        );
+        assertApproxEqAbs(
+            loanToken.balanceOf(borrower), borrowerBalanceBefore + expectedSellerAssets, 100, "borrower balance"
+        );
+        assertApproxEqAbs(loanToken.balanceOf(lender), lenderBalanceBefore - expectedBuyerAssets, 100, "lender balance");
+    }
+
+    function testTradingFeeWithMaxSellerAssetsInput() public {
```
MathisGD: todo: test the 6 cases

src/interfaces/IMorphoV2.sol
```diff
@@ -46,4 +46,9 @@ struct Seizure {
     uint256 seized;
 }
 
+struct TradingFee {
+    uint128 slope;
+    uint128 max;
```
MathisGD: maybe find better names

peyha:
The new formula looks correct, it's essentially the same as before (with the previous variable `tradingFee` renamed `slope`) while adding a `max` cap fee
Not a huge fan of the `slope` naming because it makes the formula difficult to understand (maybe it's just me)

src/interfaces/IMorphoV2.sol
MathisGD:
```suggestion
    uint128 maxCutOnInterest;
    uint128 maxCutOnCash;
```

MathisGD:
```suggestion
    uint128 maxInterestCut;
    uint128 maxCashCut;
```

MathisGD: maybe we could have something that makes it understand that it's computed on seller assets and not buyer assets

src/MorphoV2.sol
```diff
@@ -73,10 +74,11 @@ contract MorphoV2 is IMorphoV2 {
         feeSetter = newFeeSetter;
     }
 
-    function setTradingFee(bytes32 id, uint256 fee) external {
+    function setTradingFee(bytes32 id, uint128 slope, uint128 max) external {
         require(msg.sender == feeSetter, "Only feeSetter");
-        require(fee <= WAD, "Fee too high");
-        tradingFee[id] = fee;
+        require(slope <= WAD, "Slope too high");
+        require(max <= WAD, "Max too high");
```
MattLsb: max could be <= Max fee (not defined yet though)

MathisGD: yes, pending #134

src/interfaces/IMorphoV2.sol
MattLsb:
```suggestion
    uint128 growth;
    uint128 cap;
```

MattLsb: Depending on how we decide to name the global max-fee parameter: if it's max, it could make sense to name this one cap (and vice versa)

MattLsb: `interestCut` might be misleading since this is a trading fee
`growth` feels a bit more intuitive and emphasizes that we don't want the fee to be greater than interest, so needs to grow with it at the beginning

MathisGD: agree, but it's still part of the trading fee struct. I think that's fine

I don't really get "growth" tbh.

adhusson: slope/growth is the complement of a slope/growth or the complement of the inverse so it would be misleading. What about `interestCutFactor` and `priceIncreaseCap`?

src/interfaces/IMorphoV2.sol
```diff
@@ -46,4 +46,9 @@ struct Seizure {
     uint256 seized;
 }
 
+struct TradingFee {
+    uint128 interestCut;
+    uint128 cashCut;
```
adhusson: `interestCutFactor`
`priceIncreaseCap`

adhusson:
> Not a huge fan of the slope naming because it makes the formula difficult to understand (maybe it's just me)

also misleading it's a complement / complement of inverse of slope

adhusson (approved):
ok to merge anyway we can evolve naming later

src/MorphoV2.sol
```diff
@@ -7,13 +7,14 @@ import {SafeTransferLib} from "./libraries/SafeTransferLib.sol";
...
 import {IOracle} from "./interfaces/IOracle.sol";
-import {IMorphoV2, Obligation, Offer, Signature, Seizure} from "./interfaces/IMorphoV2.sol";
+import {IMorphoV2, Obligation, Offer, Signature, Seizure, TradingFee} from "./interfaces/IMorphoV2.sol";
 import {ICallbacks, IFlashLoanCallback} from "./interfaces/ICallbacks.sol";
 
 /// OBLIGATIONS
 /// @dev Obligations' collaterals must be sorted by token address.
 contract MorphoV2 is IMorphoV2 {
     using MathLib for uint256;
+    using MathLib for uint128;
```
peyha: the file compiles fine without this line

peyha (approved):
lgtm

src/MorphoV2.sol
```suggestion
```

MathisGD:
@codex please review

src/interfaces/IMorphoV2.sol
MathisGD: @codex any naming idea?

src/MorphoV2.sol
```diff
@@ -140,9 +141,19 @@ contract MorphoV2 is IMorphoV2 {
             : offer.startPrice;
         require(offerPrice <= WAD, "price too high");
 
-        uint256 _tradingFee = tradingFee[id];
-        uint256 buyerPrice = offer.buy ? offerPrice : offerPrice.mulDivDown(WAD - _tradingFee, WAD) + _tradingFee;
-        uint256 sellerPrice = offer.buy ? (offerPrice - _tradingFee).mulDivDown(WAD, WAD - _tradingFee) : offerPrice;
+        TradingFee memory _tradingFee = tradingFee[id];
```
```suggestion
        (uint128 interestCut, uint128 cashCut) = (tradingFee[id].interestCut, tradingFee[id].cashCut);
```
allows to remove the `_tradingFee.` prefix

src/interfaces/IMorphoV2.sol
QGarchery: what about (in that order, to change in the `buyerPrice` and `sellerPrice` as well):
`cashCut` -> `tradingFee`
`interestCut` -> `interestCutLimit`

Then there is the question of the name of the struct, it can be `TradingFeeParams`, but it doesn't matter much with the above suggestion

test/SettersTest.sol
```diff
@@ -33,23 +33,32 @@ contract SettersTest is BaseTest {
         morphoV2.setFeeSetter(makeAddr("newFeeSetter"));
     }
 
-    function testSetTradingFeeSuccess(bytes32 id, uint256 fee) public {
-        vm.assume(fee <= 1e18);
-        morphoV2.setTradingFee(id, fee);
-        assertEq(morphoV2.tradingFee(id), fee);
+    function testSetTradingFeeSuccess(bytes32 id, uint128 slope, uint128 max) public {
```
QGarchery: some naming to change (less important since it's in the tests)

QGarchery (approved):

src/MorphoV2.sol
MathisGD: you compute twice the slot though

Issue #190, opened 2025-10-24
check that all state is reconstructable through events


MathisGD:
just before sending to audits

adhusson:
see https://github.com/morpho-org/midnight/compare/main...sql

PR #192, opened 2025-10-24
Implement events 

### LINKED ISSUE #47: implement events

peyha:
fixes #47 

Related to #190 and #47

To be iterated on

- estimate the additional gas of logging the full offer (according to the number of collateral)


src/MorphoV2.sol
```diff
@@ -199,6 +210,10 @@ contract MorphoV2 is IMorphoV2 {
 
         require(_isHealthy(offer.obligation, seller), "Seller is unhealthy");
 
+        emit EventsLib.Take(
+            id, buyer, seller, (buyer == offer.maker), buyerAssets, sellerAssets, obligationUnits, obligationShares
+        );
```
MathisGD: I would add
- msg.sender
- the full offer (we need to group to be able to reconstruct the consumed mapping)

Also I'm thinking that it might be hard to recompute the balances, because of the fact that there are 4 cases

src/MorphoV2.sol
```diff
@@ -315,6 +338,8 @@ contract MorphoV2 is IMorphoV2 {
 
         SafeTransferLib.safeTransferFrom(obligation.loanToken, msg.sender, address(this), totalRepaid);
 
+        emit EventsLib.Liquidate(id, borrower, seizures, badDebt);
```
MathisGD: to add: msg.sender, totalRepaid

src/MorphoV2.sol
MathisGD: add an event

MathisGD, line 421: add an event

src/MorphoV2.sol
```diff
@@ -228,13 +245,17 @@ contract MorphoV2 is IMorphoV2 {
...
+        emit EventsLib.Repay(id, msg.sender, onBehalf, obligationUnits);
     }
 
     function supplyCollateral(Obligation memory obligation, address collateral, uint256 assets, address onBehalf)
         external
     {
         collateralOf[onBehalf][_id(obligation)][collateral] += assets;
         SafeTransferLib.safeTransferFrom(collateral, msg.sender, address(this), assets);
+
+        emit EventsLib.SupplyCollateral(_id(obligation), msg.sender, onBehalf, collateral, assets);
```
MathisGD: keep the id in a variable

src/MorphoV2.sol
```diff
@@ -245,6 +266,8 @@ contract MorphoV2 is IMorphoV2 {
         require(_isHealthy(obligation, onBehalf), "Unhealthy borrower");
 
         SafeTransferLib.safeTransfer(collateral, msg.sender, assets);
+
+        emit EventsLib.WithdrawCollateral(_id(obligation), msg.sender, onBehalf, collateral, assets);
```
MathisGD: id to keep in a variable

src/MorphoV2.sol
MathisGD, line 1: about the order of arguments, let's try to follow the functions parameters, or the order with which they appear in the function

src/MorphoV2.sol
peyha: I was wondering if we should log the full obligation (which is contained in the full offer) because it's quite redundant compared to the obligation id. Looks like there is no better solution for now so I'll log the offer

src/libraries/EventsLib.sol
```diff
@@ -13,19 +13,16 @@ library EventsLib {
     event SetTradingFeeRecipient(address indexed recipient);
 
     event Take(
-        bytes32 indexed obligationId,
```
peyha: don't we still want to log obligationId ?
It can be calculated by hashing offer.obligation but it would simplify indexers work if it is logged during take (especially if it is indexed)

src/libraries/EventsLib.sol
```diff
@@ -13,19 +13,16 @@ library EventsLib {
...
         uint256 buyerAssets,
         uint256 sellerAssets,
         uint256 obligationUnits,
         uint256 obligationShares,
+        address indexed taker,
         Offer offer
     );
     event Withdraw(
-        bytes32 indexed obligationId, address caller, address indexed onBehalf, uint256 obligationUnits, uint256 shares
+        address indexed caller, bytes32 indexed obligationId, address indexed onBehalf, uint256 obligationUnits, uint256 shares
```
peyha: why use this order ? 
On Morpho Blue, we logged `marketId` first and `caller` second. It feels a bit more natural to me as `obligationId` is much more informative than `caller` in practice

peyha: ah, got it, it's to keep the same order as the parameters of `withdraw()`

certora/MorphoV2.spec
```diff
@@ -49,9 +49,6 @@ rule takeInputOutputConsistency(env e, uint256 buyerAssets, uint256 sellerAssets
 strong invariant notBorrowerAndLender(bytes32 id, address user)
     sharesOf(user, id) == 0 || debtOf(user, id) == 0;
 
-strong invariant totalUnitsEqualsSumDebtPlusWithdrawable(bytes32 id)
-    totalUnits(id) == sumDebtOf[id] + withdrawable(id);
```
MathisGD, line 84: why removing

peyha: you removed it at [this commit](https://github.com/morpho-org/morpho-v2/pull/192/commits/bdb74016a45c0d12a29ab1d2859a3e32dbb63935), did you want to remove it on another PR ?

MathisGD: I think that it messed up, it was not for this pr ^^
it should be added back

peyha:
With these events, it should be possible to re-construct: `sharesOf`, `debtOf`, `withdrawable`, `totalUnits`, `totalShares`, `collateralOf`, `consumed`, `nonce`, `tradingFeeParams`, `tradingFeeRecipient`, `owner`and `feeSetter`

- `withdrawable` can be re-constructed with `Withdraw`, `Repay` and `Liquidate`
- `collateralOf` can be re-constructed with `SupplyCollateral`, `WithdrawCollateral`, `Liquidate` (though it might be difficult to link the seizure array with `SupplyCollateral` and `WithdrawCollateral`
- `consumed` can be re-constructed with `Take` and `Consume`
- `nonce` can be re-constructed with `ShuffleNonce` 
- `tradingFeeParams` can be re-constructed with `SetTradingFee`
- `tradingFeeRecipient` can be re-constructed with `SetTradingFeeRecipient`
- `owner` can be re-constructed with `Constructor` and `SetOwner`
- `feeSetter` can be re-constructed with `SetFeeSetter` 

Currently, we can't re-construct `sharesOf`, `debtOf`, `totalUnits` and `totalShares` because it depends on the path taken in `take()` and it's not logged in the `Take` event so I would suggest logging a `path` uint which would correspond to one of the path

peyha:
created #241 so we can merge this

src/MorphoV2.sol
```diff
@@ -47,6 +48,8 @@ contract MorphoV2 is IMorphoV2 {
 
     constructor() {
         owner = msg.sender;
+
```
```suggestion
```

src/MorphoV2.sol
```diff
@@ -213,6 +223,19 @@ contract MorphoV2 is IMorphoV2 {
...
+            msg.sender,
+            id,
+            buyerAssets,
+            sellerAssets,
+            obligationUnits,
+            obligationShares,
+            taker,
+            offer,
+            previousBuyerDebt,
+            previousSellerShares
```
MathisGD: an other path would be to log the final sharesOf, and totalUnits/shares so that people reconstruct this state super easily (that's what I had in mind for this option btw)

src/MorphoV2.sol
MathisGD, line 73: only functions without event, but that's expected

src/MorphoV2.sol
```diff
@@ -81,11 +85,13 @@ contract MorphoV2 is IMorphoV2 {
...
         tradingFeeParams[id] =
             TradingFeeParams({tradingFee: uint128(tradingFee), interestCutLimit: uint128(interestCutLimit)});
+        emit EventsLib.SetTradingFee(id, tradingFee, interestCutLimit);
     }
 
     function setTradingFeeRecipient(address recipient) external {
         require(msg.sender == owner, "Only owner");
         tradingFeeRecipient = recipient;
+        emit EventsLib.SetTradingFeeRecipient(recipient);
     }
```
MathisGD, line 103: msg.sender isn't logged in admin setters because it's easy to find back

src/MorphoV2.sol
```diff
@@ -213,6 +221,19 @@ contract MorphoV2 is IMorphoV2 {
...
 
+        emit EventsLib.Take(
+            msg.sender,
+            id,
+            buyerAssets,
+            sellerAssets,
+            obligationUnits,
+            obligationShares,
+            taker,
+            offer,
```
MathisGD: note that this contains the obligation. we will see this in #241 anyway

MathisGD (approved):

MathisGD:
@codex please review.

focus on two things:
- consistency between events (when they are emitted, order of parameters etc)
- the ability to reconstruct the state of the contract easily in SQL just with events

src/MorphoV2.sol
```diff
@@ -213,6 +221,19 @@ contract MorphoV2 is IMorphoV2 {
             totalUnits[id] -= obligationUnits;
         }
 
+        emit EventsLib.Take(
```
QGarchery, line 232: In the offer there is already the `callback` and the `callbackData`, why don't we log the `takerCallback` and the `takerCallbackData` as well ? This would allow to rebuild anything that is happening in the transaction

src/MorphoV2.sol
```diff
@@ -356,6 +390,8 @@ contract MorphoV2 is IMorphoV2 {
         withdrawable[id] += totalRepaid;
         debtOf[borrower][id] -= totalRepaid;
 
+        emit EventsLib.Liquidate(msg.sender, id, seizures, borrower, totalRepaid, badDebt);
+
         for (uint256 i = 0; i < seizures.length; i++) {
             Seizure memory seizure = seizures[i];
             SafeTransferLib.safeTransfer(
```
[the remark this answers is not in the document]
QGarchery: It logs the updated values, as expected. See https://github.com/morpho-org/morpho-v2/pull/265

QGarchery (approved):

src/MorphoV2.sol
QGarchery: The same thing could be said for `flashLoan`, but I believe it is slightly different because the callback of the flashloan doesn't interact with the logic of MorphoV2. I mean it could, but there is no MorphoV2 specific logic that happens after (like the `_isHealthy` check in `take`)

peyha: good point for the `takerCallback`, we never logged the callback data in morpho blue and no one ever complained about this but `takerCallback` could be useful

QGarchery:
Putting it here so we don't forget: removing `offer` from the `Take` event will not allow to reconstruct the state, it is needed to know the `maker` notably

peyha:
`offer.group` is also required to re-construct `consumed`

MathisGD (approved):

adhusson (approved):

PR #195, opened 2025-10-26
chore: use wad consistently

MathisGD:


QGarchery (approved):

peyha (approved):
WAD should also be used in tests (but it can wait because tests will need to be re-worked later on)

PR #197, opened 2025-10-26
feat: flashloan

### LINKED ISSUE #194: implement flashloans (of the collateral)

MathisGD:
fixes #194

### DIFF
--- a/src/MorphoV2.sol
+++ b/src/MorphoV2.sol
@@ -8,7 +8,7 @@ import {WAD, ORACLE_PRICE_SCALE, LIQUIDATION_INCENTIVE_FACTOR} from "./libraries
 import {MathLib} from "./libraries/MathLib.sol";
 import {IOracle} from "./interfaces/IOracle.sol";
 import {IMorphoV2, Obligation, Offer, Signature, Seizure} from "./interfaces/IMorphoV2.sol";
-import {ICallbacks} from "./interfaces/ICallbacks.sol";
+import {ICallbacks, IFlashloanCallback} from "./interfaces/ICallbacks.sol";
 
 /// OBLIGATIONS
 /// @dev Obligations' collaterals must be sorted by token address.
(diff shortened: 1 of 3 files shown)


src/MorphoV2.sol
```diff
@@ -327,6 +327,14 @@ contract MorphoV2 is IMorphoV2 {
         nonce[msg.sender] = keccak256(abi.encode(nonce[msg.sender], blockhash(block.number - 1)));
     }
 
+    function flashloan(address token, uint256 amount, address callback, bytes calldata data) external {
```
```suggestion
    function flashloan(address token, uint256 amount, bytes calldata data) external {
```
another possibility would be to make the flashloan callback on msg.sender instead of `address callback`. It's the current standard for flashloans (and what we implemented on Blue) and the logic is the same 
the main drawback is that it's not consistent with the callback behavior of `take`

peyha: nitpick but we used `flashLoan` instead of `flashloan` on Blue (though I prefer the current naming)

```suggestion
    function flashLoan(address token, uint256 amount, address callback, bytes calldata data) external {
```
as in Morpho Blue

src/MorphoV2.sol
```diff
@@ -327,6 +327,14 @@ contract MorphoV2 is IMorphoV2 {
...
+        SafeTransferLib.safeTransfer(token, msg.sender, amount);
+
+        IFlashloanCallback(callback).onFlashloan(token, amount, data);
```
```suggestion
        IFlashloanCallback(callback).onMorphoV2Flashloan(token, amount, data);
```
because `onFlashLoan` may be common so it can be a slight pain to have to dispatch in the corresponding callback contract (might be ok though). It makes we wonder about other names of callback functions

src/MorphoV2.sol
QGarchery: duplicate of https://github.com/morpho-org/morpho-v2/pull/197#discussion_r2464935916

MathisGD: ah indeed, bearish though

adhusson: `take` callbacks must be like this because callback targets will be whitelisted, it's ok for me to restrict it here.

adhusson (approved):
OK as is, but simpler with always calling back `msg.sender` as mentioned by @peyha

QGarchery (approved):
ok to merge and solve #204 later

peyha (approved):
ok to merge (also created #205)

PR #200, opened 2025-10-27
test: refactoring/simplifications

MathisGD:
first step, then I'll convert them to fuzzing


test/BaseTest.sol
```diff
@@ -12,24 +12,33 @@ import {MorphoV2} from "../src/MorphoV2.sol";
...
 
     function setUp() public virtual {
         morphoV2 = new MorphoV2();
 
         morphoV2.setFeeSetter(address(this));
 
-        (borrower, borrowerSecretKey) = makeAddrAndKey("borrower");
-        (lender, lenderSecretKey) = makeAddrAndKey("lender");
+        uint256 _privateKey;
+        (borrower, _privateKey) = makeAddrAndKey("borrower");
```
peyha, line 34: could be factorized in a function `makeAccount` which would call `makeAddrAndKey`, write the private key in `privateKey` and return an address

test/BaseTest.sol
```diff
@@ -72,10 +81,19 @@ abstract contract BaseTest is Test {
         return res;
     }
 
-    function sig(bytes32 _root, uint256 sk) internal pure returns (Signature memory) {
+    function sig(Offer[1] memory offers) internal view returns (Signature memory) {
+        bytes32 _root = root(offers);
+        bytes32 messageHash = keccak256(bytes.concat("\x19\x45thereum Signed Message:\n32", _root));
+        Signature memory signature;
+        (signature.v, signature.r, signature.s) = vm.sign(privateKey[offers[0].maker], messageHash);
```
peyha, line 88: would it make sense to require `privateKey[offers[0].maker]` ? not sure that the test would revert if the private key of the address is not defined

test/TakeTest.sol
```diff
@@ -57,20 +57,36 @@ contract TakeTest is BaseTest {
         morphoV2.supplyCollateral(obligation, obligation.collaterals[0].token, 135, borrower);
     }
 
-    function testLend() public {
+    // helpers.
+
+    // hardcodes the right root, sighature, proof, and callback (no callback)
+    function take(
```
QGarchery, line 63:
```suggestion
    function signOfferAndTake(
```
maybe this is clearer. The name could include also the fact that it takes the right root, but it's fine like that

test/TakeTest.sol
```diff
@@ -57,20 +57,36 @@ contract TakeTest is BaseTest {
         morphoV2.supplyCollateral(obligation, obligation.collaterals[0].token, 135, borrower);
     }
 
-    function testLend() public {
+    // helpers.
+
+    // hardcodes the right root, sighature, proof, and callback (no callback)
```
QGarchery:
```suggestion
    // hardcodes the right root, signature, proof, and callback (no callback)
```

QGarchery (approved):

test/TakeTest.sol
```diff
@@ -871,41 +427,113 @@ contract TakeTest is BaseTest {
...
+    function testTakeBorrowLendCallback() public {
+        vm.prank(otherLender);
+        loanToken.approve(address(morphoV2), 100);
+        lendOffer.callback = address(new LendCallback());
+        lendOffer.callbackData = abi.encode(loanToken, 100);
+        lendOffer.maker = address(otherLender);
+        deal(address(loanToken), lendOffer.callback, 100);
 
-        vm.expectRevert("invalid nonce");
         morphoV2.take(
```
peyha:
```suggestion
        take(100, 0, 0, 0, borrower, lendOffer)
```
because the callback is in the lendOffer (but it's also ok to keep as is because it's consistent with other callback tests)

test/TradingFeeTest.sol
peyha, line 83: could be great to re-use the `take` function from `TakeTest.sol` (it simplifies tests a lot)

MathisGD: it's done in #191

test/BaseTest.sol
MathisGD: yes it reverts with a clear message if the private key is zero

test/BaseTest.sol
MathisGD: I think it's fine this way

peyha (approved):

test/TakeTest.sol
MathisGD: why should it say that it signs? to make it clear that it does more things?

PR #206, opened 2025-10-28
style: auction duration constant name

MathisGD:
https://github.com/morpho-org/morpho-v2/pull/188#discussion_r2468779202

### DIFF
--- a/src/libraries/ConstantsLib.sol
+++ b/src/libraries/ConstantsLib.sol
@@ -5,4 +5,4 @@ pragma solidity ^0.8.0;
 uint256 constant WAD = 1e18;
 uint256 constant ORACLE_PRICE_SCALE = 1e36;
 uint256 constant MAX_LIF = 1.15e18; // Liquidation Incentive Factor
-uint256 constant AUCTION_DURATION = 15 minutes;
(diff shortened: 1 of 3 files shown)


QGarchery (approved):

peyha (approved):

src/libraries/ConstantsLib.sol
```diff
@@ -5,4 +5,4 @@ pragma solidity ^0.8.0;
 uint256 constant WAD = 1e18;
 uint256 constant ORACLE_PRICE_SCALE = 1e36;
 uint256 constant MAX_LIF = 1.15e18; // Liquidation Incentive Factor
-uint256 constant AUCTION_DURATION = 15 minutes;
+uint256 constant TIME_TO_LIF = 15 minutes; // Time to reach MAX_LIF
```
adhusson, line 8:
```suggestion
uint256 constant TIME_TO_MAX_LIF = 15 minutes; // Time to reach MAX_LIF
```

seems much clearer

MathisGD: agree. note that anyway this is going to change, the LIF won't be hardcoded

PR #212, opened 2025-10-29
Round in favor of the passive lenders

QGarchery:
- First commit prepares the future changes and should be a refactor with no functional change (tests go through)
- Second commit fixes the conversion units<->shares to be against the users (either taker or maker)

[Related thread](https://morpholabs.slack.com/archives/C02N7CZ088N/p1761904336547709)


src/libraries/MathLib.sol
```diff
@@ -13,6 +13,14 @@ library MathLib {
         return (x * y + (d - 1)) / d;
     }
 
+    function mulDiv(uint256 x, uint256 y, uint256 d, bool roundUp) internal pure returns (uint256) {
+        if (roundUp) {
+            return mulDivUp(x, y, d);
+        } else {
+            return mulDivDown(x, y, d);
+        }
+    }
```
MathisGD, line 18: honest work
```suggestion
    function mulDiv(uint256 x, uint256 y, uint256 d, bool roundUp) internal pure returns (uint256) {
        uint256 roundUpUint = roundUp ? 1 : 0;
        return (x * y + roundUpUint * (d - 1)) / d;
    }
```

adhusson: might as well convert bool to uint with assembly at this point

peyha: even more honest work
```suggestion
function mulDiv(uint256 x, uint256 y, uint256 d, bool roundUp) internal pure returns (uint256 result) {
    assembly { result := div(add(mul(x, y), mul(iszero(iszero(roundUp)), sub(d, 1))), d) }
}
    }
```

QGarchery: Closing since I find the new implem clearer, and it's only about optimization

QGarchery:
Closing in favor of #270

PR #217, opened 2025-10-30
style: trading fee computation

MathisGD:
motivation: https://morpholabs.slack.com/archives/C08A40KKN9W/p1761839450840299?thread_ts=1761558720.712149&cid=C08A40KKN9W

### DIFF
--- a/src/MorphoV2.sol
+++ b/src/MorphoV2.sol
@@ -139,23 +139,26 @@ contract MorphoV2 is IMorphoV2 {
             ? offer.startPrice + (offer.expiryPrice - offer.startPrice) * (block.timestamp - offer.start)
                 / (offer.expiry - offer.start)
             : offer.startPrice;
-        require(offerPrice <= WAD, "price too high");
 
         TradingFeeParams memory _tradingFeeParams = tradingFeeParams[id];
-        uint256 buyerPrice = offer.buy
-            ? offerPrice
-            : UtilsLib.min(
-                offerPrice.mulDivDown(WAD - _tradingFeeParams.interestCutLimit, WAD)
-                    + _tradingFeeParams.interestCutLimit,
-                offerPrice.mulDivDown(WAD + _tradingFeeParams.tradingFee, WAD)
-            );
-        uint256 sellerPrice = offer.buy
-            ? UtilsLib.max(
-                (offerPrice - _tradingFeeParams.interestCutLimit)
+        uint256 buyerPrice;
+        uint256 sellerPrice;
+        require(offerPrice <= WAD, "price too high");
+        if (offer.buy) {
+            buyerPrice = offerPrice;
+            sellerPrice = UtilsLib.max(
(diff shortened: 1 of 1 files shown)


src/MorphoV2.sol
```diff
@@ -139,23 +139,26 @@ contract MorphoV2 is IMorphoV2 {
...
-        require(offerPrice <= WAD, "price too high");
 
         TradingFeeParams memory _tradingFeeParams = tradingFeeParams[id];
-        uint256 buyerPrice = offer.buy
-            ? offerPrice
-            : UtilsLib.min(
-                offerPrice.mulDivDown(WAD - _tradingFeeParams.interestCutLimit, WAD)
+        uint256 buyerPrice;
+        uint256 sellerPrice;
+        require(offerPrice <= WAD, "price too high");
```
QGarchery: This makes me think: shouldn't we require the buyerPrice to be less than WAD ?

QGarchery (approved):

src/MorphoV2.sol
MathisGD: i was thinking about that. the problem is less clear to me than generally allowing price arbitrarily high. 

note that doing that would mean that selling above 1-fee is impossible

MathisGD:
certora is just timing out

src/MorphoV2.sol
```diff
@@ -139,23 +139,26 @@ contract MorphoV2 is IMorphoV2 {
...
+        if (offer.buy) {
+            buyerPrice = offerPrice;
+            sellerPrice = UtilsLib.max(
+                (buyerPrice - _tradingFeeParams.interestCutLimit)
+                .mulDivDown(WAD, WAD - _tradingFeeParams.interestCutLimit),
+                buyerPrice.mulDivDown(WAD, WAD + _tradingFeeParams.tradingFee)
```
peyha, line 152: isn't it rather 
```suggestion
            sellerPrice = UtilsLib.max(
                (buyerPrice - _tradingFeeParams.tradingFee)
                .mulDivDown(WAD, WAD - _tradingFeeParams.tradingFee),
                buyerPrice.mulDivDown(WAD, WAD + _tradingFeeParams.interestCutLimit)
```
?

MathisGD: hmm no, interestCutLimit is the "alpha", and tradingFee is the "M"

src/MorphoV2.sol
adhusson: strange to move this require below the declarations

adhusson (approved):

src/MorphoV2.sol
peyha: oh yes, got misled because in the previous formula, tradingFee was the alpha

src/MorphoV2.sol
MathisGD: #226

PR #219, opened 2025-10-30
Add default trading fee fallback per loan token

### LINKED ISSUE #175: default fee

MathisGD:
- add default trading fee parameters per loan token with activation flag
- extend trading fee setter and add default setter gated by the feeSetter role
- update matching logic and tests to cover default and deactivated fee scenarios

fixes #175 

https://chatgpt.com/codex/tasks/task_b_690393fbf2dc8321b87a232426da0913

### DIFF
--- a/src/MorphoV2.sol
+++ b/src/MorphoV2.sol
@@ -75,22 +75,26 @@ contract MorphoV2 is IMorphoV2 {
         feeSetter = newFeeSetter;
     }
 
-    function setTradingFee(bytes32 id, uint128 tradingFee, uint128 interestCutLimit, bool activated) external {
+    function setTradingFee(bytes32 id, uint256 tradingFee, uint256 interestCutLimit, bool activated) external {
         require(msg.sender == feeSetter, "Only feeSetter");
         require(tradingFee <= WAD, "Trading fee too high");
         require(interestCutLimit <= WAD, "Interest cut limit too high");
-        tradingFeeParams[id] =
-            TradingFeeParams({tradingFee: tradingFee, interestCutLimit: interestCutLimit, activated: activated});
+        // Safe cast because values are below type(uint64).max.
+        tradingFeeParams[id] = TradingFeeParams({
+            tradingFee: uint64(tradingFee), interestCutLimit: uint64(interestCutLimit), activated: activated
+        });
     }
 
-    function setDefaultTradingFee(address loanToken, uint128 tradingFee, uint128 interestCutLimit, bool activated)
+    function setDefaultTradingFee(address loanToken, uint256 tradingFee, uint256 interestCutLimit, bool activated)
         external
     {
         require(msg.sender == feeSetter, "Only feeSetter");
         require(tradingFee <= WAD, "Trading fee too high");
         require(interestCutLimit <= WAD, "Interest cut limit too high");
-        defaultTradingFeeParams[loanToken] =
-            TradingFeeParams({tradingFee: tradingFee, interestCutLimit: interestCutLimit, activated: activated});
+        // Safe cast because values are below type(uint64).max.
+        defaultTradingFeeParams[loanToken] = TradingFeeParams({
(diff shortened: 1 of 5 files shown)


src/MorphoV2.sol
```diff
@@ -142,6 +163,13 @@ contract MorphoV2 is IMorphoV2 {
         require(offerPrice <= WAD, "price too high");
 
         TradingFeeParams memory _tradingFeeParams = tradingFeeParams[id];
+        if (!_tradingFeeParams.activated) {
+            TradingFeeParams memory _defaultTradingFeeParams =
+                defaultTradingFeeParams[offer.obligation.loanToken];
+            if (_defaultTradingFeeParams.activated) {
+                _tradingFeeParams = _defaultTradingFeeParams;
```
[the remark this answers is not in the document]
MathisGD: we are developing an immutable protocol. it's not deployed yet, so we don't care about breaking changes. also once deployed it's code will never change.

src/interfaces/IMorphoV2.sol
```diff
@@ -51,6 +51,7 @@ struct Seizure {
 struct TradingFeeParams {
     uint128 tradingFee;
     uint128 interestCutLimit;
+    bool activated;
```
adhusson: Reduce to stay on 1 slot.

src/MorphoV2.sol
```diff
@@ -142,6 +154,10 @@ contract MorphoV2 is IMorphoV2 {
         require(offerPrice <= WAD, "price too high");
 
         TradingFeeParams memory _tradingFeeParams = tradingFeeParams[id];
+        if (!_tradingFeeParams.activated) {
+            TradingFeeParams memory _defaultTradingFeeParams = defaultTradingFeeParams[offer.obligation.loanToken];
+            if (_defaultTradingFeeParams.activated) _tradingFeeParams = _defaultTradingFeeParams;
+        }
```
adhusson: if it's inactive and the default is inactive too the initial value will be used

adhusson (approved):

src/MorphoV2.sol
```diff
@@ -157,26 +157,36 @@ contract MorphoV2 is IMorphoV2 {
...
+            tradingFee = params.tradingFee;
+        } else {
+            TradingFeeParams memory defaultParams = defaultTradingFeeParams[offer.obligation.loanToken];
+            if (defaultParams.activated) {
+                interestCutLimit = defaultParams.interestCutLimit;
+                tradingFee = defaultParams.tradingFee;
+            } else {
+                interestCutLimit = 0;
+                tradingFee = 0;
+            }
```
adhusson, line 174: Not needed, they are already initialized at 0. We could initialize them explicitly lines 160-161 but this else is using lines and gas for no effect

peyha: unless someone set a default trading fee with non zero interestCutLimit and tradingFee with `activated=False`
but we could also require that if `activated=Fasle` you have `tradingFee=0`and `interestCutLimit=0` (unless it makes sense somehow but I don't really see how)

src/MorphoV2.sol
```diff
@@ -74,11 +75,26 @@ contract MorphoV2 is IMorphoV2 {
...
+        // Safe cast because values are below type(uint64).max.
+        tradingFeeParams[id] = TradingFeeParams({
+            tradingFee: uint64(tradingFee), interestCutLimit: uint64(interestCutLimit), activated: activated
+        });
+    }
+
+    function setDefaultTradingFee(address loanToken, uint256 tradingFee, uint256 interestCutLimit, bool activated)
+        external
+    {
+        require(msg.sender == feeSetter, "Only feeSetter");
```
peyha, line 91: should it be the same fee setter for both ?
the trading fee would be set very regularly (one per month for monthly bonds) so the trading fee setter would probably be a bot while the default trading fee is loan-based and would operate much less regularly so it could be a safer setup

MathisGD:
implemnted in #289

PR #221, opened 2025-10-30
transferLib bubbles up revert reason

### LINKED ISSUE #203: should the transfer lib bubble up the revert reason

MathisGD:
fixes #203. some tokens revert with useful data, but we loose it. it doesn't cost much

------
https://chatgpt.com/codex/tasks/task_b_6903a1e1bbb48321bada76a0744322c7

### DIFF
--- a/src/libraries/SafeTransferLib.sol
+++ b/src/libraries/SafeTransferLib.sol
@@ -9,7 +9,11 @@ library SafeTransferLib {
         require(token.code.length > 0, "no code");
 
         (bool success, bytes memory returndata) = token.call(abi.encodeCall(IERC20.transfer, (to, value)));
-        require(success, "transfer reverted");
+        if (!success) {
+            assembly {
+                revert(add(returndata, 0x20), mload(returndata))
+            }
+        }
         require(returndata.length == 0 || abi.decode(returndata, (bool)), "transfer returned false");
     }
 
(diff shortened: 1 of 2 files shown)


adhusson:
it can create a clash if the token reverts with the same error code we use internally. Especially true with argumentless custom errors that are just 4 bytes.

MathisGD:
we could add a prefix but then it would break custom errors... not sure it's worth it

adhusson:
On the other hand what I said will apply to callbacks since they bubble up errors by default, so I'm not sure it's such a strong argument.

MathisGD:
Good point. wdyt @peyha @QGarchery

peyha:
no strong opinion on my side, the cost of adding it is very low and it could be helpful in some case

why not reverting directly in the call then ? by using `IERC20(token).transfer(to, value)`

>it can create a clash if the token reverts with the same error code we use internally. Especially true with argumentless custom errors that are just 4 bytes.

it's quite* rare actually, we will prob have about 20 error codes so clashing with a 4 bytes (4b values) error would be quite rare

MathisGD:
>why not reverting directly in the call then ? by using IERC20(token).transfer(to, value)

because we want to revert if the token returns false while not reverting if the token doesn't return a boolean (for USDT), that's the point of safetransferlib)

QGarchery (approved):

QGarchery:
All good for me, this is a very slight improvement that doesn't cost much. And I understand that missing approval for example is very common so it may be good to be more explicit. Some slight changes in tests in #259

adhusson (approved):
ok for me, changed my mind on the error conflict since we will have the issue on callback anyway.

PR #222, opened 2025-10-30
renamings

MathisGD:
rename offers parameters

------
https://chatgpt.com/codex/tasks/task_b_6903a6ff184883218c0153dfa343c295

### DIFF
--- a/src/interfaces/IMorphoV2.sol
+++ b/src/interfaces/IMorphoV2.sol
@@ -23,10 +23,10 @@ struct Offer {
     uint256 assets;
     uint256 obligationUnits;
     uint256 obligationShares;
-    uint256 start;
-    uint256 expiry;
+    uint256 validFrom;
+    uint256 validUntil;
     uint256 startPrice;
-    uint256 expiryPrice;
+    uint256 endPrice;
     bytes32 group;
     bytes32 nonce;
     address callback;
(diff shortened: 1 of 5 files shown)


src/interfaces/IMorphoV2.sol
```diff
@@ -23,10 +23,10 @@ struct Offer {
     uint256 assets;
     uint256 obligationUnits;
     uint256 obligationShares;
-    uint256 start;
-    uint256 expiry;
+    uint256 validFrom;
+    uint256 validUntil;
```
QGarchery, line 27: don't you like startDate/endDate better in the end ? The idea is that it's easier to link it to startPrice/endPrice, but you don't have the confusion of expiry and maturity

MathisGD: endDate is still confusing

adhusson: `validityStart` `validityEnd`

src/interfaces/IMorphoV2.sol
```diff
@@ -23,10 +23,10 @@ struct Offer {
...
     uint256 startPrice;
-    uint256 expiryPrice;
+    uint256 endPrice;
```
adhusson, line 29: `validityStartPrice` `validityEndPrice`

MathisGD:
at the end I prefer the current naming

PR #225, opened 2025-10-30
[pre-repay] early withdrawal

MathisGD:
auction the right to withdraw early, at the profit of the borrower who repaid early.

there are 3 ways this can be used:
- an arbitrager waits for the right prices and takes it
- the borrower buys obligations themselves withdraw (virtually with no discount)
- a lender willing to exit takes the opportunity


src/MorphoV2.sol
```diff
@@ -263,11 +266,31 @@ contract MorphoV2 is IMorphoV2 {
...
+        external
+    {
+        bytes32 id = _id(obligation);
+
+        uint256 discount = block.timestamp > start + 6 hours ? WAD : (block.timestamp - start).mulDivDown(WAD, 6 hours);
+        uint256 shares = obligationUnits.mulDivUp(totalShares[id] + 1, totalUnits[id] + 1);
+
+        auctionable[user][id][start] -= obligationUnits;
+
+        sharesOf[user][id] -= shares;
```
```suggestion
        sharesOf[msg.sender][id] -= shares;
```

MathisGD:
@codex review again

src/MorphoV2.sol
```diff
@@ -43,6 +43,8 @@ contract MorphoV2 is IMorphoV2 {
     /// @dev Address that can set trading fees.
     address public feeSetter;
 
+    mapping(address user => mapping(bytes32 id => mapping(uint256 start => uint256 units))) public auctionable;
```
```suggestion
    mapping(address user => mapping(bytes32 id => mapping(uint256 start => uint256))) public auctionable;
```

src/MorphoV2.sol
```diff
@@ -263,11 +266,31 @@ contract MorphoV2 is IMorphoV2 {
...
+
+        auctionable[user][id][start] -= obligationUnits;
+
+        sharesOf[msg.sender][id] -= shares;
+        withdrawable[id] -= obligationUnits;
+        totalShares[id] -= shares;
+        totalUnits[id] -= obligationUnits;
+
+        SafeTransferLib.safeTransfer(obligation.loanToken, msg.sender, obligationUnits.mulDivDown(discount, WAD));
+        SafeTransferLib.safeTransfer(obligation.loanToken, user, obligationUnits.mulDivDown(WAD - discount, WAD));
```
adhusson: Avoid forgetting units to roundings, especially relevant if `msg.sender == user`

```suggestion
        uint256 auctioned = obligationUnits.mulDivDown(discount, WAD);
        SafeTransferLib.safeTransfer(obligation.loanToken, msg.sender, auctioned);
        SafeTransferLib.safeTransfer(obligation.loanToken, user, obligationUnits - auctioned);
```

src/MorphoV2.sol
```diff
@@ -263,11 +266,31 @@ contract MorphoV2 is IMorphoV2 {
         SafeTransferLib.safeTransfer(obligation.loanToken, msg.sender, obligationUnits);
     }
 
+    function withdrawEarly(Obligation memory obligation, uint256 obligationUnits, address user, uint256 start)
+        external
+    {
+        bytes32 id = _id(obligation);
```
adhusson: don't let withdrawers shoot themselves in the foot

```suggestion
        require(obligation.maturity >= block.timestamp, "maturity reached");
        bytes32 id = _id(obligation);
```

src/MorphoV2.sol
```diff
@@ -349,6 +372,7 @@ contract MorphoV2 is IMorphoV2 {
         }
 
         withdrawable[id] += totalRepaid;
+        auctionable[borrower][id][block.timestamp] += totalRepaid;
         debtOf[borrower][id] -= totalRepaid;
```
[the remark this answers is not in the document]
MathisGD: that's on purpose. we want the borrower to get back their "interest" after a liquidation.

src/MorphoV2.sol
```diff
@@ -263,11 +266,31 @@ contract MorphoV2 is IMorphoV2 {
...
+        SafeTransferLib.safeTransfer(obligation.loanToken, msg.sender, obligationUnits.mulDivDown(discount, WAD));
+        SafeTransferLib.safeTransfer(obligation.loanToken, user, obligationUnits.mulDivDown(WAD - discount, WAD));
+    }
+
     function repay(Obligation memory obligation, uint256 obligationUnits, address onBehalf) external {
         bytes32 id = _id(obligation);
 
         debtOf[onBehalf][id] -= obligationUnits;
         withdrawable[id] += obligationUnits;
+        auctionable[onBehalf][id][block.timestamp] += obligationUnits;
```
peyha, line 343: in this auction, repayers would theoretically 
- repay their position
- find a sell offer at the best price
- take this sell offer
- withdraw early their own position

src/MorphoV2.sol
```diff
@@ -263,11 +266,31 @@ contract MorphoV2 is IMorphoV2 {
         SafeTransferLib.safeTransfer(obligation.loanToken, msg.sender, obligationUnits);
     }
 
+    function withdrawEarly(Obligation memory obligation, uint256 obligationUnits, address user, uint256 start)
+        external
+    {
+        bytes32 id = _id(obligation);
+
+        uint256 discount = block.timestamp > start + 6 hours ? WAD : (block.timestamp - start).mulDivDown(WAD, 6 hours);
```
adhusson: Some notes

* The range where it starts to make sense can be narrow at the low end. For a 10% bond that expires in 1 week p=0.99815, and (1-p)*6hours is 40s. If it expires in a month it's 176s, in a year 36 minutes.
* It creates time pressure for borrowers, they have 5.5 to 6 hours to react before being eaten.
* Borrowers who want to simulate #170 's `withdrawRepaid` need to split by the number of early repay / liquidations timestamps.

Because borrowers need to react they will probably subscribe to a service on top. But if they do that, there is no need for this feature in the first place.

MathisGD: - yes I think that the price "shape" should be reworked. notably it could take into account the time to maturity
- I didn't get this.
- yes, but I don't think that they would do this. they would probably let the market do it

MathisGD: why do they need to react?

src/MorphoV2.sol
```suggestion
        uint256 forBidder = obligationUnits.mulDivDown(discount, WAD);
        SafeTransferLib.safeTransfer(obligation.loanToken, msg.sender, forBidder);
        SafeTransferLib.safeTransfer(obligation.loanToken, user, obligationUnits - forBidder);
```

src/MorphoV2.sol
MathisGD: they can either do this, or wait that the free-market does it for them

src/MorphoV2.sol
```diff
@@ -263,11 +266,32 @@ contract MorphoV2 is IMorphoV2 {
...
```
```suggestion
        uint256 bidderFraction = block.timestamp > start + 6 hours ? WAD : (block.timestamp - start).mulDivDown(WAD, 6 hours);
```

src/MorphoV2.sol
peyha: yes, just wanted to highlight that borrowers are "passive" when they get liquidated so they are likely to go through the auction while borrowers repaying are active

src/MorphoV2.sol
```diff
@@ -263,11 +267,34 @@ contract MorphoV2 is IMorphoV2 {
...
+
+        auctionable[user][id][start] -= obligationUnits;
+
+        sharesOf[msg.sender][id] -= shares;
+        withdrawable[id] -= obligationUnits;
+
+        totalShares[id] -= shares;
+        totalUnits[id] -= obligationUnits;
+
+        uint256 fractionForBidder = UtilsLib.max((block.timestamp - start).mulDivDown(auctionSlope[user][id], WAD), WAD);
```
MathisGD: should it accelerate when approaching the maturity? should the curve have an other shape? the parameter could be like portion of the time to maturity to reach wad?

peyha: The curve is fine, it's predictable and I can't think of other good candidates (exponential would not work because it should go from 0 to 1 unless we tweak it)
Initially, I thought that it would be great to be sure that the auction ends before maturity but interests are usually negligible if the auction ends after maturity (when the auction only lasts a few hours). The user could take this into account when setting his slope

src/MorphoV2.sol
```diff
@@ -263,11 +267,34 @@ contract MorphoV2 is IMorphoV2 {
...
+        external
+    {
+        require(block.timestamp <= obligation.maturity, "maturity not reached");
+        bytes32 id = _id(obligation);
+
+        uint256 shares = obligationUnits.mulDivUp(totalShares[id] + 1, totalUnits[id] + 1);
+
+        auctionable[user][id][start] -= obligationUnits;
+
+        sharesOf[msg.sender][id] -= shares;
```
adhusson: should include an onBehalf param like withdraw

src/MorphoV2.sol
adhusson: `min`

src/MorphoV2.sol
```diff
@@ -273,16 +273,19 @@ contract MorphoV2 is IMorphoV2 {
...
 
-        uint256 shares = obligationUnits.mulDivUp(totalShares[id] + 1, totalUnits[id] + 1);
+        if (sharesOf[msg.sender][id] > 0) {
+            uint256 shares = obligationUnits.mulDivUp(totalShares[id] + 1, totalUnits[id] + 1);
+            sharesOf[msg.sender][id] -= shares;
+            withdrawable[id] -= obligationUnits;
+            totalShares[id] -= shares;
+            totalUnits[id] -= obligationUnits;
+        } else {
+            debtOf[msg.sender][id] += obligationUnits;
```
adhusson: An issue with this approach (compared to #256) is this scenario
1. sell 100
2. repay 60
3. buy 60 (for < 60)
4. withdraw early 60

In #256 it's fine, your debt is still high enough to support the buy 60. 

Here you only have 40 left of debt so you need to buy 40, then buy 20, then withdraw early 20, then withdraw early 40.

A solution is to add a callback to `withdrawEarly`.

MathisGD:
replaced by recovery close factor

PR #229, opened 2025-10-31
Factorize health data

QGarchery:
Adds slightly more computation cost, but that seems ok:
- `repayableDebt += collateralAmount.mulDivUp(WAD, MAX_LIF).mulDivUp(prices[i], ORACLE_PRICE_SCALE)` in health checks (except in `liquidate`), but notice that there is no storage load here
- some extra memory usage
- the "collaterals are sorted" check in `liquidate`

### DIFF
--- a/src/MorphoV2.sol
+++ b/src/MorphoV2.sol
@@ -313,10 +313,10 @@ contract MorphoV2 is IMorphoV2 {
     {
         bytes32 id = _id(obligation);
 
-        (uint256[] memory prices, uint256 maxDebt, uint256 repayableDebt) = _healthData(obligation, id, borrower);
+        (uint256 maxDebt, uint256 repayableDebt, uint256[] memory prices) = _healthData(obligation, id, borrower);
 
         uint256 originalDebt = debtOf[borrower][id];
-        require(block.timestamp > obligation.maturity || originalDebt > maxDebt, "position is not liquidatable");
+        require(originalDebt > maxDebt ||block.timestamp > obligation.maturity, "position is not liquidatable");
 
         uint256 lif = originalDebt > maxDebt
             ? MAX_LIF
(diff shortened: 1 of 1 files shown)


adhusson:
Note we considered adding prices to the [last factorization](https://github.com/morpho-org/morpho-v2/pull/97) but it was "too heavy"

peyha (approved):

QGarchery:
> Note we considered adding prices to the https://github.com/morpho-org/morpho-v2/pull/97 but it was "too heavy"

for me the blocker in the last factorization was that the prices were not reused, I missed that we could actually return the price array at the time (I don't think it's too heavy)

src/MorphoV2.sol
```diff
@@ -398,24 +389,36 @@ contract MorphoV2 is IMorphoV2 {
...
+            previousCollateralToken = currentCollateralToken;
+        }
+    }
+
     function _isHealthy(Obligation memory obligation, address borrower) internal view returns (bool) {
         bytes32 id = _id(obligation);
         uint256 debt = debtOf[borrower][id];
         if (debt == 0) {
             return true;
         } else {
```
MathisGD: why isn't this part of the other function?

MathisGD: done in https://github.com/morpho-org/morpho-v2/pull/229/commits/b3252262b0aa6fd7022ff06d589041f46983a709

MathisGD: actually no

MathisGD: https://github.com/morpho-org/morpho-v2/pull/229/commits/fb996cf0096213514317d262d1c939950e918e33

MathisGD (commented):
does this really simplify anything? plus it adds some gas..

QGarchery:
> does this really simplify anything? plus it adds some gas..

What I like is that the 2 loops were done differently (note taking the same stack variables notably), and so checking that it was the same wasn't immediate. Now this is enforced

QGarchery:
replaced by #260

PR #231, opened 2025-10-31
Show hashing in assembly

QGarchery:
Motivation was to remove lint errors.

This is very inpractical though, so this can be discarded. Feel free to close this PR.

Another idea would be to hash the memory directly. So hash the collaterals array and hash the other elements (chainId, loanToken, maturity), in memory directly. This avoids copying the memory, but it seems very ugly as well

### DIFF
--- a/foundry.toml
+++ b/foundry.toml
@@ -3,6 +3,3 @@ via_ir = true
 
 [profile.default.fmt]
 wrap_comments = true
-
-[lint]
-exclude_lints = ["asm-keccak256"]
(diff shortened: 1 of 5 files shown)


MathisGD:
do you know how much gas does it save?

MathisGD:
@codex review

MathisGD:
from snapshot --diff
image
unclear why but it says that it's worse...

QGarchery:
I think it costs more gas because this is what the compiler is doing (but better because it can remove the loop in solidity overhead). So I think the linter is pointing to a bad optimization in that case, because Obligation is a dynamic type

Issue #208, opened 2025-10-28
testing: Improve coverage


peyha:
As of #202, we have 

<img width="787" height="320" alt="Image" src="https://github.com/user-attachments/assets/51ab2e44-ed26-45fe-9a1c-bb2a9641ffd2" />

MathisGD:
I think that it's completely bugged with --ir-minimum (see for example safetransferlib)

peyha:
Ah ok, then it will be very annoying to test

peyha:
<img width="793" height="324" alt="Image" src="https://github.com/user-attachments/assets/aa9ca028-dd41-4a90-84e5-624fb79e305c" />
after fighting with the compiler, I got the coverage without 'viaIr', it's much more satisfying  (but still we are missing many branches)

peyha:
Only [this line](https://github.com/morpho-org/morpho-v2/blob/9233ee6bec1f100279cd3d680765d24c38d4f96a/src/MorphoV2.sol#L187) is not covered in the tests

peyha:
Just re-did coverage testing (refer to [this branch](https://github.com/morpho-org/midnight/tree/coverage) to have the required change for coverage to work) 

<img width="858" height="336" alt="Image" src="https://github.com/user-attachments/assets/a504dbad-e2f0-43fc-aa3a-875904f95d58" />

peyha:
Most of the missed statements in Midnight are `require(_obligationStates[id].created, "obligation not created")` or the callback check for `onBuy` callback in take or `require(buyerPos.pendingFee <= buyerPos.credit, "buyer pendingFee exceeds credit");`
A few getters like `lastAccrual` or `isLiquidatable`, `maxLif` and `obligationState` are also not tested

PR #237, opened 2025-11-04
test: consumed obligations units and shares

### LINKED ISSUE #165: Write tests for offer inputs
After #161

peyha:
Missing test as seen in #208 

fixes #165

### DIFF
--- a/test/TakeTest.sol
+++ b/test/TakeTest.sol
@@ -45,32 +45,32 @@ contract TakeTest is BaseTest {
         lenderOffer.assets = type(uint256).max;
         lenderOffer.obligation = obligation;
         lenderOffer.expiry = block.timestamp + 200;
-        lenderOffer.startPrice = 1 ether;
-        lenderOffer.expiryPrice = 1 ether;
+        lenderOffer.startPrice = 1e18;
+        lenderOffer.expiryPrice = 1e18;
 
         otherLenderOffer.buy = false;
         otherLenderOffer.maker = otherLender;
         otherLenderOffer.assets = type(uint256).max;
         otherLenderOffer.obligation = obligation;
         otherLenderOffer.expiry = block.timestamp + 200;
-        otherLenderOffer.startPrice = 1 ether;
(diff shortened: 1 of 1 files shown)


test/TakeTest.sol
MathisGD: no need to fuzz, but let's not use 1, because it creates the very special case assets=units (=shares)

MathisGD (approved):
apart from that LGTM

test/TakeTest.sol
peyha: we are supposed to have shares>units because of `createBadDebt`

MathisGD: good point

MathisGD: can you change the price also for existing tests?

peyha: i'm missing other tests but it should probably be done in another PR

MathisGD (approved):

test/TakeTest.sol
```diff
@@ -755,10 +756,10 @@ contract TakeTest is BaseTest {
         secondRevertingTake = bound(secondRevertingTake, offerAmount - assets + 1, maxAssets);
         secondPassingTake = bound(secondPassingTake, 0, offerAmount - assets);
         lenderOffer.assets = offerAmount;
-        lenderOffer.startPrice = 1 ether;
-        lenderOffer.expiryPrice = 1 ether;
+        lenderOffer.startPrice = 0.9 ether;
+        lenderOffer.expiryPrice = 0.9 ether;
         deal(address(loanToken), lender, offerAmount);
-        collateralize(obligation, borrower, offerAmount);
+        collateralize(obligation, borrower, offerAmount * 1 ether / lenderOffer.startPrice);
```
adhusson: wad?

probably a good reason to keep wad, `ether` confused me

adhusson (approved):

test/TakeTest.sol
MathisGD: actually 1e18 is better than ether

MathisGD (approved):

PR #244, opened 2025-11-05
feat: case in Take

peyha:


MathisGD (commented):
I don't like it to be honest. that seems like a hotfix. we need to find a better way

peyha:
not a huge fan in the end either, implementing a custom struct for an event is weird

src/libraries/EventsLib.sol
```diff
@@ -21,12 +21,7 @@ library EventsLib {
         uint256 obligationShares,
         address indexed taker,
         Offer offer,
-        uint256 obligationTotalUnits,
-        uint256 obligationTotalShares,
-        uint256 buyerShares,
-        uint256 buyerDebt,
-        uint256 sellerShares,
-        uint256 sellerDebt
+        TakeCase takeCase
```
adhusson, line 24: Maybe with 2 bools
- sellerEnters
- buyerEnters

PR #247, opened 2025-11-06
events log diffs

adhusson:


peyha:
It's actually more elegant than I expected but we lose a lot in terms of explainability (having a specific branch for each possible situation is great imo)

adhusson:
closed in favor of https://github.com/morpho-org/morpho-v2/pull/248

PR #248, opened 2025-11-06
feat: 2 bool in take event

peyha:


src/MorphoV2.sol
```diff
@@ -197,17 +197,19 @@ contract MorphoV2 is IMorphoV2 {
             require((consumed[offer.maker][offer.group] += obligationShares) <= offer.obligationShares, "consumed");
         }
 
-        if (debtOf[buyer][id] == 0 && sharesOf[seller][id] == 0) {
+        bool buyerEnter = (debtOf[buyer][id] == 0);
+        bool sellerEnter = (sharesOf[seller][id] == 0);
```
```suggestion
        bool buyerIsLender = (debtOf[buyer][id] == 0);
        bool sellerIsLender = (debtOf[seller][id] == 0);
```

MathisGD: other idea could be to have only uints:
- buyerSharesAddition
- buyerDebtReduction
- sellerSharesReduction
- sellerDebtAddition
- totalUnitsAddition
- totalSharesAddition
- totalUnitsReduction
- totalSharesReduction

MathisGD: oh but it's https://github.com/morpho-org/morpho-v2/pull/247/files

PR #251, opened 2025-11-06
[event 2] feat: log obligation everywhere

peyha:

PR #252, opened 2025-11-07
[event 1] feat: event NewObligation

peyha:


src/libraries/EventsLib.sol
```diff
@@ -12,6 +12,7 @@ library EventsLib {
     event SetTradingFee(bytes32 indexed obligationId, uint256 tradingFee, uint256 interestCutLimit);
     event SetTradingFeeRecipient(address indexed recipient);
 
+    event NewObligation(bytes32 indexed obligationId, Obligation obligation);
```
adhusson:
```suggestion
    event CreateObligation(bytes32 indexed obligationId, Obligation obligation);
```

peyha:
We could pack `obligationCreated` with `totalUnits` or `totalShares` or with `tradingFeeParams` to save some gas at each `take`

src/MorphoV2.sol
```diff
@@ -24,6 +24,7 @@ contract MorphoV2 is IMorphoV2 {
     mapping(bytes32 => uint256) public totalUnits;
     mapping(bytes32 => uint256) public totalShares;
     mapping(address => mapping(bytes32 => mapping(address => uint256))) public collateralOf;
+    mapping(bytes32 => bool) obligationCreated;
```
adhusson: You could do this, log an obligation id, and not have the Obligation:

- set the trading fee
- repay 0
- withdraw 0
- withdraw 0 collateral

adhusson: I think we could ask for the whole obligation in trading fee setting & do the check there as well. For the other ones it doesn't matter.

peyha: it's fine for no-ops (repay 0, withdraw 0 and withdraw 0 collateral) but it  might cause issue to not consider it in `setTradingFee`. We would have to change its interface because it currently takes an `id` as input

src/MorphoV2.sol
```diff
@@ -32,6 +24,7 @@ contract MorphoV2 is IMorphoV2 {
     mapping(bytes32 => uint256) public totalUnits;
     mapping(bytes32 => uint256) public totalShares;
     mapping(address => mapping(bytes32 => mapping(address => uint256))) public collateralOf;
+    mapping(bytes32 => bool) public obligationCreated;
```
MathisGD: nb: we could re-use this slot to store "obligationFeeActivated", which would prevent to read the obligation fee if it's not activated (then it's the default fee)

src/MorphoV2.sol
```diff
@@ -86,10 +79,17 @@ contract MorphoV2 is IMorphoV2 {
...
+    function setTradingFee(Obligation memory obligation, uint256 tradingFee, uint256 interestCutLimit) external {
         require(msg.sender == feeSetter, "Only feeSetter");
         require(tradingFee <= type(uint128).max, "Trading fee too high");
         require(interestCutLimit < WAD, "Interest cut limit too high");
+
+        bytes32 id = _id(obligation);
+        if (!obligationCreated[id]) {
+            emit EventsLib.CreateObligation(id, obligation);
+            obligationCreated[id] = true;
+        }
```
MathisGD: do we really need to do it here?

peyha: [related thread](https://morpholabs.slack.com/archives/C02N7CZ088N/p1761930946499389), it's slightly better for the indexing to have a `CreateObligation` event before any other event referencing t he `id` (otherwise the db handle NULL values). It's not breaking though so can still be discarded

src/MorphoV2.sol
MathisGD: to do later

src/MorphoV2.sol
MathisGD: now I log everywhere, but not in fees (I don't think that's useful)

MathisGD:
going for this over #251 because of [this](https://morpholabs.slack.com/archives/C02N7CZ088N/p1769098893754449)

src/MorphoV2.sol
MathisGD, line 1: not sure why it makes the certora struggle...

src/MorphoV2.sol
```diff
@@ -446,6 +452,14 @@ contract MorphoV2 is IMorphoV2 {
         SafeTransferLib.safeTransferFrom(token, msg.sender, address(this), assets);
     }
 
+    function touchObligation(Obligation memory obligation) internal {
+        bytes32 id = toId(obligation);
```
QGarchery, line 451: probably worth to take the `id` as an argument of the function, it would avoid hashing twice and hashing is costly

src/MorphoV2.sol
QGarchery, line 1: not sure, but we could add `touchObligation` to `liquidate` as well

src/MorphoV2.sol
MathisGD: better, touchObligation could return the id

adhusson (approved):

PR #253, opened 2025-11-10
[ticks]

MathisGD:
final tick implementation


src/libraries/MathLib.sol
```diff
@@ -34,4 +36,12 @@ library MathLib {
     function sort(bytes32 x, bytes32 y) internal pure returns (bytes memory) {
         return x < y ? abi.encodePacked(x, y) : abi.encodePacked(y, x);
     }
+
+    function wExp(uint256 x) internal pure returns (uint256) {
```
MathisGD: todo: 
- is it enough precision
- roundings direction

peyha: got convined by [this blog article](https://xn--2-umb.com/22/approximation/) that taylor series are quite bad to approximate functions, [solady's expWad](https://github.com/Vectorized/solady/blob/73f13dd1483707ef6b4d16cb0543570b7e1715a8/src/utils/FixedPointMathLib.sol#L207) is prob the most accurate implem

MathisGD: actually we have a big problem because delta * n is not small at all (for n = 400, it's 10)

MathisGD: solved with [range reduction](https://morpholabs.slack.com/archives/C02N7CZ088N/p1693298841708259?thread_ts=1693132156.065249&cid=C02N7CZ088N)

src/MorphoV2.sol
```diff
@@ -392,6 +394,15 @@ contract MorphoV2 is IMorphoV2 {
 
     /// INTERNAL ///
 
+    function tickToPrice(uint256 tick) public pure returns (uint256) {
+        if (tick == type(uint256).max) {
+            return WAD;
+        } else {
+            return
+                WAD.mulDivDown(WAD, WAD + (WAD.mulDivDown(WAD - P_0, P_0)).mulDivDown(MathLib.wExp(DELTA * tick), WAD));
```
peyha: this formula assumes $$(1+\delta)^n\approx e^{\delta*n}$$ introducing some approximation (in addition to the exp approximation), it might be worth quantifying it (though it should be fine because $$\delta=2.5$$%)

MathisGD: Image

MathisGD: we need to assess precision anyway, also for the [exp approximation](https://github.com/morpho-org/morpho-v2/pull/253/files#r2509898518)

src/libraries/ConstantsLib.sol
```diff
@@ -6,3 +6,5 @@ uint256 constant WAD = 1e18;
 uint256 constant ORACLE_PRICE_SCALE = 1e36;
 uint256 constant MAX_LIF = 1.15e18; // Liquidation Incentive Factor
 uint256 constant TIME_TO_MAX_LIF = 15 minutes; // Time to reach MAX_LIF
+uint256 constant DELTA = 0.025e18;
```
adhusson: `0.025` is very close to `ln(1.025)` so it works, but assuming we want the rate at tick `t` to be `R_0 * 1.025^t` and not `R_0 * e^(0.025 * t)`, this should be

```
uint256 constant LN_ONE_PLUS_DELTA = 24692612590371416; // ln(1.025)e18
```

src/MorphoV2.sol
adhusson: Some ticks with price > 1 could be useful. For instance [here](https://github.com/morpho-org/vault-v2/pull/786/files#diff-5c05bfc36c7ac98180672bfce918660bcd485af9520c6a359f56b2bbb694ad01R193) it avoids a buffer check on force deallocate by making the buyer pay any obligation fees.

We can get negative rates by returning the price that corresponds to return(tick)-1. With this change we get P_0 around tick 1305-1306.

```suggestion
                WAD.mulDivDown(WAD, (WAD.mulDivDown(WAD - P_0, P_0)).mulDivDown(MathLib.wExp(DELTA * tick), WAD));
```

adhusson: i hadn't seen this, it's free to increase precision so why not do it

adhusson: I missed that the fee was chosen such that at price 1 there is no fee.

src/libraries/MathLib.sol
peyha: wrote a short script to quantify the precision with range reduction + taylor series (on [0, ln(2)/2]), I get
  ```
     n     Approximation        Absolute Error       Relative Error
     ----------------------------------------------------------------------
     1     1.0000000000         4.14e-01             2.93e-01
     2     1.3465735903         6.76e-02             4.78e-02
     3     1.4066302170         7.58e-03             5.36e-03
     4     1.4135682306         6.45e-04             4.56e-04
```
so I would suggest either n=3 or n=4 to have sufficient prevision

Comment (MathisGD) [reply src/libraries/MathLib.sol]: note that we are between -ln(2)/2 and ln(2)/2

File: src/MorphoV2.sol (hunk 2 above)
Comment (MathisGD) [reply src/MorphoV2.sol]: maybe readability

File: src/MorphoV2.sol [hunk 4]
```diff
@@ -459,12 +458,7 @@ contract MorphoV2 is IMorphoV2 {
     /// VIEW FUNCTIONS ///
 
     function tickToPrice(uint256 tick) public pure returns (uint256) {
-        if (tick == type(uint256).max) {
-            return WAD;
-        } else {
-            return
-                WAD.mulDivDown(WAD, WAD + (WAD.mulDivDown(WAD - P_0, P_0)).mulDivDown(UtilsLib.wExp(DELTA * tick), WAD));
-        }
+        return (WAD.mulDivUp(WAD, WAD + UtilsLib.wExp(DELTA * (588 - int256(tick))))).mulDivUp(1, 1e12) * 1e12;
```
adhusson: Could be cleaner to 
- keep the uint only wExp without a branch, add back a branch here 
- center around p=0.5 to remove the `588`

```solidity
function tickToPrice(int256 tick) internal pure returns (uint256) {
      uint price;

      if (tick >= 0) {
          uint256 exp = wExp(uint256(LN_1_025 * tick));
          price = WAD * WAD / (WAD + exp);
      } else {
          uint256 exp = wExp(uint256(-LN_1_025 * tick));
          price = WAD * exp / (WAD + exp);
      }
     return price/1e12*1e12;
  }
```

src/MorphoV2.sol
MathisGD: solved anyway

src/MorphoV2.sol
```diff
@@ -448,6 +451,12 @@ contract MorphoV2 is IMorphoV2 {
 
     /// VIEW FUNCTIONS ///
 
+    function tickToPrice(uint256 tick) public pure returns (uint256) {
+        return // forge-lint: disable-next-line(unsafe-typecast) tick is always less than 1176
+                (WAD.mulDivUp(WAD, WAD + UtilsLib.wExp(LN_ONE_PLUS_DELTA * (588 - int256(tick))))).mulDivUp(1, 1e12)
```
adhusson: isn't the min price with this 1e12?

src/MorphoV2.sol
```diff
@@ -158,6 +159,7 @@ contract MorphoV2 is IMorphoV2 {
         );
         require(block.timestamp >= offer.start, "offer not started");
         require(block.timestamp <= offer.expiry, "offer expired");
+        require(offer.tick <= 1176, "start tick too high");
```
```suggestion
        require(offer.tick <= 1176, "tick too high");
```

test/UtilsLibTest.sol
```diff
@@ -102,4 +104,30 @@ contract UtilsLibTest is Test {
...
+        assertApproxEqRel(UtilsLib.wExp(-10e18), 0.000045e18, 0.1e18, "exp(-10)");
+        assertApproxEqRel(UtilsLib.wExp(-9e18), 0.000123e18, 0.1e18, "exp(-9)");
+        assertApproxEqRel(UtilsLib.wExp(-8e18), 0.000335e18, 0.001e18, "exp(-8)");
+        assertApproxEqRel(UtilsLib.wExp(-7e18), 0.000911e18, 0.03e18, "exp(-7)");
+        assertApproxEqRel(UtilsLib.wExp(-6e18), 0.002478e18, 0.003e18, "exp(-6)");
+        assertApproxEqRel(UtilsLib.wExp(-5e18), 0.006737e18, 0.05e18, "exp(-5)");
+        assertApproxEqRel(UtilsLib.wExp(-4e18), 0.018315e18, 0.006e18, "exp(-4)");
+        assertApproxEqRel(UtilsLib.wExp(-3e18), 0.049787e18, 0.07e18, "exp(-3)");
+        assertApproxEqRel(UtilsLib.wExp(-2e18), 0.135335e18, 0.01e18, "exp(-2)");
+        assertApproxEqRel(UtilsLib.wExp(-1e18), 0.367879e18, 0.1e18, "exp(-1)");
```
QGarchery: 10% error is huge ! An unknown 10% error on your return is definitely something that users can want to investigate, and then it means that this method loses the "rate meaning"

QGarchery: Not totally sure how to fix it, worth doing a few more iterations IMO if it fixes this issue (it may be due to the range reduction though)

test/UtilsLibTest.sol
```diff
@@ -102,4 +104,30 @@ contract UtilsLibTest is Test {
     function mulDivUp(uint256 x, uint256 y, uint256 d) external pure {
         UtilsLib.mulDivUp(x, y, d);
     }
+
+    function testWExp() public pure {
```
QGarchery, line 107: It's used between -14.XXXe18 and 14.XXXe18, so it makes sense to go to the full range in testing

test/UtilsLibTest.sol
MathisGD: I think that [this](https://github.com/morpho-org/morpho-v2/pull/253#discussion_r2713739958) will fix it

src/MorphoV2.sol
MathisGD: let's tackle zero in a separate pr (if it's even important)

src/libraries/UtilsLib.sol
```diff
@@ -65,15 +74,50 @@ library UtilsLib {
...
+            } else {
+                int256 ln2 = 0.693147180559945309e18;
+                int256 q = (x + ln2 / 2) / ln2;
+                int256 r = x - q * ln2;
+                int256 secondTerm = r * r / (2 * 1e18);
+                int256 thirdTerm = secondTerm * r / (3 * 1e18);
+                int256 expR = 1e18 + r + secondTerm + thirdTerm;
+                // forge-lint: disable-next-item(unsafe-typecast)
+                // - q is positive because x is positive in this branch
+                // - expR is positive?? (TODO)
```
QGarchery:
```suggestion
                // - q is non-negative because x is non-negative in this branch
                // - expR is positive because the abolute value of r is less than ln2 / 2, and secondTerm dominates thirdTerm
```

src/libraries/UtilsLib.sol
```diff
@@ -62,9 +73,51 @@ library UtilsLib {
         return x < y ? abi.encodePacked(x, y) : abi.encodePacked(y, x);
     }
 
+    function wExp(int256 x) internal pure returns (uint256) {
```
peyha: Probably not worth considering because the exp func might get removed by storing price to tick directly but not a huge fan of taylor series approximation (even when normalizing by ln2) because the approximation error grows with x on [0, ln(2)], would suggest using some of the recommandations from https://xn--2-umb.com/22/approximation/

src/libraries/UtilsLib.sol
```diff
@@ -49,6 +51,13 @@ library UtilsLib {
         return (x * y + (d - 1)) / d;
     }
 
+    /// @dev Returns (`x` + `d` - 1) / `d` rounded up, without checking for overflow.
+    function divUpUnchecked(uint256 x, uint256 d) internal pure returns (uint256) {
+        unchecked {
+            return (x + (d - 1)) / d;
+        }
+    }
+
```
QGarchery: Not clear that it is worth it to me, feels like it could be a rabbit hole if we add an operation to optimize every time

test/TickLibTest.sol
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    function _return(uint256 price) internal pure returns (uint256) {
+        return WAD.mulDivDown(WAD, price) - WAD;
+    }
+
+    function testGasBound(uint256 value) public pure {
+        bound(value, 0, 1 ether);
+    }
+
```
QGarchery, line 51:
```suggestion
```

src/libraries/UtilsLib.sol
peyha, line 1: should `priceToTick` and `tickToPrice` be implemented in its own lib in case people want to re-use it without other functions ?

test/TickTest.sol
```diff
@@ -0,0 +1,75 @@
...
+        assertGe(UtilsLib.tickToPrice(tick), price);
+        if (tick > 0) assertLe(UtilsLib.tickToPrice(tick - 1), price);
+    }
+
+    function testPriceToTickConsistency(uint256 tick) public pure {
+        tick = bound(tick, 0, TICK_RANGE);
+        uint256 price = UtilsLib.tickToPrice(tick);
+        uint256 recoveredTick = UtilsLib.priceToTick(price);
+        assertEq(UtilsLib.tickToPrice(recoveredTick), price);
+        assertLe(recoveredTick, tick);
```
QGarchery:
```suggestion    
     function testPriceToTickConsistency() public pure {
        for (uint256 tick = 0; tick <= TICK_RANGE; tick++) {
            uint256 price = UtilsLib.tickToPrice(tick);
            uint256 recoveredTick = UtilsLib.priceToTick(price);
            assertEq(UtilsLib.tickToPrice(recoveredTick), price);
            assertLe(recoveredTick, tick);
        }
```

src/libraries/UtilsLib.sol
```diff
@@ -65,15 +74,50 @@ library UtilsLib {
...
+        unchecked {
+            // forge-lint: disable-next-item(unsafe-typecast)
+            return uint256(1e36)
+                    .divUpUnchecked(1e18 + wExp(LN_ONE_PLUS_DELTA * (int256(TICK_RANGE / 2) - int256(tick))))
+                    .divUpUnchecked(1e13) * 1e13;
+        }
+    }
+
+    /// @dev Returns the lowest tick with a higher price.
+    function priceToTick(uint256 price) internal pure returns (uint256) {
```
QGarchery: Should we prevent `price > WAD` ? On top of the fact that it's an invalid value, also we don't have the property that the tick returned has a higher price than the input

test/TickTest.sol
```diff
@@ -0,0 +1,75 @@
...
+            console.log(UtilsLib.tickToPrice(i));
+        }
+    }
+
+    function testReturnJumps() public pure {
+        uint256 price = UtilsLib.tickToPrice(200);
+        uint256 previousReturn = _return(price);
+        for (uint256 i = 200; i <= 700; i++) {
+            uint256 currentReturn = _return(UtilsLib.tickToPrice(i));
+            assertApproxEqRel(currentReturn, previousReturn.mulDivDown(WAD, 1.025e18), 0.05e18, "tick i");
```
peyha:
```suggestion
            assertApproxEqRel(currentReturn, previousReturn.mulDivDown(WAD, 1.025e18), 0.01e18, "tick i");
```
seems to work with the above fix

MathisGD:
@codex review

test/TickTest.sol
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    function testTickToPriceRange() public pure {
+        for (uint256 i = 0; i <= TICK_RANGE; i++) {
+            console.log(UtilsLib.tickToPrice(i));
+        }
+    }
+
+    function testReturnJumps() public pure {
+        uint256 price = UtilsLib.tickToPrice(200);
```
peyha:
```suggestion
        uint256 price = UtilsLib.tickToPrice(199);
```

test/TickTest.sol
```diff
@@ -0,0 +1,75 @@
...
+import {UtilsLib} from "../src/libraries/UtilsLib.sol";
+import {WAD, TICK_RANGE} from "../src/libraries/ConstantsLib.sol";
+
+contract TickTest is Test {
+    using UtilsLib for uint256;
+
+    // Tick to price
+
+    function testTickToPriceMinMax() public pure {
+        assertEq(UtilsLib.tickToPrice(0), 0.00001e18, "tick 0");
```
QGarchery: So we don't have 0 price anymore, which is good because som computations could revert because of it. But I'm wondering if there wasn't a "use case" for it, in the same way that selling at 1 was essentially essentially a way to ensure that you could match yourself (as it is unattractive for other lenders). So here it would be an offer to buy at 0, notably it requires no token transfer

src/libraries/UtilsLib.sol
MathisGD: we are at precision of 0.1%, isn't it good enough? (and very simple).

src/libraries/UtilsLib.sol
MathisGD: it could be TicksLib.sol, but it's already a lib so arguably ok as is

test/TickTest.sol
QGarchery: this test is weird: we test with a delta of max 5% when the jumps should be of about ~2.5%

MathisGD: which fix?

src/libraries/UtilsLib.sol
MathisGD: agree in general, but tickToPrice is called a lot of times in priceToTick. Just by adding the unchecked, I brought the gas down from ~600 to ~200...

test/TickLibTest.sol
MathisGD: this is to be able to see how much gas bound takes, to subtract it

test/TickTest.sol
MathisGD: nb: we could retrieve it by doing a half rounding.

test/TickTest.sol
peyha: the fix on the initial price taken (` uint256 price = UtilsLib.tickToPrice(200)`)
otherwise the first comparison is with previousReturn = _return(price) and currentReturn = _return(price) so the diff can't be less than 2.5%

MathisGD: indeed...

MathisGD: ohh I just got it

MathisGD: [this](https://github.com/morpho-org/morpho-v2/pull/253#discussion_r2745720663) was the issue

test/TickTest.sol
MathisGD: nice catch

src/libraries/UtilsLib.sol
peyha: oh yes the range is [0, ln(2)/2] and not [0, ln(2)] so it's quite precise indeed. There might be some smart way to improve the precision while reducing the gas cost but probably not worth looking rn

test/TickTest.sol
peyha: I think it's worth having price = 0 because of this (implementing an obligation transfer function does not look worth it)

src/libraries/UtilsLib.sol
MathisGD: would you prefer if I inline everything?

src/libraries/UtilsLib.sol
peyha: I personally prefer to separate the different logic in different libraries (just like [uniswap does](https://github.com/Uniswap/v4-core/tree/main/src/libraries)) but not a strong pref

src/libraries/UtilsLib.sol
MathisGD: -ln(2)/2 and ln(2)/2

test/TickTest.sol
adhusson: [related discussion](https://morpholabs.slack.com/archives/C02N7CZ088N/p1769015470428889?thread_ts=1769003317.743599&cid=C02N7CZ088N)

adhusson: https://github.com/morpho-org/morpho-v2/issues/311

peyha (approved):

QGarchery (approved):

src/libraries/TickLib.sol
```diff
@@ -0,0 +1,60 @@
...
+                int256 expR = 1e18 + r + secondTerm + thirdTerm;
+                // forge-lint: disable-next-item(unsafe-typecast)
+                // - q is non-negative because x is non-negative in this branch
+                // - expR is positive because |r| < ln2 < 1e18 and |secondTerm| > |thirdTerm|
+                return uint256(expR) << uint256(q);
+            }
+        }
+    }
+
+    function tickToPrice(uint256 tick) internal pure returns (uint256) {
```
QGarchery, line 37: We should mention that this function is only meaningful for `tick <= TICK_RANGE`, similarly to the check `price <= 1e18` in `priceToTick`. It could be a comment or a `require`

test/TickLibTest.sol
```diff
@@ -0,0 +1,79 @@
...
+            console.log(TickLib.tickToPrice(i));
+        }
+    }
+
+    function testReturnJumps() public pure {
+        for (uint256 i = 207; i <= 729; i++) {
+            uint256 previousReturn = _return(TickLib.tickToPrice(i - 1));
+            uint256 currentReturn = _return(TickLib.tickToPrice(i));
+            assertApproxEqRel(
+                currentReturn.mulDivDown(1e18, previousReturn), 1.025e18, 0.1e18, string.concat("tick ", vm.toString(i))
```
QGarchery, line 38: Now this tests with up to 10% error, when the jump is of ~2.5%

src/libraries/TickLib.sol
```diff
@@ -0,0 +1,60 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Copyright (c) 2025 Morpho Association
+pragma solidity ^0.8.0;
+
+int256 constant LN_ONE_PLUS_DELTA = 0.024692612590371501e18; // ln(1 + 0.025)
+uint256 constant TICK_RANGE = 990;
```
QGarchery, line 6: I feel like `TICK_RANGE` can be in `ConstantsLib`, because it is something that users may want to import (this is used outside of this lib, in `take`)

test/TickLibTest.sol
MathisGD: but now it's 2.5%+-(0.25%)

src/libraries/TickLib.sol
MathisGD: Maybe we move the require and keep everything here ?

MathisGD: #331

src/libraries/TickLib.sol
MathisGD: #331

PR #256, opened 2025-11-10
[pre-repay] cover + auction

adhusson:


src/MorphoV2.sol
```diff
@@ -271,13 +275,42 @@ contract MorphoV2 is IMorphoV2 {
...
+
+        if (obligation.maturity < block.timestamp) require(debtOf[repayer][id] <= preRepaidOf[repayer][id], "no debt");
+        else require(_isHealthy(obligation, repayer), "Unhealthy borrower");
 
-        SafeTransferLib.safeTransferFrom(obligation.loanToken, msg.sender, address(this), obligationUnits);
+        uint256 fractionForBidder =
+            UtilsLib.min((block.timestamp - start).mulDivDown(auctionSlope[repayer][id], WAD), WAD);
+        uint256 forBidder = assets.mulDivDown(fractionForBidder, WAD);
+        SafeTransferLib.safeTransfer(obligation.loanToken, msg.sender, forBidder);
+        SafeTransferLib.safeTransfer(obligation.loanToken, repayer, assets - forBidder);
```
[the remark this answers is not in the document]
adhusson: @codex useful in theory but in practice authorizations come later

src/MorphoV2.sol
```diff
@@ -276,19 +279,38 @@ contract MorphoV2 is IMorphoV2 {
...
+            debtOf[repayer][id] -= assets;
+            uint256 shares = assets.mulDivUp(totalShares[id] + 1, totalUnits[id] + 1);
+            sharesOf[bidder][id] -= shares;
+            totalShares[id] -= shares;
+            totalUnits[id] -= assets;
         }
-        SafeTransferLib.safeTransfer(obligation.loanToken, msg.sender, assets);
+
+        if (obligation.maturity < block.timestamp) require(debtOf[repayer][id] <= preRepaidOf[repayer][id], "no debt");
+        else require(_isHealthy(obligation, repayer), "Unhealthy borrower");
```
MathisGD, line 307: at the end there is the if-else too...

src/MorphoV2.sol
```diff
@@ -276,19 +279,38 @@ contract MorphoV2 is IMorphoV2 {
...
+        if (bidder != repayer) {
+            debtOf[repayer][id] -= assets;
+            uint256 shares = assets.mulDivUp(totalShares[id] + 1, totalUnits[id] + 1);
+            sharesOf[bidder][id] -= shares;
+            totalShares[id] -= shares;
+            totalUnits[id] -= assets;
         }
-        SafeTransferLib.safeTransfer(obligation.loanToken, msg.sender, assets);
+
+        if (obligation.maturity < block.timestamp) require(debtOf[repayer][id] <= preRepaidOf[repayer][id], "no debt");
```
peyha, line 306: could you detail the motivation for this condition ?
I though it would be a `_isHealthy` check

peyha: ah ok got it, at maturity the borrower should close his full position. not sure it's perfect though, suppose I am a large borrower (larger than any lender) and I half of my position got repaid (through liquidation or pre-liquidation). at maturity, nobody can `withdrawRepaid` completely. People would have to do it partially but it's not possible because of this condition

adhusson:
replaced by recovery close factor: with rcf the healthy part of the borrow cannot be liquidated so the amount of overpaid interest is much smaller so it's not worth adding features to get it back

PR #258, opened 2025-11-10
Refactor create obligation

QGarchery:


src/MorphoV2.sol
```diff
@@ -468,6 +448,15 @@ contract MorphoV2 is IMorphoV2 {
         return tentativeSigner;
     }
 
+    function _createObligation(Obligation memory obligation) internal returns (bytes32) {
```
peyha, line 451:
```suggestion
    function _createObligation(Obligation memory obligation) internal returns (bytes32, TradingFeeParams memory) {
```
would save one SLOAD in `take` (but it's a warm one so not expensive)

peyha: could also save one SSTORE in `setTradingFee`

QGarchery: Yes, but I thought it would be ok, since the purpose of those changes is clarity (avoid code duplication) and not gas opti. But it may be fine to inline this function as well, as you did in your PR

src/MorphoV2.sol
adhusson, line 38:
```suggestion
/// @dev Trading fee parameters for a given obligation id.
mapping(bytes32 id => ObligationParams) public obligationParams;
```

Or smaller change: `.created` -> `.obligationCreated`

src/MorphoV2.sol
```diff
@@ -323,14 +310,7 @@ contract MorphoV2 is IMorphoV2 {
...
         external
     {
-        bytes32 id = _id(obligation);
-
-        TradingFeeParams memory _tradingFeeParams = tradingFeeParams[id];
-        if (!_tradingFeeParams.created) {
-            emit EventsLib.CreateObligation(id, obligation);
-            _tradingFeeParams.created = true;
-            tradingFeeParams[id] = _tradingFeeParams;
-        }
```
adhusson, line 333: I would keep it inline, just
```
        if (!tradingFeeParams[id].created) {
            emit EventsLib.CreateObligation(id, obligation);
            tradingFeeParams[id].created = true;
        }
```

without the intermediate `memory` it's clean enough. It avoids sending obligation to `_createObligation` and getting back id which is too complex.

QGarchery:
Closing because it doesn't seem to make unanimity

PR #259, opened 2025-11-11
test: improve safe transfer lib tests

QGarchery:


test/SafeTransferLibTest.sol
```diff
@@ -77,12 +71,12 @@ contract SafeTransferLibTest is Test {
...
 
     function testSafeTransferReverted() public {
-        vm.expectRevert(bytes("transfer revert reason"));
+        vm.expectRevert("transfer revert reason");
         this.safeTransfer(address(tokenRevertReason), address(1), 1);
     }
 
     function testSafeTransferRevertedNoReason() public {
-        vm.expectRevert();
+        vm.expectRevert(bytes(""));
```
QGarchery, line 79: Needed because otherwise this test would not fail if `tokenRevertNoReason` returned a revert reason

MathisGD (approved):

PR #260, opened 2025-11-11
refactor: isHealthy

MathisGD:
alternative to #229 

image

### DIFF
--- a/src/MorphoV2.sol
+++ b/src/MorphoV2.sol
@@ -7,7 +7,15 @@ import {SafeTransferLib} from "./libraries/SafeTransferLib.sol";
 import {WAD, ORACLE_PRICE_SCALE, MAX_LIF, TIME_TO_MAX_LIF} from "./libraries/ConstantsLib.sol";
 import {MathLib} from "./libraries/MathLib.sol";
 import {IOracle} from "./interfaces/IOracle.sol";
-import {IMorphoV2, Obligation, Offer, Signature, Seizure, TradingFeeParams} from "./interfaces/IMorphoV2.sol";
+import {
+    IMorphoV2,
+    Obligation,
+    Offer,
+    Signature,
+    Collateral,
+    Seizure,
+    TradingFeeParams
+} from "./interfaces/IMorphoV2.sol";
 import {ICallbacks, IFlashLoanCallback} from "./interfaces/ICallbacks.sol";
 
 /// OBLIGATIONS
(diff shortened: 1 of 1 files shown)


src/MorphoV2.sol
```diff
@@ -412,14 +421,14 @@ contract MorphoV2 is IMorphoV2 {
...
-                require(currentCollateralToken > previousCollateralToken, "collaterals not sorted");
-                uint256 price = IOracle(obligation.collaterals[i].oracle).price();
-                maxDebt += collateralOf[borrower][id][currentCollateralToken].mulDivDown(price, ORACLE_PRICE_SCALE)
-                    .mulDivDown(obligation.collaterals[i].lltv, WAD);
-                previousCollateralToken = currentCollateralToken;
+                Collateral memory _collateral = obligation.collaterals[i];
+                address collateralToken = _collateral.token;
+                require(collateralToken > previousCollateralToken, "collaterals not sorted");
+                uint256 price = IOracle(_collateral.oracle).price();
+                uint256 _collateralOf = collateralOf[borrower][id][collateralToken];
```
adhusson: This one `_collateralOf` is not needed

adhusson (approved):

src/MorphoV2.sol
MathisGD: did it such that both computations look the same

adhusson: Given how the rest of the code is (no extra variables) it looks like it's there for a reason so it raises more questions than it answers.

MathisGD: indeed, I will remove

src/MorphoV2.sol
```diff
@@ -317,11 +325,12 @@ contract MorphoV2 is IMorphoV2 {
...
 
         for (uint256 i = 0; i < obligation.collaterals.length; i++) {
-            prices[i] = IOracle(obligation.collaterals[i].oracle).price();
-            uint256 collateralAmount = collateralOf[borrower][id][obligation.collaterals[i].token];
-            maxDebt += collateralAmount.mulDivDown(prices[i], ORACLE_PRICE_SCALE)
-                .mulDivDown(obligation.collaterals[i].lltv, WAD);
-            repayableDebt += collateralAmount.mulDivUp(WAD, MAX_LIF).mulDivUp(prices[i], ORACLE_PRICE_SCALE);
+            Collateral memory _collateral = obligation.collaterals[i];
+            prices[i] = IOracle(_collateral.oracle).price();
+            uint256 price = prices[i];
```
```suggestion
            uint256 price = IOracle(_collateral.oracle).price();
            prices[i] = price;
```
(optional) so that below in the computation of maxDebt it's even more straightforward that `price` can be replaced by `IOracle(_collateral.oracle).price()`

QGarchery (approved):

src/MorphoV2.sol
MathisGD: also we touch less memory. nice!

PR #264, opened 2025-11-12
feat: isHealthy getter

### LINKED ISSUE #250: Should we add a `isHealthy` getter

MathisGD:
fixes #250


src/MorphoV2.sol
```diff
@@ -390,21 +390,14 @@ contract MorphoV2 is IMorphoV2 {
         SafeTransferLib.safeTransferFrom(token, msg.sender, address(this), amount);
     }
 
-    /// INTERNAL ///
+    /// VIEW ///
 
-    function _id(Obligation memory obligation) internal pure returns (bytes32) {
+    function toId(Obligation memory obligation) public pure returns (bytes32) {
         return keccak256(abi.encode(obligation));
     }
```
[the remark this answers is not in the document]
MathisGD: nice catch

MathisGD:
@codex please review

src/MorphoV2.sol
```diff
@@ -423,4 +416,11 @@ contract MorphoV2 is IMorphoV2 {
             return debt <= maxDebt;
         }
     }
+
```
peyha, line 428:
```suggestion
    /// INTERNAL ///
```

QGarchery: it's still view though (pure to be precise here)

src/MorphoV2.sol
```diff
@@ -390,21 +390,14 @@ contract MorphoV2 is IMorphoV2 {
         SafeTransferLib.safeTransferFrom(token, msg.sender, address(this), amount);
     }
 
-    /// INTERNAL ///
+    /// VIEW ///
 
-    function _id(Obligation memory obligation) internal pure returns (bytes32) {
```
peyha, line 395: what is the rationale for converting this function from `internal` to `public` ? (as it's not related to #250)
it's very low cost to change it but it breaks consistency with blue (not really an issue)

QGarchery (approved):

peyha (approved):

src/MorphoV2.sol
MathisGD: in general why not putting it public? if it makes sense to use it externally, why not exposing it. are you worried about the name?

peyha: was playing the devil's advocate here, I think we should do it for what you are saying but wondered if there was some extra reason linked to `isHealthy`

src/MorphoV2.sol
```diff
@@ -399,21 +399,14 @@ contract MorphoV2 is IMorphoV2 {
...
-    function _signer(bytes32 root, Signature memory signature) internal pure returns (address) {
-        bytes32 messageHash = keccak256(bytes.concat("\x19\x45thereum Signed Message:\n32", root));
-        address tentativeSigner = ecrecover(messageHash, signature.v, signature.r, signature.s);
-        require(tentativeSigner != address(0), "invalid signature");
-        return tentativeSigner;
-    }
-
-    function _isHealthy(Obligation memory obligation, address borrower) internal view returns (bool) {
-        bytes32 id = _id(obligation);
+    function isHealthy(Obligation memory obligation, address borrower) public view returns (bool) {
```
adhusson, line 408: We could also separate _isHealthy(id,obligation) and isHealthy(obligation) would save gas.

`toId` costs 2155 gas if the obligation has 2 collaterals

PR #270, opened 2025-11-13
[rounding 2] Round in favor of the maker

### LINKED ISSUE #10: Grieffing attack when rounding seller/buyer offers

QGarchery:
Improves #212 because now the maker is guaranteed its price. The taker takes the loss, and there is no guarantee on the trading fee still

fixes #10


src/Midnight.sol
```diff
@@ -212,11 +212,26 @@ contract Midnight is IMidnight {
...
+        uint256 obligationUnits = buyerIsLender ? unitsUp : unitsDown;
+
+        uint256 buyerAssets;
+        uint256 sellerAssets;
+        if (offer.buy && buyerIsLender) {
+            buyerAssets = unitsDown.mulDivDown(buyerPrice, WAD);
+            sellerAssets = unitsDown.mulDivDown(sellerPrice, WAD);
+        } else if (offer.buy && !buyerIsLender) {
+            buyerAssets = obligationUnits.mulDivDown(buyerPrice, WAD);
+            sellerAssets = obligationUnits.mulDivDown(sellerPrice, WAD);
```
```suggestion
        if (offer.buy) {
            buyerAssets = unitsDown.mulDivDown(buyerPrice, WAD);
            sellerAssets = unitsDown.mulDivDown(sellerPrice, WAD);
```
you could do that but it's not as clear why it works

src/Midnight.sol
```diff
@@ -212,11 +212,26 @@ contract Midnight is IMidnight {
...
         // used to discriminate, as the remaining two cases do not change total units and total shares.
-        uint256 obligationUnits = buyerIsLender
-            ? obligationShares.mulDivUp(_obligationState.totalUnits + 1, _obligationState.totalShares + 1)
-            : obligationShares.mulDivDown(_obligationState.totalUnits + 1, _obligationState.totalShares + 1);
-        uint256 buyerAssets = obligationUnits.mulDivDown(buyerPrice, WAD);
-        uint256 sellerAssets = obligationUnits.mulDivDown(sellerPrice, WAD);
+        uint256 unitsDown =
+            obligationShares.mulDivDown(_obligationState.totalUnits + 1, _obligationState.totalShares + 1);
+        uint256 unitsUp = obligationShares.mulDivUp(_obligationState.totalUnits + 1, _obligationState.totalShares + 1);
+        uint256 obligationUnits = buyerIsLender ? unitsUp : unitsDown;
```
MathisGD, line 218: doesn't this break "share price doesn't decrease" / passive users don't loose?

QGarchery: no it's exactly the same computation as before, I just cached the `unitsDown` and `unitsUp` variables

src/Midnight.sol
```diff
@@ -212,11 +212,26 @@ contract Midnight is IMidnight {
...
+        } else if (offer.buy && !buyerIsLender) {
+            buyerAssets = obligationUnits.mulDivDown(buyerPrice, WAD);
+            sellerAssets = obligationUnits.mulDivDown(sellerPrice, WAD);
+        } else if (!offer.buy && sellerIsBorrower) {
+            sellerAssets = obligationUnits.mulDivUp(sellerPrice, WAD);
+            buyerAssets = obligationUnits.mulDivUp(buyerPrice, WAD);
+        } else {
+            sellerAssets = unitsUp.mulDivUp(sellerPrice, WAD);
+            buyerAssets = unitsUp.mulDivUp(buyerPrice, WAD);
+        }
```
MathisGD: we can be more "generous" (not make the minimal safe rounding)
```suggestion
        if (offer.buy) {
            buyerAssets = unitsDown.mulDivDown(buyerPrice, WAD);
            sellerAssets = unitsDown.mulDivDown(sellerPrice, WAD);
        } else {
            sellerAssets = unitsUp.mulDivUp(sellerPrice, WAD);
            buyerAssets = unitsUp.mulDivUp(buyerPrice, WAD);
        }
```
becomes almost as good as in #423

MathisGD: actually it's the same thing as [this](https://github.com/morpho-org/midnight/pull/270/changes#r2884266636), no?

src/Midnight.sol
```diff
@@ -212,11 +212,13 @@ contract Midnight is IMidnight {
...
-        uint256 obligationUnits = buyerIsLender
-            ? obligationShares.mulDivUp(_obligationState.totalUnits + 1, _obligationState.totalShares + 1)
-            : obligationShares.mulDivDown(_obligationState.totalUnits + 1, _obligationState.totalShares + 1);
-        uint256 buyerAssets = obligationUnits.mulDivDown(buyerPrice, WAD);
-        uint256 sellerAssets = obligationUnits.mulDivDown(sellerPrice, WAD);
+        uint256 unitsDown =
+            obligationShares.mulDivDown(_obligationState.totalUnits + 1, _obligationState.totalShares + 1);
+        uint256 unitsUp = obligationShares.mulDivUp(_obligationState.totalUnits + 1, _obligationState.totalShares + 1);
+        uint256 obligationUnits = buyerIsLender ? unitsUp : unitsDown;
+
```
```suggestion
```

src/Midnight.sol
```diff
@@ -212,11 +212,13 @@ contract Midnight is IMidnight {
...
         // used to discriminate, as the remaining two cases do not change total units and total shares.
-        uint256 obligationUnits = buyerIsLender
-            ? obligationShares.mulDivUp(_obligationState.totalUnits + 1, _obligationState.totalShares + 1)
-            : obligationShares.mulDivDown(_obligationState.totalUnits + 1, _obligationState.totalShares + 1);
-        uint256 buyerAssets = obligationUnits.mulDivDown(buyerPrice, WAD);
-        uint256 sellerAssets = obligationUnits.mulDivDown(sellerPrice, WAD);
+        uint256 unitsDown =
+            obligationShares.mulDivDown(_obligationState.totalUnits + 1, _obligationState.totalShares + 1);
+        uint256 unitsUp = obligationShares.mulDivUp(_obligationState.totalUnits + 1, _obligationState.totalShares + 1);
+        uint256 obligationUnits = buyerIsLender ? unitsUp : unitsDown;
```
MathisGD, line 218: the fact that this doesn't depend on offer.buy shows that we don't really round in favor of the maker right? rather in favor of passive lenders, then the maker (but they could still be loosing)

src/periphery/TakeAmountsLib.sol
```diff
@@ -25,35 +25,41 @@ library TakeAmountsLib {
...
+        returns (uint256)
+    {
         uint256 offerPrice = TickLib.tickToPrice(offer.tick);
         uint256 _tradingFee = midnight.tradingFee(id, UtilsLib.zeroFloorSub(offer.obligation.maturity, block.timestamp));
         uint256 buyerPrice = offer.buy ? offerPrice : offerPrice + _tradingFee;
         require(buyerPrice <= WAD, "buyerPrice");
-        uint256 targetUnits = targetBuyerAssets.mulDivUp(WAD, buyerPrice);
-        return unitsToShares(midnight, id, taker, offer, targetUnits);
+        if (offer.buy) {
+            uint256 targetUnitsDown = targetBuyerAssets.mulDivUp(WAD, buyerPrice);
```
peyha: I find the naming `targetUnitsDown` slightly misleading

src/periphery/TakeAmountsLib.sol
```diff
@@ -25,35 +25,41 @@ library TakeAmountsLib {
...
+    // Forward: buyerAssets = offer.buy ? unitsDown.mulDivDown(buyerPrice, WAD) : unitsUp.mulDivUp(buyerPrice, WAD).
     /// @dev Should not be used if buyerPrice > WAD, because not all buyerAssets are reachable then.
-    function buyerAssetsToShares(
-        Midnight midnight,
-        bytes32 id,
-        address taker,
-        Offer memory offer,
-        uint256 targetBuyerAssets
-    ) internal view returns (uint256) {
+    function buyerAssetsToShares(Midnight midnight, bytes32 id, Offer memory offer, uint256 targetBuyerAssets)
```
peyha, line 30: just wondering, is this the reciprocal of the forward function ? 
Generally it is not but I'm wondering if it might be the case when we have buyerAssets <= units <= shares

QGarchery: Yes I think. The goal of this function is that you pass its result (some shares amount) to target the buyer assets that you want. Note that it reverts if buyerPrice > 1

src/periphery/TakeAmountsLib.sol
QGarchery: right, I meant that you target the `unitsDown` variable. Not sure what else to put here, do you have an idea ?

MathisGD: don't define the variable?

src/Midnight.sol
QGarchery: I think that it still works, in the sense that you cannot lose. Maybe you don't get the full upside of the rounding error, but you are guaranteed to have a favorable conversion of shares/units/assets.

Checking case by case for the maker:
- if you are a borrower then you don't care about shares. Either you are buying, in which case you only pay for `unitsDown <= obligationUnits`, or you are selling and you get a value corresponding to `unitsUp >= obligationUnits`
- if you are a lender, then what matters is share value, and whether you are selling or buying, you are doing so at a better value than before the trade (because the computations in `buyerAssets` and `sellerAssets` definitions are in the same direction). And for the remaining shares they gaining value (as a passive lender)

MathisGD: discussed live, it's fine

MathisGD (approved):

adhusson (approved):

PR #271, opened 2025-11-14
[rounding 1] Round in favor of passive lenders

QGarchery:
Simplification of #212, but less powerful than #270 because it doesn't protect the maker

[Related thread](https://morpholabs.slack.com/archives/C08A40KKN9W/p1771849848345719)


src/MorphoV2.sol
```diff
@@ -202,24 +202,29 @@ contract MorphoV2 is IMorphoV2 {
         uint256 sellerPrice = offer.buy ? offerPrice - _tradingFee : offerPrice;
         uint256 buyerPrice = sellerPrice + _tradingFee;
 
+        bool roundDownShares = borrowerState[id][buyer].debt == 0 && sharesOf[id][seller] == 0;
```
```suggestion
        bool roundDownShares = buyerIsSeller && sellerIsBorrower;
```

QGarchery: done in 2ad50ce

src/MorphoV2.sol
MathisGD, line 269: note:
- path 1: we want the lender to receive less shares ("deposit" parallel)
- path 2: we don't care (in this PR at least)
- path 3: same
- path 4: we want the lender to burn more shares ("withdraw" parallel)

if we put the convention that shares are rounded up on path2 and path3, then we want to round shares down if and only if buyerIsLender && sellerIsBorrower, which is what is done above

src/MorphoV2.sol
```diff
@@ -202,24 +202,31 @@ contract MorphoV2 is IMorphoV2 {
...
+        bool sellerIsBorrower = sharesOf[id][seller] == 0;
+        bool roundDownShares = buyerIsLender && sellerIsBorrower;
         if (buyerAssets > 0) {
             obligationUnits = buyerAssets.mulDivDown(WAD, buyerPrice);
             sellerAssets = buyerAssets.mulDivDown(sellerPrice, buyerPrice);
-            obligationShares =
-                obligationUnits.mulDivDown(_obligationState.totalShares + 1, _obligationState.totalUnits + 1);
+            obligationShares = obligationUnits.mulDiv(
+                _obligationState.totalShares + 1, _obligationState.totalUnits + 1, roundDownShares
+            );
```
MathisGD, line 216: (the same for each branch, achieves the same thing)
```suggestion
            obligationShares = obligationUnits.mulDiv(
                _obligationState.totalShares + 1, _obligationState.totalUnits + 1, buyerIsLender
            );
```

QGarchery: Nice that works too. I will also add a comment, because the `roundDownShares` variable name was helping understanding IMO

src/MorphoV2.sol
```diff
@@ -202,24 +202,32 @@ contract MorphoV2 is IMorphoV2 {
         uint256 sellerPrice = offer.buy ? offerPrice - _tradingFee : offerPrice;
         uint256 buyerPrice = sellerPrice + _tradingFee;
 
+        bool buyerIsLender = borrowerState[id][buyer].debt == 0;
+        bool sellerIsBorrower = sharesOf[id][seller] == 0;
+        // To ensure the share price increases, the shares should be rounded down when buyerIsLender & sellerIsBorrower,
+        // and rounded up when !buyerIsLender & !sellerIsBorrower, so buyerIsLender discriminates between the two cases.
```
```suggestion
        // To ensure the share price increases, the shares should be rounded down (units should be rounded up) when buyerIsLender & sellerIsBorrower,
        // and rounded up (resp. down) when !buyerIsLender & !sellerIsBorrower, so buyerIsLender discriminates between the two cases.
```

src/MorphoV2.sol
MathisGD: note that in path3 it's worse than "we don't care", because the obligationShares are actually never used. they are computed for nothing

test/TakeTest.sol
MathisGD, line 1: changes in tests make sense:
- path1: we only make a change when input is shares
- path2: same
- path3: no change see [this](https://github.com/morpho-org/morpho-v2/pull/271/changes#r2849062369)
- path4: we make changes on all inputs except shares input

MathisGD (approved):
🔥

peyha (approved):

PR #272, opened 2025-11-17
remove chain and contract for 712 struct

adhusson:


src/MorphoV2.sol
```diff
@@ -479,7 +479,7 @@ contract MorphoV2 is IMorphoV2 {
     }
 
     function _signer(bytes32 root, Signature memory signature) internal view returns (address) {
-        bytes32 structHash = keccak256(abi.encode(ROOT_TYPEHASH, root, block.chainid, address(this)));
+        bytes32 structHash = keccak256(abi.encode(ROOT_TYPEHASH, root));
```
QGarchery, line 482: It has the issue that a given signature could be replayed on a different contract/chain, or am I missing something ?

adhusson: What is signed is the concatenation of the domain separator and the root, and the domain separator is scoped to the chain and the contract.

QGarchery: Oh right, I got confused :sweat_smile:

PR #279, opened 2025-12-11
quick fix: allow optimizer

adhusson:
Tracking better fixes and more optimizer runs [here](https://github.com/morpho-org/morpho-v2/issues/280)


foundry.toml
```diff
@@ -1,5 +1,7 @@
 [profile.default]
 via_ir = true
+optimizer = true
+optimizer_runs = 216
```
MathisGD: 216?

src/MorphoV2.sol
```diff
@@ -468,10 +453,41 @@ contract MorphoV2 is IMorphoV2 {
...
+    /// INTERNAL ///
+
     function _signer(bytes32 root, Signature memory signature) internal pure returns (address) {
         bytes32 messageHash = keccak256(bytes.concat("\x19\x45thereum Signed Message:\n32", root));
         address tentativeSigner = ecrecover(messageHash, signature.v, signature.r, signature.s);
         require(tentativeSigner != address(0), "invalid signature");
         return tentativeSigner;
     }
+
+    // Separated to avoid stack too deep error
```
MathisGD: 💀

foundry.toml
adhusson: max value I could get, after that it stack too deeps again

src/MorphoV2.sol
adhusson: "quick fix" 🙃

foundry.toml
adhusson: updated with a better fix runs is now 999999

MathisGD:
@codex review

QGarchery (approved):

MathisGD:
replaced by #304

PR #281, opened 2025-12-19
style: name mapping keys

MathisGD:


src/MorphoV2.sol
```diff
@@ -26,12 +26,13 @@ contract MorphoV2 is IMorphoV2 {
...
-    mapping(bytes32 => uint256) public totalUnits;
-    mapping(bytes32 => uint256) public totalShares;
-    mapping(address => mapping(bytes32 => mapping(address => uint256))) public collateralOf;
+    mapping(address user => mapping(bytes32 obligationId => uint256)) public sharesOf;
+    mapping(address user => mapping(bytes32 obligationId => uint256)) public debtOf;
+    mapping(bytes32 obligationId => uint256) public withdrawable;
+    mapping(bytes32 obligationId => uint256) public totalUnits;
+    mapping(bytes32 obligationId => uint256) public totalShares;
+    mapping(address user => mapping(bytes32 obligationId => mapping(address collateralToken => uint256))) public
+        collateralOf;
```
QGarchery, line 35: This makes it clear that the first key could be always `obligationId`, and that `collateralOf` is more closely related to `sharesOf` and `debtOf`

QGarchery (approved):

PR #282, opened 2025-12-24
feat: implement min borrow amount

peyha:


peyha:
closing in favor of #320

PR #283, opened 2025-12-24
Batch liquidation

peyha:
Implementation was simpler than I expected


MathisGD:
should we close?

peyha:
yes, it was a solution against dust position accumulation/position splitting attack but min collat solves it

PR #285, opened 2026-01-08
POC Min collat + recovery close factor

peyha:


src/MorphoV2.sol
```diff
@@ -469,6 +472,22 @@ contract MorphoV2 is IMorphoV2 {
         }
     }
 
+    function hasMinCollateral(Obligation memory obligation, address borrower) public view returns (bool) {
+        bytes32 id = toId(obligation);
```
adhusson, line 482: obligation -> id is expensive, we could add it as an argument

src/interfaces/IMorphoV2.sol
```diff
@@ -8,6 +8,7 @@ struct Obligation {
     // Must be sorted by address.
     Collateral[] collaterals;
     uint256 maturity;
+    uint256 minCollateral; // quoted in loan token
```
adhusson:
```suggestion
    uint256 minCollateralValue; // quoted in loan token
```

src/MorphoV2.sol
```diff
@@ -254,6 +254,7 @@ contract MorphoV2 is IMorphoV2 {
                     buyerCallbackData
                 );
         }
+        require(hasMinCollateral(offer.obligation, buyer), "Buyer has insufficient collateral");
```
adhusson: buyer is reducing his debt so if he was in a state where his collateral value had become too low why not let him improve his situation

src/MorphoV2.sol
adhusson, line 358: min collateral only makes sense in the context of recovery close factor so maybe we could combine both in one PR

peyha: was about to create a PR on top but will do I can do it in this one

src/MorphoV2.sol
peyha: I get stack too deep when I implement it, will try to refactor before

adhusson: maybe let's wait until https://github.com/morpho-org/morpho-v2/pull/279 or similar is merged to save time on this

src/MorphoV2.sol
```diff
@@ -415,6 +417,9 @@ contract MorphoV2 is IMorphoV2 {
 
         SafeTransferLib.safeTransferFrom(obligation.loanToken, msg.sender, address(this), totalRepaid);
 
+        if (block.timestamp <= obligation.maturity) {
+            require(!isHealthy(obligation, borrower) || badDebt > 0, "Recovery close factor");
```
peyha: `badDebt > 0` allows to fully liquidate a position if it generates bad debt, it's not mandatory but otherwise seizing the full collateral/repaying all the obligation units will revert

src/MorphoV2.sol
```diff
@@ -469,6 +474,22 @@ contract MorphoV2 is IMorphoV2 {
...
     }
 
+    function hasMinCollateral(Obligation memory obligation, address borrower) public view returns (bool) {
+        bytes32 id = toId(obligation);
+        if (debtOf[borrower][id] == 0) {
+            return true;
+        }
+
+        uint256 totalCollateral = 0;
+        for (uint256 i = 0; i < obligation.collaterals.length; i++) {
```
MathisGD, line 488: probably worth checking only for the collateral that was touched (instead of looping)

src/MorphoV2.sol
MathisGD, line 358: you don't need to be able to seize multiple collaterals at the same time once you have recovery close factor. and arguably simplicity matters more than the gas saving in the case of multi-collat
```suggestion
    function liquidate(Obligation memory obligation, address collateral, address borrower, uint256 seized, uint256 repaid, bytes calldata data)
```

src/MorphoV2.sol
```diff
@@ -415,6 +417,9 @@ contract MorphoV2 is IMorphoV2 {
 
         SafeTransferLib.safeTransferFrom(obligation.loanToken, msg.sender, address(this), totalRepaid);
 
+        if (block.timestamp <= obligation.maturity) {
+            require(!isHealthy(obligation, borrower) || originalDebt >= repayableDebt, "Recovery close factor");
+        }
```
MathisGD, line 426: I think that we can do something better than this. will try

MathisGD: #303

peyha:
closing in favor of #303

PR #286, opened 2026-01-08
refactor: separate buy sell functions

MathisGD:
not convinced, wdyt?


MathisGD:
@codex review

peyha:
It's +50 lines of code and +500 bytecode size while the clarity improvement in the code is not significant so not convinced

adhusson (commented):
same, not convinced, would be worth it if it created a lot of code simplification but it's not here

PR #288, opened 2026-01-09
Chore/optimizer

peyha:


peyha:
The easiest way to remove a few variables on the stack right now is to define the callback variables at the end of the take function.

PR #238, opened 2025-11-04
[continuous fee] global, lender side

MathisGD:

Issue #226, opened 2025-10-31
buyerPrice > 1 is ok?


MathisGD:
> This makes me think: shouldn't we require the buyerPrice to be less than WAD ?  

 _Originally posted by @QGarchery in [#217](https://github.com/morpho-org/morpho-v2/pull/217/files#r2478832242)_

MathisGD:
https://morpholabs.slack.com/archives/C08A40KKN9W/p1761906060833739

peyha:
Re-opening as the interest cut limit was removed

MathisGD:
settled last week, it's ok to have a taker price above one but not a maker price above one (and the tick formula prevents it naturally)

QGarchery:
[Current state of things](https://morpholabs.slack.com/archives/C08A40KKN9W/p1772480664248809?thread_ts=1772454137.334879&cid=C08A40KKN9W)

PR #289, opened 2026-01-11
feat: new trading fees

### LINKED ISSUE #175: default fee

MathisGD:
changes:
- fees are taken on the notional/face value (instead of the traded value) ([context](https://morpholabs.slack.com/archives/C02N7CZ088N/p1767880539487179))
- default fee (fixes #175)
- piecewise linear fees


src/MorphoV2.sol
```diff
@@ -158,24 +165,11 @@ contract MorphoV2 is IMorphoV2 {
...
-        } else {
-            sellerPrice = offerPrice;
-            buyerPrice = UtilsLib.min(
-                sellerPrice.mulDivDown(WAD - _tradingFeeParams.interestCutLimit, WAD)
-                    + _tradingFeeParams.interestCutLimit,
-                sellerPrice.mulDivDown(WAD + _tradingFeeParams.tradingFee, WAD)
-            );
-        }
+        uint256 ttm = UtilsLib.zeroFloorSub(offer.obligation.maturity, block.timestamp);
+        uint256 tradingFeeIndex = ttm < 1 days ? 0 : ttm < 7 days ? 1 : ttm < 30 days ? 2 : ttm < 90 days ? 3 : 4;
```
peyha: The post-maturity trading fee is the same as the zero days trading fee, is this ok ? 
Post-maturity trades should essentially be lenders leaving at a loss so I intuitively expected a very low fee on them

MathisGD: yes, I think that this makes sense. you also expect a very low fee very close to maturity (otherwise it ruins the rate)

src/MorphoV2.sol
```diff
@@ -87,14 +79,29 @@ contract MorphoV2 is IMorphoV2 {
         emit EventsLib.SetFeeSetter(newFeeSetter);
     }
 
-    function setTradingFee(bytes32 id, uint256 tradingFee, uint256 interestCutLimit) external {
+    function setTradingFee(
```
MathisGD: maybe there is a nicer way to do

src/MorphoV2.sol
adhusson: We could use the tree of ifs we discussed to save some gas without adding complexity

adhusson (approved):

src/MorphoV2.sol
peyha: Agreed.
Just to have an overview on the magnitude of these. A 10% obligation taken for half a day would correspond to a price for which 1-P=1bps so the trading fee would have to be negligible compared to it (0.1 bps or even lower?)
Post-maturity trading prices are difficult to estimate because it would correspond to tail events but it's likely to be way lower than the "less than 1 day price" so it work (seems like there would even be an opportunity to raise the trading fee after maturity)

peyha:
Worth noting that this design incentivizes users to take at specific times (1 day, 7 day, 30 day, 90 days before maturity) because, assuming a constant rate, they will pay less fees compared to their interest

src/MorphoV2.sol
peyha: another possibility would be to use `(uint256,uint256)[]` to store (trading fee, period) tuples to potentially change the granularity of the fee. it's an option if we want to add a new time to maturity band

src/MorphoV2.sol
```diff
@@ -158,24 +165,11 @@ contract MorphoV2 is IMorphoV2 {
...
-            buyerPrice = UtilsLib.min(
-                sellerPrice.mulDivDown(WAD - _tradingFeeParams.interestCutLimit, WAD)
-                    + _tradingFeeParams.interestCutLimit,
-                sellerPrice.mulDivDown(WAD + _tradingFeeParams.tradingFee, WAD)
-            );
-        }
+        uint256 ttm = UtilsLib.zeroFloorSub(offer.obligation.maturity, block.timestamp);
+        uint256 tradingFeeIndex = ttm < 1 days ? 0 : ttm < 7 days ? 1 : ttm < 30 days ? 2 : ttm < 90 days ? 3 : 4;
+        uint256 tradingFee = tradingFees[id][tradingFeeIndex];
+        uint256 buyerPrice = offer.buy ? offerPrice : offerPrice + tradingFee;
```
adhusson: same [comment](https://github.com/morpho-org/morpho-v2/pull/287/changes#r2675502629) as chatgpt here. Or keeping fee as % of notional is ok it will constraint offered prices anyway but the final buyerPrice should be capped to 1.

src/MorphoV2.sol
MathisGD: yes, the OOM for 1 day trading fees is 0.1bps. Indeed, there is probably an opportunity to have a higher after the maturity, as there is indeed a shift in "reasons to use". Though idk we should optimize for that, as it remains a tail thing.

peyha: I don't think so neither. At a high level it seems fine to "remove" the fee for traders losing funds through bad debt

src/MorphoV2.sol
MathisGD: why is capping to 1 important actually?

adhusson: to protect users, is there a scenario where price > 1 is not a mistake?

src/MorphoV2.sol
```diff
@@ -87,14 +79,29 @@ contract MorphoV2 is IMorphoV2 {
...
+        require(ninetyDaysTradingFee <= WAD, "90days trading fee too high");
         // forge-lint: disable-next-item(unsafe-typecast) Safe cast because values are below type(uint128).max.
-        tradingFeeParams[id] =
-            TradingFeeParams({tradingFee: uint128(tradingFee), interestCutLimit: uint128(interestCutLimit)});
-        emit EventsLib.SetTradingFee(id, tradingFee, interestCutLimit);
+        tradingFees[id][0] = zeroDaysTradingFee;
+        tradingFees[id][1] = oneDaysTradingFee;
+        tradingFees[id][2] = sevenDaysTradingFee;
+        tradingFees[id][3] = thirtyDaysTradingFee;
+        tradingFees[id][4] = ninetyDaysTradingFee;
```
adhusson: to pack (in a struct for instance)

src/MorphoV2.sol
adhusson: the current signature is simple enough

src/MorphoV2.sol
MathisGD: nb: it's not that useful, because we will only read one per transaction

adhusson: not the most common case but most rolling will profit from it

MathisGD: not for the per obligation mapping, but for the default one yes

src/MorphoV2.sol
MathisGD: let's do in an other PR

src/MorphoV2.sol

src/MorphoV2.sol
```diff
@@ -156,26 +177,13 @@ contract MorphoV2 is IMorphoV2 {
...
-        } else {
-            sellerPrice = offerPrice;
-            buyerPrice = UtilsLib.min(
-                sellerPrice.mulDivDown(WAD - _tradingFeeParams.interestCutLimit, WAD)
-                    + _tradingFeeParams.interestCutLimit,
-                sellerPrice.mulDivDown(WAD + _tradingFeeParams.tradingFee, WAD)
-            );
-        }
+        uint256 ttm = UtilsLib.zeroFloorSub(offer.obligation.maturity, block.timestamp);
+        uint256 tradingFee = _obligationTradingFee[id] != 0
```
MathisGD: not fully satisfying. should we have a bool "activated" so that one can overwrite the default fee to zero?

src/MorphoV2.sol
MathisGD, line 1: >Worth noting that this design incentivizes users to take at specific times (1 day, 7 day, 30 day, 90 days before maturity) because, assuming a constant rate, they will pay less fees compared to their interest

@peyha yes and no. in terms of pure maths yes, but I don't think that this is the right way to think about the trading fee anyway. the values are supposed to stay negligible in terms of the interest.

src/MorphoV2.sol
```diff
@@ -87,14 +100,24 @@ contract MorphoV2 is IMorphoV2 {
...
-        require(interestCutLimit < WAD, "Interest cut limit too high");
-        // forge-lint: disable-next-item(unsafe-typecast) Safe cast because values are below type(uint128).max.
-        tradingFeeParams[id] =
-            TradingFeeParams({tradingFee: uint128(tradingFee), interestCutLimit: uint128(interestCutLimit)});
-        emit EventsLib.SetTradingFee(id, tradingFee, interestCutLimit);
+        require(newTradingFee <= WAD, "Trading fee too high");
+        require(index < 6, "Invalid index");
+        _defaultTradingFee[loanToken] =
+            _defaultTradingFee[loanToken] & ~(0xFFFFFFFF << (index * 32)) | (newTradingFee / 1e9 << (index * 32));
+        emit EventsLib.SetDefaultTradingFee(loanToken, index, newTradingFee);
```
QGarchery, line 109: Should it log `newTradingFee / 1e9 * 1e9` instead ? To avoid that question, IMO the best is to revert when `newTradingFee` is not a multiple of `1e9`

MathisGD: good point

src/MorphoV2.sol
```diff
@@ -156,26 +179,13 @@ contract MorphoV2 is IMorphoV2 {
...
-                sellerPrice.mulDivDown(WAD + _tradingFeeParams.tradingFee, WAD)
-            );
-        }
+        uint256 ttm = UtilsLib.zeroFloorSub(offer.obligation.maturity, block.timestamp);
+        uint256 tradingFee = _obligationTradingFee[id] != 0
+            ? obligationTradingFee(id, ttm)
+            : defaultTradingFee(offer.obligation.loanToken, ttm);
+        uint256 sellerPrice = offer.buy ? offerPrice - tradingFee : offerPrice;
+        uint256 buyerPrice = sellerPrice + tradingFee;
+        require(buyerPrice <= WAD, "cannot trade at price above one");
```
peyha, line 181: still not completely settled on this
On the one hand, I get that it would possible for a lender to have an unwanted price > 1 if a new fee has been set by the governance between the offer creation and settlement
On the other hand, fee activation should be a one time event that is done sufficiently early on the bond so lenders should be aware of it in advance. 
Moreover we can easily imagine market conditions that would lead to price > 1 such as incentives. In tradfi there are also [many reasons](https://www.mastertrust.co.in/blog/why-do-investors-buy-bonds-that-are-sold-at-a-negative-rate) that can lead to negative rate bond (even though there's usually a coupon)
In the end I worry more about missing a dimension of use-cases compared to an edge case related to the asynchrony of fee modification in the protocol
([related thread](https://morpholabs.slack.com/archives/C02N7CZ088N/p1768323663743099))

src/MorphoV2.sol
MathisGD: solved in #293

src/MorphoV2.sol
```diff
@@ -43,8 +35,15 @@ contract MorphoV2 is IMorphoV2 {
...
-    /// @dev Trading fee parameters for a given obligation id.
-    mapping(bytes32 id => TradingFeeParams) public tradingFeeParams;
+    /// @dev Obligation trading fees for a given obligation id.
+    /// @dev The slot contains the 6 trading fees packed (each takes 32 bits).
+    mapping(bytes32 obligationId => uint256) private _obligationTradingFee;
+
+    /// @dev Default trading fees per loan token.
+    /// @dev The slot contains the 6 trading fees packed (each takes 32 bits).
+    /// @dev Used when obligation fee is not set (slot is empty).
+    mapping(address loanToken => uint256) private _defaultTradingFee;
```
adhusson: just `internal` is enough

src/MorphoV2.sol
MathisGD: more on this: https://morpholabs.slack.com/archives/C02N7CZ088N/p1768414082315599

src/MorphoV2.sol
peyha: update: i'm starting to change my mind, it creates very weird situation regarding interest fees

src/MorphoV2.sol
```diff
@@ -469,10 +469,36 @@ contract MorphoV2 is IMorphoV2 {
...
+    function signer(bytes32 root, Signature memory signature) internal pure returns (address) {
         bytes32 messageHash = keccak256(bytes.concat("\x19\x45thereum Signed Message:\n32", root));
         address tentativeSigner = ecrecover(messageHash, signature.v, signature.r, signature.s);
         require(tentativeSigner != address(0), "invalid signature");
         return tentativeSigner;
     }
+
+    /// @dev Return the trading fee using piecewise linear interpolation between breakpoints.
+    /// @dev Returns 0 if neither obligation nor default fee is activated.
+    function tradingFee(bytes32 id, address loanToken, uint256 ttm) public view returns (uint256) {
```
MathisGD: open to implementation ideas, notably for things that would increase the readability. fees will be read a lot!

peyha: a way to improve readability could be to use a FeeMathLib that would abstract the bit computations.
`tradingFeeStorage & 1 == 0` could be replaced by `isFeeSet(tradingFee)`
`uint24(tradingFeeStorage >> (1 + index * 24))` could be replaced by `feeAtIndex(tradingFee, index)`

MathisGD: good point, this should work well as it doesn't contain business logic

src/MorphoV2.sol
```diff
@@ -469,10 +466,33 @@ contract MorphoV2 is IMorphoV2 {
...
+        uint256 feeStorage = _obligationTradingFeeStorage[id];
+        if (!FeeLib.isActivated(feeStorage)) {
+            feeStorage = _defaultTradingFeeStorage[loanToken];
+            if (!FeeLib.isActivated(feeStorage)) return 0;
+        }
+
+        uint256[6] memory breakpoints = [uint256(0), 1 days, 7 days, 30 days, 90 days, 180 days];
```
MathisGD: not convinced that we need this much granularity with piecewise linear, but it doesn't seems to cost much neither

src/libraries/FeeLib.sol
```diff
@@ -0,0 +1,37 @@
...
+
+/// @title FeeLib
+/// @notice Library for packing/unpacking trading fees in a single uint256.
+/// @dev Storage layout:
+///   - Bit 0: activated flag
+///   - Bits 1-144: 6 trading fees packed (24 bits each)
+///   - Fee indices: 0=0d, 1=1d, 2=7d, 3=30d, 4=90d, 5=180d
+/// @dev Fees are stored divided by 1e12 to fit in 24 bits.
+library FeeLib {
+    uint256 internal constant FEE_PRECISION = 1e12;
```
adhusson: The fee precision is 1e18-1e12, or 18-12 decimals, maybe another name?

Or simpler directly ask for fees in 1e6, or even simpler in basis points so people are used to it.

src/libraries/FeeLib.sol
```diff
@@ -0,0 +1,37 @@
...
+
+    /// @dev Returns whether the fee storage is activated.
+    function isActivated(uint256 feeStorage) internal pure returns (bool) {
+        return feeStorage & ACTIVATED_MASK != 0;
+    }
+
+    /// @dev Returns the fee at the given index, scaled back to WAD precision.
+    function getFee(uint256 feeStorage, uint256 index) internal pure returns (uint256) {
+        // forge-lint: disable-next-line(unsafe-typecast)
+        return uint256(uint24(feeStorage >> (1 + index * FEE_BITS))) * FEE_PRECISION;
```
adhusson: by making the activated mask the high bit it simplifies this and setFee because we don't need the `1 +`.

src/MorphoV2.sol
```diff
@@ -87,14 +87,26 @@ contract MorphoV2 is IMorphoV2 {
...
+        require(newTradingFee <= WAD, "Trading fee too high");
+        require(index <= 5, "Invalid index");
+        require(newTradingFee % FeeLib.FEE_PRECISION == 0, "fee should be a multiple of 1e12");
+
+        _obligationTradingFeeStorage[id] =
+            FeeLib.setFee(_obligationTradingFeeStorage[id], index, newTradingFee, activated);
+        emit EventsLib.SetObligationTradingFee(id, index, newTradingFee);
+    }
+
+    function setDefaultTradingFee(address loanToken, uint256 index, uint256 newTradingFee, bool activated) external {
```
adhusson: if we ever need to save space we can encode default trading fee as `setTradingFee(loanToken = 0, ...)`

src/MorphoV2.sol
```diff
@@ -479,10 +466,33 @@ contract MorphoV2 is IMorphoV2 {
...
+
+        uint256 index = ttm < breakpoints[1]
+            ? 0
+            : ttm < breakpoints[2] ? 1 : ttm < breakpoints[3] ? 2 : ttm < breakpoints[4] ? 3 : 4;
+
+        uint256 feeLower = FeeLib.getFee(feeStorage, index);
+        uint256 feeUpper = FeeLib.getFee(feeStorage, index + 1);
+
+        return (feeLower * (breakpoints[index + 1] - ttm) + feeUpper * (ttm - breakpoints[index]))
+            / (breakpoints[index + 1] - breakpoints[index]);
```
```suggestion
        return feeLower + (feeUpper - feeLower) * (ttm - breakpoints[index]) / (breakpoints[index + 1] - breakpoints[index]);
```
equivalent, but maybe a bit clearer that it is a linear interpolation

src/MorphoV2.sol
```diff
@@ -100,23 +87,25 @@ contract MorphoV2 is IMorphoV2 {
...
-    /// @dev Trading fees are truncated to the nearest 1e9.
-    function setObligationTradingFee(bytes32 id, uint256 index, uint256 newTradingFee) external {
+    function setObligationTradingFee(bytes32 id, uint256 index, uint256 newTradingFee, bool activated) external {
         require(msg.sender == feeSetter, "Only feeSetter");
         require(newTradingFee <= WAD, "Trading fee too high");
-        require(index < 6, "Invalid index");
-        _obligationTradingFee[id] =
-            _obligationTradingFee[id] & ~(0xFFFFFFFF << (index * 32)) | (newTradingFee / 1e9 << (index * 32));
+        require(index <= 5, "Invalid index");
+        require(newTradingFee % FeeLib.FEE_PRECISION == 0, "fee should be a multiple of 1e12");
```
QGarchery: (optional) those 2 `require` could be moved to the lib. Reasoning is that they are mentioning constants specific to the lib (FEE_PRECISION and max index)

src/MorphoV2.sol
```diff
@@ -100,23 +87,25 @@ contract MorphoV2 is IMorphoV2 {
         emit EventsLib.SetFeeSetter(newFeeSetter);
     }
 
-    /// @dev Trading fees are truncated to the nearest 1e9.
-    function setObligationTradingFee(bytes32 id, uint256 index, uint256 newTradingFee) external {
+    function setObligationTradingFee(bytes32 id, uint256 index, uint256 newTradingFee, bool activated) external {
```
QGarchery: Wondering if deactivating should not be a dedicated function, that would reset all the bits

src/MorphoV2.sol
MathisGD: but it breaks if fees are decreasing. which we don't expect to happen, but then we would need to verify that they are increasing in the setter...

src/MorphoV2.sol
MathisGD: agree especially for the second one. the index logic is more "business logic"

src/MorphoV2.sol
MathisGD: setActivated, and if false it checks that all fees are at zero?

src/MorphoV2.sol
```diff
@@ -469,10 +466,33 @@ contract MorphoV2 is IMorphoV2 {
...
+            feeStorage = _defaultTradingFeeStorage[loanToken];
+            if (!FeeLib.isActivated(feeStorage)) return 0;
+        }
+
+        uint256[6] memory breakpoints = [uint256(0), 1 days, 7 days, 30 days, 90 days, 180 days];
+        if (ttm >= breakpoints[5]) return FeeLib.getFee(feeStorage, 5);
+
+        uint256 index = ttm < breakpoints[1]
+            ? 0
+            : ttm < breakpoints[2] ? 1 : ttm < breakpoints[3] ? 2 : ttm < breakpoints[4] ? 3 : 4;
```
adhusson: not sure but probably lower gas to avoid allocating an array

```suggestion
        if (ttm >= 180 days) return FeeLib.getFee(feeStorage, 5);
        
        (uint256 index, uint start, uint end) = ttm < 1 days
            ? (0, 0 days, 1 days)
            : ttm < 7 days ? (1, 1 days, 7 days) : ttm < 30 days ? (2, 7 days, 30 days) : ttm < 90 days ? (3, 30 days, 90 days) : (4, 90 days, 180 days);
```

MathisGD: @codex how much gas will it save

src/MorphoV2.sol
MathisGD: like id = loanToken?

src/libraries/FeeLib.sol
MathisGD: I find it great that everything is in 1e18. we can just find a better name here

src/libraries/FeeLib.sol
MathisGD: alright, I can change that

src/MorphoV2.sol
MathisGD: actually I like to have all require in the same place

src/MorphoV2.sol
MathisGD: had the same intuition. I think that the current version is cleaner then.

src/MorphoV2.sol
MathisGD: good idea, will resurface if needed

certora/confs/SharePrice.conf
```diff
@@ -15,7 +15,8 @@
   "smt_timeout": 7200,
   "prover_args": [
     "-destructiveOptimizations twostage",
-    "-s [z3:def{randomSeed=1},z3:def{randomSeed=2},z3:def{randomSeed=3},z3:def{randomSeed=4},z3:def{randomSeed=5},z3:def{randomSeed=6},z3:def{randomSeed=7},z3:def{randomSeed=8},z3:def{randomSeed=9},z3:def{randomSeed=10}]"
+    "-s [z3:def{randomSeed=1},z3:def{randomSeed=2},z3:def{randomSeed=3},z3:def{randomSeed=4},z3:def{randomSeed=5},z3:def{randomSeed=6},z3:def{randomSeed=7},z3:def{randomSeed=8},z3:def{randomSeed=9},z3:def{randomSeed=10}]",
+    "-timeout 10800"
```
QGarchery:
```suggestion
    "-timeout 7200"
```
anything above is useless because the prover is always capped at 2 hours

src/MorphoV2.sol
adhusson: why require that all fees are zero first?

MathisGD: then you have the invariant deactivated => fees are zero

src/MorphoV2.sol
```diff
@@ -87,28 +87,44 @@ contract MorphoV2 is IMorphoV2 {
...
+        uint256 feeStorage = _obligationTradingFeeStorage[id];
         require(msg.sender == feeSetter, "Only feeSetter");
         require(newTradingFee <= WAD, "Trading fee too high");
         require(index <= 5, "Invalid index");
-        require(newTradingFee % FeeLib.FEE_PRECISION == 0, "fee should be a multiple of 1e12");
-
-        _obligationTradingFeeStorage[id] =
-            FeeLib.setFee(_obligationTradingFeeStorage[id], index, newTradingFee, activated);
+        require(newTradingFee % FeeLib.FEE_DIVISOR == 0, "fee should be a multiple of 1e12");
+        require(newTradingFee == 0 || FeeLib.getActivated(feeStorage), "fee must be activated");
```
```suggestion
        require(FeeLib.getActivated(feeStorage), "fee must be activated");
```
it doesn't restrict usage to do this

src/MorphoV2.sol
```diff
@@ -87,28 +87,44 @@ contract MorphoV2 is IMorphoV2 {
...
+        uint256 feeStorage = _defaultTradingFeeStorage[loanToken];
         require(msg.sender == feeSetter, "Only feeSetter");
         require(newTradingFee <= WAD, "Trading fee too high");
         require(index <= 5, "Invalid index");
-        require(newTradingFee % FeeLib.FEE_PRECISION == 0, "fee should be a multiple of 1e12");
-
-        _defaultTradingFeeStorage[loanToken] =
-            FeeLib.setFee(_defaultTradingFeeStorage[loanToken], index, newTradingFee, activated);
+        require(newTradingFee % FeeLib.FEE_DIVISOR == 0, "fee should be a multiple of 1e12");
+        require(newTradingFee == 0 || FeeLib.getActivated(feeStorage), "fee must be activated");
```
```suggestion
        require(FeeLib.getActivated(feeStorage), "fee must be activated");
```

src/libraries/FeeLib.sol
adhusson: `FEE_GRANULARITY` `FEE_STEP` `FEE_RESCALING`

src/MorphoV2.sol
adhusson: chatgpt responded to some other idea here (tree of ifs).

testing with optimizer runs at 999999 (using [this](https://github.com/morpho-org/morpho-v2/pull/279)) and 3k fuzz runs shows a 680 gas diff (well worth it imo). The code in the end is


```solidity
        if (ttm >= 180 days) return FeeLib.getFee(feeStorage, 5);

        (uint256 index, uint256 start, uint256 end) = ttm < 1 days
            ? (0, 0 days, 1 days)
            : ttm < 7 days
                ? (1, 1 days, 7 days)
                : ttm < 30 days ? (2, 7 days, 30 days) : ttm < 90 days ? (3, 30 days, 90 days) : (4, 90 days, 180 days);

        uint256 feeLower = FeeLib.getFee(feeStorage, index);
        uint256 feeUpper = FeeLib.getFee(feeStorage, index + 1);

        return (feeLower * (end - ttm) + feeUpper * (ttm - start)) / (end - start);
```

src/MorphoV2.sol
QGarchery: pushing a bit for this, still optional and it's not extremely important but it makes checking the code a bit nicer. In particular, with the activated bit being the last bit now, one needs to check that the `set*TradingFee` functions are not called with any possible index (that would risk overwriting the activated bit). If the `require` is in the lib there is nothing to check, but if it is outside you need to check every time it's called

QGarchery (approved):

src/MorphoV2.sol
peyha: re-opened #226 and resolving this as it's not the goal of the PR

src/MorphoV2.sol
adhusson: not sure it's super interesting having this invariant, there is no timelock involved so it doesn't give you guarantees. In a timelock world you could know that if you don't get charged a fee now you'll have to wait `feeChangeTimelock` and not `min(feeActivationTimelock, feeChangeTimelock)` before you get charged. Here everything can be done atomically anyway.

On the other hand it could be useful to temporarily disable a fee without having to unset the reset everything.

Not a big deal either way.

src/libraries/FeeLib.sol
```diff
@@ -0,0 +1,51 @@
...
+///   - Fee indices: 0=0d, 1=1d, 2=7d, 3=30d, 4=90d, 5=180d
+/// @dev Fees are stored divided by 1e12 to fit in 24 bits.
+library FeeLib {
+    uint256 internal constant FEE_DIVISOR = 1e12;
+    uint256 internal constant FEE_BITS = 24;
+    uint256 internal constant FEE_MASK = 0xFFFFFF;
+    uint256 internal constant ACTIVATED_MASK = 1 << 255;
+
+    /// @dev Returns whether the fee storage is activated.
+    function getActivated(uint256 feeStorage) internal pure returns (bool) {
```
peyha, line 19: I slightly prefer the naming `isActivated` for a getter returning a bool

src/libraries/FeeLib.sol
```diff
@@ -0,0 +1,51 @@
...
+// Copyright (c) 2025 Morpho Association
+pragma solidity ^0.8.0;
+
+/// @title FeeLib
+/// @notice Library for packing/unpacking trading fees in a single uint256.
+/// @dev Storage layout:
+///   - Bit 255: activated flag
+///   - Bits 0-143: 6 trading fees packed (24 bits each)
+///   - Fee indices: 0=0d, 1=1d, 2=7d, 3=30d, 4=90d, 5=180d
+/// @dev Fees are stored divided by 1e12 to fit in 24 bits.
```
peyha: shouldn't we store fees on 20 bits ? If a fee is less than 100%(<>1e18), then we need 1e6 we can store it on log2(1e6)~20 to save a few bits (for the interest fee & possibly a boolean if the obligation is active)

src/MorphoV2.sol
MathisGD: >that would risk overwriting the activated bit

as it's divided in the lib, this cannot happen. it's more that you would loose some precision

src/MorphoV2.sol
```diff
@@ -156,26 +184,11 @@ contract MorphoV2 is IMorphoV2 {
...
-            );
-        } else {
-            sellerPrice = offerPrice;
-            buyerPrice = UtilsLib.min(
-                sellerPrice.mulDivDown(WAD - _tradingFeeParams.interestCutLimit, WAD)
-                    + _tradingFeeParams.interestCutLimit,
-                sellerPrice.mulDivDown(WAD + _tradingFeeParams.tradingFee, WAD)
-            );
-        }
+        uint256 ttm = UtilsLib.zeroFloorSub(offer.obligation.maturity, block.timestamp);
```
```suggestion
        uint256 timeToMaturity = UtilsLib.zeroFloorSub(offer.obligation.maturity, block.timestamp);
```

src/libraries/FeeLib.sol
MathisGD: I used 24 before because I was using uint24, but at the end we do low level stuff so uint20 is fine

test/TradingFeeTest.sol
peyha, line 1: Should we add a test for the corner temporal cases: post-maturity & after 180 days ?

src/libraries/FeeLib.sol
peyha: we still have plenty of bit space left in the trading fee so it's not necessary either

MathisGD: done

MathisGD:
@codex review again

src/MorphoV2.sol
```diff
@@ -87,14 +87,36 @@ contract MorphoV2 is IMorphoV2 {
         emit EventsLib.SetFeeSetter(newFeeSetter);
     }
 
-    function setTradingFee(bytes32 id, uint256 tradingFee, uint256 interestCutLimit) external {
+    function setObligationTradingFee(bytes32 id, uint256 index, uint256 newTradingFee) external {
+        uint256 feeStorage = _obligationTradingFeeStorage[id];
+        require(msg.sender == feeSetter, "Only feeSetter");
+        require(newTradingFee <= WAD, "Trading fee too high");
+        require(index <= 5, "Invalid index");
+        _obligationTradingFeeStorage[id] = FeeLib.setFee(feeStorage, index, newTradingFee);
```
```suggestion
        require(msg.sender == feeSetter, "Only feeSetter");
        require(newTradingFee <= WAD, "Trading fee too high");
        require(index <= 5, "Invalid index");
        _obligationTradingFeeStorage[id] = FeeLib.setFee(_obligationTradingFeeStorage[id], index, newTradingFee);
```

src/MorphoV2.sol
```diff
@@ -43,8 +35,16 @@ contract MorphoV2 is IMorphoV2 {
     /// @dev The session can be shuffled by the user to cancel all current offers easily and efficiently.
     mapping(address user => bytes32) public session;
 
-    /// @dev Trading fee parameters for a given obligation id.
-    mapping(bytes32 id => TradingFeeParams) public tradingFeeParams;
+    /// @dev Obligation trading fees for a given obligation id.
+    /// @dev Bit 0: activated flag. Bits 1-144: 6 trading fees packed (24 bits each).
+    /// @dev Fee indices: 0=0d, 1=1d, 2=7d, 3=30d, 4=90d, 5=180d.
+    mapping(bytes32 obligationId => uint256) internal _obligationTradingFeeStorage;
```
adhusson, line 41: why "Storage"?

src/MorphoV2.sol
```diff
@@ -87,14 +87,36 @@ contract MorphoV2 is IMorphoV2 {
...
+        require(newTradingFee <= WAD, "Trading fee too high");
+        require(index <= 5, "Invalid index");
+        _obligationTradingFeeStorage[id] = FeeLib.setFee(feeStorage, index, newTradingFee);
+        emit EventsLib.SetObligationTradingFee(id, index, newTradingFee);
+    }
+
+    function setObligationTradingFeeActivated(bytes32 id, bool activated) external {
+        uint256 feeStorage = _obligationTradingFeeStorage[id];
+        require(msg.sender == feeSetter, "Only feeSetter");
+        _obligationTradingFeeStorage[id] = FeeLib.setActivated(feeStorage, activated);
```
```suggestion
        require(msg.sender == feeSetter, "Only feeSetter");
        _obligationTradingFeeStorage[id] = FeeLib.setActivated(_obligationTradingFeeStorage[id], activated);
```

src/MorphoV2.sol
```diff
@@ -87,14 +87,36 @@ contract MorphoV2 is IMorphoV2 {
...
+        emit EventsLib.SetObligationTradingFeeActivated(id, activated);
+    }
+
+    function setDefaultTradingFee(address loanToken, uint256 index, uint256 newTradingFee) external {
+        uint256 feeStorage = _defaultTradingFeeStorage[loanToken];
+        require(msg.sender == feeSetter, "Only feeSetter");
+        require(newTradingFee <= WAD, "Trading fee too high");
+        require(index <= 5, "Invalid index");
+        _defaultTradingFeeStorage[loanToken] = FeeLib.setFee(feeStorage, index, newTradingFee);
```
```suggestion
        require(msg.sender == feeSetter, "Only feeSetter");
        require(newTradingFee <= WAD, "Trading fee too high");
        require(index <= 5, "Invalid index");
        _defaultTradingFeeStorage[loanToken] = FeeLib.setFee(_defaultTradingFeeStorage[loanToken], index, newTradingFee);
```

src/MorphoV2.sol
```diff
@@ -87,14 +87,36 @@ contract MorphoV2 is IMorphoV2 {
...
+    function setDefaultTradingFeeActivated(address loanToken, bool activated) external {
+        uint256 feeStorage = _defaultTradingFeeStorage[loanToken];
         require(msg.sender == feeSetter, "Only feeSetter");
-        require(tradingFee <= type(uint128).max, "Trading fee too high");
-        require(interestCutLimit < WAD, "Interest cut limit too high");
-        // forge-lint: disable-next-item(unsafe-typecast) Safe cast because values are below type(uint128).max.
-        tradingFeeParams[id] =
-            TradingFeeParams({tradingFee: uint128(tradingFee), interestCutLimit: uint128(interestCutLimit)});
-        emit EventsLib.SetTradingFee(id, tradingFee, interestCutLimit);
+        _defaultTradingFeeStorage[loanToken] = FeeLib.setActivated(feeStorage, activated);
```
```suggestion
require(msg.sender == feeSetter, "Only feeSetter");
        _defaultTradingFeeStorage[loanToken] = FeeLib.setActivated(_defaultTradingFeeStorage[loanToken], activated);
```

adhusson (approved):

src/libraries/EventsLib.sol
```diff
@@ -9,7 +9,10 @@ library EventsLib {
 
     event SetOwner(address indexed owner);
     event SetFeeSetter(address indexed feeSetter);
-    event SetTradingFee(bytes32 indexed obligationId, uint256 tradingFee, uint256 interestCutLimit);
+    event SetObligationTradingFee(bytes32 indexed obligationId, uint256 index, uint256 newTradingFee);
```
```suggestion
    event SetObligationTradingFee(bytes32 indexed obligationId, uint256 indexed index, uint256 newTradingFee);
```
not necessary, slightly more costly in gas

src/MorphoV2.sol
MathisGD: because it's a pack of a bunch of things. it's to make it clear that you can't use the value. btw I think that we should make it public

adhusson: why not `obligationTradingFeePacked`, I think it would be clearer and I've seen that elsewhere

src/libraries/FeeLib.sol
```diff
@@ -0,0 +1,46 @@
...
+
+    /// @dev Returns the updated packed fee storage value, preserving the activated flag.
+    function setFee(uint256 feeStorage, uint256 index, uint256 fee) internal pure returns (uint256) {
+        require(fee % FEE_STEP == 0, "fee should be a multiple of 1e12");
+        uint256 shift = index * FEE_BITS;
+        uint256 cleared = feeStorage & ~(FEE_MASK << shift);
+        uint256 packedFee = (fee / FEE_STEP) << shift;
+        return cleared | packedFee;
+    }
+
```
peyha, line 36:
```suggestion
    /// @dev Returns the updated packed fee storage value, preserving the trading fee values.
```

src/MorphoV2.sol
```diff
@@ -87,14 +87,36 @@ contract MorphoV2 is IMorphoV2 {
         emit EventsLib.SetFeeSetter(newFeeSetter);
     }
 
-    function setTradingFee(bytes32 id, uint256 tradingFee, uint256 interestCutLimit) external {
+    function setObligationTradingFee(bytes32 id, uint256 index, uint256 newTradingFee) external {
+        uint256 feeStorage = _obligationTradingFeeStorage[id];
+        require(msg.sender == feeSetter, "Only feeSetter");
+        require(newTradingFee <= WAD, "Trading fee too high");
+        require(index <= 5, "Invalid index");
```
```suggestion
        require(msg.sender == feeSetter, "Only feeSetter");
        require(newTradingFee <= WAD, "Trading fee too high");
        require(index <= 5, "Invalid index");
        uint256 feeStorage = _obligationTradingFeeStorage[id];
```
slightly more costly in case of revert

src/MorphoV2.sol
MathisGD: nb: we are going to change it anyway, with #252 and #238

peyha (approved):
LGTM

QGarchery (approved):

PR #291, opened 2026-01-14
issuance fee

adhusson:
extremely simple, only depends on volume not price, but 
- could be obligation specific 
- and depend on ttm


src/MorphoV2.sol
MathisGD, line 1: this implementation is great because it doesn't impact the price. but then the price is a bit less meaningful too

MathisGD: like the interest fee (cut on interest) we wouldn't be able to completely solve this anyway

PR #293, opened 2026-01-14
refactor: fee bands

MathisGD:


lib/forge-std
MathisGD, line 1: to be able to use 0.8.31 (to have clz)

peyha:
why not go for 0.8.33 while we're at it ?

lib/forge-std
adhusson: note clz is pretty fresh and we'd have to wait for a chain to adopt it before we can deploy

src/MorphoV2.sol
```diff
@@ -100,23 +100,23 @@ contract MorphoV2 is IMorphoV2 {
...
-    /// @dev Trading fees are truncated to the nearest 1e9.
     function setObligationTradingFee(bytes32 id, uint256 index, uint256 newTradingFee) external {
         require(msg.sender == feeSetter, "Only feeSetter");
         require(newTradingFee <= WAD, "Trading fee too high");
-        require(index < 6, "Invalid index");
+        require(index <= 8, "Invalid index");
+        require(newTradingFee % 1e12 == 0, "fee should be a multiple of 1e12");
         _obligationTradingFee[id] =
-            _obligationTradingFee[id] & ~(0xFFFFFFFF << (index * 32)) | (newTradingFee / 1e9 << (index * 32));
+            _obligationTradingFee[id] & ~(0xFFFFF << (index * 24)) | (newTradingFee / 1e12 << (index * 24));
```
adhusson: we can prove we don't need the extra F (`0xFFFFFF`) but it would be easier to read & make proofs about with a 24 bit mask

src/MorphoV2.sol
```diff
@@ -55,15 +55,15 @@ contract MorphoV2 is IMorphoV2 {
     /// GETTERS ///
 
     function tradingFeeIndex(uint256 ttm) public pure returns (uint256) {
-        return ttm == 0 ? 0 : ttm < 1 days ? 1 : ttm < 7 days ? 2 : ttm < 30 days ? 3 : ttm < 90 days ? 4 : 5;
+        return ttm == 0 ? 0 : UtilsLib.min(8, UtilsLib.log2(ttm / 1 days) + 1);
```
adhusson, line 58: smart but don't know if it's worth having the jump points at unconventional durations

src/MorphoV2.sol
MathisGD: actually it's path dependance, at some point I had 20 bit long words

src/MorphoV2.sol
MathisGD: anyway we are going to move to piecewise linear (or alternative)

PR #294, opened 2026-01-14
interest fee with cost basis tracking

adhusson:
[Explainer note on Notion
](https://www.notion.so/morpho-labs/Interest-fee-idea-2edd69939e6d8068ae52f4ba3e8be39d?source=copy_link)

> [!NOTE]
> The files changed view is not updating, here is an [up-to-date compare view](https://github.com/morpho-org/morpho-v2/compare/fix/optimizer-2..feat/interest-fee)


src/MorphoV2.sol
adhusson: This is now important, without it I can start by lending to myself at price >> 1. My costOf is very high so my future exits create no interest fee.

src/MorphoV2.sol
```diff
@@ -39,12 +46,21 @@ contract MorphoV2 is IMorphoV2 {
...
     mapping(address loanToken => uint256) private _defaultTradingFee;
 
-    address public tradingFeeRecipient;
+    /// @dev Default interest fees per loan token.
+    mapping(address loanToken => uint256) public _defaultInterestFee;
+
+    address public feeRecipient;
+
+    /// @dev Track total cost paid and not yet recovered.
+    mapping(address => mapping(bytes32 => uint256)) public costOf;
```
peyha, line 64: I would prefer the naming `cashflow`

src/MorphoV2.sol
peyha: couldn't we imagine another solution against this kind of behavior ? For instance the interest fee could apply to negative borrow rates
(though preventing price >> 1 would be the simplest thing to implement)

src/MorphoV2.sol
```diff
@@ -454,6 +504,17 @@ contract MorphoV2 is IMorphoV2 {
 
     /// VIEW ///
 
+    function calculateExit(bytes32 id, address loanToken, address lender, uint256 shares, uint256 assets)
+        internal
+        view
+        returns (uint256 recoveredCost, uint256 exitFee)
+    {
+        uint256 interestFee =
```
peyha: one potential issue with this is that setting a new interest fee is retroactive: all current obligation will have to pay it on the already earned interests which was not the behavior on morpho blue/vault v1/vault v2 (because the interest fee was taken during interest accrual). This can be mitigated by setting a maximum interest fee but it's not perfect

src/MorphoV2.sol
adhusson: Interest on negative rates would have to be taken at the initial take (not when the profit is made) right? otherwise you can walk away?

peyha: not yet sure, it could be taken into the accounting of the borrower debt to be considered during liquidation
but it's starting to be a bit complex to consider

PR #295, opened 2026-01-14
refactor: trading fee formula F(t) = (max-min)*t/(t+halfLife) + min

MathisGD:


MathisGD:
we went for piecewise linear, see #289

PR #297, opened 2026-01-16
feat: receiver

### LINKED ISSUE #262: recipient parameter

MathisGD:
fixes #262


src/MorphoV2.sol
```diff
@@ -134,12 +133,13 @@ contract MorphoV2 is IMorphoV2 {
         uint256 obligationUnits,
         uint256 obligationShares,
         address taker,
+        address takerCallback,
+        bytes memory takerCallbackData,
+        address sellerRecipient,
```
adhusson: We could have a takerRecipient. If the taker sells it receives assets, if it buys it receives units.

src/MorphoV2.sol
```diff
@@ -242,38 +235,47 @@ contract MorphoV2 is IMorphoV2 {
...
+                    .onSell(
+                        offer.obligation,
+                        seller,
+                        buyerAssets,
+                        sellerAssets,
+                        obligationUnits,
+                        obligationShares,
+                        sellerCallbackData
+                    );
+            }
```
adhusson: not fantastic we should fix this stack issue properly (maybe a big struct after all)

src/MorphoV2.sol
MathisGD: not fan. also it will clash with onbehalf

adhusson: it's the same split as here:

```solidity
function withdraw(
        Obligation memory obligation,
        uint256 obligationUnits,
        uint256 shares,
        address onBehalf,
        address recipient
```

onBehalf gives, recipient receives. ok you mean it would clash with the meaning of onBehalf in `repay` where `onBehalf` receives the units. 

Agree, but it's strange to have that parameter which only means something when taker=seller + not have onBehalf + no direct equivalent for the maker.

MathisGD: we can add a recipient for the maker too, actually

src/MorphoV2.sol
MathisGD: fixed

src/MorphoV2.sol
QGarchery: `taker` is alreaty the `onBehalf` no ? It's the address that gets its position change. We just need to correctly [authorize to take on behalf of someone](https://github.com/morpho-org/morpho-v2/pull/340#discussion_r2768118948).

MathisGD: solved

src/MorphoV2.sol
MathisGD, line 1: recipient or receiver

src/MorphoV2.sol
```diff
@@ -167,10 +168,27 @@ contract MorphoV2 is IMorphoV2 {
...
-            ? (offer.maker, offer.callback, offer.callbackData, taker, takerCallback, takerCallbackData)
-            : (taker, takerCallback, takerCallbackData, offer.maker, offer.callback, offer.callbackData);
+            ? (
+                offer.maker,
+                offer.callback,
+                offer.callbackData,
+                taker,
+                takerCallback,
+                takerCallbackData,
+                recipientIfSeller
```
```suggestion
                recipientIfTakerIsSeller
```
and then below `offer.recipientIfMakerIsSeller`. I think it's worth to break the symmetry of naming for clarity. 
Without breaking symmetry it would give:
```suggestion
                takerRecipientIfSeller
```
and keep the naming `offer.recipientIfSeller` below

src/MorphoV2.sol
QGarchery: I like receiver more, feels more widely used for this, so it's nice to immediately get the meaning of this variable

QGarchery (approved):

src/MorphoV2.sol
adhusson: +1 to `recipientIfTakerIsSeller` / `recipientIfMakerIsSeller`, removes all doubt.

lib/forge-std
peyha, line 1: why change the forge version ?

src/MorphoV2.sol
peyha: we've always used receiver except for fees where it was recipient (even in vault-v2) so it's great to stay consistent (but it's not a big deal)

src/libraries/EventsLib.sol
```diff
@@ -26,19 +26,30 @@ library EventsLib {
         uint256 obligationShares,
         bool buyerIsLender,
         bool sellerIsBorrower,
+        address recipient,
```
```suggestion
        address sellerRecipient,
```
slightly less misleading

src/libraries/EventsLib.sol
```diff
@@ -26,19 +26,30 @@ library EventsLib {
...
         uint256 consumed
     );
     event Withdraw(
-        address indexed caller, bytes32 indexed id, uint256 obligationUnits, uint256 shares, address indexed onBehalf
+        address indexed caller,
+        bytes32 indexed id,
+        uint256 obligationUnits,
+        uint256 shares,
+        address indexed onBehalf,
+        address recipient
```
peyha: note that we always indexed the recipient instead of the caller on morpho blue when we had to choose between the 2
I'm not sure indexing even one of them would be useful

src/MorphoV2.sol
MathisGD: yes that's what made me change for recipient

lib/forge-std
MathisGD: not meant

src/libraries/EventsLib.sol
MathisGD: alright, can change that

test/OtherFunctionsTest.sol
```diff
@@ -159,6 +161,34 @@ contract OtherFunctionsTest is BaseTest {
         assertEq(returnedShares, shares, "returned shares");
     }
 
+    function testWithdrawToRecipient(uint256 units, uint256 withdraw) public {
```
adhusson:
```suggestion
    function testWithdrawToReceiver(uint256 units, uint256 withdraw) public {
```

test/OtherFunctionsTest.sol
```diff
@@ -159,6 +161,34 @@ contract OtherFunctionsTest is BaseTest {
...
+        address receiver = makeAddr("receiver");
+
+        vm.prank(lender);
+        morphoV2.withdraw(obligation, withdraw, 0, lender, receiver);
+
+        assertEq(loanToken.balanceOf(lender), 0, "balance of lender");
+        assertEq(loanToken.balanceOf(receiver), withdraw, "balance of receiver");
+    }
+
+    function testWithdrawCollateralToRecipient(uint256 supply, uint256 withdraw) public {
```
adhusson:
```suggestion
    function testWithdrawCollateralToReceiver(uint256 supply, uint256 withdraw) public {
```

src/MorphoV2.sol
```diff
@@ -120,10 +120,10 @@ contract MorphoV2 is IMorphoV2 {
         emit EventsLib.SetDefaultTradingFee(loanToken, index, newTradingFee);
     }
 
-    function setTradingFeeRecipient(address recipient) external {
+    function setTradingFeeRecipient(address receiver) external {
         require(msg.sender == owner, "Only owner");
-        tradingFeeRecipient = recipient;
-        emit EventsLib.SetTradingFeeRecipient(recipient);
+        tradingFeeRecipient = receiver;
```
adhusson: Some mismatch between "fee recipient" concept and "receiver" concept. OK to 
- have both concepts or 
- completely erase `recipient`.

But at the moment it's inconsistent because 'receiver' is used in some functions & tests related to 'fee recipient'.

MathisGD:
@codex review

src/MorphoV2.sol
```diff
@@ -272,7 +291,7 @@ contract MorphoV2 is IMorphoV2 {
         SafeTransferLib.safeTransferFrom(
             offer.obligation.loanToken, buyer, tradingFeeRecipient, buyerAssets - sellerAssets
         );
-        SafeTransferLib.safeTransferFrom(offer.obligation.loanToken, buyer, seller, sellerAssets);
+        SafeTransferLib.safeTransferFrom(offer.obligation.loanToken, buyer, receiver, sellerAssets);
```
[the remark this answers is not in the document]
MathisGD: we generally don't do zero checks.

test/TakeTest.sol
```diff
@@ -1191,12 +1206,13 @@ contract TakeTest is BaseTest {
...
-            sig([lenderOffer, otherOffer]),
-            root([lenderOffer, otherOffer]),
-            proof,
             address(0),
-            hex""
+            hex"",
+            borrower,
+            lenderOffer,
+            sig([lenderOffer]),
+            root([lenderOffer]),
```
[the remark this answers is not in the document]
MathisGD: nice catch!

adhusson (approved):

MathisGD:
@codex review

src/MorphoV2.sol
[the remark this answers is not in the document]
MathisGD: we don't do zero checks

test/AuthorizationTest.sol
```diff
@@ -200,7 +202,9 @@ contract AuthorizationTest is BaseTest {
 
         // Operator can take on behalf of taker
         vm.prank(operator);
-        morphoV2.take(assets, 0, 0, 0, taker, offer, sig([offer]), root([offer]), proof([offer]), address(0), hex"");
+        morphoV2.take(
+            assets, 0, 0, 0, taker, address(0), hex"", address(0), offer, sig([offer]), root([offer]), proof([offer])
```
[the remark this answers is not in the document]
MathisGD: useful but won't fix

QGarchery (approved):

PR #298, opened 2026-01-16
max optimizer

MathisGD:
opus 4.5 found that. quite ugly.. but could be worse

### DIFF
--- a/src/libraries/UtilsLib.sol
+++ b/src/libraries/UtilsLib.sol
@@ -5,34 +5,34 @@ pragma solidity ^0.8.0;
 library UtilsLib {
     /// @dev Returns true if at most one of `x` and `y` is nonzero.
     function atMostOneNonZero(uint256 x, uint256 y) internal pure returns (bool z) {
-        assembly {
+        assembly ("memory-safe") {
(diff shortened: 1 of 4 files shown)


src/libraries/UtilsLib.sol
```diff
@@ -5,34 +5,34 @@ pragma solidity ^0.8.0;
 library UtilsLib {
     /// @dev Returns true if at most one of `x` and `y` is nonzero.
     function atMostOneNonZero(uint256 x, uint256 y) internal pure returns (bool z) {
-        assembly {
+        assembly ("memory-safe") {
```
adhusson, line 8: was this necessary ? solidity docs [say](https://docs.soliditylang.org/en/latest/assembly.html#:~:text=Inline%20assembly%20that%20neither%20involves%20any%20operations%20that%20access%20memory%20nor%20assigns%20to%20any%20Solidity%20variables%20in%20memory%20is%20automatically%20considered%20memory%2Dsafe%20and%20does%20not%20need%20to%20be%20annotated.)

> Inline assembly that neither involves any operations that access memory nor assigns to any Solidity variables in memory is automatically considered memory-safe and does not need to be annotated.

PR #303, opened 2026-01-20
recovery close factor + min collat

MathisGD:
implements recovery close factor (=liquidation cannot liquidated more than what's needed to put back the position into health), and min collat in supplyCollateral / withdrawCollataral. The final behavior of min collat will be changed in an other PR later


adhusson (commented):
what is the relationship with https://github.com/morpho-org/morpho-v2/pull/285 ?

MathisGD:
alternative implém of recovery close factor

src/MorphoV2.sol
```diff
@@ -379,52 +380,50 @@ contract MorphoV2 is IMorphoV2 {
...
+                seizedAssets =
+                    repaidUnits.mulDivDown(ORACLE_PRICE_SCALE, liquidatedCollateralPrice).mulDivDown(lif, WAD);
             }
 
-            totalRepaid += seizure.repaid;
-            address collateralToken = obligation.collaterals[seizure.collateralIndex].token;
-            collateralOf[borrower][id][collateralToken] -= seizure.seized;
+            uint256 newMaxDebt = maxDebt
+                - seizedAssets.mulDivDown(liquidatedCollateralPrice, ORACLE_PRICE_SCALE)
+                    .mulDivDown(obligation.collaterals[collateralIndex].lltv, WAD);
```
QGarchery: More an intuition than something concrete, but I don't like the fact that we have multiple rouding for the same collateral. Another option is to cache the max debt without the contribution of the liquidated collateral

src/MorphoV2.sol
```diff
@@ -379,52 +380,50 @@ contract MorphoV2 is IMorphoV2 {
...
+            require(
+                block.timestamp > obligation.maturity || originalDebt - repaidUnits >= newMaxDebt,
+                "recovery close factory violated"
+            );
```
QGarchery: Nice that because of the recovery close factor we can omit the loop on multiple seizure. Apart from that, why do you like this PR more than #285 ? It seems like it still has the issue that you cannot completely close a position as well
(minor but `block.timestamp > obligation.maturity` is checked twice, I wonder if it can be simplified)

MathisGD: it's more efficient in terms of gas, because it doesn't recompute the full health.

src/MorphoV2.sol
```diff
@@ -348,29 +348,30 @@ contract MorphoV2 is IMorphoV2 {
...
-    function liquidate(Obligation memory obligation, Seizure[] memory seizures, address borrower, bytes calldata data)
-        external
-        returns (Seizure[] memory)
-    {
-        uint256 repayableDebt;
-        uint256 maxDebt;
+    /// @dev Returns repaid units and seized assets.
+    function liquidate(
+        Obligation calldata obligation,
+        uint256 collateralIndex,
```
peyha, line 408: this makes liquidating several collateral more complicated for the liquidator. 
The liquidator can still liquidate collaterals one by one but he won't seize all collaterals at the same time (which is more efficient theoretically because he has more routing possibilities)
To seize all collaterals at the same time, he can nest his liquidation calls in callbacks. It's a bit more technical to do but still manageable, especially for liquidators

src/MorphoV2.sol
peyha: it's more efficient in gas for the liquidation of a single collateral (because oracles are called only once) but oracle calls are duplicated when trying to liquidate several collaterals
it does not seem like an issue though (but it could be great to quantify it)

src/MorphoV2.sol
peyha: closing this after [this](https://morpholabs.slack.com/archives/C08A40KKN9W/p1769619441869549?thread_ts=1769619438.899839&cid=C08A40KKN9W)

src/MorphoV2.sol
MathisGD: or remove the old contribution and add the new contribution here

peyha: The previous computation (using `isHealthy`) was essentially doing `(collateralAsset-seizedAssets)*oraclePrice*lltv` while the new one is doing `collateralAssets*oraclePrice*lltv - seizedAssets*oraclePrice*lltv` where all the muls are mulDivDown. Claude says that these two expressions can't be compared in the general case (there exist variables such that one side is strictly greater than the other for both sides). I think it's safer to use `(collateralAsset-seizedAssets)*oraclePrice*lltv` because that's consistent with the next max debt value

MathisGD: fixed

src/MorphoV2.sol
```diff
@@ -378,52 +371,54 @@ contract MorphoV2 is IMorphoV2 {
...
 
-            totalRepaid += seizure.repaid;
-            address collateralToken = obligation.collaterals[seizure.collateralIndex].token;
-            collateralOf[id][borrower][collateralToken] -= seizure.seized;
+            address collateralToken = obligation.collaterals[collateralIndex].token;
+            uint256 newMaxDebt = maxDebt
+                - collateralOf[id][borrower][collateralToken].mulDivDown(liquidatedCollateralPrice, ORACLE_PRICE_SCALE)
+                    .mulDivDown(lltv, WAD)
+                + (collateralOf[id][borrower][collateralToken] - seizedAssets)
+                    .mulDivDown(liquidatedCollateralPrice, ORACLE_PRICE_SCALE).mulDivDown(lltv, WAD);
```
MathisGD: a bit ugly..

src/MorphoV2.sol
```diff
@@ -378,52 +371,52 @@ contract MorphoV2 is IMorphoV2 {
...
         }
 
-        uint256 totalRepaid;
-
-        for (uint256 i = 0; i < seizures.length; i++) {
-            Seizure memory seizure = seizures[i];
-            require(UtilsLib.atMostOneNonZero(seizure.repaid, seizure.seized), "INCONSISTENT_INPUT");
+        if (repaidUnits > 0 || seizedAssets > 0) {
+            uint256 lltv = obligation.collaterals[collateralIndex].lltv;
+            uint256 lif = originalDebt > maxDebt
```
peyha, line 441:
```suggestion
            uint256 lif = block.timestamp < obligation.maturity
```

peyha: weird that tests do not catch this

MathisGD: no I think that the current code is correct. you want the max lif if the position is not healthy

src/MorphoV2.sol
```diff
@@ -378,52 +371,52 @@ contract MorphoV2 is IMorphoV2 {
...
             _obligationState.totalUnits -= UtilsLib.toUint128(badDebt);
         }
 
-        uint256 totalRepaid;
-
-        for (uint256 i = 0; i < seizures.length; i++) {
-            Seizure memory seizure = seizures[i];
-            require(UtilsLib.atMostOneNonZero(seizure.repaid, seizure.seized), "INCONSISTENT_INPUT");
+        if (repaidUnits > 0 || seizedAssets > 0) {
+            uint256 lltv = obligation.collaterals[collateralIndex].lltv;
```
QGarchery: The `lltv` is only used when computing the `newMaxDebt`, so it could be defined in the `if (block.timestamp <= obligation.maturity)`. Also it may be worth defining `liquidatedCollateralToken` if it makes things more readable

src/MorphoV2.sol
QGarchery, line 1: Wondering about the gas issue, let's discuss [here](https://morpholabs.slack.com/archives/C08A40KKN9W/p1770134797740239?thread_ts=1769619438.899839&cid=C08A40KKN9W)

src/MorphoV2.sol
```diff
@@ -378,52 +371,52 @@ contract MorphoV2 is IMorphoV2 {
...
-            totalRepaid += seizure.repaid;
-            address collateralToken = obligation.collaterals[seizure.collateralIndex].token;
-            collateralOf[id][borrower][collateralToken] -= seizure.seized;
+            if (block.timestamp <= obligation.maturity) {
+                uint256 _collateralOf = collateralOf[id][borrower][obligation.collaterals[collateralIndex].token];
+                uint256 newMaxDebt = maxDebt
+                    - _collateralOf.mulDivDown(liquidatedCollateralPrice, ORACLE_PRICE_SCALE).mulDivDown(lltv, WAD)
+                    + (_collateralOf - seizedAssets).mulDivDown(liquidatedCollateralPrice, ORACLE_PRICE_SCALE)
+                        .mulDivDown(lltv, WAD);
+                require(originalDebt - repaidUnits >= newMaxDebt, "recovery close factory violated");
```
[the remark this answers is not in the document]
QGarchery: Badly worded by codex, but it's indeed very strange that a full liquidation is only possible in liquidation call where a bad debt is realized. So if you have multiple collaterals, this means that you cannot be fully liquidated. But with one collateral you can

src/MorphoV2.sol
QGarchery, line 1: Are we ok with the issue about not being able to fully liquidate a user ? It seems like it could lead to situations where there are many very small positions, that are not profitable to liquidate because of the gas. Also the [global min collateral](https://github.com/morpho-org/morpho-v2/pull/320) as it currently stands doesn't address this issue, because it prevents new positions but doesn't change the liquidation

src/MorphoV2.sol
```diff
@@ -346,30 +338,31 @@ contract MorphoV2 is IMorphoV2 {
...
-        uint256[] memory prices = new uint256[](obligation.collaterals.length);
 
+        uint256 repayableDebt;
+        uint256 maxDebt;
+        uint256 liquidatedCollateralPrice;
         for (uint256 i = 0; i < obligation.collaterals.length; i++) {
             Collateral memory _collateral = obligation.collaterals[i];
             uint256 price = IOracle(_collateral.oracle).price();
-            prices[i] = price;
+            if (i == collateralIndex) liquidatedCollateralPrice = price;
```
QGarchery: Minor but the liquidation will revert at unpredictable times if `collateralIndex` is out of range. If we change the implementation a bit, it may be possible to use a price of 0 because this line is not executed

src/interfaces/ICallbacks.sol
```diff
@@ -23,7 +23,14 @@ interface ICallbacks {
         uint256 obligationShares,
         bytes memory data
     ) external;
-    function onLiquidate(Seizure[] memory seizures, address borrower, address liquidator, bytes memory data) external;
```
QGarchery, line 26: We removed the `liquidator` here, was there a reason we had it before ?

MathisGD: I guess it's matching more with onBuy and onSell now?

src/MorphoV2.sol
MathisGD: cheap to add a require that collateralIndex is in range, will do

src/MorphoV2.sol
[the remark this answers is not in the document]
MathisGD: >Badly worded by codex, but it's indeed very strange that a full liquidation is only possible in liquidation call where a bad debt is realized.

actually no it's expected (at least high level) with RCF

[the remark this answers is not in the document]
MathisGD: >So if you have multiple collaterals, this means that you cannot be fully liquidated. But with one collateral you can

didn't get this

src/MorphoV2.sol
MathisGD: >Are we ok with the issue about not being able to fully liquidate a user ?

just to be sure, you are not speaking about an issue with this precise implementation of RCF, but rather challenging the feature as a whole

MathisGD, line 1: @codex how much more costly in gas will it be to liquidate multiple collaterals after this pr

QGarchery: yes, but at the same time we could potentially find workarounds in the implementation. For example if the global min collateral applies to liquidations as well (can liquidate fully if under min collateral)

src/MorphoV2.sol
```diff
@@ -379,52 +374,52 @@ contract MorphoV2 is IMorphoV2 {
...
-            address collateralToken = obligation.collaterals[seizure.collateralIndex].token;
-            collateralOf[id][borrower][collateralToken] -= seizure.seized;
+            if (block.timestamp <= obligation.maturity) {
+                uint256 lltv = obligation.collaterals[collateralIndex].lltv;
+                uint256 _collateralOf = collateralOf[id][borrower][liquidatedCollateralToken];
+                uint256 newMaxDebt = maxDebt
+                    - _collateralOf.mulDivDown(liquidatedCollateralPrice, ORACLE_PRICE_SCALE).mulDivDown(lltv, WAD)
+                    + (_collateralOf - seizedAssets).mulDivDown(liquidatedCollateralPrice, ORACLE_PRICE_SCALE)
+                        .mulDivDown(lltv, WAD);
+                require(originalDebt - repaidUnits >= newMaxDebt, "recovery close factor violated");
```
MathisGD: this computation is not very intuitive, if you take seized = collateralOf, you get newMaxDebt = 0, it seems that you can seize everything. but actually no because it would make your repay too much (would underflow later)

MathisGD: in general I find that something around #334 would be better

src/interfaces/ICallbacks.sol
QGarchery: yes, not sure why we had it in `onLiquidate` actually

src/MorphoV2.sol
MathisGD: discussed, this will be combined later with #342

MathisGD: pushed a test to get numbers

Gas 1st seizure (cold): 160732
Gas 2nd seizure (warm): 48072

src/MorphoV2.sol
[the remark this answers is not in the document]
MathisGD: if there has been bad debt, the current level of debt should be such that the liquidation doesn't improve the position's health. with roundings maybe it does a little bit, but almost everything should be liquidatable. once we implement this #343, it should be really all the time fully closed.

yet, we might want something cleaner..

src/MorphoV2.sol
peyha: it would underflow in `originalDebt - repaidUnits` so on the same line which I agree is not very intuitive but I'm not sure is breaking the approach

src/MorphoV2.sol
MathisGD, line 1: TODO: takeaways of [this call](https://morpholabs.slack.com/archives/C02N7CZ088N/p1770390239608109)

src/MorphoV2.sol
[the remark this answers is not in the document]
MathisGD: we decided at the end to not pursue this, it's not important. plus there will be the min collat

MathisGD:
@codex review

src/MorphoV2.sol
```diff
@@ -416,52 +430,52 @@ contract MorphoV2 is IMorphoV2 {
...
             debtOf[id][borrower] -= badDebt;
             _obligationState.totalUnits -= UtilsLib.toUint128(badDebt);
         }
 
-        uint256 totalRepaid;
-
-        for (uint256 i = 0; i < seizures.length; i++) {
-            Seizure memory seizure = seizures[i];
-            require(UtilsLib.atMostOneNonZero(seizure.repaid, seizure.seized), "INCONSISTENT_INPUT");
+        if (repaidUnits > 0 || seizedAssets > 0) {
```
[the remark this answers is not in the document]
MathisGD: >(e.g. lltv > 1/MAX_LIF)

this is enforced to be impossible

src/MorphoV2.sol
```diff
@@ -416,52 +430,52 @@ contract MorphoV2 is IMorphoV2 {
...
+            }
+
+            if (block.timestamp <= obligation.maturity) {
+                uint256 lltv = obligation.collaterals[collateralIndex].lltv;
+                uint256 _collateralOf = collateralOf[id][borrower][liquidatedCollateralToken];
+                uint256 newMaxDebt = maxDebt
+                    - _collateralOf.mulDivDown(liquidatedCollateralPrice, ORACLE_PRICE_SCALE).mulDivDown(lltv, WAD)
+                    + (_collateralOf - seizedAssets).mulDivDown(liquidatedCollateralPrice, ORACLE_PRICE_SCALE)
+                        .mulDivDown(lltv, WAD);
+                require(originalDebt - badDebt - repaidUnits >= newMaxDebt, "recovery close factor violated");
```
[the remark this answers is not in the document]
MathisGD: good point, but not sure if we want to fix, because it would require handling the case debt=0 separately (while in practice it never happens due to roundings I guess)

src/interfaces/IMorphoV2.sol
```diff
@@ -7,6 +7,8 @@ struct Obligation {
     // Must be sorted by address.
     Collateral[] collaterals;
     uint256 maturity;
+    // Minimum collateral value (quoted in loan token) to be left on supply & withdraw collateral.
+    uint256 minCollateral;
```
MathisGD:
```suggestion
    uint256 minCollateralValue;
```

src/MorphoV2.sol
MathisGD: precision doesn't matter here. borrowers are ok with a "rough" RCF. though, with approximation we could really simplify the work of liquidators. should we?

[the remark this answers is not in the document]
adhusson: it's not a big change to do

```solidity
require(newMaxDebt == 0 || originalDebt - badDebt - repaidUnits > newMaxDebt, "recovery close factor violated");
```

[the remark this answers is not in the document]
MathisGD: the new computation solves that

MathisGD:
@codex review

src/MorphoV2.sol
```diff
@@ -416,52 +430,50 @@ contract MorphoV2 is IMorphoV2 {
...
-                    seizure.repaid.mulDivDown(ORACLE_PRICE_SCALE, prices[seizure.collateralIndex]).mulDivDown(lif, WAD);
+                seizedAssets = repaidUnits.mulDivDown(ORACLE_PRICE_SCALE, liquidatedCollatPrice).mulDivDown(lif, WAD);
+            }
+
+            if (block.timestamp <= obligation.maturity) {
+                uint256 lltv = obligation.collaterals[collateralIndex].lltv;
+                uint256 _collateralOf = collateralOf[id][borrower][liquidatedCollatToken];
+                uint256 maxSeized = (debtOf[id][borrower].mulDivUp(ORACLE_PRICE_SCALE, liquidatedCollatPrice)
+                        - _collateralOf.mulDivDown(lltv, WAD))
+                .mulDivUp(WAD, WAD.mulDivDown(WAD, lif) - lltv);
```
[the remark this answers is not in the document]
MathisGD: later, we will enforce LLTV * LIF < 1

MathisGD:
@codex review

src/MorphoV2.sol
```diff
@@ -353,61 +345,83 @@ contract MorphoV2 is IMorphoV2 {
...
+        collateralOf[id][onBehalf][collateralToken] += assets;
 
-        collateralOf[id][onBehalf][collateral] += assets;
+        address oracle = obligation.collaterals[collateralIndex].oracle;
+        uint256 _collateralOf = collateralOf[id][onBehalf][collateralToken];
+        require(
+            _collateralOf == 0
+                || _collateralOf.mulDivDown(IOracle(oracle).price(), ORACLE_PRICE_SCALE) >= obligation.minCollatValue,
+            "Below min collateral"
+        );
```
MathisGD, line 362: should this be fully skipped if minCollatValue is zero? depending if we see some people doing such obligations or not
```suggestion
        uint256 _collateralOf = collateralOf[id][onBehalf][collateralToken];
        require(
            obligation.minCollatValue == 0 ||
            _collateralOf == 0 || _collateralOf.mulDivDown(IOracle(oracle).price(), ORACLE_PRICE_SCALE) >= obligation.minCollatValue,
            "Below min collateral"
        );
```

MathisGD: mostly against, as the gain should be minimal

MathisGD: there is a consideration for the revert behavior when the oracle reverts, but this is really negligible

src/MorphoV2.sol
```diff
@@ -416,52 +430,48 @@ contract MorphoV2 is IMorphoV2 {
...
             } else {
-                seizure.seized =
-                    seizure.repaid.mulDivDown(ORACLE_PRICE_SCALE, prices[seizure.collateralIndex]).mulDivDown(lif, WAD);
+                seizedAssets = repaidUnits.mulDivDown(ORACLE_PRICE_SCALE, liquidatedCollatPrice).mulDivDown(lif, WAD);
+            }
+
+            if (block.timestamp <= obligation.maturity) {
+                uint256 lltv = obligation.collaterals[collateralIndex].lltv;
+                uint256 maxSeized = (debtOf[id][borrower] - maxDebt)
+                .mulDivUp(WAD, (WAD.mulDivDown(WAD, lif) - lltv).mulDivDown(liquidatedCollatPrice, ORACLE_PRICE_SCALE));
```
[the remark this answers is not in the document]
MathisGD: added a temporary check, because the AI focuses all the time on this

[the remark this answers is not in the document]
MathisGD: alright, that's wanted behavior but added a comment

src/MorphoV2.sol
```diff
@@ -416,52 +430,50 @@ contract MorphoV2 is IMorphoV2 {
...
-                    seizure.repaid.mulDivDown(ORACLE_PRICE_SCALE, prices[seizure.collateralIndex]).mulDivDown(lif, WAD);
+                seizedAssets = repaidUnits.mulDivDown(ORACLE_PRICE_SCALE, liquidatedCollatPrice).mulDivDown(lif, WAD);
+            }
+
+            if (block.timestamp <= obligation.maturity) {
+                uint256 lltv = obligation.collaterals[collateralIndex].lltv;
+                // Rounded up to avoid consecutive max liquidations.
+                // Acknowledged that the position could be slightly healthy after a liquidation.
+                uint256 maxSeized = (debtOf[id][borrower] - maxDebt)
+                .mulDivUp(WAD, (WAD.mulDivDown(WAD, lif) - lltv).mulDivDown(liquidatedCollatPrice, ORACLE_PRICE_SCALE));
```
MathisGD: worth noting that `maxDebt` is calculated rounding down, probably fine

src/MorphoV2.sol
```diff
@@ -536,6 +549,7 @@ contract MorphoV2 is IMorphoV2 {
     }
 
     /// @dev This function should be called with the id corresponding to the obligation.
+    /// @dev This function does not call the oracle if debt is 0.
```
```suggestion
    /// @dev This function does not call any oracle if debt is 0.
```

src/MorphoV2.sol
```suggestion
                uint256 maxSeized = (debtOf[id][borrower] - maxDebt)
                .mulDivUp(WAD, (WAD.mulDivDown(WAD, lif) - lltv)).mulDivDown(liquidatedCollatPrice, ORACLE_PRICE_SCALE);
```
⚠️

src/MorphoV2.sol
```diff
@@ -416,52 +430,50 @@ contract MorphoV2 is IMorphoV2 {
...
-                seizure.seized =
-                    seizure.repaid.mulDivDown(ORACLE_PRICE_SCALE, prices[seizure.collateralIndex]).mulDivDown(lif, WAD);
+                seizedAssets = repaidUnits.mulDivDown(ORACLE_PRICE_SCALE, liquidatedCollatPrice).mulDivDown(lif, WAD);
+            }
+
+            if (block.timestamp <= obligation.maturity) {
+                uint256 lltv = obligation.collaterals[collateralIndex].lltv;
+                // Rounded up to avoid consecutive max liquidations.
+                // Acknowledged that the position could be slightly healthy after a liquidation.
+                uint256 maxSeized = (debtOf[id][borrower] - maxDebt)
```
QGarchery: Why not doing `maxRepaid` instead ? This would avoid one price conversion

QGarchery (changes requested):

src/MorphoV2.sol
QGarchery:
```suggestion
                uint256 maxSeized = (debtOf[id][borrower] - maxDebt)
                .mulDivUp(WAD, (WAD.mulDivDown(WAD, lif) - lltv).mulDivDown(ORACLE_PRICE_SCALE, liquidatedCollatPrice));
```
⚠️

adhusson: shouldnt a test have caught this

src/MorphoV2.sol
adhusson: agree on negligible re oracle revert, users of the obligation are supposed to be exposed to their Collateral oracles

src/MorphoV2.sol
MathisGD: If I apply both changes you requested it's an equivalent mathematical formula (but not equal because of roundings), right? Why would this version make more sense?

MathisGD: Note that the current version is closer to https://morpholabs.slack.com/archives/C02N7CZ088N/p1770996420336909?thread_ts=1770996284.097119&channel=C02N7CZ088N&message_ts=1770996420.336909

src/MorphoV2.sol
MathisGD: I can try it! But isn't the linearity great?

src/MorphoV2.sol
adhusson: I missed this.

And why not just this which is ok with maxSeized on 136 bits or less:

```solidity
uint256 maxSeized = (debtOf[id][borrower] - maxDebt)
    .mulDivUp(WAD * ORACLE_PRICE_SCALE, 
              liquidatedCollatPrice * (WAD.mulDivDown(WAD, lif) - lltv)
             );
```

MathisGD: Works too

src/MorphoV2.sol
MathisGD: tried it [here](https://github.com/morpho-org/morpho-v2/pull/365/changes)

src/MorphoV2.sol
```diff
@@ -398,6 +398,7 @@ contract MorphoV2 is IMorphoV2 {
     /// @dev At least one of `repaidUnits` or `seizedAssets` should be equal to zero.
     /// @dev Accounts are liquidatable if they are unhealthy or if the maturity is reached.
     /// @dev Before maturity, the liquidation cannot put the borrower back into health (recovery close factor).
+    /// @dev We want b - s*P/LIF >= maxDebt - s*P*LLTV <=> s <= (b-maxDebt) / (P(1/LIF-LLTV)).
```
adhusson: `b -> debt`
`s -> seized`
`P -> price`

also the comment looks like it's saying "we want A iff B" but actually A <=> B always

src/MorphoV2.sol
```diff
@@ -416,52 +431,49 @@ contract MorphoV2 is IMorphoV2 {
...
+                // Rounded up to avoid consecutive max liquidations.
+                // Acknowledged that the position could be slightly healthy after a liquidation.
+                uint256 maxRepaid = (debtOf[id][borrower] - maxDebt).mulDivUp(WAD, WAD - lif.mulDivUp(lltv, WAD));
+                require(repaidUnits <= maxRepaid, "recovery close factor violated");
             }
 
-            totalRepaid += seizure.repaid;
-            address collateralToken = obligation.collaterals[seizure.collateralIndex].token;
-            collateralOf[id][borrower][collateralToken] -= seizure.seized;
+            collateralOf[id][borrower][liquidatedCollatToken] -= seizedAssets;
```
MathisGD, line 461: nb: now that the bad debt is realized at the beginning + RCF is through maxRepaid, the limiting factor should always be the debt (we could write formal verification for that btw). is there a case to remove the seized input? or the ability to seize an exact amount of collateral is important too?

MathisGD:
@codex review

test/LiquidationTest.sol
```diff
@@ -398,14 +292,192 @@ contract LiquidationTest is BaseTest {
...
+
+        // Drop price of liquidated collateral. 0.9e36 is above critical price for lltv=0.75 (0.8625e36).
+        uint256 droppedPrice = 0.9e36;
+        Oracle(obligation.collaterals[liqIdx].oracle).setPrice(droppedPrice);
+
+        uint256 liqCollat = morphoV2.collateralOf(id, borrower, obligation.collaterals[liqIdx].token);
+        uint256 otherCollat = morphoV2.collateralOf(id, borrower, obligation.collaterals[otherIdx].token);
+        uint256 _maxDebt = liqCollat.mulDivDown(droppedPrice, ORACLE_PRICE_SCALE)
+            .mulDivDown(obligation.collaterals[liqIdx].lltv, WAD)
+        + otherCollat.mulDivDown(obligation.collaterals[otherIdx].lltv, WAD);
```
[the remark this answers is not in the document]
MathisGD: wont fix

src/MorphoV2.sol
adhusson: this is if we know a version of https://github.com/morpho-org/morpho-v2/pull/354 will not be used? otherwise you can try to seize all collat and make the position healthy

adhusson: also the gain in code is not super high, it would remove

```
require(UtilsLib.atMostOneNonZero(repaidUnits, seizedAssets), "INCONSISTENT_INPUT");
```

```
repaidUnits = seizedAssets.mulDivUp(WAD, lif).mulDivUp(liquidatedCollatPrice, ORACLE_PRICE_SCALE);
```

adhusson (approved):

src/MorphoV2.sol
```diff
@@ -353,61 +345,84 @@ contract MorphoV2 is IMorphoV2 {
         SafeTransferLib.safeTransferFrom(obligation.loanToken, msg.sender, address(this), obligationUnits);
     }
```
peyha, line 347:
```suggestion
/// @dev This function does not call any oracle if no-op.
```
not really useful

src/MorphoV2.sol
```diff
@@ -353,61 +345,84 @@ contract MorphoV2 is IMorphoV2 {
...
+        );
 
-        emit EventsLib.SupplyCollateral(msg.sender, id, collateral, assets, onBehalf);
+        emit EventsLib.SupplyCollateral(msg.sender, id, collateralToken, assets, onBehalf);
 
-        SafeTransferLib.safeTransferFrom(collateral, msg.sender, address(this), assets);
+        SafeTransferLib.safeTransferFrom(collateralToken, msg.sender, address(this), assets);
     }
 
+    /// @dev This function does not call any oracle if all the collateral is withdrawn and the borrower has no debt.
```
qdrn, line 369:
```suggestion
    /// @dev This function does not call any oracle if all the collateral is withdrawn and the borrower is healthy.
```

MathisGD: it needs to call oracles to know if the borrower is healthy

src/MorphoV2.sol
peyha: Having the seized input is easy to add and simplify liquidations a lot. Suppose we're at maturity and a position with numerous collaterals has to be completely liquidated, then it's very easy to liquidate each collateral consecutively in its full amount compared to computing the equivalent repaidDebt for each collateral (especially if the oracle price change)

src/MorphoV2.sol
```diff
@@ -353,61 +345,84 @@ contract MorphoV2 is IMorphoV2 {
...
-        returns (Seizure[] memory)
-    {
-        uint256 repayableDebt;
-        uint256 maxDebt;
+    /// @dev Returns repaid units and seized assets.
+    function liquidate(
+        Obligation calldata obligation,
+        uint256 collateralIndex,
+        uint256 repaidUnits,
+        uint256 seizedAssets,
```
peyha: We used the opposite argument order in morpho blue (first seized assets and then repaid shares cf [this](https://github.com/morpho-org/morpho-blue/blob/0427acea2f2c9066195a6824fb75967f9a9e7645/src/Morpho.sol#L350)) so it would make sense to be consistent between protocols.

src/MorphoV2.sol
MathisGD: yes, thought the same

src/MorphoV2.sol
MathisGD: >Suppose we're at maturity and a position with numerous collaterals has to be completely liquidated

because of bad debt realization is rounded up, I'm not sure that you can in the general case seize all the collateral.

QGarchery (approved):

PR #304, opened 2026-01-21
chore: enable optimizer

### LINKED ISSUE #280: Increase optimizer runs
[This PR ](https://github.com/morpho-org/morpho-v2/pull/279)is a quick fix to enable the solc optimizer (max runs 216). 

We should find a better fix and try to increase the runs if it saves meaningful gas.

MathisGD:
fixes #280

### DIFF
--- a/foundry.toml
+++ b/foundry.toml
@@ -1,5 +1,7 @@
 [profile.default]
 via_ir = true
+optimizer = true
+optimizer_runs = 4294967295
 evm_version = "osaka"
 
 [profile.default.fmt]
--- a/src/libraries/FeeLib.sol
+++ b/src/libraries/FeeLib.sol
@@ -40,7 +40,7 @@ library FeeLib {
     }
 
     function boolToUint256(bool b) internal pure returns (uint256 res) {
-        assembly {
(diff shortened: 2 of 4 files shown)


adhusson (approved):
~This is crazy did you see [this comment](https://github.com/morpho-org/morpho-v2/pull/298#discussion_r2698688235)? Solidity docs say all assembly that does not touch memory is automatically memory-safe.~

Actually only adding `memory-safe` to the safe transfer lib is enough, I missed that they were changed too, and they do touch memory (unlike in https://github.com/morpho-org/morpho-v2/pull/298 where they are not changed).

MathisGD:
Missed it. Indeed, that's weird. Maybe worth opening an issue

QGarchery (approved):
LGTM:
- [relevant thread](https://morpholabs.slack.com/archives/C08A40KKN9W/p1768573065770109) to understand why the 999999 limit is not needed anymore
- [relevant thread](https://morpholabs.slack.com/archives/C08A40KKN9W/p1769073585262819) for memory safety (we don't need to set it everywhere)

PR #306, opened 2026-01-22
ticks inverse

adhusson:
[old page](https://www.notion.so/morpho-labs/Properties-of-priceToTick-2f0d69939e6d80d793fcc26e4e4b5395?source=copy_link)

I think we can get a perfect precision in the inverse but will try later.


adhusson:
closing in favor of https://github.com/morpho-org/morpho-v2/pull/310

PR #309, opened 2026-01-23
feat: fee cannot increase

MathisGD:


src/MorphoV2.sol
```diff
@@ -452,11 +445,13 @@ contract MorphoV2 is IMorphoV2 {
         SafeTransferLib.safeTransferFrom(token, msg.sender, address(this), assets);
     }
 
-    function touchObligation(Obligation memory obligation) internal {
+    function touchObligation(Obligation memory obligation) public {
         bytes32 id = toId(obligation);
         if (!obligationCreated[id]) {
-            emit EventsLib.ObligationCreated(id, obligation);
+            _obligationTradingFeeStorage[id] = _defaultTradingFeeStorage[obligation.loanToken];
             obligationCreated[id] = true;
```
MathisGD, line 447: to pack

src/MorphoV2.sol
```diff
@@ -44,12 +44,12 @@ contract MorphoV2 is IMorphoV2 {
     mapping(address user => bytes32) public session;
 
     /// @dev Obligation trading fees for a given obligation id.
-    /// @dev Bit 0: activated flag. Bits 1-144: 6 trading fees packed (24 bits each).
+    /// @dev Bits 1-144: 6 trading fees packed (24 bits each).
```
QGarchery, line 47:
```suggestion
    /// @dev Bits 0-119: 6 trading fees packed (20 bits each).
```
there was an inconsistency before this PR

src/MorphoV2.sol
```diff
@@ -44,12 +44,12 @@ contract MorphoV2 is IMorphoV2 {
...
     /// @dev Fee indices: 0=0d, 1=1d, 2=7d, 3=30d, 4=90d, 5=180d.
     mapping(bytes32 obligationId => uint256) internal _obligationTradingFeeStorage;
 
     /// @dev Default trading fees per loan token. Used when obligation fee is not activated.
-    /// @dev Bit 0: activated flag. Bits 1-144: 6 trading fees packed (24 bits each).
+    /// @dev Bits 1-144: 6 trading fees packed (24 bits each).
```
QGarchery, line 52:
```suggestion
    /// @dev Bits 0-119: 6 trading fees packed (20 bits each).
```

src/MorphoV2.sol
```diff
@@ -95,20 +95,19 @@ contract MorphoV2 is IMorphoV2 {
...
 
+    /// @dev Overrides the fee of a specific obligation.
     function setObligationTradingFee(bytes32 id, uint256 index, uint256 newTradingFee) external {
         require(msg.sender == feeSetter, "Only feeSetter");
-        require(newTradingFee <= WAD, "Trading fee too high");
         require(index <= 5, "Invalid index");
+        require(
+            newTradingFee <= FeeLib.getFee(_obligationTradingFeeStorage[id], index),
+            "New trading fee is higher than current"
+        );
```
QGarchery, line 105: We still have the invariant that any fee is less than `WAD` but it's slightly less obvious now

src/MorphoV2.sol
QGarchery, line 1: Left to do (but this can be done in another PR):
- packing the `obligationCreated` boolean and the fees
- making it possible to have a default trading fee of zero, if this is something that we might want

QGarchery (approved):

src/MorphoV2.sol
MathisGD: >making it possible to have a default trading fee of zero, if this is something that we might want

this is possible now, no?

QGarchery, line 1: There is the issue that users could create obligations in advance, to bypass a fee switch or an increase in the default fee

QGarchery: I mean a default of zero, but that allows to have some obligations that have a non-zero fee. Right now, if the default is zero, then you can't even create obligations that have some non-zero fee

MathisGD: but then how do you guarantee that the fee cannot increase?

MathisGD: should we prevent the creation of obligation way ahead in advance?

peyha: by using a constant MAX_TTM value ? that looks reasonable given we started enshrining precise time intervals in the trading fee but I'm not sure what the value should be to have sufficient margin (something like 1 year ?)

peyha: we could introduce some kind of timelock here, when an obligation is created, the fee setter has a delay (either a fixed duration or a function of the ttm) to update the fee (the first fee set could be chosen arbitrarily and subsequent ones should be decreasing). If no fee was set, then it's set to 0

MathisGD: in my mind default=0 => no fees was accepted

peyha: the default fee could still be increased right ? so you would not have the non-increasing fee guarantee

adhusson: made an issue for it

adhusson (approved):

PR #310, opened 2026-01-26
[ticks] direct tick inverse

adhusson:
- correctness proof on the [Notion page](https://www.notion.so/morpho-labs/Correctness-of-price-tick-conversion-2f1d69939e6d804fbb91fd15acb2ed35?source=copy_link)
- price step is 1e13 instead of 1e12 (following [this message](https://morpholabs.slack.com/archives/C02N7CZ088N/p1769098613992059?thread_ts=1769003317.743599&cid=C02N7CZ088N))
- always returns a tick that maps to the closest possible tick aligned price
- 1.4k gas


test/TickLibTest.sol
```diff
@@ -0,0 +1,41 @@
...
+        }
+    }
+
+    function testPriceToTickAndTickAlignedPriceEdges() public pure {
+        // When exactly between 2 prices match the higher one.
+        for (int256 tick = MIN_TICK; tick <= MAX_TICK; tick++) {
+            uint256 price = TickLib.tickToPrice(tick);
+            if (tick < MAX_TICK) {
+                uint256 nextPrice = TickLib.tickToPrice(tick + 1);
+                if (price == nextPrice) continue;
```
QGarchery: this `continue` may skip the `prevPrice` check, when both `tick < MAX_TICK` and `tick > MIN_TICK`

src/libraries/TickLib.sol
```diff
@@ -0,0 +1,129 @@
...
+        unchecked {
+            // forge-lint: disable-next-item(unsafe-typecast) x should be small
+            int256 q = (int256(x) + LN2_WAD / 2) / LN2_WAD;
+            // forge-lint: disable-next-item(unsafe-typecast) x should be small
+            int256 r = int256(x) - q * LN2_WAD;
+            int256 secondTerm = (r * r) / (2 * WAD_INT);
+            int256 thirdTerm = (secondTerm * r) / (3 * WAD_INT);
+            int256 expR = WAD_INT + r + secondTerm + thirdTerm;
+            // forge-lint: disable-next-line(unsafe-typecast) e^r positive, q positive
+            return uint256(expR) << uint256(q);
```
QGarchery, line 139: why do we use `int256` in this function, we could stay in `uint256` no ?

src/libraries/TickLib.sol
```diff
@@ -0,0 +1,129 @@
...
+
+// Min and max useful ticks.
+// Without the PRICE_STEP rounding, the tick range is -1679 for price 1 to 1679 for price 0.
+int256 constant MIN_TICK = -495;
+int256 constant MAX_TICK = 495;
+// ln(1.025) * 1e18
+int256 constant LN_1_025_WAD = 24692612590371416;
+// ln(2) * 1e18
+int256 constant LN2_WAD = 693147180559945344;
+int256 constant WAD_INT = 1e18;
```
QGarchery, line 14: do we want the constants to be in ConstantsLib ? Notably this one is already defined there

src/libraries/TickLib.sol
```diff
@@ -0,0 +1,129 @@
...
+                // forge-lint: disable-next-item(unsafe-typecast) x is negative
+                uint256 expWad = wExp(uint256(-x));
+                price = expWad * WAD / (WAD + expWad);
+            }
+            // forge-lint: disable-next-item(divide-before-multiply) loss is the point
+            return ((price + PRICE_STEP / 2) / PRICE_STEP) * PRICE_STEP;
+        }
+    }
+
+    /// @dev Returns the closest tick-aligned price and a tick of this price.
```
QGarchery: Maybe switch the order of the return parameters to fit this description

test/TickLibTest.sol
```diff
@@ -0,0 +1,41 @@
...
+                uint256 edgePrice = price - (price - nextPrice) / 2;
+                (, uint256 alignedEdgePrice) = TickLib.priceToTickAndTickAlignedPrice(edgePrice);
+                assertEq(alignedEdgePrice, price, "lower price up to same distance");
+            }
+            if (tick > MIN_TICK) {
+                uint256 prevPrice = TickLib.tickToPrice(tick - 1);
+                if (price == prevPrice) continue;
+                uint256 edgePrice = price + (prevPrice - price - 1) / 2;
```
QGarchery: would it be possible / make sense to negate the ticks (tick -> -tick), so that `tickToPrice` is increasing ? Here notably it would make things clearer I think

src/libraries/TickLib.sol
adhusson: `r` can be negative, for instance at tick 15.

edit: under the new tick convention tick 480 is an example

src/libraries/TickLib.sol
adhusson: Not sure because the function is supposed to be used outside of the protocol + copied verbatim in other languages. It's too bad if you can't get everything you need just by reading this file.

test/TickLibTest.sol
adhusson: It makes all descriptions of what the function does slightly less simple because you add a -, and people will wonder why did they do it oh it's because now it grows with the price.

But no strong opinion on it.

src/libraries/TickLib.sol
```diff
@@ -0,0 +1,129 @@
...
+        unchecked {
+            if (price == 0) return (MAX_TICK, tickToPrice(MAX_TICK));
+            if (price >= WAD) return (MIN_TICK, tickToPrice(MIN_TICK));
+
+            tick = roiToTick(((WAD - price) << 96) / price);
+            alignedPrice = tickToPrice(tick);
+
+            if (alignedPrice > price && tick < MAX_TICK) {
+                uint256 nextAlignedPrice = tickToPrice(tick + 1);
+                if (2 * price < alignedPrice + nextAlignedPrice) {
```
adhusson, line 63: I think we could write
```suggestion
                if (price - nextAlignedPrice < alignedPrice - price) {
```

which is clearer, but it would require separately proving `price >= nextAlignedPrice` so we know there is no underflow. I think the proof looks like the following, but is it worth it?

1. Show for all ticks `t` that `p_t` the lowest price strictly below `tickToPrice(t)` such that `tick(roi(p_t)) = t` verifies `tickToPrice(t+1) <= p_t`.
2. Now any `p` below `tickToPrice(t)` that maps to `t` is at least `p_t`, so it is >= `tickToPrice(t+1)`.

src/libraries/TickLib.sol
```diff
@@ -2,59 +2,143 @@
...
             }
+            // forge-lint: disable-next-item(divide-before-multiply) loss is the point
+            return ((price + PRICE_STEP / 2) / PRICE_STEP) * PRICE_STEP;
         }
     }
 
-    function tickToPrice(uint256 tick) internal pure returns (uint256) {
+    /// @dev Converts a price to the closest tick-aligned price and a tick that maps to it.
+    /// @dev If there are two equally close prices, the higher one is returned.
+    function priceToTickAlignedPriceAndTick(uint256 price) internal pure returns (uint256, uint256) {
```
adhusson: why return both
- it's probably as common to want the tick-aligned price of a price than the tick of a price (to display in interfaces or compute roi)
- avoids the gas cost of calling tickToPrice again just after

test/TickLibTest.sol
adhusson: changed to increasing with tick

src/libraries/TickLib.sol
adhusson: it's not that hard to understand as is

src/libraries/TickLib.sol
adhusson: for now returning only tick

src/libraries/TickLib.sol
QGarchery: We don't usually optimize for the number of imports/copy that an integrator has to do, but I can see how avoiding one indirection makes it easier to read. So no strong opinion on very specific constants, but maybe we still want to avoid duplication on common variables ?

src/MorphoV2.sol
```diff
@@ -28,6 +28,7 @@ import {
 } from "./interfaces/IMorphoV2.sol";
 import {ICallbacks, IFlashLoanCallback} from "./interfaces/ICallbacks.sol";
 import {EventsLib} from "./libraries/EventsLib.sol";
+import {TickLib} from "./libraries/TickLib.sol";
```
MathisGD, line 31:
```suggestion
```

adhusson:
closing as the gas advantage matters much less now (computation)

PR #312, opened 2026-01-27
refactor: remove session

QGarchery:
See [this thread](https://morpholabs.slack.com/archives/C02N7CZ088N/p1764667868337599) for the motivation


src/MorphoV2.sol
```diff
@@ -153,16 +154,16 @@ contract MorphoV2 is IMorphoV2 {
...
-        require(
-            UtilsLib.atMostOneNonZero(offer.assets, offer.obligationUnits, offer.obligationShares),
-            "inconsistent offer input"
-        );
+        require(UtilsLib.atMostOneNonZero(groupAssets, groupUnits, groupShares), "inconsistent group input");
         require(block.timestamp >= offer.start, "offer not started");
         require(block.timestamp <= offer.expiry, "offer expired");
         require(offer.maker != taker, "buyer and seller cannot be the same");
-        require(signer(root, sig) == offer.maker, "invalid signature");
+        require(groupLoanToken == offer.obligation.loanToken, "group loan token mismatch");
```
QGarchery, line 156: We could also remove the `loanToken` from the obligation, but it's quite ugly...

src/MorphoV2.sol
MathisGD, line 1: are we ok with [this tradeoff](https://morpholabs.slack.com/archives/C02N7CZ088N/p1764668927102559?thread_ts=1764667868.337599&cid=C02N7CZ088N)?

src/MorphoV2.sol
peyha: how would that work ? The obligation id wouldn't depend on the loan token so we'd get massive id clash

QGarchery: I mean just in the `take` interface: the arguments could be everything that is in the obligation except the loan token. We could then recreate the Obligation struct

src/MorphoV2.sol
QGarchery: Yes definitely, for the reasons in that same thread. [This thead](https://morpholabs.slack.com/archives/C02N7CZ088N/p1769617927291119) is also relevant

MathisGD, line 1: raised a concern [here](https://morpholabs.slack.com/archives/C02N7CZ088N/p1770287474364369?thread_ts=1764667868.337599&cid=C02N7CZ088N)

QGarchery:
Closing because of [this](https://github.com/morpho-org/morpho-v2/pull/312#discussion_r2769067561)

PR #314, opened 2026-01-27
perf: packing

MathisGD:
trading fees fit into `uint16` because the max is only 1% (so 10000 < 2^16)


MathisGD:
@codex review

src/libraries/UtilsLib.sol
```diff
@@ -61,4 +61,10 @@ library UtilsLib {
     function sort(bytes32 x, bytes32 y) internal pure returns (bytes memory) {
         return x < y ? abi.encodePacked(x, y) : abi.encodePacked(y, x);
     }
+
+    function toUint128(uint256 x) internal pure returns (uint128) {
+        require(x <= type(uint128).max, "uint256 overflows uint128");
+        // forge-lint: disable-next-line(unsafe-typecast) as x is less than type(uint128).max
```
QGarchery, line 67: It doesn't do [this](https://morpholabs.slack.com/archives/C0A181MTG4W/p1769020211779439) anymore ?

src/MorphoV2.sol
```diff
@@ -204,28 +211,28 @@ contract MorphoV2 is IMorphoV2 {
...
         } else {
             // Borrower exits + lender exits.
-            debtOf[buyer][id] -= obligationUnits;
-            sharesOf[seller][id] -= obligationShares;
-            totalShares[id] -= obligationShares;
-            totalUnits[id] -= obligationUnits;
+            debtOf[id][buyer] -= obligationUnits;
+            sharesOf[id][seller] -= obligationShares;
+            _obligationStorage.totalShares -= UtilsLib.toUint128(obligationShares);
+            _obligationStorage.totalUnits -= UtilsLib.toUint128(obligationUnits);
```
QGarchery, line 235: Here we could theoretically have an unsafe typecast, as `obligationUnits` is less than `debOf[id][borrower]` thanks to the line above, and we know that it is less than `_obligationShares.totalUnits` than is stored in 128 bits. Probably not worth because it depends on the ordering with the line `debtOf[id][buyer] -= obligationUnits;`

src/interfaces/IMorphoV2.sol
```diff
@@ -46,4 +46,14 @@ struct Seizure {
     uint256 seized;
 }
 
+struct ObligationStorage {
+    uint128 totalUnits;
+    uint128 totalShares;
```
QGarchery, line 53: Ok to store in 128 bits, because it is already assumed that the multiplication on the 2 doesn't overflow. So if we assume that they have the same order of magnitude `x`, that assumption is ~ `x * x < 2^256` <=> `x < 2^128`.

This doesn't hold if we increase the virtual shares, in which case we should document it, as it is done [here](https://github.com/morpho-org/morpho-blue/blob/c500a8af0e91483e10c8ede2b40680056ba1c126/src/interfaces/IMorpho.sol#L121)

QGarchery (approved):

src/interfaces/IMorphoV2.sol
```diff
@@ -46,4 +46,14 @@ struct Seizure {
...
+    uint256 withdrawable;
+    bool created;
+    /// @dev Fee indices: 0=0d, 1=1d, 2=7d, 3=30d, 4=90d, 5=180d TTM buckets.
+    /// @dev Fees are stored divided by FEE_STEP (1e12) to fit in 16 bits. Max fee is 1% (0.01e18).
+    uint16[6] fees;
+}
```
adhusson, line 57: nitpick a struct field is not an allowed natspec context https://docs.soliditylang.org/en/latest/natspec-format.html#tags

```suggestion
/// @dev Fee indices: 0=0d, 1=1d, 2=7d, 3=30d, 4=90d, 5=180d TTM buckets.
/// @dev Fees are stored divided by FEE_STEP (1e12) to fit in 16 bits. Max fee is 1% (0.01e18).
struct ObligationStorage {
    uint128 totalUnits;
    uint128 totalShares;
    uint256 withdrawable;
    bool created;
    uint16[6] fees;
}
```

adhusson (approved):

src/libraries/UtilsLib.sol
MathisGD: it was because of the formatting

src/interfaces/IMorphoV2.sol
```diff
@@ -46,4 +46,14 @@ struct Seizure {
     uint256 seized;
 }
 
+struct ObligationStorage {
+    uint128 totalUnits;
+    uint128 totalShares;
+    uint256 withdrawable;
```
peyha, line 54: why not store this on 128 bits ? It would be packed with created and fees and these would be 225  (128+1+6*16) bytes in total so we could still add interest fees (depending on how we set it)

MathisGD: because we will add interest fees anyway

src/interfaces/IMorphoV2.sol
MathisGD: I would document anyway, see #38

src/MorphoV2.sol
MathisGD: think that it's not worth it too

PR #316, opened 2026-01-27
perf: return health check early

MathisGD:
to avoid looping in all the collats once the borrowing power is found.


src/MorphoV2.sol
```diff
@@ -468,18 +468,15 @@ contract MorphoV2 is IMorphoV2 {
...
-            return debt <= maxDebt;
+        uint256 maxDebt;
+        for (uint256 i = 0; i < obligation.collaterals.length; i++) {
+            if (maxDebt >= debt) return true;
+            Collateral memory _collateral = obligation.collaterals[i];
+            uint256 price = IOracle(_collateral.oracle).price();
+            maxDebt += collateralOf[borrower][id][_collateral.token].mulDivDown(price, ORACLE_PRICE_SCALE)
+                .mulDivDown(_collateral.lltv, WAD);
         }
+        return maxDebt >= debt;
```
adhusson, line 509: less clear what is going on

As long as we are ok with early returns we might as well do this:


```suggestion
        if (debt == 0) return true;
        uint256 maxDebt;
        for (uint256 i = 0; i < obligation.collaterals.length; i++) {
            Collateral memory _collateral = obligation.collaterals[i];
            uint256 price = IOracle(_collateral.oracle).price();
            maxDebt += collateralOf[borrower][id][_collateral.token].mulDivDown(price, ORACLE_PRICE_SCALE)
                .mulDivDown(_collateral.lltv, WAD);
            if (maxDebt >= debt) return true;
        }
        return false;
```

MathisGD: proposed a third thing

adhusson (approved):

QGarchery (approved):

peyha (approved):

PR #317, opened 2026-01-28
[id to obligation] SSTORE

QGarchery:
Discussion [here](https://morpholabs.slack.com/archives/C08A40KKN9W/p1769605781960269)

Using an obligation with 2 collaterals, the id to obligation is isolated from the touch obligation

Screenshot 2026-01-30 at 18 36 18

### DIFF
--- a/src/MorphoV2.sol
+++ b/src/MorphoV2.sol
@@ -38,6 +38,7 @@ contract MorphoV2 is IMorphoV2 {
     mapping(bytes32 id => mapping(address user => uint256)) public debtOf;
     mapping(bytes32 id => mapping(address user => mapping(address collateralToken => uint256))) public collateralOf;
     mapping(bytes32 id => ObligationStorage) public obligationStorage;
+    mapping(bytes32 id => Obligation) internal _idToObligation;
 
     /// @dev Groups are useful to have a global offered amount shared accross multiple offers ("OCO").
     /// @dev To work as expected, all offers in a same group should have the same assets, obligationUnits,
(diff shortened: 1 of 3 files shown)


src/MorphoV2.sol
```diff
@@ -465,13 +466,19 @@ contract MorphoV2 is IMorphoV2 {
...
         }
         return id;
     }
 
     /// VIEW FUNCTIONS ///
 
+    function idToObligation(bytes32 id) external view returns (Obligation memory) {
+        return _idToObligation[id];
+    }
+
```
MathisGD, line 481: this or the interface trick?

src/MorphoV2.sol
```diff
@@ -465,13 +466,19 @@ contract MorphoV2 is IMorphoV2 {
             obligationStorage[id].created = true;
             obligationStorage[id].fees = defaultFees[obligation.loanToken];
 
+            _idToObligation[id] = obligation;
```
MathisGD, line 469: adds 60,000 * N + 20,000 gas to each obligation creation

peyha: that's a lot

peyha (approved):

src/MorphoV2.sol
QGarchery: not sure, but if there are not so many such functions it may be best to have only one interface for simplicity

src/MorphoV2.sol
QGarchery: Seems ok to me as a one time cost

MathisGD: at which point it's smart to do a SSTORE3?

peyha: just checked with claude, SSTORE3/SSTORE2 saves massive amount of gas (from 50% to 75% depending on the nb of collat) so it's worth looking into it

QGarchery: it saves a lot of gas on reads as well, so it can be very interesting (makes me think that it would have been a nice solution for vault v1)

MathisGD: can someone open a pr to try this. ideally with minimal code

QGarchery: I'm doing it

QGarchery: done in #324

MathisGD:
#328 is 5x cheaper in gas for get (and 30% cheaper for touchObligation). a bit more complex too, but the gas gains are massive, probably worth it

QGarchery:
closing in favor of #328

PR #318, opened 2026-01-28
[ticks] ticks table

adhusson:
- tick -> price : 800 gas
- price -> tick : 2.4k gas at worst


src/libraries/TickLib.sol
```diff
@@ -0,0 +1,123 @@
...
+                    return (MID_PRICE + highDelta, -int256(highIndex));
+                } else {
+                    // forge-lint: disable-next-item(unsafe-typecast) MAX_TICK at most
+                    return (MID_PRICE + lowDelta, -int256(lowIndex));
+                }
+            }
+        }
+    }
+
+    function getDelta(bytes memory deltas, uint256 index) private pure returns (uint256 wadDelta) {
```
MathisGD, line 113: why storing deltas?

adhusson: 1 bit smaller so smaller table & values fit on 16 bits not 17 which in general makes it easier to use / reimplement elsewhere etc.

MathisGD: but doesn't it make the dichotomy super hard?

adhusson: why? i don't see it

src/libraries/TickLib.sol
```diff
@@ -0,0 +1,123 @@
...
+        hex"c30a" hex"c30b" hex"c30c" hex"c30d" hex"c30e" hex"c30f" hex"c310" hex"c311" hex"c312" hex"c313" hex"c314"
+        hex"c315" hex"c316" hex"c317" hex"c318" hex"c319" hex"c31a" hex"c31b" hex"c31c" hex"c31d" hex"c31e" hex"c31f"
+        hex"c320" hex"c321" hex"c322" hex"c323" hex"c324" hex"c325" hex"c326" hex"c327" hex"c328" hex"c329" hex"c32a"
+        hex"c32b" hex"c32c" hex"c32d" hex"c32e" hex"c32f" hex"c330" hex"c331" hex"c332" hex"c333" hex"c334" hex"c335"
+        hex"c336" hex"c337" hex"c338" hex"c339" hex"c33a" hex"c33b" hex"c33c" hex"c33d" hex"c33e" hex"c33f" hex"c340"
+        hex"c341" hex"c342" hex"c343" hex"c344" hex"c345" hex"c346" hex"c347" hex"c348" hex"c349" hex"c34a" hex"c34b"
+        hex"c34c" hex"c34d" hex"c34e" hex"c34f" hex"c350";
+
+    /// @dev Does not check price bounds
+    function tickToPrice(int256 tick) internal pure returns (uint256) {
```
peyha: shouldn't we require -MAX_TICK <= tick <= MAX_TICK in the lib ? otherwise getDelta would have weird behaviors

peyha: unless that's what you meant by "does not check price bounds" but then it should be changed to "does not change tick bounds"

src/libraries/TickLib.sol
```diff
@@ -0,0 +1,123 @@
...
+            } else {
+                // forge-lint: disable-next-item(unsafe-typecast) tick is positive
+                return WAD - (MID_PRICE + getDelta(deltas, uint256(tick)));
+            }
+        }
+    }
+
+    /// @dev Returns the closest tick-aligned price and its tick.
+    /// @dev If there are two equally close prices, returns the higher one.
+    function priceToTickAlignedPriceAndTick(uint256 price) internal pure returns (uint256, int256) {
```
peyha:
```suggestion
    function priceToAlignedPriceAndTick(uint256 price) internal pure returns (uint256, int256) {
```
small prop

peyha: or 
```suggestion
    function priceToTickAndAlignedPrice(uint256 price) internal pure returns (uint256, int256) {
```

src/libraries/TickLib.sol
adhusson: another thing is memory is byte-addressed so reading in the string is cleaner you read at startPos + 2*index. With prices either we store 3 bytes per price so the table is 50% larger instead of 496 bits larger, or we store 17 bits per price and its more ugly (need to find the byte offset of the nth 17 bit sequence then clean both left and right)

src/libraries/TickLib.sol
```diff
@@ -2,59 +2,117 @@
...
+
+uint256 constant MAX_TICK = 990;
+uint256 constant PRICE_STEP = 1e13;
 
 library TickLib {
-    using TickLib for uint256;
+    // 1e5 scaled deltas to price 0.5 for tick 495 to 990.
+    // Rounded from 1/(1+1.025^(495-tick)), but they can be shifted to keep the map injective.
+    // To generate the map: https://gist.github.com/d67bcf5c3f62e5d734f7745c95e11407
+    bytes internal constant PACKED_DELTAS = hex"0000" hex"0269" hex"04d2" hex"073b" hex"09a3" hex"0c0b" hex"0e71"
```
peyha, line 14: too bad we can't use inheritance to get `PACKED_DELTAS`, it would have been great to put the data in a specific sol file because it reduces the visibility of `tickToPrice` and `priceToTick`

adhusson:
closing as the gas advantage matters much less now (computation)

PR #319, opened 2026-01-28
refactor: trading fee can increase

MathisGD:


src/MorphoV2.sol
QGarchery, line 1: Actually, the taker could get front-run by a fee switch or a fee increase. Possible solutions:
- acknowledge it, by adding a comment. Also this could be dealt with on top with a slippage check : litterally a slippage check, or checking directly that the fee has not changed
- add a parameter to do this last check in this contract

I prefer the first solution, because in the worst case changes in the trading fee should be pretty minimal and predictable

peyha (approved):
We're missing a few tests for this (raised in #321)

src/MorphoV2.sol
MathisGD: pushed a comment

QGarchery (approved):

PR #320, opened 2026-01-29
Obligation level min collat quoted in loan

peyha:
Obligation now has a new parameter `uint256 minCollateral` which is quoted in loan token.
The min collateral is checked in `withdrawCollateral` and in `take` for the seller position.

Note that some positions with an accepted initial amount of collateral could fall under min collateral through an oracle update (that would make the collateral price decrease)

Remaining questions
- Is this the correct implementation ? (compared to collateral level min collateral)
- Should positions whose collateral fall below min collat be liquidatable ? With full liquidation ?
- Should the liquidation of these positions always fully close the position ?


src/MorphoV2.sol
```diff
@@ -488,17 +492,26 @@ contract MorphoV2 is IMorphoV2 {
         return obligationState[id].fees;
     }
 
-    /// @dev This function should be called with the id corresponding to the obligation.
-    function isHealthy(Obligation memory obligation, bytes32 id, address borrower) public view returns (bool) {
+    function toId(Obligation memory obligation) public view returns (bytes32) {
+        return keccak256(abi.encode(block.chainid, address(this), obligation));
+    }
+
```
QGarchery, line 498:
```suggestion
```

wrong merge I think

src/MorphoV2.sol
```diff
@@ -488,17 +492,26 @@ contract MorphoV2 is IMorphoV2 {
         return obligationState[id].fees;
     }
 
-    /// @dev This function should be called with the id corresponding to the obligation.
-    function isHealthy(Obligation memory obligation, bytes32 id, address borrower) public view returns (bool) {
+    function toId(Obligation memory obligation) public view returns (bytes32) {
```
qdrn, line 495: Can't you reuse IdLib?

src/MorphoV2.sol
```diff
@@ -488,17 +492,26 @@ contract MorphoV2 is IMorphoV2 {
...
             Collateral memory collateral = obligation.collaterals[i];
             uint256 price = IOracle(collateral.oracle).price();
-            maxDebt += collateralOf[id][borrower][collateral.token].mulDivDown(price, ORACLE_PRICE_SCALE)
-                .mulDivDown(collateral.lltv, WAD);
+            uint256 collateralQuoted =
+                collateralOf[id][borrower][collateral.token].mulDivDown(price, ORACLE_PRICE_SCALE);
+            collateralValue += collateralQuoted;
+            maxDebt += collateralQuoted.mulDivDown(collateral.lltv, WAD);
+            belowMinCollateral =
+                belowMinCollateral || (collateralQuoted > 0 && collateralQuoted < obligation.minCollateral);
```
QGarchery, line 512: shouldn't it be if the sum of collateral quoted is below min collat ? Otherwise I don't see the benefit compared to #339

src/MorphoV2.sol
```diff
@@ -332,7 +334,9 @@ contract MorphoV2 is IMorphoV2 {
 
         collateralOf[id][onBehalf][collateral] -= assets;
 
-        require(isHealthy(obligation, id, onBehalf), "Unhealthy borrower");
+        (bool _isHealthy, bool belowMinCollateral) = isHealthy(obligation, id, onBehalf);
+        require(_isHealthy, "Unhealthy borrower");
+        require(!belowMinCollateral, "Borrower collateral below min");
```
QGarchery, line 339: At the moment we always do those checks, so it could be factored in `isHealthy`. But this implementation is nice in case we want to implement "can fully liquidate below min collat"

peyha:
Closing it in favor of #342 which is a similar idea but implemented differently

PR #322, opened 2026-01-29
[interest fee] interest fee per borrower

MathisGD:


src/MorphoV2.sol
peyha, line 144: shouldn't the interest fee be accrued in this function ?

PR #324, opened 2026-01-29
[id to obligation] SSTORE2

QGarchery:
Motivation comes from [this](https://github.com/morpho-org/morpho-v2/pull/317#discussion_r2737586475)

### DIFF
--- a/src/MorphoV2.sol
+++ b/src/MorphoV2.sol
@@ -477,7 +477,7 @@ contract MorphoV2 is IMorphoV2 {
             require(_idToObligationContract != address(0), "obligation deploy failed");
             idToObligationContract[id] = _idToObligationContract;
 
-            emit EventsLib.ObligationCreated(id, obligation);
+            emit EventsLib.ObligationCreated(id, obligation, _idToObligationContract);
         }
         return id;
     }
--- a/src/MorphoV2.sol
+++ b/src/MorphoV2.sol
@@ -12,7 +12,8 @@ import {
     MAX_LIF,
     TIME_TO_MAX_LIF,
     EIP712_DOMAIN_TYPEHASH,
-    ROOT_TYPEHASH
+    ROOT_TYPEHASH,
+    OBLIGATION_DEPLOYER_PREFIX
 } from "./libraries/ConstantsLib.sol";
 import {IOracle} from "./interfaces/IOracle.sol";
 import {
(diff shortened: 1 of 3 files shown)


src/MorphoV2.sol
```diff
@@ -476,7 +485,19 @@ contract MorphoV2 is IMorphoV2 {
...
+        }
+        uint256 size;
+        assembly ("memory-safe") {
+            size := extcodesize(_idToObligationContract)
+        }
+        bytes memory data = new bytes(size);
+        assembly ("memory-safe") {
+            extcodecopy(_idToObligationContract, add(data, 0x20), 0, size)
+        }
+        return abi.decode(data, (Obligation));
```
QGarchery, line 500: this could be abstracted in a lib. External contracts may store `idToObligationContract[id]` instead of the id, and save gas this way: no need to fetch a cold slot, and no need to go through the v2 contract

src/MorphoV2.sol
```diff
@@ -38,7 +39,8 @@ contract MorphoV2 is IMorphoV2 {
     mapping(bytes32 id => mapping(address user => uint256)) public debtOf;
     mapping(bytes32 id => mapping(address user => mapping(address collateralToken => uint256))) public collateralOf;
     mapping(bytes32 id => ObligationState) public obligationState;
-    mapping(bytes32 id => Obligation) internal _idToObligation;
+    /// @dev Address of a contract whose bytecode is abi.encode(obligation)
+    mapping(bytes32 id => address) internal idToObligationContract;
```
QGarchery, line 43: this could be public

MathisGD, line 43: why do you need to store a contract at all? can't create2 make the job?

src/MorphoV2.sol
```diff
@@ -466,17 +468,36 @@ contract MorphoV2 is IMorphoV2 {
...
 
-            _idToObligation[id] = obligation;
+            bytes memory obligationData = abi.encode(obligation);
+            bytes memory creationCode = abi.encodePacked(OBLIGATION_DEPLOYER_PREFIX, obligationData);
+            address _idToObligationContract;
+            assembly ("memory-safe") {
+                _idToObligationContract := create(0, add(creationCode, 0x20), mload(creationCode))
+            }
+            require(_idToObligationContract != address(0), "obligation deploy failed");
+            idToObligationContract[id] = _idToObligationContract;
```
MathisGD, line 478: can probably be a function in utilsLib `sstore2`

src/MorphoV2.sol
QGarchery: that sounds great 👍

QGarchery:
closing in favor of https://github.com/morpho-org/morpho-v2/pull/328

PR #325, opened 2026-01-30
[id to obligation] SSTORE4

QGarchery:
The idea is that we want to use create2 to not have to store the contract address and being able to recompute it instead.

This means that the creation bytecode of the storing contract cannot contain the Obligation struct. This is because the init bytecode is needed to compute the address, and would defeat the purpose of the `idToObligation` getter.
The usual way to solve this (SSTORE3 pattern), is by creating the runtime bytecode by reading the necessary values in storage. But we can't write to storage, that would defeat the whole gas savings ideas. So instead we write to transient storage (hence naming this jokingly SSTORE4)

### DIFF
--- a/src/MorphoV2.sol
+++ b/src/MorphoV2.sol
@@ -12,8 +12,7 @@ import {
     MAX_LIF,
     TIME_TO_MAX_LIF,
     EIP712_DOMAIN_TYPEHASH,
-    ROOT_TYPEHASH,
-    OBLIGATION_DEPLOYER_PREFIX
+    ROOT_TYPEHASH
 } from "./libraries/ConstantsLib.sol";
 import {IOracle} from "./interfaces/IOracle.sol";
 import {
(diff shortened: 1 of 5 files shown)


src/MorphoV2.sol
MathisGD, line 1: I didn't think that create2 wouldn't work as simply. Then it's a bit painful now..

src/MorphoV2.sol
```diff
@@ -454,6 +452,47 @@ contract MorphoV2 is IMorphoV2 {
         SafeTransferLib.safeTransferFrom(token, msg.sender, address(this), assets);
     }
 
+    function setObligationCreation(Obligation memory obligation) internal {
```
QGarchery, line 455: we have to do manual transient storage because solidity doesn't allow complex transient types

QGarchery:
closing in favor of https://github.com/morpho-org/morpho-v2/pull/328

PR #327, opened 2026-01-30
refactor: default trading fee

MathisGD:


src/MorphoV2.sol
```diff
@@ -104,7 +104,15 @@ contract MorphoV2 is IMorphoV2 {
         emit EventsLib.SetObligationTradingFee(id, index, newTradingFee);
     }
 
-    /// @dev Doesn't change the fee of already created obligations.
+    /// @dev Activates or deactivates custom fees for an obligation.
+    /// @dev When deactivated, the obligation uses defaultFees[loanToken] instead.
+    function setObligationCustomFeeActivated(bytes32 id, bool customFeeActivated) external {
+        require(msg.sender == feeSetter, "Only feeSetter");
+        obligationState[id].customFeeActivated = customFeeActivated;
+        emit EventsLib.SetObligationCustomFeeActivated(id, customFeeActivated);
```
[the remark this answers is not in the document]
MathisGD: expected

src/MorphoV2.sol
QGarchery, line 1: What is the rationale for that change ? It seems like it has the issue that now the fees can increase (if the default fee increases)

MathisGD: for the trading fee yes

QGarchery: oh right my bad

PR #328, opened 2026-01-30
[id to obligation] id as hash of creation code

QGarchery:
see [this thead](https://morpholabs.slack.com/archives/C08A40KKN9W/p1769764057128329)

Using an obligation with 2 collaterals, the id to obligation is isolated from the touch obligation

Screenshot 2026-01-30 at 18 35 08


src/libraries/IdLib.sol
```diff
@@ -0,0 +1,23 @@
...
+library IdLib {
+    function toId(Obligation memory obligation, uint256 chainid, address morphoV2) internal pure returns (bytes32) {
+        bytes memory creationCode =
+            abi.encodePacked(type(ObligationDeployer).creationCode, abi.encode(obligation, chainid, morphoV2));
+        return keccak256(creationCode);
+    }
+
+    function idToObligationContract(address morphoV2, bytes32 id) internal pure returns (address) {
+        return address(uint160(uint256(keccak256(abi.encodePacked(uint8(0xff), morphoV2, bytes32(0), id)))));
+    }
```
MathisGD: we could not expose this at all (as a way to abstract what's going on too)

src/libraries/IdLib.sol
```diff
@@ -0,0 +1,20 @@
...
+// Copyright (c) 2025 Morpho Association
+pragma solidity ^0.8.0;
+
+import {ObligationDeployer} from "../ObligationDeployer.sol";
+import {Obligation} from "../interfaces/IMorphoV2.sol";
+
+library IdLib {
+    function toId(Obligation memory obligation, uint256 chainid, address morphoV2) internal pure returns (bytes32) {
+        bytes memory creationCode =
+            abi.encodePacked(type(ObligationDeployer).creationCode, abi.encode(obligation, chainid, morphoV2));
```
QGarchery: wondering if we should hardcode the creation bytecode. This way it makes it more easily reusable, no need to compile the `ObligationDeployer` contract

src/MorphoV2.sol
```diff
@@ -466,7 +467,7 @@ contract MorphoV2 is IMorphoV2 {
             obligationState[id].created = true;
             obligationState[id].fees = defaultFees[obligation.loanToken];
 
-            _idToObligation[id] = obligation;
+            new ObligationDeployer{salt: bytes32(0)}(obligation, block.chainid, address(this));
```
MathisGD: should it be abstracted in the lib?

QGarchery: It would make sense to group the logic of ids in the same place, but at the same time I don't expect anyone to use that function. So the lib would contain both useful and useless functions for integrators

src/libraries/IdLib.sol
```diff
@@ -0,0 +1,20 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Copyright (c) 2025 Morpho Association
+pragma solidity ^0.8.0;
+
+import {ObligationDeployer} from "../ObligationDeployer.sol";
+import {Obligation} from "../interfaces/IMorphoV2.sol";
+
+library IdLib {
```
peyha, line 7:
```suggestion
library ObligationLib {
```
wdyt ?

MathisGD: I don't like it, if it was ObligationParams maybe

src/MorphoV2.sol
MathisGD: yes, libs have two purpose: abstracting away things when it makes sense (rarer than people think imo) like mathslib / utilslib, and making things available for external use (like constantLib, or eventsLib). here it would be the two at the same time (thus some functions might be useless for integrators)

MathisGD: or we could have a sstore2 lib (general purpose) and an id lib

MathisGD: the issue with the current code is that this contract deployment feels a bit weird, it could be misinterpreted. while it's a deep technicality

src/libraries/IdLib.sol
adhusson: Agree, something like `bytes.concat(hex"0x385f5f39600b380360bf3",encodedObligation)`.

It copies the entire code to memory then returns bytes `11` to `codesize-11`: `codesize, push0, push0, codecopy, push1 11, codesize, sub, push1 11, return`.

This would not be etherscan verifiable. But we also would have to check that the current version with ObligationDeployer is etherscan verifiable since it uses assembly in the constructor and bypasses the normal constructor return value.

src/libraries/IdLib.sol
peyha: yes, not perfect but I struggle to find a better naming. The main issue being that this lib is specific to `Obligation` so it could be great to make it clear in the name
other prop `ObligationIdLib` but it's also pretty ugly

src/libraries/IdLib.sol
```diff
@@ -5,10 +5,13 @@ pragma solidity ^0.8.0;
 import {ObligationDeployer} from "../ObligationDeployer.sol";
 import {Obligation} from "../interfaces/IMorphoV2.sol";
 
+bytes constant OBLIGATION_DEPLOYER_BYTECODE =
```
adhusson: we should check if its verifiable with the assembly in constructor trick, if not no reason not to use some [minimal creation bytecode](https://github.com/morpho-org/morpho-v2/pull/328#discussion_r2754551866)

src/libraries/IdLib.sol
```diff
@@ -0,0 +1,27 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Copyright (c) 2025 Morpho Association
+pragma solidity ^0.8.0;
+
+import {ObligationDeployer} from "../ObligationDeployer.sol";
+import {Obligation} from "../interfaces/IMorphoV2.sol";
+
+bytes constant OBLIGATION_DEPLOYER_BYTECODE =
```
peyha: generally not a fan of putting raw bytecode in contracts but in this case it looks ok given the scope of ObligationDeployer.
Shall we put a comment to specify either an existing deployment of this contract or the solc version/compilation settings used to get this one for reproducability ?

src/libraries/IdLib.sol
QGarchery: No real reason to verify this contract right ? I had exactly the same minimal creation bytecode in a [previous iteration](https://github.com/morpho-org/morpho-v2/pull/324/files), but in the end I think the readability is better this way. But now we also copy the full bytecode and put it as a constant, so I'm less sure

src/libraries/IdLib.sol
MathisGD: actually I think that it's fine to recompile the code no? you are afraid that with solidity versions etc it's too breaking?

QGarchery: Not sure why we would need reproducibility, this will be deployed by the MorphoV2 contract only. We need the bytecode to be the one that is deployed though, and this is why we have the bytecode test (assuming tests and deployment have the same config)

src/libraries/IdLib.sol
```diff
@@ -0,0 +1,27 @@
...
+        return keccak256(creationCode);
+    }
+
+    function idToObligation(address morphoV2, bytes32 id) internal view returns (Obligation memory) {
+        address create2Address =
+            address(uint160(uint256(keccak256(abi.encodePacked(uint8(0xff), morphoV2, bytes32(0), id)))));
+        return abi.decode(create2Address.code, (Obligation));
+    }
+
+    function deployObligation(Obligation memory obligation) internal {
```
MathisGD: should it take bytes?
```suggestion
    function sstore2(bytes memory data) internal {
```

MathisGD: btw it's not exactly sstore2, because it's using create2. neither sstore3.

src/libraries/IdLib.sol
adhusson: Sorry I missed the one you did, not the same btw but same size.

Agree for me it doesn't matter if it's verifiable, but if none are the point would be moot anyway.

The minimal versions are not hard to understand worth removing a contract & having a very simple function for me.

src/ObligationDeployer.sol
```diff
@@ -0,0 +1,15 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Copyright (c) 2025 Morpho Association
+pragma solidity ^0.8.0;
+
+import {Obligation} from "./interfaces/IMorphoV2.sol";
+
+contract ObligationDeployer {
```
MathisGD:
```suggestion
contract Sstore2 {
```

src/libraries/IdLib.sol
```diff
@@ -0,0 +1,27 @@
...
+        new ObligationDeployer{salt: bytes32(0)}(obligation, block.chainid, address(this));
```
MathisGD: does more or less the same thing
```suggestion
        new ObligationDeployer{salt: keccak256(abi.encode(block.chainid, address(this))}(obligation);
```

src/ObligationDeployer.sol
```diff
@@ -0,0 +1,15 @@
...
+    constructor(Obligation memory obligation, uint256, address) {
+        bytes memory encodedObligation = abi.encode(obligation);
+
+        assembly ("memory-safe") {
+            return(add(encodedObligation, 0x20), mload(encodedObligation))
+        }
+    }
```
MathisGD:
```suggestion
    constructor(bytes memory data) {
        assembly ("memory-safe") {
            return(add(data, 0x20), mload(data))
        }
    }
```

src/libraries/IdLib.sol
QGarchery: you could also say that this lib is specific to `bytes32 id`. So unless we put both in the name as you suggested there will still be an issue

src/MorphoV2.sol
```diff
@@ -466,7 +466,7 @@ contract MorphoV2 is IMorphoV2 {
             obligationState[id].created = true;
             obligationState[id].fees = defaultFees[obligation.loanToken];
 
-            _idToObligation[id] = obligation;
+            IdLib.deployObligation(obligation);
```
MathisGD: would go for something more abstract to avoid confusion (oh there is a contract per obligation..)
```suggestion
            Sstore2Lib.sstore2(obligation);
```

src/libraries/IdLib.sol
QGarchery: it would store the chain id and the morpho v2 address in the bytecode, is that ok ?

QGarchery: Actually no, we can put those in the salt 🤯

QGarchery: Wait no, it doesn't work because then the id dosn't depend on the chainid and the morphoV2 address :/

adhusson: Does it matter? Before it did not depend on the chainid and morphoV2 address.

QGarchery: > Before it did not depend on the chainid and morphoV2 address.

it did depend on it, see [here](https://github.com/morpho-org/morpho-v2/blob/4cee515d32156237e65fcec3db327ef5cbb01f8e/src/MorphoV2.sol#L494-L496)

> Does it matter? 

Yes for [this reason](https://morpholabs.slack.com/archives/C02N7CZ088N/p1761837349904449)

src/libraries/IdLib.sol
peyha: I missed the bytecode test, then it's fine so I'm resolving the comment

src/libraries/IdLib.sol
QGarchery: not exactly because then the id dosn't depend on the chainid and the morphoV2 address

src/libraries/IdLib.sol
peyha: that's right we would need both (even though we [did not care for MarketParams](https://github.com/morpho-org/morpho-blue/blob/main/src/libraries/MarketParamsLib.sol))

src/libraries/IdLib.sol
```diff
@@ -0,0 +1,28 @@
...
+
+    function idToObligation(address morphoV2, bytes32 id) internal view returns (Obligation memory) {
+        address create2Address =
+            address(uint160(uint256(keccak256(abi.encodePacked(uint8(0xff), morphoV2, bytes32(0), id)))));
+        // The SStore2 contract code of an obligation also contains chainid and morphoV2, which are discarded.
+        return abi.decode(create2Address.code, (Obligation));
+    }
+
+    function sstore2(Obligation memory obligation) internal {
+        new SStore2{salt: bytes32(0)}(abi.encode(obligation, block.chainid, address(this)));
```
peyha: just saw Solady's implem and they append a `STOP`opcode at the beginning of the data to make sure that the SSTORE2 contract cannot be called. Not sure we should do it but still worth flagging, there would be a potential issue with the old SELFDESTRUCT opcode but now it should be ok

peyha: oh wait it could be used to DoS some obligation if its data starts with 0xFF (SELFDESTRUCT) by touching it and then calling it within the same tx

peyha: in the end it should be ok. when encoding obligation, the first bytes are 0x00 (STOP) because `loanToken` is of type address. Still worth having in mind if we ever change `Obligation`

src/libraries/IdLib.sol
```diff
@@ -0,0 +1,28 @@
...
+
+bytes constant SSTORE2_BYTECODE =
+    hex"60806040523460845760c180380380601581609c565b9283398101906020818303126084578051906001600160401b0382116084570181601f8201121560845780516001600160401b038111608857605f601f8201601f1916602001609c565b91818352602083019360208383010111608457815f926020809301865e830101525190f35b5f80fd5b634e487b7160e01b5f52604160045260245ffd5b6040519190601f01601f191682016001600160401b0381118382101760885760405256fe";
+
+library IdLib {
+    function toId(Obligation memory obligation, uint256 chainid, address morphoV2) internal pure returns (bytes32) {
+        bytes memory sstore2Data = abi.encode(obligation, chainid, morphoV2);
+        bytes memory creationCode = abi.encodePacked(SSTORE2_BYTECODE, abi.encode(sstore2Data));
+        return keccak256(creationCode);
+    }
```
MathisGD: @codex how much more costly is the new toId after this PR? this is an important consideration because each interaction with the obligation will call this, not only the first one

src/libraries/IdLib.sol
```diff
@@ -0,0 +1,28 @@
...
+import {SStore2} from "../SStore2.sol";
+import {Obligation} from "../interfaces/IMorphoV2.sol";
+
+bytes constant SSTORE2_BYTECODE =
+    hex"60806040523460845760c180380380601581609c565b9283398101906020818303126084578051906001600160401b0382116084570181601f8201121560845780516001600160401b038111608857605f601f8201601f1916602001609c565b91818352602083019360208383010111608457815f926020809301865e830101525190f35b5f80fd5b634e487b7160e01b5f52604160045260245ffd5b6040519190601f01601f191682016001600160401b0381118382101760885760405256fe";
+
+library IdLib {
+    function toId(Obligation memory obligation, uint256 chainid, address morphoV2) internal pure returns (bytes32) {
+        bytes memory sstore2Data = abi.encode(obligation, chainid, morphoV2);
+        bytes memory creationCode = abi.encodePacked(SSTORE2_BYTECODE, abi.encode(sstore2Data));
```
MathisGD: the double abi.encode is needed?

src/libraries/IdLib.sol
```diff
@@ -0,0 +1,28 @@
...
+        return keccak256(creationCode);
+    }
+
+    function idToObligation(address morphoV2, bytes32 id) internal view returns (Obligation memory) {
+        address create2Address =
+            address(uint160(uint256(keccak256(abi.encodePacked(uint8(0xff), morphoV2, bytes32(0), id)))));
+        // The SStore2 contract code of an obligation also contains chainid and morphoV2, which are discarded.
```
MathisGD: maybe we should decode them (and not return them) for clarity? I was unsure that abi.decode would not revert on this before

src/libraries/IdLib.sol
QGarchery: yes, and it took me some time to debug this 😅

src/SStore2.sol
```diff
@@ -0,0 +1,11 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Copyright (c) 2025 Morpho Association
+pragma solidity ^0.8.0;
+
+contract SStore2 {
```
MathisGD: let's add a dev comment, as there is no real sstore2 standard

src/MorphoV2.sol
MathisGD, line 1: we can make the pr point to main directly

MathisGD:
@codex review

src/libraries/IdLib.sol
```diff
@@ -0,0 +1,28 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Copyright (c) 2025 Morpho Association
+pragma solidity ^0.8.0;
+
+import {SStore2} from "../SStore2.sol";
+import {Obligation} from "../interfaces/IMorphoV2.sol";
+
+bytes constant SSTORE2_BYTECODE =
+    hex"60806040523460845760c180380380601581609c565b9283398101906020818303126084578051906001600160401b0382116084570181601f8201121560845780516001600160401b038111608857605f601f8201601f1916602001609c565b91818352602083019360208383010111608457815f926020809301865e830101525190f35b5f80fd5b634e487b7160e01b5f52604160045260245ffd5b6040519190601f01601f191682016001600160401b0381118382101760885760405256fe";
```
peyha: If we want to really gas-opti, the bytecode can be reduced by writing in assembly, we could shrink it to a dozen bytes but we would lose some explainability

src/libraries/IdLib.sol
QGarchery: Smart to put the STOP. It the current code, even if we change the Obligation type, I think the first byte is 0, because it is the length of the `data` bytes array

src/libraries/IdLib.sol
QGarchery: duplicate of [this comment](https://github.com/morpho-org/morpho-v2/pull/328#discussion_r2754661220)

src/libraries/IdLib.sol
MathisGD: you could do a create2 in assembly, and happen the chainId and contract at the end just to have them be in the creation code (to influence the id) but not in the code of the contract. but maybe useless

src/libraries/IdLib.sol
MathisGD: looking ok

src/libraries/IdLib.sol
adhusson: this would also reduce the gas overhead if i understand this answer correctly: https://github.com/morpho-org/morpho-v2/pull/328#discussion_r2755415457

src/SStore2.sol
MathisGD, line 1: ideally the contract doesn't stay in `src` (a bit distracting ^^) but no big deal

src/libraries/IdLib.sol
QGarchery: This would save ~12k gas on write. Also it would save about 200 gas on read, although that part could be saved in other ways, we don't have to read the whole code

src/SStore2.sol
MathisGD: at this point I start to like the pure bytecode approach actually

MathisGD: works

MathisGD: wait no actually, then the solc version is not fixed anymore (and one might struggle finding back the compiled code)

src/libraries/IdLib.sol
MathisGD: 12k on write sounds great, there is no real downside except maybe this piece of assembly

src/libraries/IdLib.sol
MathisGD: are we good with the current naming?

src/SStore2.sol
QGarchery: Wdyt about putting the SStore2 contract in TickLib ?

src/libraries/IdLib.sol
MathisGD: it makes reading a bit more painful. maybe we can just comment that the beginning will always be zero and that's it?

src/SStore2.sol
QGarchery: > wait no actually, then the solc version is not fixed anymore (and one might struggle finding back the compiled code)

I don't get it, in assembly there would be no compilation at all. So the only thing that an integrator would need is the assembly prefix

MathisGD: maybe not an integrator, but someone trying to double check the bytecode of this contract. they might run into issue as this is a lib (with loose pragma)

QGarchery: Ah yes, I got confused because github switched the order of the messages, so I thought you were talking about the full assembly idea
Screenshot 2026-02-02 at 19 29 39

MathisGD: yes, so weird ^^

src/libraries/IdLib.sol
```diff
@@ -0,0 +1,56 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Copyright (c) 2025 Morpho Association
+pragma solidity ^0.8.0;
+
+import {Obligation} from "../interfaces/IMorphoV2.sol";
+
+library IdLib {
+    /// @dev Minimal creation code that returns code after the prefix as runtime bytecode.
```
adhusson: Maybe i misunderstand but since we are hardcoding offsets anyway why not do

```
// CODESIZE, PUSH0, PUSH0, CODECOPY, PUSH1 75, CODESIZE, SUB, PUSH1 75, RETURN
bytes constant DEPLOYER = hex"385f5f39604b3803604bf3";
```

and 

```
creationCode = abi.encodePacked(DEPLOYER, abi.encode(chainid, morphoV2), abi.encode(obligation));
```


edit: slightly simpler
```
// CODESIZE, PUSH0, PUSH0, CODECOPY, PUSH1 63, CODESIZE, SUB, PUSH1 63, RETURN
bytes constant DEPLOYER = hex"385f5f39603f3803603ff3";
```

and 

```
creationCode = abi.encodePacked(DEPLOYER, chainid, morphoV2, abi.encode(obligation));
```

QGarchery: ah yes nice, did something similar

src/libraries/IdLib.sol
```diff
@@ -0,0 +1,49 @@
...
```
MathisGD: right?
```suggestion
    /// @dev Minimal creation code that returns code after the prefix as runtime bytecode (except the first 2 words).
```

QGarchery: yes this is updated now (note that it skips 52 bytes actually)

src/libraries/IdLib.sol
```diff
@@ -0,0 +1,49 @@
...
+        return keccak256(creationCode(obligation, chainId, morphoV2));
+    }
+
+    function idToObligation(bytes32 id, address morphoV2) internal view returns (Obligation memory) {
+        address create2Address =
+            address(uint160(uint256(keccak256(abi.encodePacked(uint8(0xff), morphoV2, bytes32(0), id)))));
+        return abi.decode(create2Address.code, (Obligation));
+    }
+
+    /// @dev The contract code begins with 0x00 (STOP), because the first word is the offset of the obligation.
```
MathisGD, line 41:
```suggestion
    /// @dev The contract code begins with 0x00 (STOP), because the first word is the length of the obligation.
```

src/libraries/IdLib.sol
```diff
@@ -0,0 +1,49 @@
...
+    /// 39        CODECOPY        [len]              mem[0:len] <- code[63:63+len]
+    /// 5f        PUSH0           [0, len]           return offset = 0
+    /// f3        RETURN          []                 mem[0:len] is returned
+    function creationCode(Obligation memory obligation, uint256 chainId, address morphoV2)
+        internal
+        pure
+        returns (bytes memory)
+    {
+        bytes memory prefix = hex"603f380380603f5f395ff3";
+        return abi.encodePacked(prefix, chainId, morphoV2, abi.encode(obligation));
```
MathisGD, line 27: I think that this would work too, but then the contract wouldn't always start with zero?
```suggestion
        return abi.encodePacked(prefix, chainId, morphoV2, obligation);
```

src/MorphoV2.sol
```diff
@@ -464,6 +465,8 @@ contract MorphoV2 is IMorphoV2 {
             obligationState[id].created = true;
             obligationState[id].fees = defaultFees[obligation.loanToken];
```
```suggestion
```

src/MorphoV2.sol
```diff
@@ -491,12 +494,8 @@ contract MorphoV2 is IMorphoV2 {
         return obligationState[id].fees;
     }
 
-    function toId(Obligation memory obligation) public view returns (bytes32) {
-        return keccak256(abi.encode(block.chainid, address(this), obligation));
-    }
-
     function isHealthy(Obligation memory obligation, address borrower) public view returns (bool) {
-        bytes32 id = toId(obligation);
+        bytes32 id = IdLib.toId(obligation, block.chainid, address(this));
```
MathisGD: isHealthy can take as input id

MathisGD (approved):
🔥

src/libraries/IdLib.sol
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    function idToObligation(bytes32 id, address morphoV2) internal view returns (Obligation memory) {
+        address create2Address =
+            address(uint160(uint256(keccak256(abi.encodePacked(uint8(0xff), morphoV2, bytes32(0), id)))));
+        return abi.decode(create2Address.code, (Obligation));
+    }
+
+    /// @dev The contract code begins with 0x00 (STOP), because the first word is the offset of the obligation.
+    function sstore2(Obligation memory obligation) internal {
```
MathisGD: think it's useful (as it's a non standard sstore2)
```suggestion
    /// @dev Deploys a contract with creationCode = prefix|chainId|morphoV2|abi.encode(obligation), and as runtimeCode = abi.encode(obligation)
    function sstore2(Obligation memory obligation) internal {
```

src/libraries/IdLib.sol
QGarchery: No it's not the length, it's the offset in memory: it's always 0x20

src/libraries/IdLib.sol
QGarchery: I get a compiler error when I try (not sure what it would mean if it did not have an error)

MathisGD: not have the size at the beginning? but ok, fine like that

src/libraries/IdLib.sol
MathisGD: offset in memory? but it's not in memory? I don't get it

src/libraries/IdLib.sol
QGarchery: added only the runtime bytecode, as it is what matters, and the creation code is clear in the `creationCode` function

src/libraries/IdLib.sol
QGarchery: sorry it's not the offset in memory, it's the offset that is added when abi encoded (in the end we store the whole thing in memory, hence the confusion). This is what is called the `head` for dynamic types [here](https://docs.soliditylang.org/en/v0.8.33/abi-spec.html#formal-specification-of-the-encoding).  Essentially, this is where to find the data in the encoded bytes for dynamic types

peyha (approved):
LGTM, very nice improvement

src/libraries/IdLib.sol
QGarchery: see [this](https://github.com/morpho-org/morpho-v2/pull/328#discussion_r2758293046), but indeed we could  skip a word. Implemented [here](https://github.com/morpho-org/morpho-v2/pull/333)

src/libraries/IdLib.sol
MathisGD: can we add a test that this is actually true?

src/libraries/IdLib.sol
adhusson: [moved comment here](https://github.com/morpho-org/morpho-v2/pull/333#discussion_r2758842217)

src/libraries/IdLib.sol
```diff
@@ -0,0 +1,50 @@
...
+
+    function idToObligation(bytes32 id, address morphoV2) internal view returns (Obligation memory) {
+        address create2Address =
+            address(uint160(uint256(keccak256(abi.encodePacked(uint8(0xff), morphoV2, bytes32(0), id)))));
+        return abi.decode(create2Address.code, (Obligation));
+    }
+
+    /// @dev Deploys a contract with runtime code = abi.encode(obligation)
+    /// @dev The contract code begins with 0x00 (STOP), because the first word is the offset of the obligation.
+    function sstore2(Obligation memory obligation) internal {
```
adhusson: what about no sstore2 at all, but `storeInCode` or `storeObligationInCode` or something like this.

MathisGD: works for me

adhusson (approved):

src/libraries/IdLib.sol
MathisGD: done

PR #329, opened 2026-01-30
test: ticks in python

MathisGD:
image

imo worth keeping in the repo


test/test_ticks.py
```diff
@@ -0,0 +1,33 @@
...
+    return 1e18 / (1 + 1.025 ** (TICK_RANGE / 2 - tick))
+
+def sol_price(tick):
+    LN_ONE_PLUS_DELTA = 0.024692612590371501
+    def wexp(x):
+        if x < 0: return int(1e36) // wexp(-x)
+        ln2, quotient = 0.693147180559945309, int((x + 0.346573590279972654) / 0.693147180559945309)
+        remainder = x - quotient * ln2
+        return int((1 + remainder + remainder*remainder/2 + remainder*remainder*remainder/6) * 1e18) << quotient
+    def div_half_down_unchecked(x, divisor): return (x + divisor // 2) // divisor
```
QGarchery:
```suggestion
    def div_half_down_unchecked(x, divisor): return (x + (divisor - 1) // 2) // divisor
```

test/test_ticks.py
QGarchery, line 1: This is very convincing, because for "real" prices, meaning prices > 0.001 (corresponding already to ~100_000% return) we have less than 0.1% relative difference.

I would be more confident with a test that reuses the actual implementation of `tickToPrice` though, this could be done with a ffi call ?

QGarchery (approved):

test/test_ticks.py
QGarchery, line 1: Another thing we can check:
```python
n = len(data)

prev_prices = [0] * n
prev_price = 0
for i in range(1, n):
    prev_prices[i] = prev_price
    sol = data[i][2]
    if sol > prev_prices[i - 1]:
        prev_price = prev_prices[i - 1]

next_prices = [10**18] * n
next_price = 10**18
for i in range(n - 2, -1, -1):
    next_prices[i] = next_price
    sol = data[i][2]
    if sol < next_prices[i + 1]:
        next_price = next_prices[i + 1]

# Check that the exact price is always between the previous and next sol price
for i in range(1, n - 1):
    exact = data[i][1]
    assert prev_prices[i] <= exact <= next_prices[i]
```

MathisGD: indeed, will try something

MathisGD: but this is not true because of the 1e13 rounding no?

test/TickLibTest.sol
```diff
@@ -76,4 +77,53 @@ contract TickLibTest is Test {
...
+
+            uint256 absErrorWad = absDiff(solPrice, exactPrice);
+            maxAbsErrorWad = max(maxAbsErrorWad, absErrorWad);
+            totalAbsErrorWad += absErrorWad;
+            uint256 relErrorWad = absDiff(solPrice, exactPrice) * 1e18 / exactPrice;
+            totalRelErrorWad += relErrorWad;
+            maxRelErrorWad = max(maxRelErrorWad, relErrorWad);
+
+            assertLe(absErrorWad, 0.00015e18, string.concat("Tick ", vm.toString(tick), " error exceeds 1.5 bps"));
+            if (tick > TICK_RANGE / 2) {
```
QGarchery:
```suggestion
            if (solPrice > 0.01e18) {
```

test/TickLibTest.sol
```diff
@@ -76,4 +77,53 @@ contract TickLibTest is Test {
...
+                assertLe(relErrorWad, 0.001e18, string.concat("Tick ", vm.toString(tick), " error exceeds 0.1%"));
+            }
+
+            // Check exact price is bracketed by adjacent sol prices (only where prices vary per-tick)
```
QGarchery, line 114: It would be nice to remove the assumption in the parenthesis, but it makes the test much more complicated, so maybe not worth it

QGarchery (approved):

test/TickLibTest.sol
```diff
@@ -76,4 +77,54 @@ contract TickLibTest is Test {
...
+
+            uint256 absErrorWad = absDiff(solPrice, exactPrice);
+            maxAbsErrorWad = max(maxAbsErrorWad, absErrorWad);
+            totalAbsErrorWad += absErrorWad;
+            uint256 relErrorWad = absDiff(solPrice, exactPrice) * 1e18 / exactPrice;
+            totalRelErrorWad += relErrorWad;
+            maxRelErrorWad = max(maxRelErrorWad, relErrorWad);
+
+            assertLe(absErrorWad, 0.00015e18, string.concat("Tick ", vm.toString(tick), " error exceeds 1.5 bps"));
+            if (solPrice > 0.01e18) {
```
peyha, line 110: corresponds to tick > 308, so 1/3 of ticks are missing from this but it's ok because the price is fairly small

peyha (approved):
all good for me, verified the json

PR #330, opened 2026-01-30
feat: max ttm

MathisGD:


src/MorphoV2.sol
```diff
@@ -122,6 +125,14 @@ contract MorphoV2 is IMorphoV2 {
         emit EventsLib.SetTradingFeeRecipient(recipient);
     }
 
+    /// @dev Sets the maximum time to maturity for new obligations. Can only be increased.
+    function setMaxTimeToMaturity(uint256 newMaxTimeToMaturity) external {
+        require(msg.sender == owner, "Only owner");
+        require(newMaxTimeToMaturity >= maxTimeToMaturity, "can only increase");
```
peyha, line 131: Why do we need this ? 
To prevent the owner from blocking the protocol by setting max TTM to 0 ? The worst case scenario in this config would be that we slowly increase the max TTM until some high value and then a new loan token starts being used so users create a lot of short-term bonds which have zero interest fee (depending on how we implement the interest fee)
Another option would be to require max TTM to be higher than some value (e.g. an year).

MathisGD:
to reopen

PR #333, opened 2026-02-03
[id to obligation] skip word byte sstore2

QGarchery:
[Motivation](https://github.com/morpho-org/morpho-v2/pull/328#discussion_r2758181498)

### DIFF
--- a/src/libraries/IdLib.sol
+++ b/src/libraries/IdLib.sol
@@ -24,7 +24,16 @@ library IdLib {
         returns (bytes memory)
     {
         bytes memory prefix = hex"603f380380603f5f395ff3";
-        return abi.encodePacked(prefix, chainId, morphoV2, abi.encode(obligation));
+        bytes memory encodedObligation = abi.encode(obligation);
+        bytes memory runtimeCode;
+        uint256 runtimeCodeLength = encodedObligation.length - 32;
+        assembly ("memory-safe") {
+            runtimeCode := mload(0x40)
+            mcopy(add(runtimeCode, 0x20), add(encodedObligation, 0x40), runtimeCodeLength)
+            mstore(runtimeCode, runtimeCodeLength)
+            mstore(0x40, add(runtimeCode, runtimeCodeLength))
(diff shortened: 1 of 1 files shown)


src/libraries/IdLib.sol
QGarchery, line 1: Not convinced because of the added complexity

src/libraries/IdLib.sol
```diff
@@ -24,7 +24,16 @@ library IdLib {
...
+        bytes memory encodedObligation = abi.encode(obligation);
+        bytes memory runtimeCode;
+        uint256 runtimeCodeLength = encodedObligation.length - 32;
+        assembly ("memory-safe") {
+            runtimeCode := mload(0x40)
+            mcopy(add(runtimeCode, 0x20), add(encodedObligation, 0x40), runtimeCodeLength)
+            mstore(runtimeCode, runtimeCodeLength)
+            mstore(0x40, add(runtimeCode, runtimeCodeLength))
+        }
+        return abi.encodePacked(prefix, chainId, morphoV2, runtimeCode);
```
adhusson, line 36: i think the contract now only starts with stop because the high bytes of addresses (here loanToken) are 0. It's becoming a bit ugly, it relies on the struct definition. But we could add a test that it always does so adding a field that can be nonzero at the beginning would be detected.

Also instead of this assembly why not just copy from (63+32) to (codesize-63-32), it strips the first word.

src/libraries/IdLib.sol
peyha: not convinced either, do we know how much gas is saved ? we shave off one byte from the deployed contract so about 200 gas but the cost from handling bytes in idToObligation might be larger than that

QGarchery: it should be ~6k gain on deployment (200 * 32)

QGarchery:
closing for now, we may come back later with more optimized versions such as [tight packing](https://morpholabs.slack.com/archives/C08A40KKN9W/p1770131186092029)

PR #334, opened 2026-02-03
[RCF] refactor

MathisGD:


test/LiquidationTest.sol
```diff
@@ -157,6 +157,7 @@ contract LiquidationTest is BaseTest {
         collateralize(obligation, borrower, units);
         setupObligation(obligation, units);
         Oracle(obligation.collaterals[0].oracle).setPrice(1e36 - 1);
+        vm.warp(obligation.maturity + TIME_TO_MAX_LIF); // Warp to post-maturity to bypass recovery close factor.
```
MathisGD, line 165: idk why it was not reverting before

peyha: it was probably reverting in `originalDebt - repaidUnits` from the recovery close factor computation `require(originalDebt - repaidUnits >= newMaxDebt, "recovery close factory violated")` so not exactly where we wanted it to revert but with the same error (arithmetic)

peyha: too bad we can't specify the FP of where we want the revert to happen on foundry

src/MorphoV2.sol
```diff
@@ -370,6 +370,10 @@ contract MorphoV2 is IMorphoV2 {
 
         uint256 originalDebt = debtOf[id][borrower];
         require(block.timestamp > obligation.maturity || originalDebt > maxDebt, "position is not liquidatable");
+        uint256 maxSeizableAssets = block.timestamp > obligation.maturity || liquidatedCollateralPrice == 0
+            ? type(uint256).max
+            : (originalDebt - maxDebt).mulDivDown(ORACLE_PRICE_SCALE, liquidatedCollateralPrice)
```
peyha, line 378: I don't have the same formula
The initial health condition is maxDebt=sum c_i l_i p_i > debt where (c_i, l_i, p_i) are respectively collateral, lltv and price of collateral i.
Supposing that j is the seized collateral and that s is the quantity seized and that r is the repaid debt, after liquidation, the health condition becomes maxDebt-s l_j p_j > debt - r which gives originalDebt-r-maxDebt<s l_i p_i
The corresponding maxSeizableAssets would be `(originalDebt - maxDebt - repaidUnits).mulDivDown(ORACLE_PRICE_SCALE, liquidatedCollateralPrice)`
It's possible to express repaidUnits as a function of the seized collateral but it adds the lif to the inequation

PR #339, opened 2026-02-04
Collateral level min collateral quoted in collat

peyha:
Much simpler implementation than #320 but each collateral should now have a min collateral so part of the complexity is pushed to the product

Remaining questions:
- Should it be possible to fall under min collateral through liquidation ?
- Should the contract require specific min collateral values (10^k,..) ?

### DIFF
--- a/src/MorphoV2.sol
+++ b/src/MorphoV2.sol
@@ -320,6 +320,19 @@ contract MorphoV2 is IMorphoV2 {
 
         collateralOf[id][onBehalf][collateral] += assets;
 
+        uint256 collateralIndex = 0;
+        for (uint256 i = 0; i < obligation.collaterals.length; i++) {
+            if (obligation.collaterals[i].token == collateral) {
+                collateralIndex = i;
+                break;
+            }
+        }
+        require(
+            (collateralOf[id][onBehalf][collateral] == 0)
+                || (collateralOf[id][onBehalf][collateral] >= obligation.collaterals[collateralIndex].minCollateral),
+            "Collateral below min"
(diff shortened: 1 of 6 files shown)


src/MorphoV2.sol
```diff
@@ -320,6 +320,19 @@ contract MorphoV2 is IMorphoV2 {
 
         collateralOf[id][onBehalf][collateral] += assets;
 
+        uint256 collateralIndex = 0;
+        for (uint256 i = 0; i < obligation.collaterals.length; i++) {
+            if (obligation.collaterals[i].token == collateral) {
+                collateralIndex = i;
+                break;
+            }
+        }
```
QGarchery, line 329:
```suggestion
        for (; collateralIndex < obligation.collaterals.length && obligation.collaterals[collateralIndex].token != collateral; collateralIndex++) {}
```

QGarchery: other option: pass the index as an extra argument, same as in the liquidation

peyha:
> It's not enforced at protocol level that minCollat should be a power of, say, 10?
We could enforce it in `touchObligation`, yes, adding it as a remaining question

src/MorphoV2.sol
MathisGD: it can replace the collateralToken address

MathisGD: in favor of this

peyha:
Closing it in favor of #342  because computing a min collat for each collateral would be tremendous work for front-end/curators

PR #342, opened 2026-02-05
Obligation level min collat quoted in loan without liquidation

peyha:
New min collat implementation. 

A few notes
- `supplyCollateral` and `withdrawCollateral` can no longer be done with a collateral that is not in the `Obligation`
- we could say that in this implementation, the min collateral enforcement is 'weak' because it's only checked during `supplyCollateral` and `withdrawCollateral` but it could fall below min collateral later through an oracle price update and no liquidation mechanism is currently mitigating this (#343)

see rationale: https://morpholabs.slack.com/archives/C02N7CZ088N/p1770296213352059

### DIFF
--- a/src/MorphoV2.sol
+++ b/src/MorphoV2.sol
@@ -320,6 +320,21 @@ contract MorphoV2 is IMorphoV2 {
 
         collateralOf[id][onBehalf][collateral] += assets;
 
+        uint256 collateralIndex = 0;
+        for (
+            ;
+            collateralIndex < obligation.collaterals.length
+                && obligation.collaterals[collateralIndex].token != collateral;
+            collateralIndex++
+        ) {}
+        uint256 collateralPrice = IOracle(obligation.collaterals[collateralIndex].oracle).price();
+        uint256 collateralValue = collateralOf[id][onBehalf][collateral].mulDivDown(collateralPrice, ORACLE_PRICE_SCALE);
+
+        require(
+            collateralOf[id][onBehalf][collateral] == 0 || collateralValue >= obligation.minCollateral,
+            "Below min collateral"
+        );
+
         emit EventsLib.SupplyCollateral(msg.sender, id, collateral, assets, onBehalf);
 
         SafeTransferLib.safeTransferFrom(collateral, msg.sender, address(this), assets);
@@ -333,6 +348,20 @@ contract MorphoV2 is IMorphoV2 {
         collateralOf[id][onBehalf][collateral] -= assets;
 
         require(isHealthy(obligation, id, onBehalf), "Unhealthy borrower");
+
+        uint256 collateralIndex = 0;
+        for (
+            ;
+            collateralIndex < obligation.collaterals.length
+                && obligation.collaterals[collateralIndex].token != collateral;
+            collateralIndex++
+        ) {}
+        uint256 collateralPrice = IOracle(obligation.collaterals[collateralIndex].oracle).price();
+        uint256 collateralValue = collateralOf[id][onBehalf][collateral].mulDivDown(collateralPrice, ORACLE_PRICE_SCALE);
+        require(
+            collateralOf[id][onBehalf][collateral] == 0 || collateralValue >= obligation.minCollateral,
+            "Below min collateral"
+        );
 
         emit EventsLib.WithdrawCollateral(msg.sender, id, collateral, assets, onBehalf);
 
--- a/src/MorphoV2.sol
+++ b/src/MorphoV2.sol
@@ -317,40 +317,40 @@ contract MorphoV2 is IMorphoV2 {
         external
     {
         bytes32 id = touchObligation(obligation);
-        address collateral = obligation.collaterals[collateralIndex].token;
+        address collateralToken = obligation.collaterals[collateralIndex].token;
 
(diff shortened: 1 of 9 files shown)


src/MorphoV2.sol
MathisGD: to avoid looping
```suggestion
function supplyCollateral(Obligation memory obligation, uint256 collateralIndex, uint256 assets, address onBehalf)
external
```

```suggestion
function withdrawCollateral(Obligation memory obligation, uint256 collateralIndex, uint256 assets, address onBehalf)
external
```

peyha: done

src/MorphoV2.sol
```diff
@@ -313,27 +313,44 @@ contract MorphoV2 is IMorphoV2 {
...
+    function supplyCollateral(Obligation memory obligation, uint256 collateralIndex, uint256 assets, address onBehalf)
         external
     {
         bytes32 id = touchObligation(obligation);
+        address collateral = obligation.collaterals[collateralIndex].token;
 
         collateralOf[id][onBehalf][collateral] += assets;
 
+        uint256 collateralPrice = IOracle(obligation.collaterals[collateralIndex].oracle).price();
+        uint256 collateralValue = collateralOf[id][onBehalf][collateral].mulDivDown(collateralPrice, ORACLE_PRICE_SCALE);
```
MathisGD: prevents to redo the keccak to find the slot
```suggestion
        uint256 _collateralOf = collateralOf[id][onBehalf][collateral];
```

src/MorphoV2.sol
```diff
@@ -313,27 +313,44 @@ contract MorphoV2 is IMorphoV2 {
         SafeTransferLib.safeTransferFrom(obligation.loanToken, msg.sender, address(this), obligationUnits);
     }
 
-    function supplyCollateral(Obligation memory obligation, address collateral, uint256 assets, address onBehalf)
+    function supplyCollateral(Obligation memory obligation, uint256 collateralIndex, uint256 assets, address onBehalf)
         external
     {
         bytes32 id = touchObligation(obligation);
+        address collateral = obligation.collaterals[collateralIndex].token;
```
MathisGD: to match with the naming elsewhere
```suggestion
        address collateralToken = obligation.collaterals[collateralIndex].token;
```

src/interfaces/IMorphoV2.sol
```diff
@@ -7,6 +7,7 @@ struct Obligation {
     // Must be sorted by address.
     Collateral[] collaterals;
     uint256 maturity;
+    uint256 minCollateral; // quoted in loan token
```
MathisGD: let's try to write what low level guarantee this gives to people, like
```suggestion
    // minimum collateral value (in loan token) that can be left after a supplyCollateral/withdrawCollateral.
    uint256 minCollateral;
```

src/MorphoV2.sol
peyha: thought solc would optimize it, I'll still fix this

MathisGD: no it doesn't

src/MorphoV2.sol
```diff
@@ -313,30 +313,44 @@ contract MorphoV2 is IMorphoV2 {
...
+        collateralOf[id][onBehalf][collateralToken] -= assets;
 
         require(isHealthy(obligation, id, onBehalf), "Unhealthy borrower");
 
-        emit EventsLib.WithdrawCollateral(msg.sender, id, collateral, assets, onBehalf);
+        uint256 collateralPrice = IOracle(obligation.collaterals[collateralIndex].oracle).price();
+        uint256 _collateralOf = collateralOf[id][onBehalf][collateralToken];
+        uint256 collateralValue = _collateralOf.mulDivDown(collateralPrice, ORACLE_PRICE_SCALE);
+
+        require(_collateralOf == 0 || collateralValue >= obligation.minCollateral, "Below min collateral");
```
MathisGD: avoids calling the oracle if the user has no debt. let's add a dev comment that it's the case, such that we notice if we break it.
```suggestion
        uint256 _collateralOf = collateralOf[id][onBehalf][collateralToken];
        require(_collateralOf == 0 || _collateralOf.mulDivDown( IOracle(obligation.collaterals[collateralIndex].oracle).price(), ORACLE_PRICE_SCALE) >= obligation.minCollateral, "Below min collateral");
```
(for consistency, I would do the same in supplyCollat)

src/MorphoV2.sol
```diff
@@ -313,30 +313,50 @@ contract MorphoV2 is IMorphoV2 {
...
+        );
+
+        emit EventsLib.SupplyCollateral(msg.sender, id, collateralToken, assets, onBehalf);
 
-        SafeTransferLib.safeTransferFrom(collateral, msg.sender, address(this), assets);
+        SafeTransferLib.safeTransferFrom(collateralToken, msg.sender, address(this), assets);
     }
 
-    function withdrawCollateral(Obligation memory obligation, address collateral, uint256 assets, address onBehalf)
+    /// @dev This function does not call the oracle if the collateral balance is 0 after the update.
```
MathisGD, line 338: and the borrower has no debt

MathisGD (approved):
nice 🔥

PR #346, opened 2026-02-05
[obligation packing] pack stored obligation in solidity

### LINKED ISSUE #337: [gas] optimize sstore2
the encoding could be custom, there is a lot to win

https://morpholabs.slack.com/archives/C08A40KKN9W/p1770131186092029

adhusson:
version of #341 with simpler code for `pack`

### DIFF
--- a/src/libraries/IdLib.sol
+++ b/src/libraries/IdLib.sol
@@ -67,6 +67,8 @@ library IdLib {
         }
     }
 
+    /// @dev Unpacks a packed representation of the obligation.
+    /// @dev Should only be used on data known to be a packed obligation.
     function unpack(bytes memory data) internal pure returns (Obligation memory obligation) {
         require(data.length > 0, "empty data");
         unchecked {
--- a/src/libraries/IdLib.sol
+++ b/src/libraries/IdLib.sol
@@ -55,40 +55,15 @@ library IdLib {
     function pack(Obligation memory obligation) internal pure returns (bytes memory result) {
         require(obligation.maturity <= type(uint48).max, "maturity too large");
         require(obligation.collaterals.length <= type(uint8).max, "collateral count too large");
-        Collateral[] memory collaterals = obligation.collaterals;
-        uint256 len = collaterals.length;
 
-        uint256 ptr;
-        assembly ("memory-safe") {
-            result := mload(0x40)
-            let size := add(28, mul(len, 48))
-            mstore(result, size)
+        result = abi.encodePacked(
+            bytes1(0x00), obligation.loanToken, uint48(obligation.maturity), uint8(obligation.collaterals.length)
+        );
 
-            ptr := add(result, 32)
-            // [stop, 1 byte][loanToken, 20 bytes][maturity, 6 bytes][collateral count, 1 byte]
-            let header := or(or(shl(88, mload(obligation)), shl(40, mload(add(obligation, 0x40)))), shl(32, len))
-            mstore(ptr, header)
-            ptr := add(ptr, 28)
-        }
-
-        for (uint256 i = 0; i < len; i++) {
-            Collateral memory c = collaterals[i];
-            require(c.lltv <= type(uint64).max, "lltv too large");
-
-            assembly ("memory-safe") {
-                // [token, 20 bytes][lltv, 8 bytes]
-                mstore(ptr, or(shl(96, mload(c)), shl(32, mload(add(c, 0x20)))))
-                ptr := add(ptr, 28)
-
-                // [oracle, 20 bytes]
-                mstore(ptr, shl(96, mload(add(c, 0x40))))
-                ptr := add(ptr, 20)
-            }
-        }
-
-        assembly ("memory-safe") {
-            // Round up to the next 32-byte word for the free memory pointer.
-            mstore(0x40, and(add(ptr, 31), not(31)))
+        for (uint256 i = 0; i < obligation.collaterals.length; i++) {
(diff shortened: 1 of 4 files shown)


src/libraries/IdLib.sol
```diff
@@ -47,4 +47,59 @@ library IdLib {
...
+        );
+
+        for (uint256 i = 0; i < obligation.collaterals.length; i++) {
+            Collateral memory collateral = obligation.collaterals[i];
+            require(collateral.lltv <= type(uint64).max, "lltv too large");
+            result = abi.encodePacked(result, collateral.token, uint64(collateral.lltv), collateral.oracle);
+        }
+    }
+
+    function unpack(bytes memory data) internal pure returns (Obligation memory obligation) {
```
MathisGD, line 72: didn't have in mind that this still required some assembly for unpack..

adhusson: redid it isolating the assembly (don't know if it's much clearer)

src/libraries/IdLib.sol
```diff
@@ -47,4 +47,47 @@ library IdLib {
...
+            for (uint256 i = 0; i < len; i++) {
+                uint256 offset = 28 + i * 48;
+                obligation.collaterals[i].token = address(uint160(get(data, offset) >> 96));
+                obligation.collaterals[i].lltv = uint64(get(data, offset + 20) >> 192);
+                obligation.collaterals[i].oracle = address(uint160(get(data, offset + 28) >> 96));
+            }
+        }
+    }
+
+    function get(bytes memory data, uint256 offset) private pure returns (uint256 w) {
```
MathisGD: wdyt
```suggestion
    function get(bytes memory data, uint256 offset, uint256 size) private pure returns (uint256 w) {
```

src/libraries/IdLib.sol
MathisGD: yes I like it better

src/libraries/IdLib.sol
adhusson: you mean something like this?

```solidity
function get(bytes memory data, uint256 offset, uint256 size) private pure returns (uint256 w) {
    uint shift = 256-size;
    assembly ("memory-safe") {
        w := shr(shift,mload(add(add(data, 32), offset)))
    }
}

/// ...

obligation.collaterals[i].oracle = address(uint160(get(data, offset + 28,160)));
```

src/libraries/IdLib.sol
```diff
@@ -47,4 +47,59 @@ library IdLib {
...
+
+        for (uint256 i = 0; i < obligation.collaterals.length; i++) {
+            Collateral memory collateral = obligation.collaterals[i];
+            require(collateral.lltv <= type(uint64).max, "lltv too large");
+            result = abi.encodePacked(result, collateral.token, uint64(collateral.lltv), collateral.oracle);
+        }
+    }
+
+    function unpack(bytes memory data) internal pure returns (Obligation memory obligation) {
+        require(data.length > 0, "empty data");
```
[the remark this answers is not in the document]
adhusson: added a warning

src/libraries/IdLib.sol
```diff
@@ -47,4 +47,59 @@ library IdLib {
...
+
+        result = abi.encodePacked(
+            bytes1(0x00), obligation.loanToken, uint48(obligation.maturity), uint8(obligation.collaterals.length)
+        );
+
+        for (uint256 i = 0; i < obligation.collaterals.length; i++) {
+            Collateral memory collateral = obligation.collaterals[i];
+            require(collateral.lltv <= type(uint64).max, "lltv too large");
+            result = abi.encodePacked(result, collateral.token, uint64(collateral.lltv), collateral.oracle);
+        }
```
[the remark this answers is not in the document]
adhusson: - At 1 collateral any readable version I tried is more expensive than this.
- With more collaterals you can be cheaper but only significantly if you go full assembly which is https://github.com/morpho-org/morpho-v2/pull/341.

src/libraries/IdLib.sol
MathisGD: yes

src/libraries/IdLib.sol
```diff
@@ -47,4 +47,50 @@ library IdLib {
...
+                obligation.collaterals[i].lltv = uint64(get(data, offset + 20, 64));
+                obligation.collaterals[i].oracle = address(uint160(get(data, offset + 28, 160)));
+            }
+        }
+    }
+
+    function get(bytes memory data, uint256 offset, uint256 size) private pure returns (uint256 w) {
+        uint256 shift = 256 - size;
+        assembly ("memory-safe") {
+            w := shr(shift, mload(add(add(data, 32), offset)))
```
MathisGD, line 93: actually the shift is not necessary, right? because then we cast to the desired type which does the shift. but maybe it's more prone to errors to not do it

QGarchery: You have to shift to both remove the junk at the right (leat significant bits) whereas casting removes bits at the left, and also to have it at the right place. So for an address, you are getting: 
```
AAAAAAAAAAAAAAAAAAAAxxxxxxxxxxxx
^^^^^^^^^^^^^^^^^^^^
   address bytes
```
when you want
```
000000000000AAAAAAAAAAAAAAAAAAAA
            ^^^^^^^^^^^^^^^^^^^^
                address bytes
```

MathisGD: oh yes!

adhusson:
closing because too complex for the win

PR #348, opened 2026-02-05
[Certora] Solvency

jhoenicke:
This adds the specification that proves token balances cover collateral and withdrawable.

Assumptions:
* ERC-20 tokens behave naturally on transfer: no fee taking for sender or receiver, no rebalancing.
* This doesn't check that no approvals are set, only that the transfer/transferFrom the contract calls respect the rules.
* The main invariant needs the assumption that the obligation.maker and taker is never the MorphoV2 contract itself.

For flash loans and liquidate we hook on the callback to tell the formal verification about the size of the flash loan.

We also prove that flash loans are paid back.

### DIFF
--- a/certora/helpers/FlashLiquidateCallback.sol
+++ b/certora/helpers/FlashLiquidateCallback.sol
@@ -2,16 +2,13 @@
 // Copyright (c) 2025 Morpho Association
 pragma solidity ^0.8.0;
 
-import {
-    Seizure
-} from "../../src/interfaces/IMorphoV2.sol";
+import {Seizure} from "../../src/interfaces/IMorphoV2.sol";
 
 interface DummyCallback {
     function doSomething() external;
 }
 
 contract FlashLiquidateCallback {
-
     function startFlashloan(address token, uint256 amount) internal {
         // Dummy function to insert the flashloan logic in the spec.
     }
@@ -28,7 +25,7 @@ contract FlashLiquidateCallback {
         // Dummy function to insert the flashloan logic in the spec.
     }
 
-    function onLiquidate(Seizure[] memory seizures, address borrower, address liquidator, bytes memory data) external {
+    function onLiquidate(Seizure[] memory seizures, address, address, bytes memory data) external {
         uint256 totalAmount;
         for (uint256 i = 0; i < seizures.length; i++) {
             totalAmount += seizures[i].repaid;
--- a/certora/helpers/FlashLiquidateCallback.sol
+++ b/certora/helpers/FlashLiquidateCallback.sol
@@ -4,8 +4,8 @@ pragma solidity ^0.8.0;
 
 import {Seizure} from "../../src/interfaces/IMorphoV2.sol";
 
-interface DummyCallback {
-    function doSomething() external;
+interface IHavoc {
+    function havoc() external;
 }
 
 contract FlashLiquidateCallback {
@@ -32,14 +32,14 @@ contract FlashLiquidateCallback {
         }
         startFlashloanForLiquidity(totalAmount);
         address account = abi.decode(data, (address));
-        (DummyCallback(account)).doSomething();
+        IHavoc(account).havoc();
         endFlashloanForLiquidity(totalAmount);
     }
 
     function onFlashLoan(address token, uint256 amount, bytes calldata data) external {
         startFlashloan(token, amount);
         address account = abi.decode(data, (address));
-        (DummyCallback(account)).doSomething();
+        IHavoc(account).havoc();
         endFlashloan(token, amount);
     }
 }
--- a/certora/helpers/FlashLiquidateCallback.sol
+++ b/certora/helpers/FlashLiquidateCallback.sol
@@ -2,7 +2,7 @@
 // Copyright (c) 2025 Morpho Association
 pragma solidity ^0.8.0;
 
-import {Seizure} from "../../src/interfaces/IMorphoV2.sol";
+import {Obligation} from "../../src/interfaces/IMorphoV2.sol";
 
 interface IHavoc {
     function havoc() external;
@@ -17,23 +17,18 @@ contract FlashLiquidateCallback {
         // Dummy function to insert the flashloan logic in the spec.
     }
 
-    function startFlashloanForLiquidity(uint256 amount) internal {
-        // Dummy function to insert the flashloan logic in the spec.
-    }
-
-    function endFlashloanForLiquidity(uint256 amount) internal {
-        // Dummy function to insert the flashloan logic in the spec.
-    }
-
-    function onLiquidate(Seizure[] memory seizures, address, address, bytes memory data) external {
-        uint256 totalAmount;
-        for (uint256 i = 0; i < seizures.length; i++) {
-            totalAmount += seizures[i].repaid;
-        }
-        startFlashloanForLiquidity(totalAmount);
+    function onLiquidate(
+        Obligation memory obligation,
+        uint256 collateralIndex,
+        uint256 seizedAssets,
+        uint256 repaidUnits,
+        address borrower,
+        bytes memory data
+    ) external {
+        startFlashloan(obligation.loanToken, repaidUnits);
         address account = abi.decode(data, (address));
         IHavoc(account).havoc();
-        endFlashloanForLiquidity(totalAmount);
+        endFlashloan(obligation.loanToken, repaidUnits);
     }
 
     function onFlashLoan(address token, uint256 amount, bytes calldata data) external {
--- a/certora/helpers/FlashLiquidateCallback.sol
+++ b/certora/helpers/FlashLiquidateCallback.sol
@@ -1,0 +1,48 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Copyright (c) 2025 Morpho Association
+pragma solidity ^0.8.0;
+
+import {
+    Seizure
+} from "../../src/interfaces/IMorphoV2.sol";
+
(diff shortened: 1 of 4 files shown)


QGarchery (approved):

certora/helpers/FlashLiquidateCallback.sol
```diff
@@ -0,0 +1,40 @@
...
+        Obligation memory obligation,
+        uint256 collateralIndex,
+        uint256 seizedAssets,
+        uint256 repaidUnits,
+        address borrower,
+        bytes memory data
+    ) external {
+        startFlashloan(obligation.loanToken, repaidUnits);
+        address account = abi.decode(data, (address));
+        IHavoc(account).havoc();
```
MathisGD, line 30: what is the havoc call useful for?

certora/specs/Solvency.spec
```diff
@@ -0,0 +1,155 @@
...
+    function IdLib.toId(MorphoV2.Obligation memory obligation, uint256 chainId, address morphoV2) internal returns (bytes32) => CVL_toId(obligation, chainId, morphoV2);
+
+    // Hook on callbacks, this adds no assumption: see FlashLiquidateCallback.sol and the summaries below.
+    function _.onFlashLoan(address token, uint256 amount, bytes data) external => DISPATCHER(true);
+    function _.onLiquidate(MorphoV2.Obligation obligation, uint256 collateralIndex, uint256 seizedAssets, uint256 repaidUnits, address borrower, bytes data) external => DISPATCHER(true);
+    function FlashLiquidateCallback.startFlashloan(address token, uint256 amount) internal => CVL_flashLoanStart(token, amount);
+    function FlashLiquidateCallback.endFlashloan(address token, uint256 amount) internal => CVL_flashLoanEnd(token, amount);
+
+    // Assume ERC20 tokens transfer correctly: no fee taking from sender or receiver, no rebasing, no blacklisting, no transfer limits.
+    function _.decimals() external => CVL_decimals(calledContract) expect(uint8);
```
MathisGD: doesn't matter but not mandatory
```suggestion
```

certora/specs/Solvency.spec
```diff
@@ -0,0 +1,155 @@
...
+        }
+        preserved take(uint256 buyerAssets, uint256 sellerAssets, uint256 obligationUnits, uint256 obligationShares, address taker, address takerCallback, bytes takerCallbackData, address receiverIfTakerIsSeller, MorphoV2.Offer offer, MorphoV2.Signature signature, bytes32 root, bytes32[] proof) with (env e) {
+            require taker != currentContract, "no trading with contract";
+            require offer.maker != currentContract, "no trading with contract";
+        }
+    }
+
+// For any token, the flash loans before and after a call is the same.
+// This rule is useful to prove that using persistent ghost for the flashloans mapping is sound.
+rule flashLoansPaidBack(method f, address token) {
```
MathisGD, line 143: maybe I'm missing something but today if we remove the second transferFrom in flashloan the rule would pass, because paid back only requires that the callback arrives to the end of the execution. so what are we verifying exactly?

certora/specs/Solvency.spec
```diff
@@ -0,0 +1,155 @@
...
+hook Sstore obligationState[KEY bytes32 id].withdrawable uint256 newWithdrawable (uint256 oldWithdrawable) {
+    withdrawableMirror[id][loantoken[id]] = newWithdrawable;
+}
+
+/// INVARIANTS AND RULES ///
+
+// For any token, the balance of the contract is always greater than or equal to the sum of all collateral and withdrawable amounts for that token minus the flash loaned amount.
+// Note: this invariant is strong, so it also holds before each external call.
+strong invariant tokenBalanceCorrect(address token)
+    tokenBalances[token][currentContract] >= collateralSum(token) + withdrawableSum(token) - flashloans[token]
```
MathisGD, line 130: I don't get how this can be true. at the moment of the call of the callback in flashloan, `flashloans[token]` is not updated yet but the balance has already been sent

certora/helpers/FlashLiquidateCallback.sol
jhoenicke: It's an external call and this tells the prover that it needs to prove the validity of strong invariants at this point.  It also allows the external call to change variables as long as the invariants are preserved (simulating reentrant calls).

MathisGD: oh ok because `strong` proves that the invariant holds at the moment at which we call unresolved contracts only?

MathisGD: this answers my question [here](https://github.com/morpho-org/morpho-v2/pull/348#discussion_r2824789051)

jhoenicke: Yes, only for unresolved calls the check is included.

certora/specs/Solvency.spec
jhoenicke: I think I tried to mutate the flash loan to pay back to little and the `tokenBalanceCorrect` rule caught the bug.  One path the solver checks is that if the havoc in doFlashloan returns with some balance that still satisfies the invariant with the flashloan subtracted, that the balance satisfies the invariant when the flashloan variable is decremented.

Because of the persistent ghost all rules assume that flashloan has the same value before and after each external callbacks.  This rule shows that this assumption is sound, because every function of MorphoV2 preserves the value of flashloan (as long as every subcall also does, but that is a sound inductive argument)

QGarchery: The `flashLoanPaidBack` is assuming that the summarization of the callbacks is sound. Notably, instead of the transfers to modify the `flashloans` amount, it uses `startFlashLoan` and `endFlashLoan` (and as Jochen mentioned, if you remove one of them, or modify the amount in them, the rule above fails). 
The `flashLoanPaidBack`rule is useful to show that the `persistent` keyword on `flashLoans` is sound, because it shows that the variable `flashLoans` doesn't change in one call. Actually it shows that it doesn't change in one call assuming that sub-calls doesn't change it either (because precisely of the `persistent` keyword), but by induction this is the same thing

certora/specs/Solvency.spec
```diff
@@ -0,0 +1,159 @@
...
+}
+
+/// HELPERS ///
+
+// ERC20 summaries.
+
+// Token balances: token => user => balance.
+ghost mapping(address => mapping(address => uint256)) tokenBalances;
+
+ghost CVL_decimals(address) returns uint8;
```
MathisGD:
```suggestion
```

certora/specs/Solvency.spec
MathisGD: we spoke about it live

MathisGD (approved):

certora/specs/Solvency.spec
```diff
@@ -0,0 +1,154 @@
...
+        tokenBalances[token][src] = assert_uint256(tokenBalances[token][src] - value);
+        tokenBalances[token][dest] = assert_uint256(tokenBalances[token][dest] + value);
+    }
+    return success;
+}
+
+// UtilsLib summaries.
+
+ghost CVL_mulDivDown(uint256, uint256, uint256) returns uint256;
+
```
bhargavbh, line 50: here mulDivDown is being summarised to an uninterpreted function. Confused why the solvency proofs go thru without using the definition of mulDiv. 
The rule FlashLoansPaidBack is check over all methods `f`, and some of them call mulDiv.

certora/specs/Solvency.spec
```diff
@@ -0,0 +1,154 @@
...
+// This rule is useful to prove that using persistent ghost for the flashloans mapping is sound.
+rule flashLoansPaidBack(method f, address token) {
+    env e;
+    calldataarg args;
+    mathint oldFlashLoan = flashloans[token];
+    f(e, args);
+    assert flashloans[token] == oldFlashLoan, "flashloan repaid";
+}
+
+// For any token, the amount of flash loans after a transaction is 0.
```
bhargavbh, line 151: IIUC, the weak invariant shows flashloan is zero after any external call. Additionally, we have `flashLoansPaidback` which states flashloans before and after are equal. Hence, we get flashloans before a call is also zero. Does this mean the entries in persistent ghost mapping is always zero?

bhargavbh (approved):
some clarification questions, otherwise LGTM

certora/specs/Solvency.spec
MathisGD: the rule kind of verifies that withdrawable + collateral updates are the same as balance updates (they use the same variable), but nothing about the variable itself (which is often computed using mulDivDown)

certora/specs/Solvency.spec
QGarchery: No, because you can be inside a call, where the weak invariant doesn't hold. If you try to prove the strong invariant `flashloans[token] == 0` you will get a counter-example

certora/specs/Solvency.spec
bhargavbh: but then why do we need the summaries at all? is it because of Certora's implementation detail, i.e., some preprocessing or static analysis is done which blows up with mulDiv

MathisGD: @jhoenicke @QGarchery can maybe answer this one

QGarchery: Yes this is only for performance reasons. When you have a lot on non linear arithmetics in your execution path the solver might get confused. This is very counter-intuitive, but I know that it happens for many SMT solvers, so I don't think it's specific to the preprocessing

PR #349, opened 2026-02-05
Taker assets

QGarchery:
Rationale:
- the buyer cares about the `buyerAssets` being exact, for example to invest precisely an amount (could be his exact balance)
- the seller cares about the `sellerAssets` being exact, for exact accounting (which could then be useful to invest precisely)

### DIFF
--- a/src/MorphoV2.sol
+++ b/src/MorphoV2.sol
@@ -132,8 +132,7 @@ contract MorphoV2 is IMorphoV2 {
     /// @dev Neither the taker nor the maker can pass from having shares to having debt in one take.
     /// @dev The taker might not get the price they expected if the trading fee was just changed.
     function take(
-        uint256 buyerAssets,
-        uint256 sellerAssets,
+        uint256 takerAssets,
         uint256 obligationUnits,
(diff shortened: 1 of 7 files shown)


adhusson:
If a taker goes through a sequence of offers denominated in assets, he will be trying to empty each offer in turn (since the next one is worse than the current one). He can preconvert into his own assets but it won't be exact/clean.

src/MorphoV2.sol
```diff
@@ -135,8 +135,7 @@ contract MorphoV2 is IMorphoV2 {
     /// @dev Neither the taker nor the maker can pass from having shares to having debt in one take.
     /// @dev The taker might not get the price they expected if the trading fee was just changed.
     function take(
-        uint256 buyerAssets,
-        uint256 sellerAssets,
+        uint256 takerAssets,
```
MathisGD, line 138: what if you want to "finish" an offer? The limiting factor could be the makerAssets, or the liquidity of the callback (in assets too)

QGarchery:
Closing because of the above comments which I didn't have in mind

PR #351, opened 2026-02-06
test: zero price

adhusson:
test conditions when zero price trading succeeds / fails (and also exact amounts)

It succeeds when there is no division by zero & no underflow: 
- any offer / unit or share take input / 0 trading fee
- sell offer / unit, share or buyer take input / > 0 trading fee.

### DIFF
--- a/test/TakeTest.sol
+++ b/test/TakeTest.sol
@@ -1311,7 +1311,7 @@ contract TakeTest is BaseTest {
         assertEq(LendCallback(callback).recordedData(), abi.encode(address(loanToken), assets));
     }
 
-    // Summary:
+    // Summary of zero price tests:
     // - trading fee = 0 with buy or sell offer:
     //   - buyer or seller assets input reverts because of division by a zero price.
     //   - obligation share or unit inputs succeeds.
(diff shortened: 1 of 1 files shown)


test/TakeTest.sol
```diff
@@ -1310,6 +1310,132 @@ contract TakeTest is BaseTest {
...
+        uint256 shares = units.mulDivDown(initialShares + 1, initialUnits + 1);
+        uint256 snap = vm.snapshotState();
+
+        // fee=0, sell, buyer assets
+        borrowerOffer.tick = 0;
+        borrowerOffer.assets = 1e18;
+        borrowerOffer.obligationUnits = 0;
+        deal(address(loanToken), lender, 1e18);
+        collateralize(obligation, borrower, units);
+        vm.expectRevert();
```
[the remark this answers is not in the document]
adhusson: fine as is

test/TakeTest.sol
```diff
@@ -1310,6 +1310,141 @@ contract TakeTest is BaseTest {
...
+    //   - obligation share or unit input succeeds.
+    function testTakeAtPriceZero() public {
+        assertEq(TickLib.tickToPrice(0), 0, "tick 0 should map to price 0");
+
+        uint256 units = 1e18;
+        uint256 shares = units.mulDivDown(initialShares + 1, initialUnits + 1);
+        uint256 snap = vm.snapshotState();
+
+        // fee=0, sell, buyer assets
+        borrowerOffer.tick = 0;
```
MathisGD, line 1326: you can do this once and for all at the top

test/TakeTest.sol
```diff
@@ -1310,6 +1310,141 @@ contract TakeTest is BaseTest {
...
+        borrowerOffer.assets = 1e18;
+        borrowerOffer.obligationUnits = 0;
```
MathisGD:
```suggestion
```

test/TakeTest.sol
```diff
@@ -1310,6 +1310,141 @@ contract TakeTest is BaseTest {
...
+        deal(address(loanToken), lender, 1e18);
+        collateralize(obligation, borrower, units);
+        vm.expectRevert();
+        take(1e18, 0, 0, 0, lender, borrowerOffer);
```
MathisGD, line 1331: no assert here?

test/TakeTest.sol
```diff
@@ -1310,6 +1310,141 @@ contract TakeTest is BaseTest {
...
+        vm.revertToState(snap);
```
MathisGD: could be separated tests, there isn't much setup

MathisGD (approved):

test/TakeTest.sol
adhusson: wdym? it expects a revert

test/TakeTest.sol
adhusson: explicit is good no? in case this gets reorganized.

could be separate tests then it's not needed

test/TakeTest.sol
adhusson: ok with separated tests will change

MathisGD (approved):

PR #352, opened 2026-02-06
Feat/global min collat with liquidation case

peyha:


MathisGD:
will implement directly in #303

PR #356, opened 2026-02-10
opti health check and limit max number of collaterals

MathisGD:
- limits the number of active collat per borrower to 10 to avoid gas issues.
- limits the max number of collats per obligation to 128 as a side effect of the implementation.
- perf: avoids reading all collat balances

[useful](https://github.com/morpho-org/morpho-v2/pull/364#discussion_r2817002111) for intuition on the bit tricks

### DIFF
--- a/foundry.toml
+++ b/foundry.toml
@@ -9,4 +9,4 @@ fs_permissions = [{ access = "read", path = "test/ticks_exact.json" }]
 wrap_comments = true
 
 [lint]
-exclude_lints = ["asm-keccak256", "unsafe-cheatcode"]
+exclude_lints = ["asm-keccak256", "unsafe-cheatcode", "incorrect-shift"]
--- a/src/MorphoV2.sol
+++ b/src/MorphoV2.sol
@@ -364,7 +364,7 @@ contract MorphoV2 is IMorphoV2 {
         uint256 newCollateralOf = collateralOf[id][onBehalf][collateralToken] + assets;
         collateralOf[id][onBehalf][collateralToken] = newCollateralOf;
 
-        if (newCollateralOf != assets && assets != 0) {
+        if (newCollateralOf == assets && assets > 0) {
             // forge-lint: disable-next-item(unsafe-typecast) as collateralIndex < MAX_COLLATERALS (128)
             uint128 newBitmap = borrowerState[id][onBehalf].activatedCollaterals | uint128(1 << collateralIndex);
             borrowerState[id][onBehalf].activatedCollaterals = newBitmap;
@@ -399,7 +399,7 @@ contract MorphoV2 is IMorphoV2 {
         uint256 newCollateralOf = collateralOf[id][onBehalf][collateralToken] - assets;
         collateralOf[id][onBehalf][collateralToken] = newCollateralOf;
 
-        if (newCollateralOf == 0 && assets != 0) {
+        if (newCollateralOf == 0) {
             // forge-lint: disable-next-item(unsafe-typecast) as collateralIndex < MAX_COLLATERALS (128)
             borrowerState[id][onBehalf].activatedCollaterals &= ~uint128(1 << collateralIndex);
         }
--- a/src/MorphoV2.sol
+++ b/src/MorphoV2.sol
@@ -13,6 +13,8 @@ import {
     MAX_FEE,
     MAX_LIF,
     TIME_TO_MAX_LIF,
+    MAX_COLLATERALS,
+    MAX_COLLATERALS_PER_BORROWER,
     EIP712_DOMAIN_TYPEHASH,
     ROOT_TYPEHASH
 } from "./libraries/ConstantsLib.sol";
@@ -31,6 +33,7 @@ contract MorphoV2 is IMorphoV2 {
     mapping(bytes32 id => mapping(address user => uint256)) public sharesOf;
     mapping(bytes32 id => mapping(address user => uint256)) public debtOf;
     mapping(bytes32 id => mapping(address user => mapping(address collateralToken => uint256))) public collateralOf;
+    mapping(bytes32 id => mapping(address user => uint256)) public activatedCollaterals;
     mapping(bytes32 id => ObligationState) public obligationState;
 
     /// @dev Groups are useful to have a global offered amount shared accross multiple offers ("OCO").
@@ -331,6 +334,12 @@ contract MorphoV2 is IMorphoV2 {
 
         collateralOf[id][onBehalf][collateralToken] += assets;
 
+        if (collateralOf[id][onBehalf][collateralToken] != 0) {
+            uint256 newBitmap = activatedCollaterals[id][onBehalf] | (1 << collateralIndex);
+            activatedCollaterals[id][onBehalf] = newBitmap;
+            require(UtilsLib.countBits(newBitmap) <= MAX_COLLATERALS_PER_BORROWER, "too many collaterals per borrower");
+        }
+
         address oracle = obligation.collaterals[collateralIndex].oracle;
         uint256 _collateralOf = collateralOf[id][onBehalf][collateralToken];
         require(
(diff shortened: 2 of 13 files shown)


src/MorphoV2.sol
```diff
@@ -331,6 +334,12 @@ contract MorphoV2 is IMorphoV2 {
 
         collateralOf[id][onBehalf][collateralToken] += assets;
 
+        if (collateralOf[id][onBehalf][collateralToken] != 0) {
```
adhusson: could save a cold sstore if collateral was already > 0 (although not if we disable on liq below min collat)

foundry.toml
```diff
@@ -9,4 +9,4 @@ fs_permissions = [{ access = "read", path = "test/ticks_exact.json" }]
 wrap_comments = true
 
 [lint]
-exclude_lints = ["asm-keccak256", "unsafe-cheatcode"]
+exclude_lints = ["asm-keccak256", "unsafe-cheatcode", "incorrect-shift"]
```
adhusson, line 12: better to disable items individually?

MathisGD: in general I agree, but I disagree with this warning at all (why shifting by variable should be bad?)

src/MorphoV2.sol
MathisGD: what if I replace with assets>0?

adhusson: this would likely save 100 gas of reading warm but not save the 2200 of a cold noop sstore

foundry.toml
adhusson: Because it's more common to do `var << 128` than `1 << var`, and maybe also because in assembly `shl(x,y)` is `y << x`. But okfm either way.

src/MorphoV2.sol

src/interfaces/IMorphoV2.sol
```diff
@@ -40,6 +40,11 @@ struct Signature {
     bytes32 s;
 }
 
+struct BorrowerState {
+    uint128 debt;
+    uint128 activatedCollaterals;
```
adhusson, line 45: It would take 88 bits to store `[count][index1]...[index10]`, or if max 7 collats it would take 32 bits, but I think the code would be more complex (surprisingly).

adhusson (approved):

src/libraries/UtilsLib.sol
```diff
@@ -67,4 +67,21 @@ library UtilsLib {
...
+        unchecked {
+            x = x - ((x >> 1) & 0x5555555555555555555555555555555555555555555555555555555555555555);
+            x = (x & 0x3333333333333333333333333333333333333333333333333333333333333333)
+                + ((x >> 2) & 0x3333333333333333333333333333333333333333333333333333333333333333);
+            x = (x + (x >> 4)) & 0x0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f;
+            return (x * 0x0101010101010101010101010101010101010101010101010101010101010101) >> 248;
+        }
+    }
+
+    function ctz(uint256 bitmap) internal pure returns (uint256 res) {
```
peyha: why use ctz when it's possible to do clz and bitmap -= (1 << i) ? gas savings ?

src/MorphoV2.sol
```diff
@@ -378,15 +396,20 @@ contract MorphoV2 is IMorphoV2 {
         bytes32 id = touchObligation(obligation);
         address collateralToken = obligation.collaterals[collateralIndex].token;
 
-        collateralOf[id][onBehalf][collateralToken] -= assets;
+        uint256 newCollateralOf = collateralOf[id][onBehalf][collateralToken] - assets;
+        collateralOf[id][onBehalf][collateralToken] = newCollateralOf;
 
-        require(isHealthy(obligation, id, onBehalf), "Unhealthy borrower");
+        if (newCollateralOf == 0) {
```
```suggestion
        if (newCollateralOf == 0 && assets > 0) {
```
(optional) for symmetry with `supplyCollateral` mainly

src/MorphoV2.sol
```diff
@@ -351,13 +361,21 @@ contract MorphoV2 is IMorphoV2 {
         bytes32 id = touchObligation(obligation);
         address collateralToken = obligation.collaterals[collateralIndex].token;
 
-        collateralOf[id][onBehalf][collateralToken] += assets;
+        uint256 newCollateralOf = collateralOf[id][onBehalf][collateralToken] + assets;
+        collateralOf[id][onBehalf][collateralToken] = newCollateralOf;
+
+        if (newCollateralOf == assets && assets > 0) {
```
peyha, line 367:
```suggestion
        uint256 oldCollateralOf = collateralOf[id][onBehalf][collateralToken];
        collateralOf[id][onBehalf][collateralToken] = oldCollateralOf + assets;

        if (oldCollateralOf == 0 && assets > 0) {
```
a bit cleaner to me but no big deal

MathisGD: agree, but we need it later too for the min collat check

src/MorphoV2.sol
```suggestion
        if (newCollateralOf == 0 && assets > 0) {
```
reduce gas cost if no-op (not really useful but a tiny bit cleaner)

src/MorphoV2.sol
```diff
@@ -454,13 +481,17 @@ contract MorphoV2 is IMorphoV2 {
                 uint256 lltv = obligation.collaterals[collateralIndex].lltv;
                 // Rounded up to avoid consecutive max liquidations.
                 // Acknowledged that the position could be slightly healthy after a liquidation.
-                uint256 maxRepaid = (debtOf[id][borrower] - maxDebt).mulDivUp(WAD, WAD - lif.mulDivUp(lltv, WAD));
+                uint256 maxRepaid = (uint256(_state.debt) - maxDebt).mulDivUp(WAD, WAD - lif.mulDivUp(lltv, WAD));
                 require(repaidUnits <= maxRepaid, "recovery close factor violated");
             }
 
             collateralOf[id][borrower][liquidatedCollatToken] -= seizedAssets;
+            if (collateralOf[id][borrower][liquidatedCollatToken] == 0) {
```
```suggestion
            if (collateralOf[id][borrower][liquidatedCollatToken] == 0 && seizedAssets > 0) {
```
same as before, feel free to discard

src/libraries/ConstantsLib.sol
```diff
@@ -10,3 +10,5 @@ uint256 constant MAX_LIF = 1.15e18; // Liquidation Incentive Factor
 uint256 constant TIME_TO_MAX_LIF = 15 minutes; // Time to reach MAX_LIF
 bytes32 constant EIP712_DOMAIN_TYPEHASH = keccak256("EIP712Domain(uint256 chainId,address verifyingContract)");
 bytes32 constant ROOT_TYPEHASH = keccak256("Root(bytes32 root)");
+uint256 constant MAX_COLLATERALS = 128;
+uint256 constant MAX_COLLATERALS_PER_BORROWER = 10;
```
peyha, line 14: we'll probably have to fine tune these values because it should be set according to the gas cost of some functions of the contract and to lender/borrower requirements. For instance, separated matching could improve the gas cost of the function so it would make sense to lower `MAX_COLLATERALS_PER_BORROWER` if we end up implementing it 
Opened #376

test/BaseTest.sol
QGarchery, line 1: some other possible tests related to this PR:
- check both max collaterals and max collaterals per borrower
- on an empty bitmap, when adding a collateral with index `i`, check that ctz of the new bitmap gives `i`
- when adding a few different collaterals (say `k` of them)
  - check that countBits returns `k`
  - check that maxDebt returns the same thing than if iterating over all collaterals

QGarchery (approved):

src/libraries/UtilsLib.sol
MathisGD: good ida

src/MorphoV2.sol
adhusson: agree but less useful (supplying on a nonzero position will happen often withdrawing from an already zero position will happen rarely)

src/libraries/UtilsLib.sol
MathisGD: it's indeed cheaper because you naturally don't have the underflow check

peyha (approved):

PR #359, opened 2026-02-11
[Certora] Solvency: some minor tweaks

QGarchery:
Changes:
- used the CVL formatter
- removed unnecessary config (such as the debt sum, the envfree declarations)
- make style closer to our conventions (reordering, document some assumptions, inline comments)
- parametric contract only has MorphoV2


jhoenicke (approved):
looks good to me

PR #360, opened 2026-02-12
[min collat] variable LIF to clear all collat

adhusson:
if liq can go below minCollat (with RCF), seize all collat
- liquidators will always capture all of min_collateral by seizing `collateral-minCollateral+1`
- liquidatable users with `collat < minCollat` already due to the price can have their collat seized at no cost

### DIFF
--- a/src/MorphoV2.sol
+++ b/src/MorphoV2.sol
@@ -395,7 +395,7 @@ contract MorphoV2 is IMorphoV2 {
         SafeTransferLib.safeTransfer(collateralToken, receiver, assets);
     }
 
-    /// @dev At least one of `repaidUnits` or `seizedAssets` should be equal to zero.
+    /// @dev At least one of `repaidUnits` or `requestedSeizedAssets` should be equal to zero.
     /// @dev Accounts are liquidatable if they are unhealthy or if the maturity is reached.
     /// @dev Before maturity, the liquidation cannot put the borrower back into health (recovery close factor).
     /// @dev If an account is healthy, the LIF grows linearly from 1 at maturity to MAX_LIF at maturity +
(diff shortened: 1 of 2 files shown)


adhusson:
> The pre-maturity recovery-close-factor guard allows `debtAfter == newMaxDebt` because it uses `>=`, but `isHealthy` treats `maxDebt >= debt` as healthy. This means a liquidation can pass this check while bringing a borrower exactly back to the health threshold, which violates the stated requirement that pre-maturity liquidations must not restore health.

pending [discussion](https://github.com/morpho-org/morpho-v2/pull/303#discussion_r2788057014)

> **P2 Badge Skip oracle reads when minCollateral is zero**

liquidations are supposed to rely on oracles anyway


> **P2 Badge Approve collateral before supplying in zero-oracle test**

The test does not revert and does not expect a revert

src/MorphoV2.sol
```diff
@@ -405,11 +405,11 @@ contract MorphoV2 is IMorphoV2 {
         Obligation calldata obligation,
         uint256 collateralIndex,
         uint256 repaidUnits,
-        uint256 seizedAssets,
+        uint256 requestedSeizedAssets,
```
QGarchery, line 408: I don't understand what does it change to have this new variable ? It seems equivalent to me

adhusson: It's for clarity: even if you set a > 0 `requestedSeizedAssets` the actual seized assets may end up higher if the lif gets increased.

adhusson:
@QGarchery : Currently broken as it can create bad debt

qdrn:
Aside from the current implementation, as described it will create bad debt for positions that are collaterized with collat ~ min collat and close to lltv as their collat will be seized for 0.

adhusson:
replaced by https://github.com/morpho-org/morpho-v2/pull/391

PR #366, opened 2026-02-16
feat: use idlib for isCreated

adhusson:


src/MorphoV2.sol
MathisGD, line 1: why does that save gas? the slot would be touched anyway in take (for fees)

adhusson: it's not for saving gas it's for simplicity

src/libraries/IdLib.sol
```diff
@@ -31,10 +31,16 @@ library IdLib {
...
-        return abi.decode(create2Address.code, (Obligation));
+    function toAddress(bytes32 id, address morphoV2) internal pure returns (address) {
+        return address(uint160(uint256(keccak256(abi.encodePacked(uint8(0xff), morphoV2, bytes32(0), id)))));
+    }
+
+    function isCreated(bytes32 id, address morphoV2) internal view returns (bool) {
+        return toAddress(id, morphoV2).code.length > 0;
+    }
+
+    function toObligation(bytes32 id, address morphoV2) internal view returns (Obligation memory) {
```
adhusson, line 42: also updated names so it's all of the form `toY` inside idlib instead of sometimes `toY` sometimes `XToY`

QGarchery (approved):

src/MorphoV2.sol
MathisGD: should we test the increase in gas cost?

adhusson: It's +2600 if the sstore2 contract would never have been read otherwise (but with id inputs that never happens).

t's +500 if the obligation params would never have been read otherwse because cold SLOAD is 2100 and cold EXTCODESIZE is 2600.

It's nothing or +100 if both SLOAD and EXTCODE would have been used otherwise.

MathisGD: then maybe let's wait to settle on id as input

adhusson:
optimistically closing (subsumed by https://github.com/morpho-org/morpho-v2/pull/367 )

PR #367, opened 2026-02-16
[gas] id based entry points + simpler offer creation naming

adhusson:
[[paused until glamsterdam numbers are out](https://morpholabs.slack.com/archives/C08A40KKN9W/p1771508404612859)]

### DIFF
--- a/src/MorphoV2.sol
+++ b/src/MorphoV2.sol
@@ -496,7 +496,7 @@ contract MorphoV2 is IMorphoV2 {
     /// @dev Returns the obligation id and creates the obligation if it doesn't exist yet.
     function touchObligation(Obligation memory obligation) public returns (bytes32) {
         bytes32 id = IdLib.toId(obligation, block.chainid, address(this));
-        if (!obligationState[id].created) {
+        if (!IdLib.isCreated(id, address(this))) {
             address previousCollateralToken;
             for (uint256 i = 0; i < obligation.collaterals.length; i++) {
                 address collateralToken = obligation.collaterals[i].token;
@@ -504,7 +504,6 @@ contract MorphoV2 is IMorphoV2 {
                 previousCollateralToken = collateralToken;
             }
 
-            obligationState[id].created = true;
             obligationState[id].fees = defaultFees[obligation.loanToken];
             IdLib.storeInCode(obligation);
 
@@ -524,7 +523,7 @@ contract MorphoV2 is IMorphoV2 {
     }
 
     function obligationCreated(bytes32 id) external view returns (bool) {
-        return obligationState[id].created;
+        return IdLib.isCreated(id, address(this));
     }
 
     function withdrawable(bytes32 id) external view returns (uint256) {
--- a/src/MorphoV2.sol
+++ b/src/MorphoV2.sol
@@ -483,19 +483,23 @@ contract MorphoV2 is IMorphoV2 {
     }
 
     /// @dev Returns the obligation id and creates the obligation if it doesn't exist yet.
-    function createObligation(Obligation memory obligation) external returns (bytes32) {
-        bytes32 id = IdLib.createCode(obligation);
-        address previousCollateralToken;
-        for (uint256 i = 0; i < obligation.collaterals.length; i++) {
-            address collateralToken = obligation.collaterals[i].token;
-            require(collateralToken > previousCollateralToken, "collaterals not sorted");
-            require(obligation.collaterals[i].lltv < WAD.mulDivDown(WAD, MAX_LIF), "lltv too high or LIF too high"); // temporary.
-            previousCollateralToken = collateralToken;
+    function createObligation(Obligation memory obligation) public returns (bytes32) {
+        bytes memory encoded = abi.encode(obligation);
+        (bool created, bytes32 id) = IdLib.codeIsCreated(encoded, block.chainid, address(this));
+        if (!created) {
+            address previousCollateralToken;
+            for (uint256 i = 0; i < obligation.collaterals.length; i++) {
+                address collateralToken = obligation.collaterals[i].token;
+                require(collateralToken > previousCollateralToken, "collaterals not sorted");
+                require(obligation.collaterals[i].lltv < WAD.mulDivDown(WAD, MAX_LIF), "lltv too high or LIF too high"); // temporary.
+                previousCollateralToken = collateralToken;
+            }
+
+            obligationState[id].fees = defaultFees[obligation.loanToken];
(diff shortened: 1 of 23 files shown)


src/MorphoV2.sol
```diff
@@ -353,10 +337,8 @@ contract MorphoV2 is IMorphoV2 {
         SafeTransferLib.safeTransferFrom(obligation.loanToken, msg.sender, address(this), obligationUnits);
     }
 
-    function supplyCollateral(Obligation memory obligation, address collateral, uint256 assets, address onBehalf)
-        external
-    {
-        bytes32 id = touchObligation(obligation);
+    function supplyCollateral(bytes32 id, address collateral, uint256 assets, address onBehalf) external {
+        require(IdLib.codeIsCreated(id, address(this)), "obligation not created");
```
adhusson: This one is +500 compared to `obligationState.created`, worth not doing it just for this, or worth it for eliminating the redundancy?

test/TakeTest.sol
```diff
@@ -902,11 +908,14 @@ contract TakeTest is BaseTest {
         lenderOffer.obligationUnits = firstFill + secondFill;
         lenderOffer.assets = 0;
         lenderOffer.tick = TICK_RANGE;
+        Obligation memory obligation2 = abi.decode(abi.encode(obligation), (Obligation));
+        obligation2.maturity = obligation.maturity + 100;
+        bytes32 id2 = morphoV2.createObligation(obligation2);
         Offer memory lenderOffer2 = lenderOffer;
-        lenderOffer2.obligation.maturity = obligation.maturity + 100;
```
adhusson, line 907: before this would also update `lenderOffer.obligation.maturity` due to references, fine in context but very misleading

src/MorphoV2.sol
```diff
@@ -515,6 +491,10 @@ contract MorphoV2 is IMorphoV2 {
 
     /// VIEW FUNCTIONS ///
 
+    function idToObligation(bytes32 id) public view returns (Obligation memory) {
+        return IdLib.toObligation(id, address(this));
```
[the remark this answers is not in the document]
adhusson: existence implicitly checked in toObligation

src/MorphoV2.sol
```diff
@@ -524,7 +504,7 @@ contract MorphoV2 is IMorphoV2 {
     }
 
     function obligationCreated(bytes32 id) external view returns (bool) {
-        return obligationState[id].created;
+        return IdLib.codeIsCreated(id, address(this));
     }
```
QGarchery: I would remove the getter now, it shouldn't be called as it's more efficient to call the lib directly

src/MorphoV2.sol
```diff
@@ -535,8 +515,11 @@ contract MorphoV2 is IMorphoV2 {
         return obligationState[id].fees;
     }
 
-    /// @dev This function should be called with the id corresponding to the obligation.
-    function isHealthy(Obligation memory obligation, bytes32 id, address borrower) public view returns (bool) {
+    function isHealthy(bytes32 id, address borrower) external view returns (bool) {
+        return _isHealthy(idToObligation(id), id, borrower);
+    }
```
QGarchery: this one we can remove too, it should be called by a contract anyway

src/MorphoV2.sol
```diff
@@ -515,6 +491,10 @@ contract MorphoV2 is IMorphoV2 {
 
     /// VIEW FUNCTIONS ///
 
+    function idToObligation(bytes32 id) public view returns (Obligation memory) {
+        return IdLib.toObligation(id, address(this));
+    }
```
QGarchery: and this one too, although we could keep it internal to avoid inlining and having the `address(this)` but I think it's not worth it

src/MorphoV2.sol
```diff
@@ -494,19 +471,18 @@ contract MorphoV2 is IMorphoV2 {
     }
 
     /// @dev Returns the obligation id and creates the obligation if it doesn't exist yet.
-    function touchObligation(Obligation memory obligation) public returns (bytes32) {
+    function createObligation(Obligation memory obligation) public returns (bytes32) {
```
```suggestion
    function createObligation(Obligation memory obligation) external returns (bytes32) {
```

src/MorphoV2.sol
adhusson: And make `isHealthy(obligation)` public? If called with `eth_call` no gas, if called by a contract no calldata cost.

src/MorphoV2.sol
adhusson: this is like blue's [`idToMarketParams`](https://github.com/morpho-org/morpho-blue/blob/main/src/Morpho.sol#L70), isn't it worth the convenience + consistency?

test/TradingFeeTest.sol
```diff
@@ -242,13 +251,13 @@ contract TradingFeeTest is BaseTest {
...
 
         obligation.maturity = block.timestamp + 3 days;
-        id = toId(obligation);
-        lenderOffer.obligation = obligation;
-        borrowerOffer.obligation = obligation;
 
         // Set fees at breakpoints for linear interpolation (3 days is between 1 and 7 days)
         morphoV2.setDefaultTradingFee(address(loanToken), 1, fee1Day);
         morphoV2.setDefaultTradingFee(address(loanToken), 2, fee7Days);
+        id = morphoV2.createObligation(obligation);
```
MathisGD: could be in the helpers collateralize and/or take

src/MorphoV2.sol
adhusson: going for this

src/MorphoV2.sol
QGarchery: I find it more convenient to have only one way to do something, than to add another worse way to do it (plus you already need the address of MorphoV2 anyway to call it)

src/MorphoV2.sol
QGarchery, line 1: some concern [here](https://morpholabs.slack.com/archives/C08A40KKN9W/p1771267527454759?thread_ts=1771025176.189189&cid=C08A40KKN9W): about the fact that it will require to duplicate the bundler functions (and other contracts on top)

src/MorphoV2.sol
adhusson: worse in terms of gas but what about external callers that use it through eth_call? maybe i'm inventing use cases here, not sure

src/MorphoV2.sol
adhusson: - and I think from our last discussion one option is a bundler flag so 1 function can either receive the obligation or not depending on whether it exists already

PR #368, opened 2026-02-16
Refactor perf create obligation

QGarchery:


src/MorphoV2.sol
QGarchery, line 1: The goal is to avoid doing `abi.encode(obligation)` twice. I tried a couple of things, but nothing felt nice because it is either too complex, or it breaks the IdLib abstraction (like this PR)

adhusson: what do you think about this? https://github.com/morpho-org/morpho-v2/pull/367/changes/90609d49d002a7d0fa7980d7dca0c8e4c4fcefc3 (updated to maintain idempotent behaviour)

QGarchery: I had something similar did not manage to have something that:
- prevents to reset to the default fee when being called again (as the code is currently doing)
- ideally, in the no-op case, does not include the cost of creating the contract

adhusson: indeed was just looking at default fee reset, will look at it again

adhusson: actually in the noop case my attempt is worse than a revert, it consumes all remaining gas

adhusson: ok tried something like this but with maybe less abstraction breaking, and it allows for completely skipping the obligation encoding step by reading it directly from calldata for anyone who wants to : https://github.com/morpho-org/morpho-v2/pull/367/changes/9bb27b82dba0d0a0a485168898ba3c3e42402e1b

but also looks less clean that this one

QGarchery:
closed in favor of #379

PR #369, opened 2026-02-16
perf: pack collaterals

MathisGD:
note that in obligations with high number of collaterals, the 10 collaterals of the borrower have low chances of being in the same slot. it only makes a difference in smaller obligations. but it's almost free so..


adhusson:
Looking a 2nd time I like it less because of 1) tiny complexity increase and 2) the small indirection when trying to read a users collateral (must sort collaterals first) for a hard to quantify win. But still OK with it.

MathisGD:
good points. note that supplyCollateral, withdrawCollateral and liquidate take the index as input

adhusson (approved):

src/MorphoV2.sol
```diff
@@ -396,8 +396,8 @@ contract MorphoV2 is IMorphoV2 {
         bytes32 id = touchObligation(obligation);
         address collateralToken = obligation.collaterals[collateralIndex].token;
 
-        uint256 newCollateralOf = collateralOf[id][onBehalf][collateralToken] - assets;
-        collateralOf[id][onBehalf][collateralToken] = newCollateralOf;
+        uint256 newCollateralOf = uint256(collateralOf[id][onBehalf][collateralIndex]) - assets;
+        collateralOf[id][onBehalf][collateralIndex] = UtilsLib.toUint128(newCollateralOf);
```
QGarchery, line 400: We can do an unsafe cast safely here since `newCollateralOf <= collateralOf[id][onBehalf][collateralIndex] < type(uint128).max`

src/MorphoV2.sol
```diff
@@ -485,8 +485,8 @@ contract MorphoV2 is IMorphoV2 {
                 require(repaidUnits <= maxRepaid, "recovery close factor violated");
             }
 
-            collateralOf[id][borrower][liquidatedCollatToken] -= seizedAssets;
-            if (collateralOf[id][borrower][liquidatedCollatToken] == 0 && seizedAssets > 0) {
+            collateralOf[id][borrower][collateralIndex] -= UtilsLib.toUint128(seizedAssets);
+            if (collateralOf[id][borrower][collateralIndex] == 0 && seizedAssets > 0) {
```
QGarchery, line 488: We can now inline `liquidatedCollatToken` in the line `        SafeTransferLib.safeTransfer(liquidatedCollatToken, msg.sender, seizedAssets);` below

QGarchery (approved):

src/MorphoV2.sol
MathisGD: but then we need a comment etc.. worth it?

src/MorphoV2.sol
MathisGD: note that it will revert a bit later when index > length, but it's ok

PR #371, opened 2026-02-17
Bad debt overestimation

peyha:
cf [this](https://morpholabs.slack.com/archives/C02N7CZ088N/p1771259050345039)

Main points for this PR
- overestimating means that the collateral cannot be completely seized (while underestimating means that the debt cannot be completely repaid)
- with underestimation, it's possible that the bad debt is realized with consecutive liquidation (which we're not sure is a potential vector of attack but is undoubtedly less clean)


test/LiquidationTest.sol
```diff
@@ -486,6 +526,17 @@ contract LiquidationTest is BaseTest {
 
     // helpers.
 
+    /// @dev Repayable debt as computed in liquidate: sum over all collaterals with mulDivDown.
+    function _repayableDebt() internal view returns (uint256) {
```
MathisGD: I would rather compute the bad debt, I find it more natural

src/MorphoV2.sol
```diff
@@ -451,7 +451,7 @@ contract MorphoV2 is IMorphoV2 {
             if (i == collateralIndex) liquidatedCollatPrice = price;
             uint256 _collateralOf = collateralOf[id][borrower][_collateral.token];
             maxDebt += _collateralOf.mulDivDown(price, ORACLE_PRICE_SCALE).mulDivDown(_collateral.lltv, WAD);
-            repayableDebt += _collateralOf.mulDivUp(WAD, MAX_LIF).mulDivUp(price, ORACLE_PRICE_SCALE);
+            repayableDebt += _collateralOf.mulDivDown(WAD, MAX_LIF).mulDivDown(price, ORACLE_PRICE_SCALE);
```
MathisGD: the name is not satisfying anymore (in general this concept is bof). so I propose this:
```suggestion
            badDebt = badDebt.zeroFloorSub(_collateralOf.mulDivDown(WAD, MAX_LIF).mulDivDown(price, ORACLE_PRICE_SCALE));
```
and at the beginning of the loop:
```solidity
badDebt = originalDebt;
```

test/LiquidationTest.sol
```diff
@@ -108,6 +108,10 @@ contract LiquidationTest is BaseTest {
         Oracle(obligation.collaterals[0].oracle).setPrice(liquidationOraclePrice);
         vm.warp(obligation.maturity + TIME_TO_MAX_LIF); // Warp to post-maturity to bypass recovery close factor.
 
+        uint256 badDebt = _badDebt();
+        uint256 debtAfterBadDebt = units - badDebt;
+        repaid = bound(repaid, 0, debtAfterBadDebt);
```
MathisGD: there shouldn't be any bad debt (see the liqPrice bounds)

test/LiquidationTest.sol
```diff
@@ -131,6 +135,11 @@ contract LiquidationTest is BaseTest {
         Oracle(obligation.collaterals[0].oracle).setPrice(liquidationOraclePrice);
         vm.warp(obligation.maturity + TIME_TO_MAX_LIF); // Warp to post-maturity to bypass recovery close factor.
 
+        uint256 badDebt = _badDebt();
+        uint256 debtAfterBadDebt = units - badDebt;
+        uint256 maxSeized = debtAfterBadDebt.mulDivDown(ORACLE_PRICE_SCALE, 1e36 - 1).mulDivDown(MAX_LIF, WAD);
+        seized = bound(seized, 0, maxSeized > initialCollateral ? initialCollateral : maxSeized);
```
MathisGD: same

test/LiquidationTest.sol
```diff
@@ -156,6 +165,10 @@ contract LiquidationTest is BaseTest {
         Oracle(obligation.collaterals[0].oracle).setPrice(liquidationOraclePrice);
         vm.warp(obligation.maturity + TIME_TO_MAX_LIF); // Warp to post-maturity to bypass recovery close factor.
 
+        uint256 badDebt = _badDebt();
+        uint256 debtAfterBadDebt = units - badDebt;
+        repaid = bound(repaid, 0, debtAfterBadDebt);
```
MathisGD: same

test/LiquidationTest.sol
```diff
@@ -196,8 +209,8 @@ contract LiquidationTest is BaseTest {
         collateralize(obligation, borrower, units);
         setupObligation(obligation, units);
         Oracle(obligation.collaterals[0].oracle).setPrice(liquidationOraclePrice);
-        uint256 repayable = uint256(morphoV2.collateralOf(id, borrower, 0)).mulDivUp(WAD, MAX_LIF)
-            .mulDivUp(liquidationOraclePrice, ORACLE_PRICE_SCALE);
+        uint256 repayable = uint256(morphoV2.collateralOf(id, borrower, 0)).mulDivDown(WAD, MAX_LIF)
+            .mulDivDown(liquidationOraclePrice, ORACLE_PRICE_SCALE);
         uint256 expectedBadDebt = units - repayable;
```
MathisGD:
```suggestion
uint256 expectedBadDebt = _badDebt();
```

test/LiquidationTest.sol
peyha: actually it can, we overestimate it so there's sometimes a few wei of bad debt at badDebtPrice

test/LiquidationTest.sol
```diff
@@ -279,9 +314,13 @@ contract LiquidationTest is BaseTest {
 
         uint256 initialCollateral = morphoV2.collateralOf(id, borrower, 0);
 
+        uint256 badDebt = _badDebt();
+
+        repaid = bound(repaid, 0, units - badDebt);
```
MathisGD: same

test/LiquidationTest.sol
peyha: I added these because the check `units-repaid=debtOf` was failing

test/LiquidationTest.sol
```diff
@@ -213,11 +226,20 @@ contract LiquidationTest is BaseTest {
...
+        uint256 repayable = units - expectedBadDebt;
+        uint256 maxDebt = initialCollateral.mulDivDown(liquidationOraclePrice, ORACLE_PRICE_SCALE)
+            .mulDivDown(obligation.collaterals[0].lltv, WAD);
+        vm.assume(repayable > maxDebt); // So that some repayment is allowed by recovery close factor.
+        uint256 maxRepaidUncapped =
+            (repayable - maxDebt).mulDivUp(WAD, WAD - MAX_LIF.mulDivUp(obligation.collaterals[0].lltv, WAD));
+        uint256 maxRepaid = maxRepaidUncapped > repayable ? repayable : maxRepaidUncapped; // Cannot repay more than
+        // debt.
+        uint256 maxSeized = maxRepaid.mulDivDown(MAX_LIF, WAD).mulDivDown(ORACLE_PRICE_SCALE, liquidationOraclePrice);
+        seized = bound(seized, 0, maxSeized > initialCollateral ? initialCollateral : maxSeized);
```
MathisGD: there shouldn't need all this complexity (same for the 2 other tests)

src/MorphoV2.sol
```diff
@@ -438,10 +438,13 @@ contract MorphoV2 is IMorphoV2 {
...
         ObligationState storage _obligationState = obligationState[id];
 
-        uint256 repayableDebt;
         uint256 maxDebt;
         uint256 liquidatedCollatPrice;
+
         BorrowerState storage _state = borrowerState[id][borrower];
+        uint256 originalDebt = _state.debt;
+        uint256 badDebt = originalDebt;
+
```
```suggestion
        BorrowerState storage _state = borrowerState[id][borrower];
        uint256 originalDebt = _state.debt;
        uint256 badDebt = originalDebt;
```

src/MorphoV2.sol
```diff
@@ -480,7 +481,8 @@ contract MorphoV2 is IMorphoV2 {
                 uint256 lltv = obligation.collaterals[collateralIndex].lltv;
                 // Rounded up to avoid consecutive max liquidations.
                 // Acknowledged that the position could be slightly healthy after a liquidation.
-                uint256 maxRepaid = (uint256(_state.debt) - maxDebt).mulDivUp(WAD, WAD - lif.mulDivUp(lltv, WAD));
+                uint256 maxRepaid =
+                    uint256(_state.debt).zeroFloorSub(maxDebt).mulDivUp(WAD, WAD - lif.mulDivUp(lltv, WAD));
```
MathisGD: zeroFloorSub is introduced because now debt - badDebt >= maxDebt can break (I guess for debt very small)

test/LiquidationTest.sol
```diff
@@ -101,13 +101,18 @@ contract LiquidationTest is BaseTest {
...
         setupObligation(obligation, units);
+
         uint256 initialCollateral = morphoV2.collateralOf(id, borrower, 0);
+        liquidationOraclePrice =
+            bound(liquidationOraclePrice, badDebtPrice(units, initialCollateral), ORACLE_PRICE_SCALE);
         Oracle(obligation.collaterals[0].oracle).setPrice(liquidationOraclePrice);
         vm.warp(obligation.maturity + TIME_TO_MAX_LIF); // Warp to post-maturity to bypass recovery close factor.
 
+        repaid = bound(repaid, 0, units);
+
```
MathisGD:
```suggestion
```

test/LiquidationTest.sol
```diff
@@ -485,10 +539,27 @@ contract LiquidationTest is BaseTest {
...
     function badDebtPrice() internal view returns (uint256) {
         uint256 lltv = obligation.collaterals[0].lltv;
-        return lltv.mulDivUp(MAX_LIF, WAD) * (ORACLE_PRICE_SCALE / WAD);
+        return lltv.mulDivUp(MAX_LIF, WAD).mulDivUp(ORACLE_PRICE_SCALE, WAD);
+    }
+
+    /// @dev Minimum oracle price for collateral[0] such that there won't be bad debt.
+    function badDebtPrice(uint256 debt, uint256 collateral) public pure returns (uint256) {
+        return debt.mulDivUp(ORACLE_PRICE_SCALE, collateral.mulDivDown(WAD, MAX_LIF));
     }
```
MathisGD: I think that you made this for roundings? I would rather have badDebtPriceUp (when you don't want bad debt) and badDebtPriceDown (when you want bad debt). it would be simpler

test/LiquidationTest.sol
```diff
@@ -269,16 +317,20 @@ contract LiquidationTest is BaseTest {
...
+        liquidationOraclePrice = bound(
+            liquidationOraclePrice, badDebtPrice(units, morphoV2.collateralOf(id, borrower, 0)), ORACLE_PRICE_SCALE
+        );
         Oracle(obligation.collaterals[0].oracle).setPrice(liquidationOraclePrice);
         vm.warp(obligation.maturity + TIME_TO_MAX_LIF + delay);
 
         uint256 initialCollateral = morphoV2.collateralOf(id, borrower, 0);
 
+        repaid = bound(repaid, 0, units);
+
```
MathisGD: to revert

test/LiquidationTest.sol
```diff
@@ -149,13 +159,18 @@ contract LiquidationTest is BaseTest {
...
         collateralize(obligation, borrower, units);
         setupObligation(obligation, units);
+        liquidationOraclePrice = bound(
+            liquidationOraclePrice, badDebtPrice(units, morphoV2.collateralOf(id, borrower, 0)), ORACLE_PRICE_SCALE
+        );
         Oracle(obligation.collaterals[0].oracle).setPrice(liquidationOraclePrice);
         vm.warp(obligation.maturity + TIME_TO_MAX_LIF); // Warp to post-maturity to bypass recovery close factor.
 
+        repaid = bound(repaid, 0, units);
+
```
MathisGD: to revert

test/LiquidationTest.sol
```diff
@@ -123,14 +128,19 @@ contract LiquidationTest is BaseTest {
...
         setupObligation(obligation, units);
         uint256 initialCollateral = morphoV2.collateralOf(id, borrower, 0);
+        liquidationOraclePrice =
+            bound(liquidationOraclePrice, badDebtPrice(units, initialCollateral), ORACLE_PRICE_SCALE);
         seized = bound(seized, 0, units.mulDivDown(ORACLE_PRICE_SCALE, liquidationOraclePrice).mulDivDown(MAX_LIF, WAD));
         Oracle(obligation.collaterals[0].oracle).setPrice(liquidationOraclePrice);
         vm.warp(obligation.maturity + TIME_TO_MAX_LIF); // Warp to post-maturity to bypass recovery close factor.
 
+        uint256 maxSeized = units.mulDivDown(ORACLE_PRICE_SCALE, 1e36 - 1).mulDivDown(MAX_LIF, WAD);
+        seized = bound(seized, 0, maxSeized > initialCollateral ? initialCollateral : maxSeized);
```
MathisGD: let's try to not bound twice the same variable.

test/LiquidationTest.sol
```diff
@@ -123,14 +128,19 @@ contract LiquidationTest is BaseTest {
...
     function testLiquidateCollateralInput(uint256 units, uint256 seized, uint256 liquidationOraclePrice) public {
         units = bound(units, 1, MAX_TEST_AMOUNT);
-        liquidationOraclePrice = bound(liquidationOraclePrice, badDebtPrice(), ORACLE_PRICE_SCALE - 1);
+
         collateralize(obligation, borrower, units);
         setupObligation(obligation, units);
         uint256 initialCollateral = morphoV2.collateralOf(id, borrower, 0);
+        liquidationOraclePrice =
+            bound(liquidationOraclePrice, badDebtPrice(units, initialCollateral), ORACLE_PRICE_SCALE);
         seized = bound(seized, 0, units.mulDivDown(ORACLE_PRICE_SCALE, liquidationOraclePrice).mulDivDown(MAX_LIF, WAD));
```
MathisGD: we should take the min with the collat

test/LiquidationTest.sol
```diff
@@ -485,10 +539,27 @@ contract LiquidationTest is BaseTest {
 
     // helpers.
 
+    /// @dev Bad debt as computed in liquidate: debt minus repayable (sum over all collaterals with mulDivDown).
```
MathisGD: let's do exactly like in the code

test/LiquidationTest.sol
```diff
@@ -209,55 +227,85 @@ contract LiquidationTest is BaseTest {
...
         assertEq(morphoV2.totalShares(id), units, "total shares");
     }
 
     function testLiquidateWithBadDebtRepaidInput(uint256 units, uint256 repaid, uint256 liquidationOraclePrice) public {
         units = bound(units, 10, MAX_TEST_AMOUNT); // if the amount is too small, no bad debt is created.
-        liquidationOraclePrice = bound(liquidationOraclePrice, 1, badDebtPrice() / 2);
         collateralize(obligation, borrower, units);
         setupObligation(obligation, units);
+
+        liquidationOraclePrice = bound(liquidationOraclePrice, 1, badDebtPrice());
```
MathisGD: to revert to the old position (so that we understand precisely the impact of this change)

test/LiquidationTest.sol
```diff
@@ -485,10 +539,27 @@ contract LiquidationTest is BaseTest {
 
     // helpers.
```
MathisGD, line 540: we could have a maxRepaid helper

test/LiquidationTest.sol
```diff
@@ -209,55 +227,85 @@ contract LiquidationTest is BaseTest {
...
-        uint256 expectedBadDebt = units - repayableDebt;
+
+        uint256 debtAfterBadDebt = units - _badDebt();
+        uint256 collatAmount = morphoV2.collateralOf(id, borrower, 0);
+        uint256 maxDebt = collatAmount.mulDivDown(liquidationOraclePrice, ORACLE_PRICE_SCALE)
+            .mulDivDown(obligation.collaterals[0].lltv, WAD);
+        uint256 maxRepaid =
+            (debtAfterBadDebt - maxDebt).mulDivUp(WAD, WAD - MAX_LIF.mulDivUp(obligation.collaterals[0].lltv, WAD));
+        repaid = bound(repaid, 0, debtAfterBadDebt);
+        repaid = bound(repaid, 0, maxRepaid);
```
MathisGD: isn't maxRepaid always smaller?

test/LiquidationTest.sol
```diff
@@ -209,55 +227,85 @@ contract LiquidationTest is BaseTest {
...
+        .mulDivUp(WAD, WAD - MAX_LIF.mulDivUp(obligation.collaterals[0].lltv, WAD));
+        uint256 maxCollateralRepayable =
+            collatAmount.mulDivDown(liquidationOraclePrice, ORACLE_PRICE_SCALE).mulDivDown(WAD, MAX_LIF);
+
+        uint256 repaidAmount = debtAfterBadDebt;
+        repaidAmount = repaidAmount > maxRepaid ? maxRepaid : repaidAmount; // capped by the recovery close factor
+        repaidAmount = repaidAmount > maxCollateralRepayable ? maxCollateralRepayable : repaidAmount; // capped by the
+        // collateral amount
+
+        morphoV2.liquidate(obligation, 0, 0, repaidAmount, borrower, "");
```
MathisGD: ideally should be like that
```suggestion
        morphoV2.liquidate(obligation, 0, 0, units, borrower, "");
```

test/LiquidationTest.sol
```diff
@@ -358,24 +410,26 @@ contract LiquidationTest is BaseTest {
     /// so repaying all remaining debt is allowed.
     function testMaxRepaidWithBadDebt(uint256 units, uint256 liquidationOraclePrice) public {
```
MathisGD: same as testLiquidateWithBadDebtRepayAll actually

test/LiquidationTest.sol
peyha: theoretically yes, maxRepaid as the RCF amount should be capped by the remaining debt but it's not the case when I fuzz, my guess is that it's due to roundings

test/LiquidationTest.sol
peyha: I was about to delete this test because it's no longer possible to repay all the debt when there is bad debt with bad debt overestimation

MathisGD: isn't it the opposite?

MathisGD: so this test is relevant

test/LiquidationTest.sol
```diff
@@ -375,50 +402,26 @@ contract LiquidationTest is BaseTest {
         assertLe(remainingDebt, newMaxDebt + 3, "position should be approximately just healthy after max repayment");
     }
 
-    /// @dev When price is low enough to create bad debt, maxRepaid >= debtAfterBadDebt,
-    /// so repaying all remaining debt is allowed.
-    function testMaxRepaidWithBadDebt(uint256 units, uint256 liquidationOraclePrice) public {
```
MathisGD, line 380: removed because duplicate

src/MorphoV2.sol
QGarchery: I don't understand this, I think it should still hold:
- we know that `originalDebt > maxDebt`, so without bad debt realization it's clear that `debt > maxDebt`
- in case of bad debt realization, then the `debt` is set to the "repayableDebt" (this concept doesn't exist in the code anymore, but it's useful for understanding here), which is the sum of `_collateralOf.mulDivDown(WAD, MAX_LIF).mulDivDown(price, ORACLE_PRICE_SCALE)`, which is greater than `maxDebt` which is the sum of `_collateralOf.mulDivDown(price, ORACLE_PRICE_SCALE).mulDivDown(_collateral.lltv, WAD)`. This is because each term of `repayableDebt` is greater than each term of `maxDebt` if there was no rounding error + the rounding errors are in the right direction

QGarchery: wait nevermind, the rounding don't go in the right direction necessarily, but if we switch the order then it should hold again

QGarchery: Done in https://github.com/morpho-org/morpho-v2/pull/414

MathisGD: how do we know that x.mulDivDown(lltv, WAD) <= x.mulDivDown(WAD, lif)? even if `lltv*lif <= WAD*WAD`, it's not obvious to me

peyha: I think we can prove it mathematically, x.mulDivDown(lltv, WAD) = floor(x*lltv/WAD) and x.mulDivDown(WAD, lif)=floor(x*WAD/lif). We have `lltv*lif <= WAD*WAD` so `lltv/WAD <= WAD/lif` so `x*lltv/WAD <= x*WAD/lif` and the floor function is increasing over real (rational in our case) numbers so we get the inequality

MathisGD: great, thanks. maybe worth a comment somewhere about all this?

QGarchery: yes let's not forget it, we came back to this topic already (see #129)

QGarchery (approved):

src/MorphoV2.sol

MathisGD (approved):

PR #373, opened 2026-02-17
[Certora] Added rule liquidateRequireUnhealthy

jhoenicke:
Check that if liquidate succeeds, the position must have been unhealthy before or reached maturity.

The rule needs a few summaries:

* `price` is deterministic
* `mulDiv*` are deterministic
* `toId` summary to figure out the id of an obligation.

The rule currently takes the guess and verify approach: We guess some non-deterministic id for isHealthy and after calling liquidate (which computes the id), we verify that we guessed the id correctly.

Alternatively one could add a harness to compute the Id, or make a deterministic summary for toId.

### DIFF
--- a/.github/workflows/certora.yml
+++ b/.github/workflows/certora.yml
@@ -15,6 +15,7 @@ jobs:
       matrix:
         conf:
           - Hasher
+          - Liquidate
           - MorphoV2
           - SharePrice
           - Solvency
--- a/certora/confs/Liquidate.conf
+++ b/certora/confs/Liquidate.conf
@@ -1,0 +1,19 @@
+{
+  "files": [
+    "src/MorphoV2.sol"
+  ],
+  "verify": "MorphoV2:certora/specs/Liquidate.spec",
+  "solc": "solc-0.8.31",
+  "solc_via_ir": true,
+  "solc_evm_version": "osaka",
+  "optimistic_loop": true,
+  "loop_iter": 2,
+  "optimistic_hashing": true,
+  "hashing_length_bound": 2048,
+  "prover_args": [
+    "-depth 5",
+    "-mediumTimeout 20",
+    "-timeout 3600"
+  ],
+  "msg": "MorphoV2 Liquidation"
+}
--- a/certora/specs/Liquidate.spec
+++ b/certora/specs/Liquidate.spec
@@ -1,0 +1,48 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+methods {
+    function multicall(bytes[]) external => HAVOC_ALL DELETE;
+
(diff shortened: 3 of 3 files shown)


MathisGD (approved):

certora/specs/MorphoV2.spec
QGarchery, line 1: There are some non-trivial conficts to resolve

certora/specs/MorphoV2.spec
```diff
@@ -9,11 +9,12 @@ methods {
...
+    function isHealthy(MorphoV2.Obligation obligation, bytes32 id, address borrower) external returns (bool) envfree;
 
-    function _.price() external => NONDET;
-    function IdLib.toId(MorphoV2.Obligation memory, uint256, address) internal returns (bytes32) => NONDET;
-    function UtilsLib.mulDivDown(uint256, uint256, uint256) internal returns (uint256) => NONDET;
-    function UtilsLib.mulDivUp(uint256, uint256, uint256) internal returns (uint256) => NONDET;
+    function _.price() external => CVL_price(calledContract) expect uint256;
+    function IdLib.toId(MorphoV2.Obligation memory obligation, uint256 chainId, address morphoV2) internal returns (bytes32) => CVL_toId(obligation, chainId, morphoV2);
+    function UtilsLib.mulDivDown(uint256 a, uint256 b, uint256 denominator) internal returns (uint256) => CVL_mulDivDown(a, b, denominator);
+    function UtilsLib.mulDivUp(uint256 a, uint256 b, uint256 denominator) internal returns (uint256) => CVL_mulDivUp(a, b, denominator);
```
QGarchery: We should put the rule in another file, because we don't want to make the assumption that the price cannot change for the other rules (if we don't need to)

jhoenicke: Okay, I will move it in a separate file.  This should also resolve conflicts.

It may also be good to check if it works with the new isHealthy implementation.   I had tried yesterday the Healthiness spec on the new isHealthy implementation.  There was a false positive in supplyCollateral, since xor is approximated and it assumed the xor in the loop could just return 0.  But here we only compare two pieces of code executing the same xor, so it should be able to reason that they do the same.

jhoenicke:
I did a force push, because the code moved to a different file anyway and cross-file diffs do not work well.

The changes to previous version are minimal:  
* bytes32 -> bytes20 for obligation id.
* deterministic summary for msb() to avoid loop unrolling vacuity.

QGarchery (approved):
LGTM (assuming it goes through)

MathisGD (approved):

bhargavbh (approved):

PR #375, opened 2026-02-17
test: improve liquidations tests

MathisGD:
fuzz prices

### DIFF
--- a/test/LiquidationTest.sol
+++ b/test/LiquidationTest.sol
@@ -41,45 +41,45 @@ contract LiquidationTest is BaseTest {
         morphoV2.liquidate(obligation, 2, 0, 0, borrower, "");
     }
 
-    function testLiquidateHealthyPreMaturity(uint256 units, uint256 oraclePrice) public {
+    function testLiquidateHealthyPreMaturity(uint256 units, uint256 liquidationOraclePrice) public {
         units = bound(units, 1, MAX_TEST_AMOUNT);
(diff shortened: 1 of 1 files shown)


test/LiquidationTest.sol
```diff
@@ -41,48 +41,45 @@ contract LiquidationTest is BaseTest {
         morphoV2.liquidate(obligation, 2, 0, 0, borrower, "");
     }
 
-    function testLiquidateHealthyPreMaturity(uint256 units) public {
+    function testLiquidateHealthyPreMaturity(uint256 units, uint256 oraclePrice) public {
```
adhusson:
```suggestion
    function testLiquidateHealthyPreMaturity(uint256 units, uint256 liquidationOraclePrice) public {
```

adhusson (approved):

test/LiquidationTest.sol
MathisGD: to distinguish with the opening one? as we are in "liquidationTest" it's not really ambiguous no?

adhusson: it's because of this ordering of operations:

```
oraclePrice = bound(oraclePrice, ORACLE_PRICE_SCALE, 10 * ORACLE_PRICE_SCALE);
Oracle(obligation.collaterals[0].oracle).setPrice(oraclePrice);
```

so the "oracle price" is in fact not the oracle price at borrow time, it becomes the oracle price after, which can be confusing.

PR #377, opened 2026-02-17
POC feat/ per user lending fee

MattLsb:
- Add a time-based Lending Fee for lenders with per-obligation/per-user state (fee, lastUpdate, accrued) 
- Implement maturity-clamped accrual, blended fee update on additional lender entries, and settlement on lender decreases (take/withdraw) with partial-payment carryover

### DIFF
--- a/src/MorphoV2.sol
+++ b/src/MorphoV2.sol
@@ -26,6 +26,12 @@ import {EventsLib} from "./libraries/EventsLib.sol";
 contract MorphoV2 is IMorphoV2 {
     using UtilsLib for uint256;
 
+    struct LenderFeeState {
+        uint192 accrued;
+        uint40 lastUpdate;
+        uint16 fee;
+    }
+
     /// STORAGE ///
 
     mapping(bytes32 id => mapping(address user => uint256)) public sharesOf;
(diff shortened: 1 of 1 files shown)


src/MorphoV2.sol
```diff
@@ -10,7 +10,8 @@ import {
     WAD,
     ORACLE_PRICE_SCALE,
     FEE_STEP,
-    MAX_FEE,
+    MAX_TRADING_FEE,
```
peyha, line 13: I think you forgot to stage the changes for ConstantsLib & EventsLib

src/MorphoV2.sol
```diff
@@ -281,9 +311,10 @@ contract MorphoV2 is IMorphoV2 {
         }
 
         SafeTransferLib.safeTransferFrom(
-            offer.obligation.loanToken, buyer, tradingFeeRecipient, buyerAssets - sellerAssets
+            offer.obligation.loanToken, buyer, feeRecipient, buyerAssets - sellerAssets + lendingFeePaid
```
QGarchery, line 314: This change breaks:
- the fact that the assets of the buyer are exact if `buyerAssets` is given as input of the protocol
- the fact that the maker price is guaranteed, if the maker is the buyer

MathisGD:
closing as we are converging for #469 vs #470

PR #378, opened 2026-02-18
[count bits] remove assumption

QGarchery:
The assumption is that the max number of collaterals is **strictly less than 255** (not 256), which is an invariant of the protocol but not immediately intuitive. Indeed, before this PR, if `MAX_COLLATERALS_PER_BORROWER` is 255 or more, then the check `UtilsLib.countBits(newBitmap) <= MAX_COLLATERALS_PER_BORROWER` is always true

Another option is to more explicitly rely on this assumption, by adding a comment. It also allows to make the `countBits` function slightly cheaper, see [this branch](https://github.com/morpho-org/morpho-v2/compare/main...refactor/faster-limit-collateral)

### DIFF
--- a/src/libraries/UtilsLib.sol
+++ b/src/libraries/UtilsLib.sol
@@ -68,14 +68,15 @@ library UtilsLib {
         return uint128(x);
     }
 
-    /// @dev Returns the number of set bits in x < 2^256-1, 0 otherwise.
+    /// @dev Returns the number of set bits.
     function countBits(uint256 x) internal pure returns (uint256) {
         unchecked {
             x = x - ((x >> 1) & 0x5555555555555555555555555555555555555555555555555555555555555555);
             x = (x & 0x3333333333333333333333333333333333333333333333333333333333333333)
                 + ((x >> 2) & 0x3333333333333333333333333333333333333333333333333333333333333333);
             x = (x + (x >> 4)) & 0x0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f;
-            return (x * 0x0101010101010101010101010101010101010101010101010101010101010101) >> 248;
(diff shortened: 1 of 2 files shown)


adhusson:
nice indeed in principle the check should be done before the increase

but active collats are stored on 128 bits, so we can't increase the max borrower collats enough

adhusson:
this https://github.com/morpho-org/morpho-v2/compare/main...refactor/faster-limit-collateral is great + a test that the fn is correct at `MAX_COLLATERALS_PER_BORROWER`, or that `MAX_COLLATERALS_PER_BORROWER` is not too big.

src/libraries/UtilsLib.sol
```diff
@@ -68,14 +68,15 @@ library UtilsLib {
...
-    /// @dev Returns the number of set bits in x < 2^256-1, 0 otherwise.
+    /// @dev Returns the number of set bits.
     function countBits(uint256 x) internal pure returns (uint256) {
         unchecked {
             x = x - ((x >> 1) & 0x5555555555555555555555555555555555555555555555555555555555555555);
             x = (x & 0x3333333333333333333333333333333333333333333333333333333333333333)
                 + ((x >> 2) & 0x3333333333333333333333333333333333333333333333333333333333333333);
             x = (x + (x >> 4)) & 0x0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f;
-            return (x * 0x0101010101010101010101010101010101010101010101010101010101010101) >> 248;
+            x = (x + (x >> 8)) & 0x00ff00ff00ff00ff00ff00ff00ff00ff00ff00ff00ff00ff00ff00ff00ff00ff;
```
MathisGD, line 78: with this new operation, I think that the previous version (with the while) was cheaper

MathisGD: why don't we just change the input to be uint128

QGarchery:
closed in favor of #383

PR #380, opened 2026-02-18
[count bits] faster limit collateral (WIP)

QGarchery:
See #378 for context

### DIFF
--- a/src/MorphoV2.sol
+++ b/src/MorphoV2.sol
@@ -365,10 +365,10 @@ contract MorphoV2 is IMorphoV2 {
         collateralOf[id][onBehalf][collateralToken] = newCollateralOf;
 
         if (newCollateralOf == assets && assets > 0) {
+            uint128 bitmap = borrowerState[id][onBehalf].activatedCollaterals;
+            require(UtilsLib.countBits(bitmap) < MAX_COLLATERALS_PER_BORROWER, "too many collaterals per borrower");
             // forge-lint: disable-next-item(unsafe-typecast) as collateralIndex < MAX_COLLATERALS (128)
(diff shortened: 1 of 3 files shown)


src/MorphoV2.sol
```diff
@@ -365,10 +365,10 @@ contract MorphoV2 is IMorphoV2 {
         collateralOf[id][onBehalf][collateralToken] = newCollateralOf;
 
         if (newCollateralOf == assets && assets > 0) {
+            uint128 bitmap = borrowerState[id][onBehalf].activatedCollaterals;
+            require(UtilsLib.countBits(bitmap) < MAX_COLLATERALS_PER_BORROWER, "too many collaterals per borrower");
             // forge-lint: disable-next-item(unsafe-typecast) as collateralIndex < MAX_COLLATERALS (128)
-            uint128 newBitmap = borrowerState[id][onBehalf].activatedCollaterals | uint128(1 << collateralIndex);
-            borrowerState[id][onBehalf].activatedCollaterals = newBitmap;
-            require(UtilsLib.countBits(newBitmap) <= MAX_COLLATERALS_PER_BORROWER, "too many collaterals per borrower");
```
MathisGD, line 371: why this change? to be able to have one more collat?

MathisGD (approved):

src/MorphoV2.sol
QGarchery: Main reason is that it is more natural to audit: you only need to check that `countBit` works on valid states, not on valid states that are potentially changed by one more bit

MathisGD: ok, didn't think about this

MathisGD: but then do we really want this change? like this makes things a bit harder, for like 10gas ?

adhusson (approved):

QGarchery:
closed in favor of #383

PR #383, opened 2026-02-18
refactor: safer count bits

MathisGD:
safer than #380


test/UtilsLibTest.sol
```diff
@@ -6,18 +6,15 @@ import {UtilsLib} from "../src/libraries/UtilsLib.sol";
...
-            assertEq(UtilsLib.countBits(bitmap), 0);
-        } else {
-            uint256 actual = UtilsLib.countBits(bitmap);
-            uint256 expected;
-            while (bitmap != 0) {
-                bitmap &= bitmap - 1;
-                expected++;
-            }
-            assertEq(actual, expected);
+    function testFuzzCountBits(uint128 bitmap) public pure {
```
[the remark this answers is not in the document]
MathisGD: this is usually fine, foundry fuzzes especially bounds

adhusson (approved):

test/LiquidationTest.sol
```diff
@@ -258,7 +258,7 @@ contract LiquidationTest is BaseTest {
         morphoV2.liquidate(obligation, 0, initialCollateral, 0, borrower, "");
 
         assertEq(morphoV2.collateralOf(id, borrower, obligation.collaterals[0].token), 0);
-        assertEq(UtilsLib.countBits(morphoV2.activatedCollaterals(id, borrower)), 0, "no bits should be set");
+        assertEq(UtilsLib.countBits(uint128(morphoV2.activatedCollaterals(id, borrower))), 0, "no bits should be set");
```
QGarchery: (optional) Wondering if `morphoV2.activatedCollaterals` should not return a `uint128`, to make it clear that the highest bits are 0, and that you can call `countBits` on it without the fear of an unsafe typecast

src/libraries/UtilsLib.sol
```diff
@@ -68,14 +68,12 @@ library UtilsLib {
...
         unchecked {
-            x = x - ((x >> 1) & 0x5555555555555555555555555555555555555555555555555555555555555555);
-            x = (x & 0x3333333333333333333333333333333333333333333333333333333333333333)
-                + ((x >> 2) & 0x3333333333333333333333333333333333333333333333333333333333333333);
-            x = (x + (x >> 4)) & 0x0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f;
-            return (x * 0x0101010101010101010101010101010101010101010101010101010101010101) >> 248;
+            x = x - ((x >> 1) & 0x55555555555555555555555555555555);
+            x = (x & 0x33333333333333333333333333333333) + ((x >> 2) & 0x33333333333333333333333333333333);
+            x = (x + (x >> 4)) & 0x0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f;
+            return (x * 0x01010101010101010101010101010101) >> 120;
```
QGarchery, line 76: Interestingly this last multiplication can't overflow anymore. So we don't rely on the behavior of `MUL` in case of an overflow anymore, which can be undefined behavior in some programming languages

QGarchery (approved):

PR #384, opened 2026-02-18
[refactor] generic create2 for main

adhusson:
to consider after #382

PR #387, opened 2026-02-19
generic create2 v2

adhusson:
alternative to #384

### DIFF
--- a/src/MorphoV2.sol
+++ b/src/MorphoV2.sol
@@ -551,7 +551,7 @@ contract MorphoV2 is IMorphoV2 {
 
             obligationState[id].created = true;
             obligationState[id].fees = defaultFees[obligation.loanToken];
-            UtilsLib.storeInCode(abi.encode(obligation), block.chainid);
+            IdLib.storeInCode(abi.encode(obligation), block.chainid);
 
             emit EventsLib.ObligationCreated(id, obligation);
         }
--- a/src/MorphoV2.sol
+++ b/src/MorphoV2.sol
@@ -538,8 +538,9 @@ contract MorphoV2 is IMorphoV2 {
 
     /// @dev Returns the obligation id and creates the obligation if it doesn't exist yet.
     function touchObligation(Obligation memory obligation) public returns (bytes32) {
-        bytes32 id = IdLib.toId(obligation, block.chainid, address(this));
-        if (!obligationState[id].created) {
+        bytes memory creationCode = UtilsLib.sstore2Code(abi.encode(obligation));
+        bytes32 id = UtilsLib.create2Hash(address(this), block.chainid, creationCode);
+        if (address(uint160(uint256(id))).code.length == 0) {
             require(obligation.collaterals.length <= MAX_COLLATERALS, "too many collaterals");
             address previousCollateralToken;
             for (uint256 i = 0; i < obligation.collaterals.length; i++) {
@@ -551,7 +552,8 @@ contract MorphoV2 is IMorphoV2 {
 
             obligationState[id].created = true;
             obligationState[id].fees = defaultFees[obligation.loanToken];
-            IdLib.storeInCode(obligation);
+            // The deployed code begins with 0x00 (STOP), because the first word is the offset of the obligation.
+            UtilsLib.create2Deploy(creationCode, block.chainid);
 
             emit EventsLib.ObligationCreated(id, obligation);
         }
@@ -559,6 +561,10 @@ contract MorphoV2 is IMorphoV2 {
     }
 
     /// VIEW FUNCTIONS ///
+
+    function idToObligation(bytes32 id) public view returns (Obligation memory) {
+        return IdLib.toObligation(id);
+    }
 
     function debtOf(bytes32 id, address user) external view returns (uint256) {
         return borrowerState[id][user].debt;
--- a/src/MorphoV2.sol
+++ b/src/MorphoV2.sol
@@ -551,8 +551,7 @@ contract MorphoV2 is IMorphoV2 {
 
             obligationState[id].created = true;
             obligationState[id].fees = defaultFees[obligation.loanToken];
-            // The deployed code begins with 0x00 (STOP), because the first word is the offset of the obligation.
-            UtilsLib.create2Deploy(abi.encode(obligation), block.chainid);
+            UtilsLib.storeInCode(abi.encode(obligation), block.chainid);
 
             emit EventsLib.ObligationCreated(id, obligation);
         }
@@ -561,8 +560,14 @@ contract MorphoV2 is IMorphoV2 {
 
     /// VIEW FUNCTIONS ///
 
-    function idToObligation(bytes32 id) public view returns (Obligation memory) {
+    /// @dev Returns the obligation with the given id if it was touched before.
+    /// @dev Reverts otherwise.
+    function toObligation(bytes32 id) public view returns (Obligation memory) {
         return IdLib.toObligation(id);
+    }
+
+    function toId(Obligation memory obligation) public view returns (bytes32) {
+        return IdLib.toId(obligation, block.chainid, address(this));
     }
 
     function debtOf(bytes32 id, address user) external view returns (uint256) {
--- a/src/MorphoV2.sol
+++ b/src/MorphoV2.sol
@@ -538,8 +538,7 @@ contract MorphoV2 is IMorphoV2 {
 
     /// @dev Returns the obligation id and creates the obligation if it doesn't exist yet.
     function touchObligation(Obligation memory obligation) public returns (bytes32) {
-        bytes memory creationCode = UtilsLib.sstore2Code(abi.encode(obligation));
-        bytes32 id = UtilsLib.create2Hash(address(this), block.chainid, creationCode);
+        bytes32 id = IdLib.toId(obligation, block.chainid, address(this));
         if (address(uint160(uint256(id))).code.length == 0) {
             require(obligation.collaterals.length <= MAX_COLLATERALS, "too many collaterals");
             address previousCollateralToken;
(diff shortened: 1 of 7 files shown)


src/libraries/UtilsLib.sol
```diff
@@ -2,6 +2,8 @@
 // Copyright (c) 2025 Morpho Association
 pragma solidity ^0.8.0;
 
+import {SSTORE2_PREFIX} from "./ConstantsLib.sol";
```
MathisGD: makes more sense to define it here imo

src/libraries/UtilsLib.sol
```diff
@@ -88,4 +90,14 @@ library UtilsLib {
         }
         return 0;
     }
+
+    /// @dev Deploys a contract with the given creation code and salt.
+    /// @dev It is recommended to give data that starts with STOP (0x00).
+    function create2Deploy(bytes memory data, uint256 salt) internal returns (address addr) {
```
MathisGD:
```suggestion
    function storeInCode(bytes memory data, uint256 salt) internal returns (address addr) {
```

src/MorphoV2.sol
```diff
@@ -551,7 +551,8 @@ contract MorphoV2 is IMorphoV2 {
 
             obligationState[id].created = true;
             obligationState[id].fees = defaultFees[obligation.loanToken];
-            IdLib.storeInCode(obligation);
+            // The deployed code begins with 0x00 (STOP), because the first word is the offset of the obligation.
```
MathisGD: we don't really care about this in MorphoV2 imo

src/MorphoV2.sol
```diff
@@ -560,6 +561,10 @@ contract MorphoV2 is IMorphoV2 {
 
     /// VIEW FUNCTIONS ///
 
+    function idToObligation(bytes32 id) public view returns (Obligation memory) {
```
MathisGD: why not the exact same name as the lib function

src/libraries/IdLib.sol
```diff
@@ -3,41 +3,18 @@
...
-    /// 60 0b     PUSH1 0x0b      [11, len, len]     code offset = 11
-    /// 5f        PUSH0           [0, 11, len, len]  mem offset = 0
-    /// 39        CODECOPY        [len]              mem[0:len] <- code[11:11+len]
-    /// 5f        PUSH0           [0, len]           return offset = 0
-    /// f3        RETURN          []                 mem[0:len] is returned
-    function creationCode(Obligation memory obligation) internal pure returns (bytes memory) {
-        return abi.encodePacked(hex"600b380380600b5f395ff3", abi.encode(obligation));
-    }
-
     function toId(Obligation memory obligation, uint256 chainId, address morphoV2) internal pure returns (bytes32) {
```
MathisGD, line 23: should MorphoV2 have a function for this too? not sure

src/libraries/UtilsLib.sol
adhusson: not too bad to break the constants in constantlib? for people looking for it

src/libraries/IdLib.sol
```diff
@@ -3,41 +3,18 @@
...
-        return keccak256(abi.encodePacked(uint8(0xff), morphoV2, bytes32(chainId), keccak256(creationCode(obligation))));
+        return keccak256(
+            abi.encodePacked(
+                uint8(0xff), morphoV2, chainId, keccak256(abi.encodePacked(SSTORE2_PREFIX, abi.encode(obligation)))
+            )
+        );
     }
 
-    function idToObligation(bytes32 id) internal view returns (Obligation memory) {
+    function toObligation(bytes32 id) internal view returns (Obligation memory) {
```
MathisGD, line 29: maybe explain that it revert if the obligation doesn't exist yet

src/libraries/IdLib.sol
adhusson: `toObligation` needs the state so you can't just reimplement it offchain you need to do rpc calls, or make an external call/code read.

`toId` is a pure function

so for me it's much less important to have, but ok with both

MathisGD: agree

src/MorphoV2.sol
```diff
@@ -560,6 +561,10 @@ contract MorphoV2 is IMorphoV2 {
...
+        return IdLib.toObligation(id);
```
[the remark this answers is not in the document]
MathisGD: maybe worth a comment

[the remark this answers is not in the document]
MathisGD: actually why not taking a `bytes20`

src/libraries/UtilsLib.sol
MathisGD: no I don't think so. you always prefer self contained stuff, just we want morphoV2 constants to be accessible from the outside so they are in a lib, but we would prefer to have them in morpho V2

src/MorphoV2.sol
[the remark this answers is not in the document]
QGarchery: We don't have an `obligationCreated` function anymore, so not sure this is really relevant. We have the event `ObligationCreated` that logs an `id` but I'm not sure how users can find that confusing. 
In any case a comment or taking a bytes20 works too (although we are getting pretty close to an address...)

src/libraries/UtilsLib.sol
adhusson: let me know if the update is what you had in mind

src/MorphoV2.sol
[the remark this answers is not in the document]
MathisGD: yes it's more to fix the weird behavior that multiple id can give the same obligation (while in reality no)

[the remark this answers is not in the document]
adhusson: - we do not care about the lost bytes since we already rely on create2 addresses not conflicting
- taking a bytes20 takes the upper bits which does not map to the create2 address (unless you mean defining the id as a bytes20 instead of bytes32, which works but is unusual)

src/MorphoV2.sol
```diff
@@ -551,7 +551,7 @@ contract MorphoV2 is IMorphoV2 {
 
             obligationState[id].created = true;
             obligationState[id].fees = defaultFees[obligation.loanToken];
-            IdLib.storeInCode(obligation);
+            UtilsLib.storeInCode(abi.encode(obligation), block.chainid);
```
QGarchery: I'm not sure to get it, don't we do twice `abi.encode(obligation)` in this PR ? The goal of 384 was to avoid that (at the cost of breaking the abstraction of the IdLib), so what do we gain here ?

QGarchery (changes requested):

src/MorphoV2.sol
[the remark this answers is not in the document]
MathisGD: we could do it by hand and take the last 160 bits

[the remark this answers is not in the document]
QGarchery: > taking a bytes20 takes the upper bits which does not map to the create2 address
 
I think the suggestion was to take the lower bits, so something like `bytes20(uint160(uint256(keccak256(...)))`

src/MorphoV2.sol
adhusson: the goal of the PR is to simplify idlib at the acknowledged cost of encoding twice

MathisGD: we said that it was fine (the opti is small compared to the cost of storeInCode anyway). the goal of this PR is to make idLib simpler by having a generic storeInCode

adhusson: we could also do nothing

src/libraries/UtilsLib.sol
MathisGD: yes that's great

src/libraries/IdLib.sol
```diff
@@ -3,41 +3,21 @@
...
+                morphoV2,
+                chainId,
+                keccak256(abi.encodePacked(UtilsLib.SSTORE2_PREFIX, abi.encode(obligation)))
+            )
+        );
     }
 
-    function idToObligation(bytes32 id) internal view returns (Obligation memory) {
+    function toObligation(bytes32 id) internal view returns (Obligation memory) {
         return abi.decode(address(uint160(uint256(id))).code, (Obligation));
```
peyha, line 30: what is the evm error raised when calling a non contract ? it could be worth implementing our own one such as "obligation not stored"

peyha: after testing on foundry, the error raised when there is no code is "EvmError: Revert" which is not really self explanatory

peyha:
```suggestion
        require(address(uint160(uint256(id))).code.length > 0, "obligation not stored");
        return abi.decode(address(uint160(uint256(id))).code, (Obligation));
```

adhusson: for me the error (which comes from abi decoding) is fine as is. If the function is called with an id that maps to an address with nonzero code, the check will pass but the ABI error will still occur, so the check would only return a special error message for
- valid ids of not yet stored obligations
- invalid ids that truncate to a contract with no code

peyha: good point, so there would be 3 situations which could make ABI decoding fail
1. The corresponding address has code but it can't be decoded as an Obligation
2. a. The corresponding address does not have code but it corresponds to a valid obligation (that has not been deployed yet)
2. b. The corresponding address does not have code and it's not a valid obligation (wrong chain, wrong contract or random id)

I was refering to 2.a. which is not likely to happen as long as `touchObligation` has to be done on all interactions. Might be worth re-considering if we go for #367 though

src/MorphoV2.sol
QGarchery: I don't find it simpler tbh. All else being equal the abstraction of the IdLib is a net benefit to me

MathisGD (approved):

src/MorphoV2.sol
MathisGD: @adhusson reverted this change now

src/libraries/IdLib.sol
```diff
@@ -5,7 +5,7 @@ pragma solidity ^0.8.0;
 import {Obligation} from "../interfaces/IMorphoV2.sol";
 
 library IdLib {
-    /// @dev Creation code that returns the code after the prefix as runtime bytecode.
+    /// @dev Creation code that deploys data as runtime bytecode.
```
QGarchery:
```suggestion
    /// @dev Used as a prefix to some data, to give a creation code that deploys the data as runtime bytecode.
```

QGarchery (approved):

PR #388, opened 2026-02-19
[Certora] Check that isHealthy is preserved if price does not change

jhoenicke:
This adds a new rule that isHealthy for some particular obligation,id, and user is preserved.  The callbacks are summarized and they check that isHealthy() also holds when giving control to the callbacks.  They then allow arbitrary state change as long as isHealthy() holds on return.

There are some manual struct copies, because the CVL language cannot store the obligation in ghost variables.  Also some interesting assumptions like that lltv is not too high (I got a counterexample in liquidate without the assumption).

[corresponding thread](https://morpholabs.slack.com/archives/C0ACSA8TNMA/p1771950444383039)


jhoenicke:
Prover struggles with the non-linear math here.  If I use mulDiv without summary the function liquidate() times out:
https://prover.certora.com/output/74728/59ee88526d7c4c1ab8e0dad29a88560a/?anonymousKey=943b5c20bdcbd7a6348d83fd9c05f84d68b7c66d

So I added axiomatic summaries for mulDivUp/Down; the math is really tricky here, especially for `liquidate()`

There are six mulDivs that need to be reasoned about to prove that liquidation doesn't create unhealthy positions, which makes the non-linear solvers really struggle.  Each isHealthy has two mulDivs, and the seizedAsset/repaidUnits computation has another two mulDivs.   And for the latter the ratio is different (WAD/lif instead of lltv/WAD) and the order they are applied differs.  So the solution was to insert just the right axioms that I needed for this case and do it separately from the other cases.

I also had to separate out the case where a different user/obligation is liquidated.

There is a new problem in supplyCollateral, which would require precise bitvector reasoning (which is infeasible  in combination with 256bit multiplications).  The problem is the new bitmap for the isHealthy check.  If we don't reason about xor precisely, the prover finds a counterexample where the xor in the iteration of the new collateral clears the bit of the previous collateral, so that it won't be iterated.   To avoid this, I added a new isHealthyNoBitmap() function with the old implementation.  We can prove that this is equivalent to isHealthy() later.

I excluded take() for now as it has a counterexample because of the callback bug.

certora/specs/Healthiness.spec
```diff
@@ -0,0 +1,261 @@
...
+
+function summaryMulDivUp(uint256 a, uint256 b, uint256 d) returns uint256 {
+    bool overflow;
+    if (overflow || d == 0) {
+        revert();
+    }
+    return require_uint256(summaryMulDivUpM(a, b, d));
+}
+
+// global variable indicating whether to use the optimized isHealthy() or the bitmap-less implementation
```
QGarchery, line 92: Let's explain that this can be chosen freely, because we prove the equivalence

certora/helpers/MulDiv.sol
QGarchery, line 1: We could put those in `certora/helpers/Utils.sol`

certora/specs/Healthiness.spec
```diff
@@ -0,0 +1,261 @@
...
+
+definition ORACLE_PRICE_SCALE() returns uint256 = 10 ^ 36;
+
+persistent ghost summaryPrice(address) returns uint256;
+
+persistent ghost summaryMulDivDownM(mathint, mathint, mathint) returns mathint {
+    /* mulDiv always returns an unsigned integer */
+    axiom forall mathint a. forall mathint b. forall mathint d. a >= 0 && b >= 0 && d > 0 => summaryMulDivDownM(a, b, d) >= 0;
+
+    /* proved in mulDivZero in MulDiv.spec */
```
QGarchery, line 48: I think it would make sense to have a `definition` for all of them (I see that there are a few below already). That definition would be used in the `assert`s in `MulDiv.spec` and in the `axiom`s here

certora/specs/Healthiness.spec
```diff
@@ -0,0 +1,261 @@
...
+function genericCallback() {
+    address dummy;
+    env e;
+    Midnight.Obligation obligation;
+
+    require obligation.loanToken == globalObligationLoanToken;
+    require obligation.collaterals.length == globalObligationCollateralLength;
+    require collateralMatches(obligation, 0);
+    require collateralMatches(obligation, 1);
+    require collateralMatches(obligation, 2);
```
QGarchery: IIUC, this essentially is the same as constraining the obligation to be the global obligation right ?

certora/specs/Healthiness.spec
```diff
@@ -0,0 +1,261 @@
...
+}
+
+rule stayHealthyLiquidateSameBorrower(env e, uint256 someCollateralIndex, uint256 someSeizedAssets, uint256 someRepaidUnits, bytes someData) {
+    Midnight.Obligation obligation;
+    useIsHealthyNoBitmap = true;
+
+    // reset the ghost variable that tracks whether the user was healthy before the callbacks.
+    healthyBeforeCallback = true;
+
+    require globalObligationCollateralLLTV[someCollateralIndex] * globalObligationCollateralMaxLif[someCollateralIndex] < WAD() * WAD(), "collateral lltv must be less then 1/maxLif";
```
QGarchery:
```suggestion
    require globalObligationCollateralLLTV[someCollateralIndex] * globalObligationCollateralMaxLif[someCollateralIndex] <= WAD() * WAD(), "collateral lltv must be less then 1/maxLif";
```
we proved this in `lifTimesLltvIsLessThanOrEqualToOne`, would it work here ?

certora/specs/Healthiness.spec
```diff
@@ -0,0 +1,261 @@
...
+    // require collateralMatches(obligation, 1);
+    // require collateralMatches(obligation, 2);
+
+    require someBorrower != globalBorrower || someObligation.loanToken != globalObligationLoanToken || someObligation.collaterals.length != globalObligationCollateralLength || !collateralMatches(someObligation, 0) || !collateralMatches(someObligation, 1) || !collateralMatches(someObligation, 2), "either user or obligation in the liquidation call is different";
+
+    require callIsHealthy(obligation, globalId, globalBorrower), "user is healthy before call";
+
+    uint256 seizedAssets;
+    uint256 repaidUnits;
+
```
QGarchery:
```suggestion
```

certora/specs/Healthiness.spec
```diff
@@ -0,0 +1,261 @@
...
+    assert healthyBeforeCallback, "user is healthy before callbacks";
+    assert callIsHealthy(obligation, globalId, globalBorrower), "user is healthy after call";
+}
+
+rule stayHealthy(env e, method f, calldataarg args) filtered { f -> f.selector != sig:liquidate(Midnight.Obligation, uint256, uint256, uint256, address, bytes).selector && f.selector != sig:take(uint256, address, address, bytes, address, Midnight.Offer, Midnight.Signature, bytes32, bytes32[]).selector } {
+    Midnight.Obligation obligation;
+    useIsHealthyNoBitmap = (f.selector != sig:withdrawCollateral(Midnight.Obligation, uint256, uint256, address, address).selector);
+
+    // reset the ghost variable that tracks whether the user was healthy before the callbacks.
```
QGarchery: This comment is not very informative, as it's describing what the next line does, not why it does it. I think what we want to say that it is only changed to `false` (and never back to true) whenever the global borrower is not healthy on the global obligation before any callback

certora/specs/Healthiness.spec
```diff
@@ -0,0 +1,261 @@
...
+    uint256 seizedAssets;
+    uint256 repaidUnits;
+
+    seizedAssets, repaidUnits = liquidate(e, someObligation, someCollateralIndex, someSeizedAssets, someRepaidUnits, someBorrower, someData);
+
+    assert healthyBeforeCallback, "user is healthy before callbacks";
+    assert callIsHealthy(obligation, globalId, globalBorrower), "user is healthy after call";
+}
+
+rule stayHealthy(env e, method f, calldataarg args) filtered { f -> f.selector != sig:liquidate(Midnight.Obligation, uint256, uint256, uint256, address, bytes).selector && f.selector != sig:take(uint256, address, address, bytes, address, Midnight.Offer, Midnight.Signature, bytes32, bytes32[]).selector } {
```
QGarchery, line 257: it's not clear to me what prevents us from writing it as a strong invariant

jhoenicke: One is Aesthetics:  stating the strong invariant "user is always healthy" feels wrong.  Of course, it only holds if price doesn't change.  It would also be an invariant that you cannot just require, as this would be unsound.

The other is that this would make splitting the cases for liquidate much harder.  

Finally, the proof currently depends on delaying the invariant check before the callbacks, by using the auxiliary ghost variable.  The reason we need that is that we have to require our axioms before the assert, but we cannot require them before liquidate (or we would need to compute seizedAssets/repaidDebt in CVL before).  Writing it as a strong invariant would immediately assert the invariant before the callbacks and thus before the axioms are required.

certora/specs/Healthiness.spec
jhoenicke: Yes, the problem is that CVL does not support storing solidity structs in ghost variables or ghost mappings.   Thus, I stored every field separately and for the array I had to compare every index separately (I had problems using forall here).

certora/specs/Healthiness.spec
jhoenicke: Yes that should should work.  I just rechecked MulDiv with the weaker precondition.

certora/specs/Healthiness.spec
jhoenicke: You can't `assert` the definition in MulDiv.spec, because the definition is about the CVL summaryMulDiv and the assert needs to be about the solidity mulDiv.

I will move these to their own definitions, though.

Do you want to make the MulDiv asserts syntactically closer, i.e., writing the precondition into the assert as an implication?

certora/specs/Healthiness.spec
```diff
@@ -0,0 +1,264 @@
...
+persistent ghost summaryPrice(address) returns uint256;
+
+persistent ghost summaryMulDivDownM(mathint, mathint, mathint) returns mathint;
+
+persistent ghost summaryMulDivUpM(mathint, mathint, mathint) returns mathint;
+
+/* Axioms that are proved by MulDiv.spec */
+
+/* proved in mulDivZero */
+definition axiomDownZero(mathint b, mathint d) returns bool = 0 <= b && 0 < d => summaryMulDivDownM(0, b, d) == 0;
```
lilCertora: this is never used no ?

certora/specs/Healthiness.spec
```diff
@@ -0,0 +1,264 @@
...
+definition axiomUpMonotoneA(mathint a1, mathint a2, mathint b, mathint d) returns bool = 0 <= a1 && a1 <= a2 && 0 <= b && 0 < d => summaryMulDivUpM(a1, b, d) <= summaryMulDivUpM(a2, b, d);
+
+/* proved in mulDivMonotoneB */
+definition axiomDownMonotoneB(mathint a, mathint b1, mathint b2, mathint d) returns bool = 0 <= a && 0 <= b1 && b1 <= b2 && 0 < d => summaryMulDivDownM(a, b1, d) <= summaryMulDivDownM(a, b2, d);
+
+/* proved in mulDivMonotoneD */
+definition axiomUpMonotoneD(mathint a, mathint b, mathint d1, mathint d2) returns bool = 0 <= a && 0 <= b && 0 < d1 && d1 <= d2 => summaryMulDivUpM(a, b, d1) >= summaryMulDivUpM(a, b, d2);
+
+/* proved in mulDivAddDownUp in MulDiv.spec */
+definition axiomAddDownUp(mathint a1, mathint a2, mathint b, mathint d) returns bool = d > 0 => summaryMulDivDownM(a1, b, d) + summaryMulDivUpM(a2, b, d) >= summaryMulDivDownM(a1 + a2, b, d);
```
lilCertora: do we need to have a1 >= 0, a2 >= 0, b >= 0 or a1 + a2 in uint256 ? this is fine because solidity would revert in such a case but just curious on your reflexion

lilCertora (approved):

certora/specs/Healthiness.spec
QGarchery: let's also require that the maturity is the same, as it makes it easier to think about the rules in this spec

certora/specs/Healthiness.spec
QGarchery: much better now !

certora/specs/CollateralBitmap.spec
```diff
@@ -0,0 +1,53 @@
...
+
+methods {
+    function multicall(bytes[]) external => HAVOC_ALL DELETE;
+
+    function collateralOf(bytes32 id, address user, uint256) external returns (uint128) envfree;
+    function collateralBitSet(bytes32, address, uint256) external returns (bool) envfree;
+    function isHealthy(Midnight.Obligation, bytes32, address) external returns (bool) envfree;
+    function isHealthyNoBitmap(Midnight.Obligation, bytes32, address) external returns (bool) envfree;
+
+    function _.price() external => summaryPrice(calledContract) expect(uint256);
```
QGarchery: We could have PER_CALLEE_CONSTANT here I think. Let's explain why we make this assumption (it's only to allow to have a deterministic execution to be able to compare with and without bitmap)

certora/specs/CollateralBitmap.spec
```diff
@@ -0,0 +1,53 @@
...
+    function TickLib.tickToPrice(uint256 tick) internal returns (uint256) => NONDET;
+    function IdLib.toId(Midnight.Obligation memory obligation, uint256 chainId, address morpho) internal returns (bytes32) => NONDET;
+    function UtilsLib.mulDivDown(uint256 x, uint256 y, uint256 d) internal returns (uint256) => summaryMulDivDown(x, y, d);
+    function UtilsLib.mulDivUp(uint256 x, uint256 y, uint256 d) internal returns (uint256) => summaryMulDivUp(x, y, d);
```
QGarchery, line 24: Let's also explain why we have summaries here: simplify the run, while still having some axioms about the mulDiv

certora/specs/CollateralBitmap.spec
```diff
@@ -0,0 +1,53 @@
...
+
+// Check that a collateral bit is set exactly when there is collateral for that index.
+invariant bitsetIffCollateral(bytes32 id, address borrower, uint256 idx)
+    idx < 128 => (collateralBitSet(id, borrower, idx) <=> collateralOf(id, borrower, idx) != 0);
+
+// This shows that the real isHealthy returns true if and only if the isHealthy function
+// that does not use collateral bitmap returns true.  We also check that the latter function
+// does not revert if isHealthy does not revert.
+rule isHealthyEquivalent(Midnight.Obligation obligation, bytes32 id, address borrower) {
+    // We restrict to at most three collaterals
```
QGarchery:
```suggestion
```

certora/specs/Healthiness.spec
```diff
@@ -0,0 +1,264 @@
...
+    function _.onSell(Midnight.Obligation obligation, address seller, uint256 buyerAssets, uint256 sellerAssets, uint256 obligationUnits, uint256 obligationShares, bytes data) external => genericCallback() expect void;
+    function _.onLiquidate(Midnight.Obligation obligation, uint256 collateralIndex, uint256 seizedAssets, uint256 repaidUnits, address borrower, bytes data) external => genericCallback() expect void;
+    function _.onFlashLoan(address token, uint256 amount, bytes data) external => genericCallback() expect void;
+}
+
+/// ASSUMPTIONS ///
+
+// price does not change (isHealthy() can be violated if price changes)
+// isHealthy() and isHealthyNoBitmap() are equivalent (proved in CollateralBitmap.spec)
+// mulDivDown/Up() fulfill all the axioms defined here (proved in MulDiv.spec)
```
QGarchery: Let's put those assumptions closer to where they are used/defined:
- the price comment can be in the method's block, on the `_.price` summary
- the mulDiv comment can also be in the method's block, on the respective summaries
- the isHealthy one can be removed, as it's already written above `callIsHealthy`

certora/specs/Healthiness.spec
```diff
@@ -0,0 +1,264 @@
...
+// and then we have a final rule for all other functions of the contract.
+
+// Show that the user stays healthy on liquidate, if the user gets liquidated (can occur if blocktime exceeds maturity)
+rule stayHealthyLiquidateSameBorrower(env e, uint256 someCollateralIndex, uint256 someSeizedAssets, uint256 someRepaidUnits, bytes someData) {
+    useIsHealthyNoBitmap = true;
+
+    // This variable is set to false whenever isHealthy() is violated before a callback.  Initially we set it to true to indicate no violations detected.
+    healthyBeforeCallback = true;
+
+    require globalObligationCollateralLLTV[someCollateralIndex] * globalObligationCollateralMaxLif[someCollateralIndex] <= WAD() * WAD(), "collateral lltv must be less then 1/maxLif";
```
QGarchery: Let's mention the rule that proves it in the reason string

certora/specs/Healthiness.spec
```diff
@@ -0,0 +1,264 @@
...
+// Show that the user stays healthy on liquidate, if another user gets liquidated or obligation differs.
+rule stayHealthyLiquidateOtherBorrower(env e, Midnight.Obligation someObligation, uint256 someCollateralIndex, uint256 someSeizedAssets, uint256 someRepaidUnits, address someBorrower, bytes someData) {
+    useIsHealthyNoBitmap = true;
+
+    // This variable is set to false whenever isHealthy() is violated before a callback.  Initially we set it to true to indicate no violations detected.
+    healthyBeforeCallback = true;
+
+    require globalObligationCollateralLength <= 1, "too many collaterals for the spec to handle";
+
+    Midnight.Obligation obligation = getGlobalObligation();
```
QGarchery: nitpick
```suggestion
    Midnight.Obligation globalObligation = getGlobalObligation();
```
and then that allows to remove all the `some` prefixes

QGarchery: if we do it here, then let's do it everywhere else

certora/specs/Healthiness.spec
```diff
@@ -0,0 +1,264 @@
...
+
+persistent ghost summaryMulDivDownM(mathint, mathint, mathint) returns mathint;
+
+persistent ghost summaryMulDivUpM(mathint, mathint, mathint) returns mathint;
+
+/* Axioms that are proved by MulDiv.spec */
+
+/* proved in mulDivZero */
+definition axiomDownZero(mathint b, mathint d) returns bool = 0 <= b && 0 < d => summaryMulDivDownM(0, b, d) == 0;
+
```
QGarchery:
```suggestion
```
this one seems unused

certora/specs/Healthiness.spec
```diff
@@ -0,0 +1,264 @@
...
+/* proved in mulDivLifLLTV in MulDiv.spec */
+definition axiomLifLLTV(mathint a, mathint lif, mathint lltv) returns bool = lltv * lif <= WAD() * WAD() => summaryMulDivUpM(a, lltv, WAD()) <= summaryMulDivUpM(a, WAD(), lif);
+
+function summaryMulDivDown(uint256 a, uint256 b, uint256 d) returns uint256 {
+    bool overflow;
+    if (overflow || d == 0) {
+        revert();
+    }
+    return require_uint256(summaryMulDivDownM(a, b, d));
```
QGarchery, line 81: We should document that this is a real assumption, that reduces the search space (we don't consider cases where it would overflow in the spec and not in the code), same as the [comment on the isHealthy equivalence](https://github.com/morpho-org/midnight/pull/493#discussion_r2911838315). It's still a very interesting rule to have for sure !

certora/specs/Healthiness.spec
```diff
@@ -0,0 +1,264 @@
...
+    require callIsHealthy(obligation, globalId, globalBorrower), "user is healthy before call";
+
+    liquidate(e, someObligation, someCollateralIndex, someSeizedAssets, someRepaidUnits, someBorrower, someData);
+
+    assert healthyBeforeCallback, "user is healthy before callbacks";
+    assert callIsHealthy(obligation, globalId, globalBorrower), "user is healthy after call";
+}
+
+// Show that the user stays healthy on any other function than liquidate or take.
+rule stayHealthy(env e, method f, calldataarg args) filtered { f -> f.selector != sig:liquidate(Midnight.Obligation, uint256, uint256, uint256, address, bytes).selector && f.selector != sig:take(uint256, address, address, bytes, address, Midnight.Offer, Midnight.Signature, bytes32, bytes32[]).selector } {
```
QGarchery, line 257: ok to merge and to keep in mind that we should [remove the filtering on `take` later](https://github.com/morpho-org/midnight/issues/496#issuecomment-4046689866)

certora/specs/Healthiness.spec
```diff
@@ -0,0 +1,264 @@
...
+
+definition axiomUpMonotoneA(mathint a1, mathint a2, mathint b, mathint d) returns bool = 0 <= a1 && a1 <= a2 && 0 <= b && 0 < d => summaryMulDivUpM(a1, b, d) <= summaryMulDivUpM(a2, b, d);
+
+/* proved in mulDivMonotoneB */
+definition axiomDownMonotoneB(mathint a, mathint b1, mathint b2, mathint d) returns bool = 0 <= a && 0 <= b1 && b1 <= b2 && 0 < d => summaryMulDivDownM(a, b1, d) <= summaryMulDivDownM(a, b2, d);
+
+/* proved in mulDivMonotoneD */
+definition axiomUpMonotoneD(mathint a, mathint b, mathint d1, mathint d2) returns bool = 0 <= a && 0 <= b && 0 < d1 && d1 <= d2 => summaryMulDivUpM(a, b, d1) >= summaryMulDivUpM(a, b, d2);
+
+/* proved in mulDivAddDownUp in MulDiv.spec */
```
QGarchery:
```suggestion
/* proved in mulDivAddDownUp */
```
and same in the ones below, it's already written above

certora/specs/Healthiness.spec
QGarchery: duplicate of[ this](https://github.com/morpho-org/midnight/pull/388#discussion_r2923472110)

certora/specs/Healthiness.spec
jhoenicke: I added that all numbers are non-negative.  The same for the other axioms.

certora/specs/Healthiness.spec
jhoenicke: Not any longer.  It turned out monotonicity is enough to prove that adding a new collateral is fine.

It's still required for the CollateralBitmap spec to prove the equivalence of the two implementations of isHealthy().

certora/specs/Healthiness.spec
jhoenicke: What is the assumption?  The function can non-deterministically overflow and revert, so it should cover all the real behavior.   The `require_uint256` just eliminates the path where the `overflow` value was incorrectly guessed by the prover or where the value of summaryMulDivDownM was incorrectly chosen by the prover.

QGarchery: oh right good point, it's covered by the overflow 💯

certora/specs/Healthiness.spec
QGarchery: since it is already defined in CollateralBitmap, we can remove it from here I think

QGarchery (approved):

certora/helpers/MulDiv.sol
QGarchery: let's not do it for now, as this PR is ready

QGarchery (approved):

certora/specs/CollateralBitmap.spec
```diff
@@ -0,0 +1,58 @@
...
+// This shows that the real isHealthy returns true if and only if the isHealthy function
+// that does not use collateral bitmap returns true.  We also check that the latter function
+// does not revert if isHealthy does not revert.
+rule isHealthyEquivalent(Midnight.Obligation obligation, bytes32 id, address borrower) {
+    require obligation.collaterals.length <= 3, "restrict to three collaterals";
+    requireInvariant bitsetIffCollateral(id, borrower, 0);
+    requireInvariant bitsetIffCollateral(id, borrower, 1);
+    requireInvariant bitsetIffCollateral(id, borrower, 2);
+
+    // We make no claim about isHealthyNoBitmap() if isHealthy() reverts.
```
bhargavbh, line 49: why is this sufficient? should we also not show that the revert conditions of isHealthy() and isHealthyBitmap() are the same to claim they are equivalent?

certora/specs/Healthiness.spec
```diff
@@ -0,0 +1,266 @@
...
+
+    // require all the axioms that are needed to prove the healthiness after liquidation. These are the same axioms that are proved in the MulDiv.spec
+    require forall mathint a1. forall mathint a2. forall mathint b. forall mathint d. axiomDownMonotoneA(a1, a2, b, d), "axiom";
+    require forall mathint a1. forall mathint a2. forall mathint b. forall mathint d. axiomUpMonotoneA(a1, a2, b, d), "axiom";
+    require forall mathint a. forall mathint b1. forall mathint b2. forall mathint d. axiomDownMonotoneB(a, b1, b2, d), "axiom";
+    require forall mathint a. forall mathint b. forall mathint d1. forall mathint d2. axiomUpMonotoneD(a, b, d1, d2), "axiom";
+    require axiomInverseUpDown(repaidUnitsOut, globalObligationCollateralMaxLif[collateralIndex], WAD()), "axiom";
+    require axiomInverseUpDown(summaryMulDivDownM(repaidUnitsOut, globalObligationCollateralMaxLif[collateralIndex], WAD()), ORACLE_PRICE_SCALE(), price), "axiom";
+    require axiomLifLLTV(summaryMulDivUpM(seizedAssetsOut, price, ORACLE_PRICE_SCALE()), globalObligationCollateralMaxLif[collateralIndex], globalObligationCollateralLLTV[collateralIndex]), "axiom";
```
bhargavbh, line 227: lif * lltv < WAD*WAD is not proven nor defined in this spec. So how is the axiomLifLLTV useful here? the prover realises lif*lltv<1 automatically?

bhargavbh (approved):
some clarification comments

certora/specs/CollateralBitmap.spec
QGarchery: [Relevant thread](https://github.com/morpho-org/midnight/pull/493#discussion_r2911838315). The way I understand it we don't prove the equivalence, which can reduce the search space more than we'd like if we call `isHealthyNoBitmap` before calling a function that calls `isHealthy`.
So it's not perfect but overall the rule is still useful to catch bugs

certora/specs/Healthiness.spec
QGarchery: It's proven in another spec, and required here so the prover knows that `lif * lltv < 1` which is useful to bound the lif (otherwise you might be seizing so much that you make the position unhealthy)

certora/helpers/MidnightWrapper.sol
```diff
@@ -0,0 +1,35 @@
...
+            uint256 price = IOracle(collateral.oracle).price();
+            maxDebt += collateralOf[id][borrower][i].mulDivDown(price, ORACLE_PRICE_SCALE)
+                .mulDivDown(collateral.lltv, WAD);
+        }
+        return maxDebt >= debt;
+    }
+
+    function collateralBitSet(bytes32 id, address borrower, uint256 idx) external view returns (bool) {
+        return (borrowerState[id][borrower].activatedCollaterals & uint128(1 << idx)) != 0;
+    }
```
MathisGD:
```suggestion
    function collateralBitSet(bytes32 id, address borrower, uint256 index) external view returns (bool) {
        return (borrowerState[id][borrower].activatedCollaterals & uint128(1 << index)) != 0;
    }
```

MathisGD:
@codex review

certora/specs/CollateralBitmap.spec
```diff
@@ -0,0 +1,55 @@
...
+     */
+    function UtilsLib.mulDivDown(uint256 x, uint256 y, uint256 d) internal returns (uint256) => summaryMulDivDown(x, y, d);
+    function UtilsLib.mulDivUp(uint256 x, uint256 y, uint256 d) internal returns (uint256) => summaryMulDivUp(x, y, d);
+}
+
+/// SUMMARY ///
+
+persistent ghost summaryMulDivDown(uint256, uint256, uint256) returns uint256 {
+    /* proved in mulDivZero in MulDiv.spec */
+    axiom forall uint256 b. forall uint256 d. d > 0 => summaryMulDivDown(0, b, d) == 0;
```
MathisGD, line 31: (minor) what's the difference between this? is there a reason to use one or the other, other than readability?
```suggestion
    /* proved in mulDivZero in MulDiv.spec */
    uint256 res;
    require d > 0 => (x == 0 => res == 0);
```

certora/specs/CollateralBitmap.spec
```diff
@@ -0,0 +1,55 @@
...
+}
+
+persistent ghost summaryMulDivUp(uint256, uint256, uint256) returns uint256;
+
+// Check that a collateral bit is set exactly when there is collateral for that index.
+invariant bitsetIffCollateral(bytes32 id, address borrower, uint256 idx)
+    idx < 128 => (collateralBitSet(id, borrower, idx) <=> collateralOf(id, borrower, idx) != 0);
```
MathisGD: (minor)
```suggestion
invariant bitsetIffCollateral(bytes32 id, address borrower, uint256 index)
    index < 128 => (collateralBitSet(id, borrower, index) <=> collateralOf(id, borrower, index) != 0);
```

certora/specs/CollateralBitmap.spec
```diff
@@ -0,0 +1,55 @@
...
+
+persistent ghost summaryMulDivDown(uint256, uint256, uint256) returns uint256 {
+    /* proved in mulDivZero in MulDiv.spec */
+    axiom forall uint256 b. forall uint256 d. d > 0 => summaryMulDivDown(0, b, d) == 0;
+}
+
+persistent ghost summaryMulDivUp(uint256, uint256, uint256) returns uint256;
+
+// Check that a collateral bit is set exactly when there is collateral for that index.
+invariant bitsetIffCollateral(bytes32 id, address borrower, uint256 idx)
```
MathisGD: how does this invariant overlap with `nonZeroCollateralsAreActivated`?

certora/specs/Healthiness.spec
```diff
@@ -0,0 +1,266 @@
...
+    require callIsHealthy(globalObligation, globalId, globalBorrower), "user is healthy before call";
+
+    liquidate(e, obligation, collateralIndex, seizedAssets, repaidUnits, borrower, data);
+
+    assert healthyBeforeCallback, "user is healthy before callbacks";
+    assert callIsHealthy(globalObligation, globalId, globalBorrower), "user is healthy after call";
+}
+
+// Show that the user stays healthy on any other function than liquidate or take.
+rule stayHealthy(env e, method f, calldataarg args) filtered { f -> f.selector != sig:liquidate(Midnight.Obligation, uint256, uint256, uint256, address, bytes).selector && f.selector != sig:take(uint256, address, address, bytes, address, Midnight.Offer, Midnight.Signature, bytes32, bytes32[]).selector } {
```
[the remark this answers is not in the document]
jhoenicke: The property is still violated by take().  Waiting for the fix.

certora/specs/CollateralBitmap.spec
jhoenicke: It's the same invariant.  The only difference is that this is proved using bit-precise reasoning.

We may merge the files; I'm thinking of making BitmapSummaries just containing the summaries, importing it here and then only have the invariant in this file.

certora/specs/Healthiness.spec
```diff
@@ -0,0 +1,266 @@
...
+    Midnight.Obligation globalObligation = getGlobalObligation();
+
+    // check that isHealthy holds before the callback.  We remember any violation and check that none occurred at the end of each rule.
+    if (!callIsHealthy(globalObligation, globalId, globalBorrower)) {
+        healthyBeforeCallback = false;
+    }
+
+    callback.callHavoc(e, dummy);
+
+    require callIsHealthy(globalObligation, globalId, globalBorrower), "user is healthy after callback";
```
MathisGD, line 175: I don't get why is that useful?

certora/specs/Healthiness.spec
```diff
@@ -0,0 +1,266 @@
...
+function genericCallbackBool() returns (bool) {
+    bool result;
+    genericCallback();
+    return result;
+}
+
+//// RULES //////
+
+// The remaining rules show that a healthy borrower cannot get unhealthy by calling any function of the contract.
+// Since we have a ghost summary for price(), we assume the price will not change during the call.
```
MathisGD, line 188: we never call twice the oracle in any function, so why is that useful?

certora/specs/Healthiness.spec
```diff
@@ -0,0 +1,266 @@
...
+
+    liquidate(e, obligation, collateralIndex, seizedAssets, repaidUnits, borrower, data);
+
+    assert healthyBeforeCallback, "user is healthy before callbacks";
+    assert callIsHealthy(globalObligation, globalId, globalBorrower), "user is healthy after call";
+}
+
+// Show that the user stays healthy on any other function than liquidate or take.
+rule stayHealthy(env e, method f, calldataarg args) filtered { f -> f.selector != sig:liquidate(Midnight.Obligation, uint256, uint256, uint256, address, bytes).selector && f.selector != sig:take(uint256, address, address, bytes, address, Midnight.Offer, Midnight.Signature, bytes32, bytes32[]).selector } {
+    // for withdraw collateral we choose isHealthy() for all others the isHealthyNoBitmap function.
```
MathisGD, line 258: what is the intuition for why this?

certora/specs/Healthiness.spec
```diff
@@ -0,0 +1,266 @@
...
+}
+
+// Show that the user stays healthy on liquidate, if another user gets liquidated or obligation differs.
+rule stayHealthyLiquidateOtherBorrower(env e, Midnight.Obligation obligation, uint256 collateralIndex, uint256 seizedAssets, uint256 repaidUnits, address borrower, bytes data) {
+    useIsHealthyNoBitmap = true;
+
+    // This variable is set to false whenever isHealthy() is violated before a callback.  Initially we set it to true to indicate no violations detected.
+    healthyBeforeCallback = true;
+
+    require globalObligationCollateralLength <= 1, "too many collaterals for the spec to handle";
```
MathisGD: curious that the prover doesn't see easily that we don't touch the state of other borrowers. is there anything we could do to help it?

MathisGD (approved):
🔥

certora/specs/Healthiness.spec
jhoenicke: it looks like it works with 2 again.  I had problems in the past, but the code changed a lot in between.

certora/specs/MulDiv.spec
```diff
@@ -0,0 +1,50 @@
...
+    function mulDivUp(uint256 a, uint256 b, uint256 d) external returns (uint256) envfree;
+}
+
+/// RULES ///
+
+/* these proves the axiom used in the other specs */
+
+rule mulDivZero(uint256 b, uint256 d) {
+    assert mulDivDown(0, b, d) == 0;
+    assert mulDivUp(0, b, d) == 0;
```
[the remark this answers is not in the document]
jhoenicke: This is okay.  CVL semantics mean that the asserts are only checked for non-reverting runs.

certora/specs/Healthiness.spec
jhoenicke: We assume here that the callback can do anything on the morpho contract, including adding/removing collateral, withdrawing funds, etc.  We model this by callHavoc, which is HAVOC_ALL and therefore changes all variables.   But then we would not be able to prove that the user is healthy at the end, e.g. after a flashloan.  But by induction we have already proved that everything the callback can call will not change the property isHealthy().  So we can require that the user is healthy here and use it to prove that he is still healthy in the end.

certora/specs/Healthiness.spec
jhoenicke: The rule itself calls the oracle multiple times by calling isHealthy. All these calls must return the same price.   We also use the price in the liquidate rule to help the prover and here we also need that the price is consistent.

certora/specs/Healthiness.spec
jhoenicke: For withdraw collateral, the function itself calls isHealthy() (once take works, it would also be the same).  So here it's much simpler to prove that calling isHealthy() again yields the same result.

For supply collateral it is the other way round: with bitmap reasoning we would need to show that adding the bit to the collateral bitmap will not clear previously set bit and would require bit-precise reasoning (now we may be able to use the summaries instead).  With isHealthyNoBitmap() it's obvious that all collaterals are iterated.

certora/specs/CollateralBitmap.spec
jhoenicke: The ghost summaryMulDivDown is not an executable function, but a logical function.   The main feature is that it will always return the same value, which is important for the verification.   Using the axiom one can restrict the behaviour of the function.

The syntax you posted is for CVL functions and would be a nondeterministic return value that can be different for every function call even if the input arguments are the same.

certora/specs/CollateralBitmap.spec
jhoenicke: I looked into merging BitmapSummaries into this file here:  https://github.com/morpho-org/midnight/pull/569

This makes this change obsolete.

certora/specs/Healthiness.spec
```diff
@@ -0,0 +1,266 @@
...
+     */
+    function UtilsLib.mulDivDown(uint256 x, uint256 y, uint256 d) internal returns (uint256) => summaryMulDivDown(x, y, d);
+    function UtilsLib.mulDivUp(uint256 x, uint256 y, uint256 d) internal returns (uint256) => summaryMulDivUp(x, y, d);
+    function _.havocAll() external => HAVOC_ALL;
+
+    function _.transferFrom(address from, address to, uint256 amount) external with(env e) => genericCallbackBool() expect(bool);
+    function _.transfer(address to, uint256 amount) external with(env e) => genericCallbackBool() expect(bool);
+    function _.onBuy(Midnight.Obligation obligation, address buyer, uint256 buyerAssets, uint256 sellerAssets, uint256 units, bytes data) external => genericCallback() expect void;
+    function _.onSell(Midnight.Obligation obligation, address seller, uint256 buyerAssets, uint256 sellerAssets, uint256 units, bytes data) external => genericCallback() expect void;
+    function _.onLiquidate(Midnight.Obligation obligation, uint256 collateralIndex, uint256 seizedAssets, uint256 repaidUnits, address borrower, bytes data) external => genericCallback() expect void;
```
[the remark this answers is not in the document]
MathisGD: maybe we could have a sanity check that the summary is used (usable) when the summary is useful for verification

certora/specs/Healthiness.spec
```diff
@@ -0,0 +1,266 @@
...
+// Show that the user stays healthy on liquidate, if the user gets liquidated (can occur if blocktime exceeds maturity)
+rule stayHealthyLiquidateSameBorrower(env e, uint256 collateralIndex, uint256 seizedAssetsIn, uint256 repaidUnitsIn, bytes data) {
+    useIsHealthyNoBitmap = true;
+
+    // This variable is set to false whenever isHealthy() is violated before a callback.  Initially we set it to true to indicate no violations detected.
+    healthyBeforeCallback = true;
+
+    require globalObligationCollateralLLTV[collateralIndex] * globalObligationCollateralMaxLif[collateralIndex] <= WAD() * WAD(), "Proved in lifTimesLltvIsLessThanOrEqualToOne in ExactMath.spec: maxLif is at most 1/lltv";
+
+    require globalObligationCollateralLength <= 1, "too many collaterals for the spec to handle";
```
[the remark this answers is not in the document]
MathisGD: maybe 2 passes too here?

certora/specs/Healthiness.spec
MathisGD: ohh sorry it's a require not an assert (always get confused ^^)

certora/specs/Healthiness.spec
MathisGD: we could have had this one be a strong invariant and avoid [this](https://github.com/morpho-org/midnight/pull/388/changes/1008d5edbd3d6702bb34e5cf41bc5834ae0b4c6b..2d91b6720bc3b24b92dc3f313681f3daa56b7094#r2988755877), but the liquidation one be a weak invariant. doesn't matter that much anyway

certora/specs/CollateralBitmap.spec
MathisGD: right, got mistaken between ghost and function

certora/specs/Healthiness.spec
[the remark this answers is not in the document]
jhoenicke: I just tried and now get a counterexample.

I think it is related to bitmap reasoning.  Thee liquidatedCollateralPrice was not set correctly in the counterexample.  I will have to retry with the bitmap summaries.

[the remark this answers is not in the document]
MathisGD: we could also merge the PR without this. as you feel

certora/specs/Healthiness.spec
```diff
@@ -0,0 +1,271 @@
...
+/* proved in mulDivInverseUpDown */
+definition axiomInverseUpDown(mathint a, mathint b, mathint d) returns bool = a >= 0 && b > 0 && d > 0 => summaryMulDivUpM(summaryMulDivDownM(a, b, d), d, b) <= a;
+
+/* proved in mulDivLifLLTV */
+definition axiomLifLLTV(mathint a, mathint lif, mathint lltv) returns bool = a >= 0 && lltv * lif <= WAD() * WAD() => summaryMulDivUpM(a, lltv, WAD()) <= summaryMulDivUpM(a, WAD(), lif);
+
+function summaryMulDivDown(uint256 a, uint256 b, uint256 d) returns uint256 {
+    bool overflow;
+    if (overflow || d == 0) {
+        revert();
```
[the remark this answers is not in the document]
jhoenicke: The solver will check both possibilities of revert or return.  So it will not lose any runs due to revert.  This is a sound overapproximation.

certora/specs/CollateralBitmap.spec
```diff
@@ -0,0 +1,56 @@
...
+    collateralIndex < 128 => (collateralOf(id, user, collateralIndex) != 0 <=> summaryGetBit(currentContract.position[id][user].activatedCollaterals, collateralIndex));
+
+// This shows that the real isHealthy returns true if and only if the isHealthy function
+// that does not use collateral bitmap returns true.  We also check that the latter function
+// does not revert if isHealthy does not revert.
+rule isHealthyEquivalent(Midnight.Obligation obligation, bytes32 id, address borrower) {
+    require obligation.collaterals.length <= 3, "restrict to three collaterals";
+    requireInvariant nonZeroCollateralsAreActivated(id, borrower, 0);
+    requireInvariant nonZeroCollateralsAreActivated(id, borrower, 1);
+    requireInvariant nonZeroCollateralsAreActivated(id, borrower, 2);
```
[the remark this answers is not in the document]
jhoenicke: This is a known limitation, we can only show the equivalence for up to three collaterals and instantiating all 128 instances would overload the solver.   There are also similar limitations for checking whether two obligations are equal.

certora/specs/Healthiness.spec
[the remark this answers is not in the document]
jhoenicke: The prover already warns about unlinked functions, but that doesn't reflect the verification result.  At the moment it warns against canLiquidate (the new gate).

Maybe one can make the `ghost healthyBeforeCallback` non-persistent, so if there is a summary of a non-view function missing, it would be allowed to break the check.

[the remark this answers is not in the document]
jhoenicke: I think this works:
```
@@ -94,7 +94,7 @@ function summaryMulDivUp(uint256 a, uint256 b, uint256 d) returns uint256 {
 persistent ghost bool useIsHealthyNoBitmap;
 
 // global variable to track whether the user was healthy before the callbacks.
-persistent ghost bool healthyBeforeCallback;
+ghost bool healthyBeforeCallback;
 
 // global variable to track which obligation and borrower we're testing.
 persistent ghost address globalObligationLoanToken;
@@ -161,12 +161,13 @@ function genericCallback() {
     Midnight.Obligation globalObligation = getGlobalObligation();
 
     // check that isHealthy holds before the callback.  We remember any violation and check that none occurred at the end of each rule.
-    if (!callIsHealthy(globalObligation, globalId, globalBorrower)) {
-        healthyBeforeCallback = false;
-    }
+    bool savedHealthyBefore = healthyBeforeCallback && callIsHealthy(globalObligation, globalId, globalBorrower);
 
     callback.callHavoc(e, dummy);
 
+    // the callback havocs the global variable healthyBeforeCallback, so we restore it to the value before.
+    healthyBeforeCallback = savedHealthyBefore;
+
     require callIsHealthy(globalObligation, globalId, globalBorrower), "user is healthy after callback";
 }
 
```

I'm now rerunning to see if it catches the wrong summary definition and that everything still works.

certora/confs/Healthiness.conf
```diff
@@ -0,0 +1,23 @@
...
+  ],
+  "parametric_contracts": [
+    "MidnightWrapper"
+  ],
+  "verify": "MidnightWrapper:certora/specs/Healthiness.spec",
+  "solc": "solc-0.8.31",
+  "solc_via_ir": true,
+  "solc_evm_version": "osaka",
+  "optimistic_loop": true,
+  "loop_iter": 2,
```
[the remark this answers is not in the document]
jhoenicke: We tried with loopIter 3, but there are timeout problems.  Having a larger number of collaterals then loop_iter is still useful to detect more bugs, as the code does not always iterate over all collaterals.

certora/specs/Healthiness.spec
```diff
@@ -0,0 +1,266 @@
...
+
+persistent ghost address globalBorrower;
+
+// helper function to check if one of the collaterals of an obligation matches the global variables.
+// It checks for the length and also returns true if the index is out of bounds. This allows us to require this for every index.
+definition collateralMatches(Midnight.Obligation obligation, uint256 index) returns bool = (index < globalObligationCollateralLength => obligation.collaterals[index].oracle == globalObligationCollateralOracle[index] && obligation.collaterals[index].token == globalObligationCollateralToken[index] && obligation.collaterals[index].lltv == globalObligationCollateralLLTV[index] && obligation.collaterals[index].maxLif == globalObligationCollateralMaxLif[index]);
+
+function equalsGlobalObligation(Midnight.Obligation obligation) returns (bool) {
+    return obligation.loanToken == globalObligationLoanToken && obligation.collaterals.length == globalObligationCollateralLength && collateralMatches(obligation, 0) && collateralMatches(obligation, 1) && collateralMatches(obligation, 2) && obligation.maturity == globalObligationMaturity && obligation.rcfThreshold == globalObligationRcfThreshold;
```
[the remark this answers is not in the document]
jhoenicke: fixed

certora/confs/CollateralBitmap.conf
```diff
@@ -0,0 +1,19 @@
+{
+  "files": [
+    "certora/helpers/MidnightWrapper.sol"
+  ],
+  "verify": "MidnightWrapper:certora/specs/CollateralBitmap.spec",
+  "solc": "solc-0.8.31",
+  "solc_via_ir": true,
+  "solc_evm_version": "osaka",
+  "optimistic_loop": true,
+  "loop_iter": 2,
```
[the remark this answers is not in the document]
jhoenicke: Same problem with timeout.

MathisGD (approved):

PR #389, opened 2026-02-19
Rename id in event

peyha:
Cf [this](https://morpholabs.slack.com/archives/C08EACEEYD6/p1771490783606919)

### DIFF
--- a/src/libraries/EventsLib.sol
+++ b/src/libraries/EventsLib.sol
@@ -38,9 +38,15 @@ library EventsLib {
         address indexed onBehalf,
         address indexed receiver
     );
-    event Repay(address indexed caller, bytes32 indexed obligationId, uint256 obligationUnits, address indexed onBehalf);
+    event Repay(
+        address indexed caller, bytes32 indexed obligationId, uint256 obligationUnits, address indexed onBehalf
+    );
     event SupplyCollateral(
-        address caller, bytes32 indexed obligationId, address indexed collateral, uint256 assets, address indexed onBehalf
+        address caller,
+        bytes32 indexed obligationId,
+        address indexed collateral,
+        uint256 assets,
+        address indexed onBehalf
     );
 
     event WithdrawCollateral(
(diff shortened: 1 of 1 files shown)


adhusson (approved):

src/libraries/EventsLib.sol
```diff
@@ -9,14 +9,14 @@ library EventsLib {
 
     event SetOwner(address indexed owner);
     event SetFeeSetter(address indexed feeSetter);
-    event SetObligationTradingFee(bytes32 indexed id, uint256 indexed index, uint256 newTradingFee);
+    event SetObligationTradingFee(bytes32 indexed obligationId, uint256 indexed index, uint256 newTradingFee);
```
MathisGD: just to not introduce an other concept
and we can put a comment at the top for why 
```suggestion
    event SetObligationTradingFee(bytes32 indexed id_, uint256 indexed index, uint256 newTradingFee);
```

peyha: I find it quite ugly tbh

peyha: The comment I would write is essentially "/// @dev id_ is the obligation id." so it would be cleaner to stay with `obligationId`

MathisGD: i was more thinking of something like "id is used in some indexers, so we use id_"

QGarchery (approved):

src/libraries/EventsLib.sol
```diff
@@ -4,19 +4,20 @@ pragma solidity ^0.8.0;
 
 import {Obligation} from "../interfaces/IMorphoV2.sol";
 
+/// @dev id is used by indexers so we use id_ to avoid naming conflicts.
```
MathisGD:
```suggestion
/// @dev id is used by indexers so id_ is used to avoid naming conflicts.
```

MathisGD (approved):

src/libraries/EventsLib.sol
```diff
@@ -4,19 +4,20 @@ pragma solidity ^0.8.0;
 
 import {Obligation} from "../interfaces/IMorphoV2.sol";
 
+/// @dev id is used by indexers so id_ is used to avoid naming conflicts.
```
MathisGD:
```suggestion
/// @dev id_ is used to avoid naming conflicts in indexers.
```

MathisGD (approved):

PR #392, opened 2026-02-20
[min collat] rcf threshold

MathisGD:
https://morpholabs.slack.com/archives/C02N7CZ088N/p1771584258512549

### DIFF
--- a/src/MorphoV2.sol
+++ b/src/MorphoV2.sol
@@ -405,9 +405,10 @@ contract MorphoV2 is IMorphoV2 {
 
     /// @dev At least one of `seizedAssets` or `repaidUnits` should be equal to zero.
     /// @dev Accounts are liquidatable if they are unhealthy or if the maturity has passed.
-    /// @dev Before maturity, the liquidation cannot put the borrower back into health (recovery close factor).
-    /// @dev In that case, we want debtOf - repaidUnits >= maxDebt - repaidUnits*LIF*LLTV, which is equivalent to
-    /// repaidUnits <= (debtOf-maxDebt) / (1 - LIF*LLTV).
+    /// @dev Before maturity, the liquidation cannot put the borrower back into health (recovery close factor),
+    /// @dev unless doing so would bring the debt below the rcfThreshold.
+    /// @dev Recovery close factor means that debtOf - repaidUnits >= maxDebt - repaidUnits*LIF*LLTV, which is
+    /// equivalent to repaidUnits <= (debtOf-maxDebt) / (1 - LIF*LLTV).
     /// @dev If an account is healthy, the LIF grows linearly from 1 at maturity to MAX_LIF at maturity +
     /// TIME_TO_MAX_LIF.
     /// @dev Returns the seized assets and the repaid units.
@@ -463,13 +464,12 @@ contract MorphoV2 is IMorphoV2 {
 
             if (block.timestamp <= obligation.maturity) {
                 uint256 lltv = obligation.collaterals[collateralIndex].lltv;
-                uint256 debt = _state.debt;
                 // Rounded up to avoid consecutive max liquidations.
                 // Acknowledged that the position could be slightly healthy after a liquidation.
-                uint256 maxRepaid = (debt - maxDebt).mulDivUp(WAD, WAD - lif.mulDivUp(lltv, WAD));
+                uint256 maxRepaid = (_state.debt - maxDebt).mulDivUp(WAD, WAD - lif.mulDivUp(lltv, WAD));
                 require(
-                    repaidUnits <= maxRepaid || debt.zeroFloorSub(maxRepaid) < obligation.rcfThreshold,
-                    "recovery close factor violated"
+                    repaidUnits <= maxRepaid || uint256(_state.debt).zeroFloorSub(maxRepaid) < obligation.rcfThreshold,
+                    "recovery close factor conditions violated"
                 );
             }
 
--- a/src/MorphoV2.sol
+++ b/src/MorphoV2.sol
@@ -371,14 +371,6 @@ contract MorphoV2 is IMorphoV2 {
             require(UtilsLib.countBits(newBitmap) <= MAX_COLLATERALS_PER_BORROWER, "too many collaterals per borrower");
         }
 
-        require(
-            newCollateralOf == 0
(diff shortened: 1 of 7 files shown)


src/MorphoV2.sol
```diff
@@ -478,10 +463,14 @@ contract MorphoV2 is IMorphoV2 {
 
             if (block.timestamp <= obligation.maturity) {
                 uint256 lltv = obligation.collaterals[collateralIndex].lltv;
+                uint256 debt = _state.debt;
                 // Rounded up to avoid consecutive max liquidations.
                 // Acknowledged that the position could be slightly healthy after a liquidation.
-                uint256 maxRepaid = (uint256(_state.debt) - maxDebt).mulDivUp(WAD, WAD - lif.mulDivUp(lltv, WAD));
-                require(repaidUnits <= maxRepaid, "recovery close factor violated");
+                uint256 maxRepaid = (debt - maxDebt).mulDivUp(WAD, WAD - lif.mulDivUp(lltv, WAD));
```
MathisGD: slightly annoying that this depends on the collat. new reason why liquidators should choose carefully their first collat to liquidate

MathisGD:
@codex review

src/interfaces/IMorphoV2.sol
```diff
@@ -7,8 +7,8 @@ struct Obligation {
     // Must be sorted by address.
     Collateral[] collaterals;
     uint256 maturity;
-    // Minimum collateral value (quoted in loan token) to be left on supply & withdraw collateral.
-    uint256 minCollatValue;
+    // Recovery close factor threshold: below this remaining debt, full liquidation is allowed.
```
adhusson: "below this remaining debt" debt is not fully defined (debt after what)

```suggestion
    // Recovery close factor threshold: if liquidating back to health brings the debt below this threshold, full liquidation is allowed.
```

```suggestion
    // If liquidating back to health brings the debt below this threshold, full liquidation is allowed.
```

src/MorphoV2.sol
```diff
@@ -478,10 +463,14 @@ contract MorphoV2 is IMorphoV2 {
 
             if (block.timestamp <= obligation.maturity) {
                 uint256 lltv = obligation.collaterals[collateralIndex].lltv;
+                uint256 debt = _state.debt;
```
adhusson: not optimized by the compiler?

src/MorphoV2.sol
```diff
@@ -478,10 +463,14 @@ contract MorphoV2 is IMorphoV2 {
...
                 // Rounded up to avoid consecutive max liquidations.
                 // Acknowledged that the position could be slightly healthy after a liquidation.
-                uint256 maxRepaid = (uint256(_state.debt) - maxDebt).mulDivUp(WAD, WAD - lif.mulDivUp(lltv, WAD));
-                require(repaidUnits <= maxRepaid, "recovery close factor violated");
+                uint256 maxRepaid = (debt - maxDebt).mulDivUp(WAD, WAD - lif.mulDivUp(lltv, WAD));
+                require(
+                    repaidUnits <= maxRepaid || debt.zeroFloorSub(maxRepaid) < obligation.rcfThreshold,
+                    "recovery close factor violated"
```
```suggestion
                    "recovery close factor conditions violated"
```

src/MorphoV2.sol
```suggestion
    /// @dev Before maturity, the liquidation cannot put the borrower back into health (recovery close factor),
    /// @dev unless doing so would bring the debt below a set threshold.
```

```suggestion
    /// @dev In the first case, we want debtOf - repaidUnits >= maxDebt - repaidUnits*LIF*LLTV, which is equivalent to
```

adhusson, line 436: cache `liquidatedCollateralLltv` here for `if (block.timestamp <= obligation.maturity) {` ?

src/MorphoV2.sol
MathisGD: i'm pretty sure that it's not, but let me double check

MathisGD: actually it is!

src/MorphoV2.sol
qdrn:
```
address liquidatedCollatToken = obligation.collaterals[collateralIndex].token;
uint256 _collateralOf = collateralOf[id][borrower][liquidatedCollatToken];
uint256 maxRepaid = (debt - maxDebt).mulDivUp(WAD, WAD - lif.mulDivUp(lltv, WAD));
uint256 collateralValue = _collateralOf.mulDivUp(liquidatedCollatPrice, ORACLE_PRICE_SCALE);
require(
    repaidUnits <= maxRepaid
        || collateralValue.zeroFloorSub(maxRepaid.mulDivUp(lif, WAD)) < obligation.rcfThreshold,
    "recovery close factor violated"
);
```
another possibility

src/MorphoV2.sol
MathisGD: checking how much do we loose, but I don't think that it's worth it

MathisGD: indeed it's very cheap

src/MorphoV2.sol
peyha: I think it's fine, it makes the maxRepaid formula very explicit compared to just checking `isHealthy` after the liquidation

adhusson: Maybe simpler collateralOf.zeroFloorSub(maxSeized) < threshold where maxSeized = maxRepaid.muldivUp(lif * ORACLE_PRICE_SCALE, WAD * liquidatedCollatPrice)

adhusson: if can we define the collat threshold so it depends on the lltv, the collat threshold can imply a debt threshold and there is no need to check the debt (maybe this is what you had in mind @qdrn): `threshold = rcfThreshold.mulDivDown(WAD,lltv)` so at health debt >= threshold.

peyha: in this case the rcfThreshold would be quoted in collateral token ?

qdrn: In my suggestion it's still in loan asset.

peyha: Yes, was talking about @adhusson's proposal `collateralOf.zeroFloorSub(maxSeized) < rcfThreshold.mulDivDown(WAD,lltv)`

MathisGD:
replaced by #397

PR #393, opened 2026-02-20
[test] liquidate inactive collateral

adhusson:


test/LiquidationTest.sol
```diff
@@ -46,6 +46,27 @@ contract LiquidationTest is BaseTest {
         morphoV2.liquidate(obligation, 2, 0, 0, borrower, "");
     }
 
+    function testLiquidateInactiveCollateralIndex(uint256 units) public {
```
MathisGD, line 49: inactive <=> price=0?

adhusson: inactive i meant zero supply / bit set to zero here. Then in the liquidation it follows what happens when the price is zero, but it's very implicit, I wanted a more explicit test that it does what we want

MathisGD (approved):

QGarchery (approved):

Issue #109, opened 2025-09-26
[Certora] verif wish list


MathisGD:
todo:
- [x] integrate https://github.com/morpho-org/morpho-v2/pull/30
- [x] ~~prove that collaterals cannot be seized twice in the same liquidation~~ not relevant now since liquidation only targets one collateral
- [x] prove the properties of #106 
  - [x] trading fee = units * fee (done in #612)
  - [x] makerAssets = units * offerPrice (done in #612)
  - [x] fee = 0 => feeAmount = 0 (already done very early in takeInputOutputConsistency)
  - [x] buyerPrice >= sellerPrice (abandoned, because it's subsumed by the one below and because the buyerPrice&sellerPrice only live as local variables in `take`)
  - [x] buyerAssets >= sellerAssets (done in #612, is a result of the rule feeIsNotBypassed)
- [x] properties of https://github.com/morpho-org/morpho-v2/pull/169 (done in #374)
- [x] additional properties of `touchObligation`, see [this](https://github.com/morpho-org/morpho-v2/pull/374#pullrequestreview-3840937735). Done in #672 . No sensible way to prove roundtrip `toObligation(toId(obligation)) == obligation` in certora because of CREATE2 usage, hence dropped. 
- [x] show that maker price is lower than or equal to 1 ~~prices are lower than 1, see #226~~ buyer price can be greater than 1 (done it in https://github.com/morpho-org/midnight/pull/880)
- [ ] ([stuck](https://github.com/morpho-org/midnight/pull/609)) properties related to #303 
  - [x] health can be restored by a max liquidation #1083 
  - [ ] prove the rounding bound of maxRepaid
- [x] token balance should cover all collateral, and all withdrawable funds. (done in #348)
     $\mathit{token}.\mathtt{balanceOf}(\mathtt{this}) ≥ \sum_{\mathit{obligation} | \mathit{obligation}.\mathtt{loantoken} = \mathit{token}} \mathtt{withdrawable}(\mathtt{toId}(obligation)) + \sum_{\mathit{owner}, \mathit{obligation}} \mathtt{collateralOf}(\mathit{owner}, \mathit{obligation}, \mathit{token})$
- [x] $\mathtt{totalUnits}(\mathit{id}) = (\sum_{\mathit{owner}} \mathtt{debtOf}(\mathit{owner},\mathit{id})) + \mathtt{withdrawable}(\mathit{id})$
[...]

PR #395, opened 2026-02-20
Callback execution fix without Reentrancy Guard

peyha:
[Rationale](https://morpholabs.slack.com/archives/C02N7CZ088N/p1771603477009009)

### DIFF
--- a/src/MorphoV2.sol
+++ b/src/MorphoV2.sol
@@ -224,6 +224,18 @@ contract MorphoV2 is IMorphoV2 {
             sellerAssets = obligationUnits.mulDivDown(sellerPrice, WAD);
         }
 
+        uint256 newConsumed;
+        if (offer.assets > 0) {
+            newConsumed = consumed[offer.maker][offer.group] += offer.buy ? buyerAssets : sellerAssets;
+            require(newConsumed <= offer.assets, "consumed");
+        } else if (offer.obligationUnits > 0) {
+            newConsumed = consumed[offer.maker][offer.group] += obligationUnits;
+            require(newConsumed <= offer.obligationUnits, "consumed");
+        } else {
+            newConsumed = consumed[offer.maker][offer.group] += obligationShares;
+            require(newConsumed <= offer.obligationShares, "consumed");
+        }
+
         if (buyerCallback != address(0)) {
             ICallbacks(buyerCallback)
                 .onBuy(
@@ -253,18 +265,6 @@ contract MorphoV2 is IMorphoV2 {
                     obligationShares,
                     sellerCallbackData
                 );
-        }
-
-        uint256 newConsumed;
-        if (offer.assets > 0) {
-            newConsumed = consumed[offer.maker][offer.group] += offer.buy ? buyerAssets : sellerAssets;
-            require(newConsumed <= offer.assets, "consumed");
-        } else if (offer.obligationUnits > 0) {
-            newConsumed = consumed[offer.maker][offer.group] += obligationUnits;
-            require(newConsumed <= offer.obligationUnits, "consumed");
-        } else {
-            newConsumed = consumed[offer.maker][offer.group] += obligationShares;
-            require(newConsumed <= offer.obligationShares, "consumed");
         }
 
         bool buyerIsLender = (borrowerState[id][buyer].debt == 0);
--- a/src/MorphoV2.sol
+++ b/src/MorphoV2.sol
@@ -196,6 +196,8 @@ contract MorphoV2 is IMorphoV2 {
                 offer.receiverIfMakerIsSeller
             );
 
+
+
         uint256 offerPrice = TickLib.tickToPrice(offer.tick);
         uint256 timeToMaturity = UtilsLib.zeroFloorSub(offer.obligation.maturity, block.timestamp);
         uint256 _tradingFee = tradingFee(id, timeToMaturity);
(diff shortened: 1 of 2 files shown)


src/Midnight.sol
QGarchery, line 1: (minor) do we want to emit the `Take` event before the callbacks, to keep the same order of events ?

src/Midnight.sol
```diff
@@ -236,6 +236,37 @@ contract MorphoV2 is IMorphoV2 {
             require(newConsumed <= offer.obligationShares, "consumed");
         }
 
+        if (buyerCallback != address(0)) {
```
QGarchery, line 245: Let's move the callbacks before the `consume` block, since it does some state update

src/Midnight.sol
QGarchery, line 1: Would be good to have some guarantee that we are not missing anything, that there is no risk of reentrancy, but formal verif is probably better than tests for that. I added it to the formal verif todo #109

peyha: I see good reasons to do both. 
On Uniswap V4, [some hooks](https://github.com/Uniswap/v4-core/blob/d153b048868a60c2403a3ef5b2301bb247884d46/src/PoolManager.sol#L202) are executed before events are emitted but these hooks are named "beforeSwap" or "afterSwap" so one idea could be to consider the callback as executed before Take (as the state is not updated) and rename it "beforeBuy" and "beforeSell".
Or we can still consider the callback as executed after the take (because the tokens are allocated) and then it can be moved before the callback

peyha: Good point, I don't yet have a strong argument for this. My intuition is that the situation is now similar to flash accounting: if you execute a bunch of callbacks with take, all tokens will be transferred directly but the state will be updated at the end as well as the health check

QGarchery: Good point, now the from the point of view of the state it makes more sense to have the event after. Isn't it more complex for some queries though ? I thought that it was difficult to find events before a given action and somewhat easy to find events after that action, but I may be mis-remembering

QGarchery (approved):

peyha:
@codex review

peyha:
@codex review

src/Midnight.sol
```diff
@@ -224,6 +224,37 @@ contract MorphoV2 is IMorphoV2 {
             sellerAssets = obligationUnits.mulDivDown(sellerPrice, WAD);
         }
 
+        if (buyerCallback != address(0)) {
+            ICallbacks(buyerCallback)
+                .onBuy(
```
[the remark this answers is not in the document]
MathisGD: ~~not correct, all the amounts are already computed. there are only state updates remaining to do~~

src/Midnight.sol
```diff
@@ -224,6 +224,37 @@ contract MorphoV2 is IMorphoV2 {
...
+                    obligationUnits,
+                    obligationShares,
+                    buyerCallbackData
+                );
+        }
+
+        SafeTransferLib.safeTransferFrom(
+            offer.obligation.loanToken, buyer, tradingFeeRecipient, buyerAssets - sellerAssets
+        );
+        SafeTransferLib.safeTransferFrom(offer.obligation.loanToken, buyer, receiver, sellerAssets);
```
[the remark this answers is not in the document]
MathisGD: worth noting but expected

src/MorphoV2.sol
MathisGD, line 279: imagine we do this check before but the state update after (might look correct), we would have a bug (a take in the callback could double-consume)

in general it would be great to have a stronger model for what we can and cannot do. maybe formal verification can help here (as @QGarchery said)

MathisGD: actually any state update that looks like s = newValue should be suspicious (vs a s += diff or s-= diff)

src/Midnight.sol
MathisGD, line 1: let's say you have nested calls take A -> take B -> action C.

before this change the way to think about their order is A, B, C, but after it's actually C, B, A.

Not sure that it breaks anything but worth noting.

src/MorphoV2.sol
MathisGD: it is probably not enough to be safe actually. units/shares conversions depend on the state (totalUnits/totalShares) but by now the state is not the same...

MathisGD: maybe something convincing:
- take a nested call take A -> action B.
- imagine this is a problem, and a user can take advantage of the state changes made by B to trick the computations of A.
- as action B happens before any state change of A (see [this](https://github.com/morpho-org/morpho-v2/pull/395#discussion_r2853101806)), there is an argument that they could have made a multi call [B, A] and get the same end state?

QGarchery: I think the last argument doesn't hold because there can be cases where you benefit from the initial states twice with B embedded in A when you should benefit only once with [B; A]. The original message of this thread gives an example of this with the double consume. I'm wondering if we don't have the same thing currently with the units/shares conversion

src/Midnight.sol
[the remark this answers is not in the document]
QGarchery: I think this is correct and a real worry, it's the same idea as [this message.](https://github.com/morpho-org/morpho-v2/pull/395#discussion_r2853112840)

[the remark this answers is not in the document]
MathisGD: yes, I saw it after

src/MorphoV2.sol
QGarchery: So I found a concrete issue with the current code actually: you can have a share price above one.
Counter-example: assume 11 shares and 10 units in total (including virtual share and virtual unit). Then you sell 2 shares with the share as argument, it will round down to 1 unit. In the callback you do the same thing, and since the state has not been updated, it will also round down to 1 unit. So in the end you have 7 shares and 8 units

QGarchery (changes requested):
See [this](https://github.com/morpho-org/morpho-v2/pull/395#discussion_r2854128223)

src/MorphoV2.sol
QGarchery: Note that this particular bug will be fixed with #271, but it still highlights the general idea and why it may not be safe to compute the amounts before the callbacks

MathisGD:
replaced by #465

PR #397, opened 2026-02-23
[min collat] rcf threshold per collat

### LINKED ISSUE #87: Should the loan asset be usable as collateral

### LINKED ISSUE #343: liquidation min collat
discussed, this will be combined later with #342

_Originally posted by @MathisGD in https://github.com/morpho-org/morpho-v2/pull/303#discussion_r2769320296_

MathisGD:
- [initial thread](https://morpholabs.slack.com/archives/C02N7CZ088N/p1771584258512549)
- we just want to break the collat value reduction loop caused by RCF liquidations. with this change, it shouldn't drop below the threshold (in OOM).
- more precisely, we want the RCF to deactivate if c * P / LIF <= T. we could multiply by LIF and/or by LLTV to have a nicer "LI equivalent", but the easy property for the borrower matters more.
- to avoid double liquidations, we "pre-shot" the max liquidation:
  - the condition becomes (c - maxSeized) * P / LIF <= T
  - <=> c * P / LIF - maxRepaid <= T (because maxRepaid = maxSeized * P / LIF)
- borrower guarantees:
  - no deactivation if their collat are all enough to repay rcfThreshold.
  - in single-collat, d > t => no rcf deactivation ([proof](https://morpholabs.slack.com/archives/C02N7CZ088N/p1771855886129019?thread_ts=1771854302.609329&cid=C02N7CZ088N))

### DIFF
--- a/src/MorphoV2.sol
+++ b/src/MorphoV2.sol
@@ -405,8 +405,8 @@ contract MorphoV2 is IMorphoV2 {
 
     /// @dev At least one of `seizedAssets` or `repaidUnits` should be equal to zero.
     /// @dev Accounts are liquidatable if they are unhealthy or if the maturity has passed.
-    /// @dev Before maturity, the liquidation cannot put the borrower back into health (recovery close factor),
-    /// @dev unless doing so would bring the debt below the rcfThreshold.
+    /// @dev Before maturity, the liquidation cannot put the borrower back into health (recovery close factor), unless
+    /// the collateral's value is not enough to repay rcfThreshold.
     /// @dev Recovery close factor means that debtOf - repaidUnits >= maxDebt - repaidUnits*LIF*LLTV, which is
     /// equivalent to repaidUnits <= (debtOf-maxDebt) / (1 - LIF*LLTV).
     /// @dev If an account is healthy, the LIF grows linearly from 1 at maturity to MAX_LIF at maturity +
--- a/src/MorphoV2.sol
+++ b/src/MorphoV2.sol
@@ -35,6 +35,7 @@ import {EventsLib} from "./libraries/EventsLib.sol";
 /// @dev Obligations' collaterals must be sorted by token address.
 contract MorphoV2 is IMorphoV2 {
     using UtilsLib for uint256;
+    using UtilsLib for uint128;
 
     /// STORAGE ///
 
@@ -361,7 +362,7 @@ contract MorphoV2 is IMorphoV2 {
         bytes20 id = touchObligation(obligation);
         address collateralToken = obligation.collaterals[collateralIndex].token;
 
-        uint256 newCollateralOf = uint256(collateralOf[id][onBehalf][collateralIndex]) + assets;
+        uint256 newCollateralOf = collateralOf[id][onBehalf][collateralIndex] + assets;
         collateralOf[id][onBehalf][collateralIndex] = UtilsLib.toUint128(newCollateralOf);
 
         if (newCollateralOf == assets && assets > 0) {
@@ -388,7 +389,7 @@ contract MorphoV2 is IMorphoV2 {
         bytes20 id = touchObligation(obligation);
         address collateralToken = obligation.collaterals[collateralIndex].token;
 
-        uint256 newCollateralOf = uint256(collateralOf[id][onBehalf][collateralIndex]) - assets;
+        uint256 newCollateralOf = collateralOf[id][onBehalf][collateralIndex] - assets;
         collateralOf[id][onBehalf][collateralIndex] = UtilsLib.toUint128(newCollateralOf);
 
         if (newCollateralOf == 0 && assets > 0) {
@@ -464,14 +465,14 @@ contract MorphoV2 is IMorphoV2 {
 
             if (block.timestamp <= obligation.maturity) {
                 uint256 lltv = obligation.collaterals[collateralIndex].lltv;
-                uint256 _collateralOf = collateralOf[id][borrower][collateralIndex];
                 // Rounded up to avoid consecutive max liquidations.
                 // Acknowledged that the position could be slightly healthy after a liquidation.
                 uint256 maxRepaid = (_state.debt - maxDebt).mulDivUp(WAD, WAD - lif.mulDivUp(lltv, WAD));
                 require(
                     repaidUnits <= maxRepaid
-                        || _collateralOf.mulDivDown(liquidatedCollatPrice, ORACLE_PRICE_SCALE).mulDivDown(WAD, lif)
-                            .zeroFloorSub(maxRepaid) < obligation.rcfThreshold,
+                        || collateralOf[id][borrower][collateralIndex].mulDivDown(
+                                liquidatedCollatPrice, ORACLE_PRICE_SCALE
+                            ).mulDivDown(WAD, lif).zeroFloorSub(maxRepaid) < obligation.rcfThreshold,
                     "recovery close factor conditions violated"
                 );
             }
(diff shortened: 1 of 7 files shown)


src/MorphoV2.sol
```diff
@@ -478,10 +464,16 @@ contract MorphoV2 is IMorphoV2 {
 
             if (block.timestamp <= obligation.maturity) {
                 uint256 lltv = obligation.collaterals[collateralIndex].lltv;
+                uint256 _collateralOf = collateralOf[id][borrower][collateralIndex];
```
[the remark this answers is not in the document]
MathisGD: gas golfer

src/interfaces/IMorphoV2.sol
```diff
@@ -7,8 +7,8 @@ struct Obligation {
     // Must be sorted by address.
     Collateral[] collaterals;
     uint256 maturity;
-    // Minimum collateral value (quoted in loan token) to be left on supply & withdraw collateral.
-    uint256 minCollatValue;
+    // If a collateral's value is not enough to repay rcfThreshold, RCF is deactivated for this collateral.
```
peyha:
```suggestion
    // If a collateral's value quoted in loan is not enough to repay rcfThreshold, RCF is deactivated for this collateral.
```

MathisGD: value is always quoted in loan, no?

peyha: that's what we assume internally yes, not sure it's clear for external people (especially in the interface)

MathisGD:
@codex review

src/interfaces/IMorphoV2.sol
```diff
@@ -7,8 +7,9 @@ struct Obligation {
     // Must be sorted by address.
     Collateral[] collaterals;
     uint256 maturity;
-    // Minimum collateral value (quoted in loan token) to be left on supply & withdraw collateral.
-    uint256 minCollatValue;
+    // If a collateral's value (quoted in loan token) is not enough to repay rcfThreshold, RCF is deactivated for this
+    // collateral.
```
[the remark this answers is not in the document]
MathisGD: I'll try to clarify that

src/interfaces/IMorphoV2.sol
```diff
@@ -7,8 +7,9 @@ struct Obligation {
     // Must be sorted by address.
     Collateral[] collaterals;
     uint256 maturity;
-    // Minimum collateral value (quoted in loan token) to be left on supply & withdraw collateral.
-    uint256 minCollatValue;
+    // The recovery close factor is deactivated for a collateral if the collateral's value is not enough to repay
+    // rcfThreshold, and if doing a max liquidation would lead to such a situation.
```
peyha:
```suggestion
    // The recovery close factor is deactivated for a given collateral if the remaining collateral value after
    // liquidating at the RCF is not enough to repay rcfThreshold
```
a bit clearer imo

MathisGD: don't you find the parallel with if it was a simple require at the beginning of liquidate useful?

src/MorphoV2.sol
```diff
@@ -420,9 +406,10 @@ contract MorphoV2 is IMorphoV2 {
 
     /// @dev At least one of `seizedAssets` or `repaidUnits` should be equal to zero.
     /// @dev Accounts are liquidatable if they are unhealthy or if the maturity has passed.
-    /// @dev Before maturity, the liquidation cannot put the borrower back into health (recovery close factor).
-    /// @dev In that case, we want debtOf - repaidUnits >= maxDebt - repaidUnits*LIF*LLTV, which is equivalent to
-    /// repaidUnits <= (debtOf-maxDebt) / (1 - LIF*LLTV).
+    /// @dev Before maturity, the liquidation cannot put the borrower back into health (recovery close factor), unless
+    /// the collateral's value is not enough to repay rcfThreshold.
```
peyha: not sure how to phrase it properly but it should be the collateral value after liquidating at the RCF 
phrased like this, I think that the corresponding require is `collateralOf[id][borrower][collateralIndex].mulDivDown(liquidatedCollatPrice, ORACLE_PRICE_SCALE).mulDivDown(WAD, lif)< obligation.rcfThreshold`

MathisGD: will change for "unless the liquidation could leave a collateral value that is not enough..."

src/MorphoV2.sol
```diff
@@ -480,8 +467,14 @@ contract MorphoV2 is IMorphoV2 {
...
                 // Rounded up to avoid consecutive max liquidations.
                 // Acknowledged that the position could be slightly healthy after a liquidation.
-                uint256 maxRepaid = (uint256(_state.debt) - maxDebt).mulDivUp(WAD, WAD - lif.mulDivUp(lltv, WAD));
-                require(repaidUnits <= maxRepaid, "recovery close factor violated");
+                uint256 maxRepaid = (_state.debt - maxDebt).mulDivUp(WAD, WAD - lif.mulDivUp(lltv, WAD));
+                require(
+                    repaidUnits <= maxRepaid
+                        || collateralOf[id][borrower][collateralIndex].mulDivDown(
+                                liquidatedCollatPrice, ORACLE_PRICE_SCALE
+                            ).mulDivDown(WAD, lif).zeroFloorSub(maxRepaid) < obligation.rcfThreshold,
```
peyha, line 475:
```suggestion
                            ).mulDivDown(WAD, lif) < obligation.rcfThreshold,
```
this would be equivalent in term of feature (the only edge case is that liquidators would have to split their liquidation to first liquidate at maxRepaid and then liquidate all the remaining collateral when reaching rcfThreshold) but is a bit easier in explainability

MathisGD: yes exactly. worth doing imo (to make sure that liquidators do the second one). should there be an explicit comment about that?

peyha: not sure it's worth being explicit as it's a well-known liquidation trick but at the same time it's low cost to write smth

QGarchery (approved):

src/MorphoV2.sol
QGarchery: I think that this comment is fine in the interface (see [this reasoning](https://github.com/morpho-org/morpho-v2/pull/397/changes#r2847263226)), but in the code I agree that it is a bit misleading

src/MorphoV2.sol
MathisGD: actually it's not even equivalent, because if a liquidation brings you to something very small the liquidator might not do the second one. let's forget about this

peyha: it's a corner case no ? supposing the liquidator is rationale, it would only happen if the recovery close factor is very high (>50%) otherwise the profit from a double liquidation is way better.
But at the same time, gas is a heavy cost for positions of this size so it's better to have a single liquidation to close the pos so let's stay with the current version

MathisGD: anyway we would just add a comment right?

peyha (approved):

PR #400, opened 2026-02-24
[min collat] max repaid offset

MathisGD:
https://morpholabs.slack.com/archives/C02N7CZ088N/p1771927455123689


MathisGD:
we can reopen

PR #404, opened 2026-02-25
 Share/asset ratio  + Share Price does not decrease

lilCertora:
Still in Draft 

- `sharePriceBelowOrEqOne` should work against main code
- `sharePricePreservedByLiquidateNoBadDebt` should work against main code
- `sharePriceDoesNotDecrease` catch rounding issue on case 4 `muldivDown` [we discussed here](https://certora.slack.com/archives/C09EAFEJUMV/p1771844203301919)


lilCertora:
I will split it in 2 PRs so that it's easier to merge it when it works

PR #405, opened 2026-02-25
shares input everywhere

MathisGD:
https://morpholabs.slack.com/archives/C08A40KKN9W/p1772031812000379

### DIFF
--- a/src/Midnight.sol
+++ b/src/Midnight.sol
@@ -144,6 +144,8 @@ contract Midnight is IMidnight {
     /// position at the end.
     /// @dev Neither the taker nor the maker can pass from having shares to having debt in one take.
     /// @dev The taker might not get the price they expected if the trading fee was just changed.
+    /// @dev All sellerAssets are reachable with the obligationShares input, and all buyerAssets are reachable only if
+    /// buyerPrice <= WAD.
     function take(
         uint256 obligationShares,
         address taker,
@@ -198,7 +200,6 @@ contract Midnight is IMidnight {
         uint256 _tradingFee = tradingFee(id, timeToMaturity);
         uint256 sellerPrice = offer.buy ? offerPrice - _tradingFee : offerPrice;
         uint256 buyerPrice = sellerPrice + _tradingFee;
-        require(buyerPrice <= WAD, "price");
 
         bool buyerIsLender = borrowerState[id][buyer].debt == 0;
         bool sellerIsBorrower = sharesOf[id][seller] == 0;
--- a/src/Midnight.sol
+++ b/src/Midnight.sol
@@ -19,14 +19,14 @@ import {
 } from "./libraries/ConstantsLib.sol";
 import {IOracle} from "./interfaces/IOracle.sol";
 import {
-    IMorphoV2,
+    IMidnight,
     Obligation,
     Offer,
     Signature,
     Collateral,
     BorrowerState,
     ObligationState
-} from "./interfaces/IMorphoV2.sol";
+} from "./interfaces/IMidnight.sol";
 import {ICallbacks, IFlashLoanCallback} from "./interfaces/ICallbacks.sol";
 import {EventsLib} from "./libraries/EventsLib.sol";
 
@@ -40,7 +40,7 @@ import {EventsLib} from "./libraries/EventsLib.sol";
 /// @dev Post-maturity, the trading fee is the fee at the 0d breakpoint.
 /// @dev Trading fees are stored divided by FEE_STEP (1e12) to fit in 16 bits.
 /// @dev Max trading fee is defined per index (see maxTradingFee function).
-contract MorphoV2 is IMorphoV2 {
+contract Midnight is IMidnight {
     using UtilsLib for uint256;
     using UtilsLib for uint128;
 
@@ -431,9 +431,9 @@ contract MorphoV2 is IMorphoV2 {
                 );
 
             if (seizedAssets > 0) {
-                repaidUnits = seizedAssets.mulDivUp(WAD, lif).mulDivUp(liquidatedCollatPrice, ORACLE_PRICE_SCALE);
+                repaidUnits = seizedAssets.mulDivUp(liquidatedCollatPrice, ORACLE_PRICE_SCALE).mulDivUp(WAD, lif);
             } else {
-                seizedAssets = repaidUnits.mulDivDown(ORACLE_PRICE_SCALE, liquidatedCollatPrice).mulDivDown(lif, WAD);
+                seizedAssets = repaidUnits.mulDivDown(lif, WAD).mulDivDown(ORACLE_PRICE_SCALE, liquidatedCollatPrice);
             }
 
             if (block.timestamp <= obligation.maturity) {
--- a/src/Midnight.sol
+++ b/src/Midnight.sol
@@ -52,7 +52,8 @@ contract Midnight is IMidnight {
     mapping(bytes20 id => ObligationState) public obligationState;
 
     /// @dev Groups are useful to have a global offered amount shared accross multiple offers ("OCO").
-    /// @dev To work as expected, all offers in a same group should have the same obligationShares and loan token.
+    /// @dev To work as expected, all offers in a same group should have the same obligationShares, obligationUnits, and
+    /// loan token.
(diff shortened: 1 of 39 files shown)


src/Midnight.sol
```diff
@@ -216,43 +205,12 @@ contract MorphoV2 is IMorphoV2 {
...
-        } else {
-            newConsumed = consumed[offer.maker][offer.group] += obligationShares;
-            require(newConsumed <= offer.obligationShares, "consumed");
-        }
+        uint256 obligationUnits =
+            obligationShares.mulDiv(_obligationState.totalUnits + 1, _obligationState.totalShares + 1, !buyerIsLender);
+        uint256 buyerAssets = obligationUnits.mulDivDown(buyerPrice, WAD);
+        uint256 sellerAssets = obligationUnits.mulDivDown(sellerPrice, WAD);
+        uint256 newConsumed = consumed[offer.maker][offer.group] += obligationShares;
+        require(newConsumed <= offer.obligationShares, "consumed");
```
peyha: Are we sure we want to keep consume in shares ?
Suppose that I sign an offer for a number of shares (corresponding to a number of units that I want to borrow/lend) when the sharePrice is 1:1 and that bad debt is realized between the offer is signed and the offer is settled on-chain such that the sharePrice is 0.5:1. Intuitively, realizing the bad debt shouldn't change the number of units I want to borrow so it would make sense to specify consume in `obligationUnits` (and we could add an offer parameter for `maxConsumedUnits`)
Note that bad debt realization should be quite rare so it's an edge case

src/Midnight.sol
```diff
@@ -210,49 +199,19 @@ contract Midnight is IMidnight {
         uint256 _tradingFee = tradingFee(id, timeToMaturity);
         uint256 sellerPrice = offer.buy ? offerPrice - _tradingFee : offerPrice;
         uint256 buyerPrice = sellerPrice + _tradingFee;
+        require(buyerPrice <= WAD, "price");
 
         bool buyerIsLender = borrowerState[id][buyer].debt == 0;
         bool sellerIsBorrower = sharesOf[id][seller] == 0;
         // To ensure that the share price does not decrease, shares should be rounded down (units should be rounded up)
         // when buyerIsLender & sellerIsBorrower, and rounded up (units should be rounded down) when !buyerIsLender &
         // !sellerIsBorrower. The variable buyerIsLender is used to discriminate between the two cases.
```
```suggestion
        // To ensure that the share price does not decrease, units should be rounded up when buyerIsLender &
        // sellerIsBorrower, and rounded down when !buyerIsLender & !sellerIsBorrower. The variable buyerIsLender is
        // used to discriminate, as the remaining two cases do not change total units and total shares.   
```

src/Midnight.sol
adhusson, line 1: In a way, surprising how little it removes

```suggestion
/// @dev To work as expected, all offers in a same group should have the same obligationShares and loan token.
```

adhusson: the point is to stop mentioning the offer's assets and units
(weird preview bug, should be pushed directly to be safe)

MathisGD: not that much LOC, but divides the number of paths in take by 4x4 (huge)

adhusson, line 1: UtilsLib.atMostOneNonZero(a,b,c,d) can be removed

src/Midnight.sol
```diff
@@ -210,49 +199,19 @@ contract Midnight is IMidnight {
         uint256 _tradingFee = tradingFee(id, timeToMaturity);
         uint256 sellerPrice = offer.buy ? offerPrice - _tradingFee : offerPrice;
         uint256 buyerPrice = sellerPrice + _tradingFee;
+        require(buyerPrice <= WAD, "price");
```
adhusson: I don't think there is a test for this

test/TakeTest.sol
```diff
@@ -80,99 +80,7 @@ contract TakeTest is BaseTest {
...
-
-        take(buyerAssets, 0, 0, 0, lender, borrowerOffer);
-
-        assertEq(midnight.sharesOf(id, lender), expectedShares, "lender shares");
-        assertEq(midnight.debtOf(id, borrower), expectedUnits, "borrower debt");
-        assertEq(midnight.totalUnits(id), initialUnits + expectedUnits, "total units");
-        assertEq(midnight.totalShares(id), initialShares + expectedShares, "total shares");
-        assertEq(loanToken.balanceOf(borrower), buyerAssets, "borrower balance");
-        assertEq(loanToken.balanceOf(lender), 0, "lender balance");
-        assertEq(midnight.consumed(borrower, 0), buyerAssets, "borrower consumed");
```
adhusson, line 103: the `consumed` correct update is not tested anywhere anymore

src/interfaces/IMidnight.sol
```diff
@@ -22,8 +22,6 @@ struct Offer {
     Obligation obligation;
     bool buy;
     address maker;
-    uint256 assets;
-    uint256 obligationUnits;
```
adhusson, line 26: Obsoleted by [423](https://github.com/morpho-org/midnight/pull/423) but just in case: with this the actual value of offers depends on bad debt and there is no way to specify a fixed economic value.

test/TakeAmountsTest.sol
QGarchery, line 1: This tests the conversion in case the the offer is a sell offer and in case the offer is a buy offer. But instead it should test in case the buyer is a lender or not. Said otherwise: the `TakeAmountLib` functions are tested always when the last argument in `true`

src/Midnight.sol
adhusson: The paths are "shallow" (adjusts values in 4x3 variations easy to characterize, then resumes common path). Still seems better for now, but not 100% convinced it's worth the usability loss if [continuous fees](https://morpholabs.slack.com/archives/C08A40KKN9W/p1772135054037529?thread_ts=1772031812.000379&cid=C08A40KKN9W) are there.

test/TakeAmountsTest.sol
```diff
@@ -0,0 +1,201 @@
...
+        fee0 = bound(fee0, 0, midnight.maxTradingFee(0)) / 1e12 * 1e12;
+        fee1 = bound(fee1, 0, midnight.maxTradingFee(1)) / 1e12 * 1e12;
+        targetUnits = bound(targetUnits, 1, 1e30);
+        tick = bound(tick, 1, TICK_RANGE);
+
+        midnight.setObligationTradingFee(id, 0, fee0);
+        midnight.setObligationTradingFee(id, 1, fee1);
+        uint256 _tradingFee = midnight.tradingFee(id, obligation.maturity - block.timestamp);
+        vm.assume(TickLib.tickToPrice(tick) >= _tradingFee);
+        uint256 shares = TakeAmountsLib.unitsToShares(targetUnits, initialUnits, initialShares, true);
```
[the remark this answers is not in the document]
MathisGD: yes, same as [here](https://github.com/morpho-org/midnight/pull/405#discussion_r2871964956)

src/interfaces/IMidnight.sol
MathisGD: solved by #431

src/Midnight.sol
MathisGD: solved by #431

src/Midnight.sol
```diff
@@ -210,49 +198,19 @@ contract Midnight is IMidnight {
...
-            require(newConsumed <= offer.obligationUnits, "consumed");
-        } else {
-            newConsumed = consumed[offer.maker][offer.group] += obligationShares;
-            require(newConsumed <= offer.obligationShares, "consumed");
-        }
+        // To ensure that the share price does not decrease, units should be rounded up when buyerIsLender &
+        // sellerIsBorrower, and rounded down when !buyerIsLender & !sellerIsBorrower. The variable buyerIsLender is
+        // used to discriminate, as the remaining two cases do not change total units and total shares.
+        uint256 obligationUnits =
+            obligationShares.mulDiv(_obligationState.totalUnits + 1, _obligationState.totalShares + 1, !buyerIsLender);
```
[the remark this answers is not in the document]
MathisGD: roundings will be handled later

adhusson (approved):

QGarchery (commented):
LGTM

peyha (commented):
LGTM

PR #408, opened 2026-02-26
[Certora] share price below one

lilCertora:
Proves that the share/asset ratio (`totalUnits/totalShares`) never exceeds 1, split across four take parameter cases (to make prover life easier).

[last run ](https://prover.certora.com/output/7508195/cb9379e3cab54bdda0ab5415b3081640/?anonymousKey=6a4993b36dde81f08245ba1be0ad4cf82111708a)


certora/specs/SharePrice.spec
```diff
@@ -12,7 +12,57 @@ methods {
...
 }
 
-// strong invariant sharePriceBelowOne(bytes20 id)
-//     totalShares(id) >= totalUnits(id);
+/// Virtual share price = (totalUnits+1)/(totalShares+1) monotonicity.
+rule sharePriceDoesNotDecrease(bytes20 id, method f) filtered { f -> f.selector != sig:multicall(bytes[]).selector && f.selector != sig:liquidate(Midnight.Obligation, uint256, uint256, uint256, address, bytes).selector && !f.isView } {
+
+    // We need it otherwise rounding down to 0 creates shares with no backing units
+    // for withdraw +1 virtual liquidity makes exchange rate differ from actual pool ratio when totalShares > totalUnits
+    requireInvariant sharePriceBelowOrEqOne(id);
```
MathisGD: not sure to understand why we need this?

lilCertora: we do not need it anymore that's right, i will remove it and push new fix
[latest run without goes through ](https://prover.certora.com/output/7508195/3511935c35a343fd872be92262b9f47a/?anonymousKey=c14d38fa9984fd687212cb65491adc52347ae4ec)

MathisGD (approved):

certora/specs/SharePrice.spec
```diff
@@ -12,7 +12,50 @@ methods {
...
+    function TickLib.tickToPrice(uint256) internal returns (uint256) => NONDET;
+    function TickLib.wExp(int256) internal returns (uint256) => NONDET;
+    function UtilsLib.isLeaf(bytes32, bytes32, bytes32[] memory) internal returns (bool) => NONDET;
+    function UtilsLib.msb(uint256) internal returns (uint256) => NONDET;
+
+    function isHealthy(Midnight.Obligation memory, bytes20, address) internal returns (bool) => NONDET;
+
+    // Callback summaries for take, liquidate, and flashLoan.
+    function _.onBuy(Midnight.Obligation, address, uint256, uint256, uint256, uint256, bytes) external => NONDET;
+    function _.onSell(Midnight.Obligation, address, uint256, uint256, uint256, uint256, bytes) external => NONDET;
```
QGarchery: Those can make us miss real reasons why the properties are not verified, see [this issue](https://github.com/morpho-org/midnight/pull/395#discussion_r2854128223) (that was there before changing the rounding direction in #271). Are they absolutely necessary ?

certora/confs/SharePrice.conf
```diff
@@ -6,18 +6,16 @@
...
   "multi_assert_check": true,
   "smt_timeout": 7200,
   "prover_args": [
     "-destructiveOptimizations twostage",
-    "-s [z3:def{randomSeed=1},z3:def{randomSeed=2},z3:def{randomSeed=3},z3:def{randomSeed=4},z3:def{randomSeed=5},z3:def{randomSeed=6},z3:def{randomSeed=7},z3:def{randomSeed=8},z3:def{randomSeed=9},z3:def{randomSeed=10}]",
-    "-timeout 7200"
+    "-s [z3:def{randomSeed=1},z3:def{randomSeed=2},z3:def{randomSeed=3},z3:def{randomSeed=4},z3:def{randomSeed=5},z3:def{randomSeed=6},z3:def{randomSeed=7},z3:def{randomSeed=8},z3:def{randomSeed=9},z3:def{randomSeed=10}]"
   ],
-  "msg": "Midnight Share Price"
+  "msg": "Midnight Share Price Below One"
```
QGarchery:
```suggestion
  "msg": "Midnight Share Price"
```

certora/confs/SharePrice.conf
```diff
@@ -6,18 +6,16 @@
   "verify": "Midnight:certora/specs/SharePrice.spec",
   "solc": "solc-0.8.31",
   "solc_via_ir": true,
-  "solc_evm_version": "osaka",
```
QGarchery, line 9: curious why this is removed

certora/specs/SharePrice.spec
```diff
@@ -12,7 +12,50 @@ methods {
...
+    function _.onFlashLoan(address, uint256, bytes) external => NONDET;
 }
 
-// strong invariant sharePriceBelowOne(bytes20 id)
-//     totalShares(id) >= totalUnits(id);
+// Check the ratio of units over shares is below or equal to 1.
+strong invariant sharePriceBelowOrEqOne(bytes20 id)
+    totalShares(id) >= totalUnits(id);
+
+/// Liquidation without bad debt preserves virtual share price.
```
QGarchery: I think we can show instead that:
1. liquidation does not change the shares
2. liquidation does not increase the units

Then the current `sharePriceDoesNotDecreaseByLiquidateNoBadDebt` is a corollary, and we get another corollary that is used a lot when designing the protocol: liquidation does not increase the share price

certora/confs/SharePrice.conf
lilCertora: no reason, adding it back

certora/specs/SharePrice.spec
lilCertora: no they are not necessary at all finally, it was used before but no longer needed , I will remove them

certora/specs/SharePrice.spec
lilCertora: indeed, i like the idea, i will modify

QGarchery (approved):

PR #410, opened 2026-02-26
[Certora] Only authorized users can decrease the share of a lender

lilCertora:
Only authorized users can decrease the share of a lender. 

[last run](https://prover.certora.com/output/7508195/d7da77c4e6ea436396a24b97addf7203/?anonymousKey=f2240d6401e999d7dc3c76945b671bc6f0401db4)


certora/confs/OnlyAuthorizedCanDecreaseShares.conf
QGarchery, line 1: let's add the conf file to the CI, to see the runs as this gets updated

.github/workflows/certora.yml
```diff
@@ -20,6 +20,7 @@ jobs:
           - Midnight
           - SharePrice
           - Solvency
+          - OnlyAuthorizedCanDecreaseShares
```
QGarchery: We usually keep this list alphabetically sorted

certora/specs/OnlyAuthorizedCanDecreaseShares.spec
```diff
@@ -0,0 +1,36 @@
...
+    require user != e.msg.sender;
+    require !isAuthorized(user, e.msg.sender);
+
+    calldataarg args;
+    f(e, args);
+
+    assert sharesOf(id, user) >= sharesBefore || f.selector == sig:take(uint256, address, address, bytes, address, Midnight.Offer, Midnight.Signature, bytes32, bytes32[]).selector;
+}
+
+/// In take, the caller must be authorized by the taker and only the lender shares can decrease
```
QGarchery: This implicitly assume no reentrancy (a callback could take again and decrease the shares of another user). Probably fine for this rule, but I think it makes sense to document it

certora/specs/OnlyAuthorizedCanDecreaseShares.spec
```diff
@@ -0,0 +1,36 @@
...
+rule takeOnlyAuthorizedSellerSharesDecrease(env e, uint256 obligationShares, address taker, address takerCallback, bytes takerCallbackData, address receiverIfTakerIsSeller, Midnight.Offer offer, Midnight.Signature signature, bytes32 root, bytes32[] proof, bytes32 id, address user) {
+    address seller = offer.buy ? taker : offer.maker;
+    bool takerUnauthorized = e.msg.sender != taker && !isAuthorized(taker, e.msg.sender);
+
+    uint256 sharesBefore = sharesOf(id, user);
+
+    take@withrevert(e, obligationShares, taker, takerCallback, takerCallbackData, receiverIfTakerIsSeller, offer, signature, root, proof);
+
+    assert takerUnauthorized => lastReverted;
+    assert sharesOf(id, user) < sharesBefore => user == seller;
```
QGarchery: To go further, we could also show that the seller always authorized the action in some way. The missing case is when the seller is the maker, and the authorization came from signing the offer

certora/specs/OnlyAuthorizedCanDecreaseShares.spec
```diff
@@ -0,0 +1,36 @@
...
+}
```
QGarchery, line 44: Could we also verify that no other shares where changed ?

```suggestion
    assert otherUser != seller => sharesOf(id, otherUser) != sharesOfOtherUserBefore;
}
```

certora/specs/OnlyAuthorizedCanDecreaseShares.spec
QGarchery: I put that as a separate property in #109

certora/specs/OnlyAuthorizedCanDecreaseShares.spec
lilCertora: that's true

certora/specs/OnlyAuthorizedCanDecreaseShares.spec
```diff
@@ -0,0 +1,47 @@
...
+
+    uint256 sharesBefore = sharesOf(id, user);
+
+    take@withrevert(e, obligationShares, taker, takerCallback, takerCallbackData, receiverIfTakerIsSeller, offer, signature, root, proof);
+
+    bool reverted = lastReverted;
+    uint256 sharesAfter = sharesOf(id, user);
+
+    assert takerUnauthorized => reverted;
+    assert !reverted && offer.buy => (e.msg.sender == seller || isAuthorized(seller, e.msg.sender));
```
QGarchery:
```suggestion
```
because it is a particular case of the previous one when `offer.buy`

certora/specs/OnlyAuthorizedCanDecreaseShares.spec
```diff
@@ -0,0 +1,47 @@
...
+
+methods {
+    function multicall(bytes[]) external => HAVOC_ALL DELETE;
+
+    function sharesOf(bytes32 id, address user) external returns (uint256) envfree;
+    function isAuthorized(address authorizer, address authorized) external returns (bool) envfree;
+
+    function _.price() external => NONDET;
+    function _.onBuy(Midnight.Obligation obligation, address buyer, uint256 buyerAssets, uint256 sellerAssets, uint256 obligationUnits, uint256 obligationShares, bytes data) external => HAVOC_ECF;
+    function _.onSell(Midnight.Obligation obligation, address seller, uint256 buyerAssets, uint256 sellerAssets, uint256 obligationUnits, uint256 obligationShares, bytes data) external => HAVOC_ECF;
```
QGarchery:
```suggestion
```
this will be automatically summarized this way right ?

certora/specs/OnlyAuthorizedCanDecreaseShares.spec
```diff
@@ -0,0 +1,47 @@
...
+rule onlyAuthorizedCanDecreaseSharesExceptTake(env e, method f, bytes32 id, address user) {
+    uint256 sharesBefore = sharesOf(id, user);
+
+    require user != e.msg.sender;
+    require !isAuthorized(user, e.msg.sender);
+
+    calldataarg args;
+    f(e, args);
+
+    assert sharesOf(id, user) >= sharesBefore || f.selector == sig:take(uint256, address, address, bytes, address, Midnight.Offer, Midnight.Signature, bytes32, bytes32[]).selector;
```
QGarchery, line 23: we could filter out the `liquidate` function instead

certora/confs/OnlyAuthorizedCanDecreaseShares.conf
```diff
@@ -0,0 +1,28 @@
...
+  "smt_timeout": 7200,
+  "prover_args": [
+    "-destructiveOptimizations",
+    "twostage",
+    "-s",
+    "[z3:def{randomSeed=1},z3:def{randomSeed=2},z3:def{randomSeed=3},z3:def{randomSeed=4},z3:def{randomSeed=5}]",
+    "-splitParallel",
+    "true",
+    "-dontStopAtFirstSplitTimeout",
+    "true"
```
QGarchery:
```suggestion
```
I understand that this is only useful for if we expect the rules to have counter-examples

certora/confs/OnlyAuthorizedCanDecreaseShares.conf
```diff
@@ -0,0 +1,28 @@
...
+  "rule_sanity": "basic",
+  "multi_assert_check": true,
+  "smt_timeout": 7200,
+  "prover_args": [
+    "-destructiveOptimizations",
+    "twostage",
+    "-s",
+    "[z3:def{randomSeed=1},z3:def{randomSeed=2},z3:def{randomSeed=3},z3:def{randomSeed=4},z3:def{randomSeed=5}]",
+    "-splitParallel",
+    "true",
```
QGarchery:
```suggestion
    "-splitParallel true",
```
(optional) nitpick as I find it more readable

certora/confs/OnlyAuthorizedCanDecreaseShares.conf
```diff
@@ -0,0 +1,28 @@
...
+  "optimistic_loop": true,
+  "loop_iter": 2,
+  "optimistic_hashing": true,
+  "hashing_length_bound": 1024,
+  "rule_sanity": "basic",
+  "multi_assert_check": true,
+  "smt_timeout": 7200,
+  "prover_args": [
+    "-destructiveOptimizations",
+    "twostage",
```
QGarchery: same. Or even better : can we remove it ?

certora/confs/OnlyAuthorizedCanDecreaseShares.conf
```diff
@@ -0,0 +1,28 @@
+{
+  "files": [
+    "src/Midnight.sol",
+    "certora/dispatch/ERC20Standard.sol"
```
QGarchery:
```suggestion
```
I think this is unused

certora/confs/OnlyAuthorizedCanDecreaseShares.conf
lilCertora: yes we can

certora/specs/OnlyAuthorizedCanDecreaseShares.spec
lilCertora: this would weaken the rule in case of future changes no ? i'm not sure i understand why you want to do that

certora/specs/OnlyAuthorizedCanDecreaseShares.spec
lilCertora: yes that's right

certora/specs/OnlyAuthorizedCanDecreaseShares.spec
QGarchery: Yes, it's just for improved readability, but ok to keep as well

QGarchery (approved):

jhoenicke (approved):
Looks fine.

I'm wondering about the `@withrevert` in take.  You could just write it without and then `assert !takerUnauthorized`.   This is probably a minor thing.  It is interesting that it still works to assert that sharesAfter/sharesBefore behave like this, but I guess the prover knows that for a revert the shares don't change.

bhargavbh (approved):

certora/specs/OnlyAuthorizedCanDecreaseShares.spec
MathisGD, line 1: after #473, this can become OnlyAuthorizedCanChangeShares

PR #411, opened 2026-02-26
[Certora] only authorized user can increase debt of a borrower 

lilCertora:
Property trying to verify : Only authorized users can increase the debt of a borrower.

[last run](https://prover.certora.com/output/7508195/aa14672a6bfa4f6081d90438976f9c18/?anonymousKey=7b7a350b48c7e841d2f3fe004d2b79457e23f48c)

PR #412, opened 2026-02-26
[Certora] In take, only the seller side can newly become a borrower; the buyer can only reduce existing debt.

lilCertora:
Trying to prove property "In take, only the seller side can newly become a borrower; the buyer can only reduce existing debt."

[last run](https://prover.certora.com/output/7508195/825d2c6f847a4a7abe19f7bdc5522a66/?anonymousKey=a48b7696a645eddcf6537c3698f573b9359fb870)


certora/confs/TakeOnlySellerBecomesBorrower.conf
QGarchery, line 1: let's add this conf file to the CI, so that we can see the runs as this PR gets updated

lilCertora:
closed because merged manually into #411

PR #413, opened 2026-02-26
[Certora] Improve create obligation spec

QGarchery:


MathisGD (approved):

jhoenicke (approved):
Looks good to me. It mainly adds the check that the user-specific entries (shares, collateral and debt) are also zero, right?

certora/specs/CreatedObligations.spec
bhargavbh, line 55: We can add an invariant `createdObligationsMaxLifLltvRelation`. Basically verify the below require:
https://github.com/morpho-org/midnight/blob/95127694051575fbb4b2fb283bac1c08c5a3f8c1/src/Midnight.sol#L539

bhargavbh (approved):
LGTM. 
Noticed we have `loop_iter:2` and  `"hashing_length_bound": 2048` in the config. 
We may want to add a comment to justify the configs/ or the limitations arising from it for this spec.

lilCertora (approved):
lgtm

certora/specs/CreatedObligations.spec
QGarchery: We have something already like this in #442

PR #415, opened 2026-02-26
LLTV/LIF mulDiv order with price

peyha:
After #129, the mental model was that LLTV should be multiplied against a value quoted in loan while LIF would be against a value quoted in collateral. Because of the RCF maxRepaid formula involves both LLTV and LIF multiplied against a value quoted in loan, we could change the mental model to both LLTV/LIF should be multiplied against a value in loan asset.


src/MorphoV2.sol
QGarchery, line 1: related PR: https://github.com/morpho-org/morpho-v2/pull/129

and related thread: https://morpholabs.slack.com/archives/C0ACSA8TNMA/p1771950444383039

MathisGD (approved):
nice catch, indeed we would rather have the same order everywhere

PR #414, opened 2026-02-26
reorder bad debt

QGarchery:
Motivation: https://github.com/morpho-org/morpho-v2/pull/371#discussion_r2839742652

PR #416, opened 2026-02-26
underestimate max repaid

MathisGD:


src/MorphoV2.sol
```diff
@@ -466,7 +468,8 @@ contract MorphoV2 is IMorphoV2 {
                 uint256 lltv = obligation.collaterals[collateralIndex].lltv;
                 // Rounded up to avoid consecutive max liquidations.
                 // Acknowledged that the position could be slightly healthy after a liquidation.
```
peyha, line 470:
```suggestion
// Rounded down.
```

src/MorphoV2.sol
```diff
@@ -466,7 +468,8 @@ contract MorphoV2 is IMorphoV2 {
...
-                uint256 maxRepaid = _state.debt.zeroFloorSub(maxDebt).mulDivUp(WAD, WAD - lif.mulDivUp(lltv, WAD));
+                uint256 maxRepaid =
+                    _state.debt.zeroFloorSub(maxDebtOverestimated).mulDivDown(WAD, WAD - lif.mulDivDown(lltv, WAD));
```
peyha, line 472: contrary to #414, we could have `_state.debt < maxDebtOverestimated` in some cases so `zeroFloorSub` is necessary

peyha: Is it problematic if `_state.debt.zeroFloorSub(maxDebtOverestimated)=0` ? the position cannot be liquidated

PR #419, opened 2026-02-26
[Certora] Offers cannot be replayed

lilCertora:
Property trying to verify : 

- consume is monotonically non decreasing across all functions
- offers cannot be replayed

[Last run](https://prover.certora.com/output/7508195/1c4354684dd24264b16e5e5a83ec32a2/)


lilCertora:
closing this one because will be covered in #461

PR #423, opened 2026-02-27
refactor bad debt realization

### LINKED ISSUE #10: Grieffing attack when rounding seller/buyer offers

### LINKED ISSUE #210: Should offers have a way to specify if they allow net lender / net borrower positions?
With #161 it's possible for both lenders and borrowers to make sure they never become a borrower / lender. But it forces lenders to denominate their sell offers in shares and borrowers to denominate their buy offers in units, and it's easy to get it wrong and accidentally leave enough open offers that you become crossable.

### LINKED ISSUE #211: realization mechanism reconsideration
maybe we could have a lighter mechanism, that relies on people updating their prices?

### LINKED ISSUE #301: crossing prevention is painful for router

MathisGD:
fixes #301 
fixes #211 
fixes #210
fixes #10


src/Midnight.sol
```diff
@@ -46,14 +38,14 @@ contract Midnight is IMidnight {
 
     /// STORAGE ///
 
-    mapping(bytes20 id => mapping(address user => uint256)) public sharesOf;
-    mapping(bytes20 id => mapping(address user => BorrowerState)) public borrowerState;
+    mapping(bytes20 id => mapping(address user => int256)) public balanceOf;
```
adhusson: surprised by the single `int` mapping here (vs 2 uints). Not against it but why did you prefer it this time?

src/Midnight.sol
```diff
@@ -200,40 +193,21 @@ contract Midnight is IMidnight {
...
-            _obligationState.totalUnits -= UtilsLib.toUint128(obligationUnits);
-        }
+
+        uint256 newConsumed = consumed[offer.maker][offer.group] += obligationUnits;
+        require(newConsumed <= offer.obligationUnits, "consumed");
+
+        int256 oldBuyerBalance = balanceOf[id][buyer];
+        int256 oldSellerBalance = balanceOf[id][seller];
+        // forge-lint: disable-next-line(unsafe-typecast)
+        int256 newBuyerBalance = balanceOf[id][buyer] = oldBuyerBalance + int256(obligationUnits);
```
adhusson: not worth the code golf to have `decl = a = b`

adhusson: note: Before a user could not go lender <-> borrower in a single take. Now it's possible. It was not a protection, more an artefact of the code.

src/Midnight.sol
```diff
@@ -200,40 +193,21 @@ contract Midnight is IMidnight {
...
+        // forge-lint: disable-next-line(unsafe-typecast)
+        int256 newSellerBalance = balanceOf[id][seller] = oldSellerBalance - int256(obligationUnits);
+        _obligationState.totalUnits = _obligationState.totalUnits
+            + UtilsLib.toUint128(UtilsLib.negativePart(newSellerBalance) - UtilsLib.negativePart(oldSellerBalance))
+            - UtilsLib.toUint128(UtilsLib.negativePart(oldBuyerBalance) - UtilsLib.negativePart(newBuyerBalance));
```
adhusson: hard to read
1) what about remove all old debts and add all new debts
```
_obligationState.totalUnits = uint128(
      uint256(_obligationState.totalUnits)
      - UtilsLib.negativePart(oldSellerBalance) - UtilsLib.negativePart(oldBuyerBalance)
      + UtilsLib.negativePart(newSellerBalance) + UtilsLib.negativePart(newBuyerBalance)
  );
```

or 2) what about remove cleared debt, add created debt
```
_obligationState.totalUnits = uint128(                                                  
      uint256(_obligationState.totalUnits)                                              
      - UtilsLib.min(obligationUnits, UtilsLib.negativePart(oldBuyerBalance))             
      + UtilsLib.min(obligationUnits, UtilsLib.negativePart(newSellerBalance))
```

src/Midnight.sol
```diff
@@ -341,8 +290,8 @@ contract Midnight is IMidnight {
 
         if (oldCollateralOf == 0 && assets > 0) {
             // forge-lint: disable-next-item(unsafe-typecast) as collateralIndex < MAX_COLLATERALS (128)
-            uint128 newBitmap = borrowerState[id][onBehalf].activatedCollaterals | uint128(1 << collateralIndex);
-            borrowerState[id][onBehalf].activatedCollaterals = newBitmap;
+            uint128 newBitmap = activatedCollaterals[id][onBehalf] | uint128(1 << collateralIndex);
+            activatedCollaterals[id][onBehalf] = newBitmap;
             require(UtilsLib.countBits(newBitmap) <= MAX_COLLATERALS_PER_BORROWER, "too many collaterals per borrower");
```
adhusson: probably ok to do

```suggestion
            activatedCollaterals[id][onBehalf] = activatedCollaterals[id][onBehalf] | uint128(1 << collateralIndex);
            require(UtilsLib.countBits(activatedCollaterals[id][onBehalf]) <= MAX_COLLATERALS_PER_BORROWER, "too many collaterals per borrower");
```

src/Midnight.sol
```diff
@@ -537,22 +506,15 @@ contract Midnight is IMidnight {
...
 
-    function debtOf(bytes20 id, address user) external view returns (uint256) {
-        return borrowerState[id][user].debt;
-    }
-
-    function activatedCollaterals(bytes20 id, address user) external view returns (uint128) {
-        return borrowerState[id][user].activatedCollaterals;
+    function debtOf(bytes20 id, address user) public view returns (uint256) {
+        if (balanceOf[id][user] >= 0) return 0;
+        return uint256(-balanceOf[id][user]);
```
adhusson: should reuse negativePart

src/Midnight.sol
```diff
@@ -517,13 +471,28 @@ contract Midnight is IMidnight {
...
             IdLib.storeInCode(obligation);
 
             emit EventsLib.ObligationCreated(id, obligation);
         }
         return id;
     }
 
+    /// INTERNAL FUNCTIONS ///
+
+    function _slash(bytes20 id, address user) internal {
```
adhusson: "exact amount" offers that are used through a slashing can turn lenders into borrowers and there is no longer the option of denominating in shares to avoid the problem

src/Midnight.sol
```diff
@@ -517,13 +471,28 @@ contract Midnight is IMidnight {
...
+        uint256 lastIndex = lastSlashingIndex[id][user];
+        uint256 currentIndex = obligationState[id].slashingIndex;
+        if (lastIndex == currentIndex) return;
```
adhusson: early return not great no?

src/Midnight.sol
```diff
@@ -517,13 +471,28 @@ contract Midnight is IMidnight {
...
+        int256 balance = balanceOf[id][user];
+        if (balance > 0 && lastIndex > 0) {
+            // forge-lint: disable-next-line(unsafe-typecast)
+            balanceOf[id][user] = int256(uint256(balance).mulDivDown(currentIndex, lastIndex));
+        }
+        lastSlashingIndex[id][user] = currentIndex;
```
adhusson: (also broken when currentIndex == 0)

```suggestion
        uint256 lastIndex = lastSlashingIndex[id][user];
        if (lastIndex == 0) lastIndex = WAD;
        uint256 currentIndex = obligationState[id].slashingIndex;
        int256 balance = balanceOf[id][user];
        if (lastIndex != currentIndex) {
          // forge-lint: disable-next-line(unsafe-typecast)
          if (balance > 0) balanceOf[id][user] = int256(uint256(balance).mulDivDown(currentIndex, lastIndex));
          lastSlashingIndex[id][user] = currentIndex;
        }
```

src/Midnight.sol
```diff
@@ -419,7 +367,11 @@ contract Midnight is IMidnight {
         require(block.timestamp > obligation.maturity || originalDebt > maxDebt, "position is not liquidatable");
 
         if (badDebt > 0) {
-            _state.debt -= UtilsLib.toUint128(badDebt);
+            // forge-lint: disable-next-line(unsafe-typecast)
+            balanceOf[id][borrower] += int256(badDebt);
+            uint256 oldTotalUnits = _obligationState.totalUnits;
+            _obligationState.slashingIndex =
+                _obligationState.slashingIndex.mulDivDown(oldTotalUnits - badDebt, oldTotalUnits);
```
adhusson: there is no possible recovery if bad debt = total units, slashingIndex will always be zero. I think future lenders are then not slashed at all due to the way `_slash` is written right now

src/Midnight.sol
```diff
@@ -517,13 +471,28 @@ contract Midnight is IMidnight {
...
             emit EventsLib.ObligationCreated(id, obligation);
         }
         return id;
     }
 
+    /// INTERNAL FUNCTIONS ///
+
+    function _slash(bytes20 id, address user) internal {
+        uint256 lastIndex = lastSlashingIndex[id][user];
+        uint256 currentIndex = obligationState[id].slashingIndex;
```
```
- lastIndex -> currentUserIndex
- currentIndex -> currentObligationIndex
```

also

```
Index -> SlashingLevel
```

src/Midnight.sol
MathisGD: it simplifies the state updates in take (crossing is seamless)

src/Midnight.sol
MathisGD: in the previous version of the code, it gave a share price or zero which stuck the obligation anyway. not sure that this is worse

src/Midnight.sol
MathisGD: 1) is great

src/Midnight.sol
MathisGD: ok to recompute the mapping?

src/Midnight.sol
MathisGD: should they have a bool "noCross" ?

src/Midnight.sol
adhusson: Before it worked? If `total units = 0` and `total shares = 100` then a buy of 50 units `new total units = 50` and `new total shares = 100 + 50*(100+1)/1`. Some loss to rounding that disappears on bigger amounts but fine.

adhusson: On new obligations it may not be too difficult to trigger 100% bad debt on a 1 wei borrow by sandwiching oracle updates.

src/Midnight.sol
adhusson: moved to issue

src/Midnight.sol
adhusson: missing a `slashedBalanceOf` getter. Or maybe : `_balanceOf` mapping, and `balanceOf()` returns the slashed balance.

src/Midnight.sol
adhusson: No cross could be reused after the crossing but a bool "as borrower" or "as lender" (or other name) yes.

src/Midnight.sol
MathisGD: I'm not sure, you need the price to jump by a lot

src/libraries/UtilsLib.sol
```diff
@@ -77,4 +77,9 @@ library UtilsLib {
         }
         return 0;
     }
+
+    function negativePart(int256 x) internal pure returns (uint256) {
+        // forge-lint: disable-next-line(unsafe-typecast)
+        return x < 0 ? uint256(-x) : 0;
```
peyha:
```suggestion
        return mul(shr(255, x), sub(0, x))
```
optimized version

src/Midnight.sol
```diff
@@ -526,13 +477,29 @@ contract Midnight is IMidnight {
...
+    function _slash(bytes20 id, address user) internal {
+        uint256 currentUserIndex = lastSlashingLevel[id][user];
+        uint256 currentObligationIndex = obligationState[id].slashingLevel;
+        if (currentUserIndex != currentObligationIndex) {
+            int256 balance = balanceOf[id][user];
+            if (balance > 0 && currentUserIndex > 0) {
+                // forge-lint: disable-next-line(unsafe-typecast)
+                balanceOf[id][user] = int256(uint256(balance).mulDivDown(currentObligationIndex, currentUserIndex));
+            }
+            lastSlashingLevel[id][user] = currentObligationIndex;
```
adhusson: this writes the user index to WAD on first call when there has not been bad debt yet but it could be [avoided](https://github.com/morpho-org/midnight/pull/423#discussion_r2872552309) (sstore 0 -> nonzero is expensive)

MathisGD: your suggestion is to not write the level if the user has a zero balance? but if the balance gets unzeroed later in the execution?

src/Midnight.sol
MathisGD: actually this is a [real concern](https://morpholabs.slack.com/archives/C08A40KKN9W/p1772637874589059). will be fixed by #457

src/Midnight.sol
```diff
@@ -206,48 +198,28 @@ contract Midnight is IMidnight {
...
+
+        int256 oldBuyerBalance = balanceOf[id][buyer];
+        int256 oldSellerBalance = balanceOf[id][seller];
+        // forge-lint: disable-next-line(unsafe-typecast)
+        int256 newBuyerBalance = oldBuyerBalance + int256(obligationUnits);
+        // forge-lint: disable-next-line(unsafe-typecast)
+        int256 newSellerBalance = oldSellerBalance - int256(obligationUnits);
+        balanceOf[id][buyer] = newBuyerBalance;
+        balanceOf[id][seller] = newSellerBalance;
+        if (offer.asLender) require((offer.buy ? newBuyerBalance : newSellerBalance) >= 0, "crossed");
```
MathisGD: the flag is broken actually now. it should be called "notCross" and check if the balance changed sign

adhusson: doesn't this mean the offer can be reused after?

src/Midnight.sol
```diff
@@ -206,48 +198,28 @@ contract Midnight is IMidnight {
...
+        uint256 sellerAssets =
+            offer.buy ? obligationUnits.mulDivDown(sellerPrice, WAD) : obligationUnits.mulDivUp(sellerPrice, WAD);
+
+        uint256 newConsumed = consumed[offer.maker][offer.group] += obligationUnits;
+        require(newConsumed <= offer.obligationUnits, "consumed");
+
+        int256 oldBuyerBalance = balanceOf[id][buyer];
+        int256 oldSellerBalance = balanceOf[id][seller];
+        // forge-lint: disable-next-line(unsafe-typecast)
+        int256 newBuyerBalance = oldBuyerBalance + int256(obligationUnits);
```
adhusson: this is ok only because earlier we do a `obligationUnits.mulDivDown(buyerPrice, WAD);` that would overflow with `obligationUnits > int256.max` even at the minimum price, but it's very implicit. (same in newSellerBalance update)

src/Midnight.sol
```diff
@@ -440,7 +385,11 @@ contract Midnight is IMidnight {
         require(block.timestamp > obligation.maturity || originalDebt > maxDebt, "position is not liquidatable");
 
         if (badDebt > 0) {
-            _state.debt -= UtilsLib.toUint128(badDebt);
+            // forge-lint: disable-next-line(unsafe-typecast)
+            balanceOf[id][borrower] += int256(badDebt);
+            uint256 oldTotalUnits = _obligationState.totalUnits;
+            _obligationState.lossIndex =
+                WAD - (WAD - _obligationState.lossIndex).mulDivDown(oldTotalUnits - badDebt, oldTotalUnits);
```
peyha: is this the correct rounding ?
we currently underestimate bad debt meaning that we tend to minimize the lender slashing
If we use the same rounding direction here, it would mean that the loss factor (1 - lossIndex) should be maximized which would correspond to `mulDivUp` here

adhusson: Underestimating bad debt cannot create a bank runnable state, but not slashing lenders enough could since it could result in their total claim after everyone being slashed being > total debt.

Whereas if we overslash a bit we get `sum units after slash < totalUnits`.

adhusson: (bad debt understimate could create positions that are not profitable to liquidate but in principle it's recoverable)

peyha: ok I'm convinced that the current rounding is good

src/Midnight.sol
```diff
@@ -50,14 +42,14 @@ contract Midnight is IMidnight {
 
     /// STORAGE ///
 
-    mapping(bytes32 id => mapping(address user => uint256)) public sharesOf;
-    mapping(bytes32 id => mapping(address user => BorrowerState)) public borrowerState;
+    mapping(bytes32 id => mapping(address user => int256)) public balanceOf;
+    mapping(bytes32 id => mapping(address user => uint256)) public userLossIndex;
+    mapping(bytes32 id => mapping(address user => uint128)) public activatedCollaterals;
```
adhusson: index on 128 bits so this can be 1 slot?

src/Midnight.sol
```suggestion
/// @dev The max amount of debt, totalUnits, and collateral is type(uint128).max (~1e38).
```

src/periphery/TakeBundler.sol
```diff
@@ -11,73 +11,18 @@ contract TakeBundler {
...
-        uint256 obligationShares;
+        uint256 obligationUnits;
         Offer offer;
         Signature sig;
         bytes32 root;
         bytes32[] proof;
     }
 
-    /// @dev Iterates through orders, filling up to `targetShares` obligation shares total.
+    /// @dev Iterates through orders, filling up to targetUnits obligation unitstotal.
```
adhusson:
```suggestion
    /// @dev Iterates through orders, filling up to targetUnits obligation units total.
```

src/Midnight.sol
MathisGD: actually why would people need crossing at all

MathisGD: double offers are harder to handle in terms of router

src/Midnight.sol
```diff
@@ -560,6 +511,19 @@ contract Midnight is IMidnight {
...
+    function slash(bytes32 id, address user) public {
+        uint256 _userLossIndex = userLossIndex[id][user];
+        uint256 lossIndex = obligationState[id].lossIndex;
+        if (_userLossIndex != lossIndex) {
+            int256 balance = balanceOf[id][user];
+            if (balance > 0) {
+                // forge-lint: disable-next-line(unsafe-typecast)
+                balanceOf[id][user] = int256(uint256(balance).mulDivDown(WAD - lossIndex, WAD - _userLossIndex));
+            }
+            userLossIndex[id][user] = lossIndex;
```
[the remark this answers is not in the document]
MathisGD: @adhusson is on it

src/Midnight.sol
MathisGD: fixed now, with one boolean we have what we want

src/Midnight.sol
```diff
@@ -206,49 +198,28 @@ contract Midnight is IMidnight {
...
+        // forge-lint: disable-next-line(unsafe-typecast)
+        int256 newSellerBalance = oldSellerBalance - int256(obligationUnits);
+        balanceOf[id][buyer] = newBuyerBalance;
+        balanceOf[id][seller] = newSellerBalance;
+        if (offer.exitOnly) require(offer.buy ? newBuyerBalance <= 0 : newSellerBalance >= 0, "crossed");
+        _obligationState.totalUnits = UtilsLib.toUint128(
+            uint256(_obligationState.totalUnits) + UtilsLib.negativePart(newSellerBalance)
+                + UtilsLib.negativePart(newBuyerBalance) - UtilsLib.negativePart(oldSellerBalance)
+                - UtilsLib.negativePart(oldBuyerBalance)
+        );
```
adhusson, line 232:
```
 uint256 sellerBorrowedUnits =
      UtilsLib.negativePart(newSellerBalance) - UtilsLib.negativePart(oldSellerBalance);
  uint256 buyerRepaidUnits =
      UtilsLib.negativePart(oldBuyerBalance) - UtilsLib.negativePart(newBuyerBalance);
totalUnits = ...
```

not sure

src/Midnight.sol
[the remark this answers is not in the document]
MathisGD: this is fine, the market is arguably broken if a bad debt made people loose everything

src/Midnight.sol
```diff
@@ -206,49 +198,28 @@ contract Midnight is IMidnight {
...
+        int256 oldBuyerBalance = balanceOf[id][buyer];
+        int256 oldSellerBalance = balanceOf[id][seller];
+        // forge-lint: disable-next-line(unsafe-typecast)
+        int256 newBuyerBalance = oldBuyerBalance + int256(obligationUnits);
+        // forge-lint: disable-next-line(unsafe-typecast)
+        int256 newSellerBalance = oldSellerBalance - int256(obligationUnits);
+        balanceOf[id][buyer] = newBuyerBalance;
+        balanceOf[id][seller] = newSellerBalance;
+        if (offer.exitOnly) require(offer.buy ? newBuyerBalance <= 0 : newSellerBalance >= 0, "crossed");
+        _obligationState.totalUnits = UtilsLib.toUint128(
```
peyha, line 230: totalUnits might be difficult to re-compute purely with events.
If we go for #517, then computing it requires parsing both Slash event emitted at the beginning of take and comparing it to the obligationUnits of Take. Might be difficult to achieve if there are multiple take within the same tx or if the order of events is weird because of callbacks
a simple solution could be to log the new totalUnits or some data about `newSellerBalance` or `newBuyerBalance`

foundry.toml
```diff
@@ -10,3 +10,4 @@ wrap_comments = true
 
 [lint]
 exclude_lints = ["asm-keccak256", "unsafe-cheatcode", "incorrect-shift"]
+ignore = ['test/**/*.sol']
```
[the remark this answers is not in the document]
MathisGD: we are ok with that

src/Midnight.sol
MathisGD: the current version makes it clear that we are adding the new balances and removing the old ones. I think that it's fine

src/Midnight.sol
MathisGD: we could log the new totalUnit, indeed

certora/specs/OnlyAuthorizedCanChange.spec
```diff
@@ -3,91 +3,59 @@
...
-    uint256 sharesAfter = sharesOf(id, user);
-
-    assert takerIsAuthorized;
-    assert user == seller => sharesAfter <= sharesBefore;
-    assert user == buyer => sharesAfter >= sharesBefore;
-    assert user != buyer && user != seller => sharesAfter == sharesBefore;
+function CVL_signer() returns address {
+    address result;
+    signed[result] = true;
+    return result;
```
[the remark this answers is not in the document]
MathisGD: it's enough to prove what we want

src/Midnight.sol
```diff
@@ -287,58 +252,44 @@ contract Midnight is IMidnight {
...
+        return obligationUnits;
     }
 
     function repay(Obligation memory obligation, uint256 obligationUnits, address onBehalf) external {
         require(onBehalf == msg.sender || isAuthorized[onBehalf][msg.sender], "unauthorized");
         bytes32 id = touchObligation(obligation);
 
-        borrowerState[id][onBehalf].debt -= UtilsLib.toUint128(obligationUnits);
+        // forge-lint: disable-next-line(unsafe-typecast) as obligationUnits <= totalUnits <= uint128.max
+        position[id][onBehalf].balance += int256(obligationUnits);
```
adhusson: this is not a safe cast

adhusson: if we repay a uint > `int256.max` it will set balance to a negative value and withdrawable to the repaid

for instance with `repaid = 2^255` from a balance of zero, at the end `balance == -2^255`.

adhusson: a spec repayDoesNotDecreaseBalance would be nice

src/interfaces/IMidnight.sol
```diff
@@ -33,6 +32,7 @@ struct Offer {
     address callback;
     bytes callbackData;
     address receiverIfMakerIsSeller;
+    bool exitOnly;
```
peyha, line 35: I don't find the naming super clean but at the same time I have no better idea

test/BundlerTest.sol
```diff
@@ -309,10 +222,10 @@ contract BundlerTest is BaseTest {
         _authorizeBundler();
 
         // Splitting across offers can cause up to 1 extra share of debt due to rounding.
```
peyha:
```suggestion
        // Splitting across offers can cause up to 1 extra unit of debt due to rounding.
```

src/interfaces/IMidnight.sol
MathisGD: I hesitated with secondaryOnly

src/Midnight.sol
```diff
@@ -463,24 +423,25 @@ contract Midnight is IMidnight {
...
+            _position.collateral[collateralIndex] = newCollateralOf;
             if (newCollateralOf == 0 && seizedAssets > 0) {
                 // forge-lint: disable-next-item(unsafe-typecast) as collateralIndex < MAX_COLLATERALS (128)
-                _state.activatedCollaterals &= ~uint128(1 << collateralIndex);
+                _position.activatedCollaterals &= ~uint128(1 << collateralIndex);
             }
             _obligationState.withdrawable += repaidUnits;
-            _state.debt -= UtilsLib.toUint128(repaidUnits);
+            // forge-lint: disable-next-line(unsafe-typecast) as repaidUnits <= debt <= totalUnits <= uint128.max
+            _position.balance += int256(repaidUnits);
```
adhusson: I think this is not safe either, nothing forces repaidUnits to be `<= int256.max`

certora/specs/SlashBeforeBalance.spec
```diff
@@ -0,0 +1,51 @@
...
+}
+
+/// GHOSTS ///
+
+persistent ghost mapping(bytes32 => mapping(address => bool)) slashed {
+    init_state axiom (forall bytes32 id. forall address user. !slashed[id][user]);
+}
+
+function slashSummary(bytes32 id, address user) {
+    slashed[id][user] = true;
```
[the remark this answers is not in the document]
MathisGD: interesting. we could also prove that after a slash we don't retouch positive balances

src/interfaces/IMidnight.sol
peyha: the thing puzzling me is that the related require (`require(offer.buy ? newBuyerBalance <= 0 : newSellerBalance >= 0, "crossed")`) correspond to the ability of the offer to cross to the other side rather than purely "exiting" the position
but at the same time crossing lender -> collateral does not look feasible because of the lack of collateral. Crossing borrower -> lender could work though

src/periphery/TakeAmountsLib.sol
```diff
@@ -10,56 +10,30 @@ import {WAD} from "../libraries/ConstantsLib.sol";
...
-    /// @dev Should not be used if buyerPrice > WAD, because not all buyerAssets are reachable then.
-    function buyerAssetsToShares(Midnight midnight, bytes32 id, Offer memory offer, uint256 targetBuyerAssets)
-        internal
-        view
-        returns (uint256)
-    {
-        uint256 offerPrice = TickLib.tickToPrice(offer.tick);
-        uint256 _tradingFee = midnight.tradingFee(id, UtilsLib.zeroFloorSub(offer.obligation.maturity, block.timestamp));
-        uint256 buyerPrice = offer.buy ? offerPrice : offerPrice + _tradingFee;
+        uint256 makerPrice = TickLib.tickToPrice(offer.tick);
```
QGarchery:
```suggestion
        uint256 offerPrice = TickLib.tickToPrice(offer.tick);
```
(optional) I don't see a big reason to change the naming here

src/Midnight.sol
```diff
@@ -206,49 +203,26 @@ contract Midnight is IMidnight {
...
+        int256 oldSellerBalance = position[id][seller].balance;
+        int256 newBuyerBalance = oldBuyerBalance + UtilsLib.toInt256(obligationUnits);
+        int256 newSellerBalance = oldSellerBalance - UtilsLib.toInt256(obligationUnits);
+        position[id][buyer].balance = newBuyerBalance;
+        position[id][seller].balance = newSellerBalance;
+        if (offer.exitOnly) require(offer.buy ? newBuyerBalance <= 0 : newSellerBalance >= 0, "crossed");
+        _obligationState.totalUnits = UtilsLib.toUint128(
+            uint256(_obligationState.totalUnits) + UtilsLib.negativePart(newSellerBalance)
+                + UtilsLib.negativePart(newBuyerBalance) - UtilsLib.negativePart(oldSellerBalance)
+                - UtilsLib.negativePart(oldBuyerBalance)
```
```suggestion
        _obligationState.totalUnits = UtilsLib.toUint128(
            uint256(_obligationState.totalUnits) + UtilsLib.positivePart(newSellerBalance)
                + UtilsLib.positivePart(newBuyerBalance) - UtilsLib.positivePart(oldSellerBalance)
                - UtilsLib.positivePart(oldBuyerBalance)
```
should be equivalent, feels more consistent with the fact that totalUnits is decreased in `withdraw` (it is also decreased in `liquidate` when there is bad debt but it's kind of an edge case)

MathisGD: actually I find it better with negative parts, because of the slashing. the invariant that we have is totalUnit == withdrawable + sum negative balances

yum0e:
👀

certora/specs/BalanceEffects.spec
```diff
@@ -0,0 +1,231 @@
...
+    function UtilsLib.mulDivDown(uint256 x, uint256 y, uint256 d) internal returns (uint256) => summaryMulDiv(x, y, d);
+    function UtilsLib.mulDivUp(uint256 x, uint256 y, uint256 d) internal returns (uint256) => summaryMulDiv(x, y, d);
+
+    // Assume no reentrancy: callbacks and token transfers do not re-enter Midnight.
+    function _.onBuy(Midnight.Obligation, address, uint256, uint256, uint256, bytes) external => NONDET;
+    function _.onSell(Midnight.Obligation, address, uint256, uint256, uint256, bytes) external => NONDET;
+    function _.onLiquidate(Midnight.Obligation, uint256, uint256, uint256, address, bytes) external => NONDET;
+    function _.onFlashLoan(address, uint256, bytes) external => NONDET;
+    function _.transfer(address, uint256) external => NONDET;
+    function _.transferFrom(address, address, uint256) external => NONDET;
```
adhusson: aren't some of these too strong? we do have reentrancy

certora/specs/Midnight.spec
```diff
@@ -96,7 +98,25 @@ rule liquidateInputOutputConsistency(env e, Midnight.Obligation obligation, uint
...
+    uint128 lossIndexAfter = userLossIndex(id, user);
+    assert lossIndexAfter >= lossIndexBefore;
+}
+
 /// INVARIANTS ///
 
 strong invariant totalUnitsEqualsSumNegativeBalancePlusWithdrawable(bytes32 id)
     to_mathint(totalUnits(id)) == sumNegativeBalanceOf[id] + to_mathint(withdrawable(id));
+
+invariant userLossIndexLeqObligationLossIndex(bytes32 id, address user)
```
adhusson: not strong?

certora/specs/OnlyAuthorizedCanChange.spec
```diff
@@ -3,91 +3,59 @@
...
-rule onlyAuthorizedCanChangeDebtExceptTakeAndLiquidate(env e, method f, calldataarg args, bytes32 id, address user) filtered { f -> f.selector != sig:liquidate(Midnight.Obligation, uint256, uint256, uint256, address, bytes).selector && f.selector != sig:take(uint256, address, address, bytes, address, Midnight.Offer, Midnight.Signature, bytes32, bytes32[]).selector } {
+/// An unauthorized caller cannot change a user's balance except via liquidate and slash.
+/// Assumes no reentrancy: callbacks (onBuy, onSell) and token transfers are not modeled as re-entering Midnight, so re-entrant balance changes are not covered.
+rule onlyAuthorizedCanChangeBalanceExceptLiquidateAndSlash(env e, method f, calldataarg args, bytes32 id, address user) filtered { f -> f.selector != sig:liquidate(Midnight.Obligation, uint256, uint256, uint256, address, bytes).selector && f.selector != sig:slash(bytes32, address).selector } {
     bool userIsAuthorized = user == e.msg.sender || isAuthorized(user, e.msg.sender);
+    bool userIsUnHealthy = !isHealthy(e, toObligation(e, id), id, user);
```
QGarchery: Why do we need to see if the user is unhealthy ? We are filtering out liquidate already

certora/specs/SlashBeforeBalance.spec
```diff
@@ -0,0 +1,68 @@
...
+
+function slashSummary(bytes32 id, address user) {
+    slashedAtLossIndex[id][user] = currentLossIndex[id];
+}
+
+/// HOOKS ///
+
+// Keep the ghost in sync with obligationState.lossIndex.
+hook Sload uint128 value obligationState[KEY bytes32 id].lossIndex {
+    require currentLossIndex[id] == value;
```
QGarchery:
```suggestion
   currentLossIndex[id] == value;
```
otherwise you are discarding all paths where the loss index is not up to date, but what you want to do is to update the current loss index

src/Midnight.sol
```diff
@@ -528,12 +522,14 @@ contract Midnight is IMidnight {
         if (_userLossIndex != lossIndex) {
             int256 balance = _position.balance;
             if (balance > 0) {
-                // forge-lint: disable-next-item(unsafe-typecast) as result <= uint256(balance) <= int256.max
-                _position.balance = int256(
+                balance = UtilsLib.toInt256(
+                    // forge-lint: disable-next-line(unsafe-typecast)
```
adhusson: needs explanation

adhusson (approved):
casts lgtm

certora/specs/SlashBeforeBalance.spec
QGarchery: btw, I don't think you need to have this `currentLossIndex`, you can just use the loss index of the obligation directly (the hook that matters is on `position[id][user].balance`, not on `obligationState[id].lossIndex`)

certora/specs/SlashBeforeBalance.spec
```diff
@@ -0,0 +1,68 @@
...
+        && f.selector != sig:isHealthy(Midnight.Obligation, bytes32, address).selector
+} {
+    f(e, args);
+    assert true;
+}
+
+rule balanceWrittenAfterSlash(method f, env e, calldataarg args) {
+    f(e, args);
+    assert true;
+}
```
QGarchery: Those 2 rules prove the same thing, except one filters more functions out

certora/specs/SlashBeforeBalance.spec
```diff
@@ -0,0 +1,68 @@
...
+    function IdLib.toId(Midnight.Obligation memory, uint256, address) internal returns (bytes32) => NONDET;
+
+    function slash(bytes32 id, address user) internal => slashSummary(id, user);
+}
+
+/// GHOSTS ///
+
+// Track the obligation's current lossIndex.
+persistent ghost mapping(bytes32 => uint128) currentLossIndex {
+    init_state axiom (forall bytes32 id. currentLossIndex[id] == 0);
```
QGarchery:
```suggestion
```
this is only useful for invariants
-> same for the other ghost variable below

certora/specs/OnlyAuthorizedCanChange.spec
MathisGD: we shouldn't be filtering liquidate actually

certora/specs/SlashBeforeBalance.spec
MathisGD: indeed! but then how can the second pass?

certora/specs/BalanceEffects.spec
```diff
@@ -0,0 +1,224 @@
...
+
+/// repay only changes position[id][onBehalf].balance.
+rule repayOnlyChangesTargetBalance(env e, Midnight.Obligation obligation, uint256 obligationUnits, address onBehalf, bytes32 anyId, address anyUser) {
+    bytes32 id = toId(e, obligation);
+    require anyUser != onBehalf || anyId != id;
+    int256 balanceBefore = balanceOf(anyId, anyUser);
+    repay(e, obligation, obligationUnits, onBehalf);
+    int256 balanceAfter = balanceOf(anyId, anyUser);
+    assert balanceAfter == balanceBefore;
+}
```
QGarchery: It would make sense to factorize those rules into 1. I find it easier to read, and that way we don't have to make sure that the inputs of the rule and the setup is sane multiple times

certora/specs/BalanceEffects.spec
```diff
@@ -0,0 +1,224 @@
...
+    bytes32 id = toId(e, obligation);
+    repay(e, obligation, obligationUnits, onBehalf);
+    int256 balanceAfter = balanceOf(id, onBehalf);
+    assert balanceAfter <= 0;
+}
+
+/// repay only changes position[id][onBehalf].balance.
+rule repayOnlyChangesTargetBalance(env e, Midnight.Obligation obligation, uint256 obligationUnits, address onBehalf, bytes32 anyId, address anyUser) {
+    bytes32 id = toId(e, obligation);
+    require anyUser != onBehalf || anyId != id;
```
QGarchery: let's add a reason string, to avoid having a warning. 
Better yet: put those as a premise of the property. So it would remove this line, and change the last line into:
```
assert anyUser != onBehalf || anyId != id => balanceAfter == balanceBefore;
```
-> also applies to `require`s below

certora/specs/BalanceEffects.spec
```diff
@@ -0,0 +1,224 @@
...
+/// SLASH ///
+
+/// slash can only decrease balances (or keep them unchanged).
+/// Requires the system invariant that the obligation's lossIndex >= the user's lossIndex.
+rule slashOnlyDecreasesBalance(env e, bytes32 id, address user) {
+    require userLossIndex(id, user) <= currentContract.obligationState[id].lossIndex, "TODO prove this";
+    int256 balanceBefore = balanceOf(id, user);
+    slash(e, id, user);
+    int256 balanceAfter = balanceOf(id, user);
+    assert to_mathint(balanceAfter) <= to_mathint(balanceBefore);
```
QGarchery: We can also check that `balanceAfter >= 0` (not a big thing but no costly to add)

certora/specs/BalanceEffects.spec
```diff
@@ -0,0 +1,224 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+methods {
+    function multicall(bytes[]) external => HAVOC_ALL DELETE;
+
+    function balanceOf(bytes32 id, address user) external returns (int256) envfree;
+    function balanceOfAfterSlashing(bytes32 id, address user) external returns (int256) envfree;
+    function userLossIndex(bytes32 id, address user) external returns (uint128) envfree;
+    function toId(Midnight.Obligation) external returns (bytes32);
```
QGarchery:
```suggestion
```

certora/specs/BalanceEffects.spec
```diff
@@ -0,0 +1,224 @@
...
+    function UtilsLib.mulDivDown(uint256 x, uint256 y, uint256 d) internal returns (uint256) => summaryMulDiv(x, y, d);
+    function UtilsLib.mulDivUp(uint256 x, uint256 y, uint256 d) internal returns (uint256) => summaryMulDiv(x, y, d);
+
+    // Assume no reentrancy: callbacks and token transfers do not re-enter Midnight.
+    function _.onBuy(Midnight.Obligation, address, uint256, uint256, uint256, bytes) external => NONDET;
+    function _.onSell(Midnight.Obligation, address, uint256, uint256, uint256, bytes) external => NONDET;
+    function _.onLiquidate(Midnight.Obligation, uint256, uint256, uint256, address, bytes) external => NONDET;
+    function _.onFlashLoan(address, uint256, bytes) external => NONDET;
+    function _.transfer(address, uint256) external => NONDET;
+    function _.transferFrom(address, address, uint256) external => NONDET;
```
QGarchery:
```suggestion
```
useless because we already summarize `SafeTransferLib`

certora/specs/BalanceEffects.spec
```diff
@@ -0,0 +1,224 @@
...
+    function _.price() external => NONDET;
+
+    // Summarize internals irrelevant to balance tracking.
+    function IdLib.storeInCode(Midnight.Obligation memory) internal returns (address) => NONDET;
+    function SafeTransferLib.safeTransfer(address, address, uint256) internal => NONDET;
+    function SafeTransferLib.safeTransferFrom(address, address, address, uint256) internal => NONDET;
+    function UtilsLib.isLeaf(bytes32, bytes32, bytes32[] memory) internal returns (bool) => NONDET;
+    function UtilsLib.msb(uint256) internal returns (uint256) => NONDET;
+    function TickLib.tickToPrice(uint256) internal returns (uint256) => NONDET;
+    function TickLib.wExp(int256) internal returns (uint256) => NONDET;
```
QGarchery:
```suggestion
```
useless because we already summarize `tickToPrice`

certora/specs/BalanceEffects.spec
```diff
@@ -0,0 +1,224 @@
...
+
+/// take changes taker's balance by +/- obligationUnits relative to its post-slash balance.
+rule takeChangesTakerBalance(env e, uint256 obligationUnits, address taker, address takerCallback, bytes takerCallbackData, address receiver, Midnight.Offer offer, Midnight.Signature signature, bytes32 root, bytes32[] proof) {
+    bytes32 id = toId(e, offer.obligation);
+    int256 takerPostSlash = balanceOfAfterSlashing(id, taker);
+
+    take(e, obligationUnits, taker, takerCallback, takerCallbackData, receiver, offer, signature, root, proof);
+
+    int256 takerAfter = balanceOf(id, taker);
+    mathint delta = offer.buy ? -to_mathint(obligationUnits) : to_mathint(obligationUnits);
```
QGarchery:
```suggestion
    mathint delta = offer.buy ? -obligationUnits : obligationUnits;
```
actually we can remove `to_mathint` everywhere (but not `mathint`), because the arithmetic operations will convert to mathint automatically when needed

src/Midnight.sol
```diff
@@ -440,7 +387,13 @@ contract Midnight is IMidnight {
         require(block.timestamp > obligation.maturity || originalDebt > maxDebt, "position is not liquidatable");
 
         if (badDebt > 0) {
-            _state.debt -= UtilsLib.toUint128(badDebt);
+            _position.balance += UtilsLib.toInt256(badDebt);
+            uint256 oldTotalUnits = _obligationState.totalUnits;
+            _obligationState.lossIndex = UtilsLib.toUint128(
```
peyha, line 400: worth noting that with this formula, when ignoring roundings and with consecutive bad debt realizations of a constant proportion epsilon of the total units , the loss index decreases exponentially (after n events, the loss index will be `type(uint128).max*(1-(1-epsilon)^n`). The fact that it's exponentially approaching the max could be worrying though in practice when checking various value it seems ok

certora/specs/BalanceEffects.spec
```diff
@@ -0,0 +1,224 @@
...
+    int256 balanceAfter = balanceOf(anyId, anyUser);
+    assert balanceAfter == balanceBefore;
+}
+
+/// WITHDRAW ///
+
+/// withdraw decreases onBehalf's post-slash balance by exactly obligationUnits.
+rule withdrawDecreasesBalanceExactly(env e, Midnight.Obligation obligation, uint256 obligationUnits, address onBehalf, address receiver) {
+    bytes32 id = toId(e, obligation);
+    int256 balanceBeforeSlash = balanceOfAfterSlashing(id, onBehalf);
```
QGarchery:
```suggestion
    int256 balanceAfterSlash = balanceOfAfterSlashing(id, onBehalf);
```

certora/confs/BalanceEffects.conf
QGarchery, line 1: Weirdly this file is not checked in the CI, I don't understand why for the moment:

Image

certora/specs/BalanceEffects.spec
QGarchery: Agree that it is too strong to not assume reentrancy. I still think that the current assumption is fine but it deserves a longer comment. The reasoning is the following:
- for checking if the balance can or cannot change (when checking for a different user, or a different obligation id), the property that is interesting is if the function itself can change the balance. So yes the callback of a take can change the balance of another random user, but it's still interesting to show that the body of `take` itself cannot change it
- for the positive/negative balances, the corresponding functions don't have callbacks except `liquidate`. For `liquidate` it's also not true that the balance of the borrower after is <= 0, but it's still interesting to show that the body of liquidate doesn't make the borrower a lender. So it's the same reason as the point above
- for exact balances change, for example `assert makerAfter == makerPostSlash + delta;` the important point is that the callback cannot affect the delta. So even with reentry and multiple interactions, the changes expected are the cumulative expected changes of all the interactions. To make that clearer we could also define the delta before the corresponding function is even called

src/Midnight.sol
QGarchery: > with consecutive bad debt realizations of a constant proportion epsilon of the total units

but with this assumption, doesn't it also mean that the price drops exponentially ? It would make sense to have a property that bounds the price loss index change compared to a price drop. I know we have a high-level formula that bounds the bad debt for a given price drop, but here it would be about the loss index and taking into account rounding and other implementation details

src/periphery/TakeAmountsLib.sol
```diff
@@ -10,56 +10,30 @@ import {WAD} from "../libraries/ConstantsLib.sol";
...
-            return targetBuyerAssets.mulDivUp(WAD, buyerPrice)
-                .mulDivUp(midnight.totalShares(id) + 1, midnight.totalUnits(id) + 1);
-        } else {
-            return targetBuyerAssets.mulDivDown(WAD, buyerPrice)
-                .mulDivDown(midnight.totalShares(id) + 1, midnight.totalUnits(id) + 1);
-        }
+        return offer.buy ? targetBuyerAssets.mulDivUp(WAD, buyerPrice) : targetBuyerAssets.mulDivDown(WAD, buyerPrice);
     }
 
-    // Forward: sellerAssets = offer.buy ? unitsDown.mulDivDown(sellerPrice, WAD) : unitsUp.mulDivUp(sellerPrice, WAD).
```
adhusson, line 48: why remove this?

certora/specs/SlashBeforeBalance.spec
QGarchery: It's because the `require`s in the hooks should be `assert`s !

src/Midnight.sol
peyha: Not necessarily, I was thinking about a dust position attack where an attacker would have numerous small positions in bad debt. If all position are created at the same time, a single price drop could lead to consecutive bad debt realization

peyha: Asymptotically, it's not so worrying, when epsilon << 1, the loss index is approx equal to `type(uint128).max*n*epsilon`so it grows linearly so it's fine for dust position attacks. Consecutive realization of 50% of the obligation would be more annoying but it seems quite unlikely

src/periphery/TakeAmountsLib.sol
MathisGD: didn't find it very useful. but can re-add if you prefer

adhusson: yes i think its good for readers

certora/confs/BalanceEffects.conf
QGarchery: fixed in https://github.com/morpho-org/midnight/pull/521

src/Midnight.sol
```diff
@@ -206,49 +203,26 @@ contract Midnight is IMidnight {
...
+        uint256 newConsumed = consumed[offer.maker][offer.group] += obligationUnits;
+        require(newConsumed <= offer.obligationUnits, "consumed");
+
+        int256 oldBuyerBalance = position[id][buyer].balance;
+        int256 oldSellerBalance = position[id][seller].balance;
+        int256 newBuyerBalance = oldBuyerBalance + UtilsLib.toInt256(obligationUnits);
+        int256 newSellerBalance = oldSellerBalance - UtilsLib.toInt256(obligationUnits);
+        position[id][buyer].balance = newBuyerBalance;
+        position[id][seller].balance = newSellerBalance;
+        if (offer.exitOnly) require(offer.buy ? newBuyerBalance <= 0 : newSellerBalance >= 0, "crossed");
```
peyha: Why did we implement crossing like this ? It prevents partial fills, would have expected something looking like
````
if (!offer.canCross) require(offer.buy ? newBuyerBalance >= 0 : newSellerBalance <= 0, "crossed");
```

certora/specs/BalanceEffects.spec
adhusson: so maybe add an issue to check with callbacks later?

adhusson: adding optimistically that i understood your comment

src/Midnight.sol
QGarchery: I think that it works as expected currently, and that it doesn't prevent partial fills. Assume that the boolean is activated (meaning `exitOnly` or equivalently `!canCross`):
- if `offer.buy`, then the balance goes up after the trade, but it is required to stay non-positive. So you could have a partial fill and end up with a negative balance
- if `!offer.buy`, then the balance goes down after the trade, but it is required to stay non-negative. So you could have a partial fii and end up with a positive balance

src/Midnight.sol
```diff
@@ -574,22 +563,27 @@ contract Midnight is IMidnight {
...
-        return borrowerState[id][user].debt;
+    function balanceOfAfterSlashing(bytes32 id, address user) public view returns (int256) {
+        Position storage _position = position[id][user];
+        int256 balance = _position.balance;
+        uint128 _userLossIndex = _position.lossIndex;
+        uint128 lossIndex = obligationState[id].lossIndex;
+        if (balance > 0 && _userLossIndex != lossIndex) {
+            // forge-lint: disable-next-item(unsafe-typecast) as result <= uint256(balance) <= int256.max
+            return
+                int256(uint256(balance).mulDivDown(type(uint128).max - lossIndex, type(uint128).max - _userLossIndex));
```
peyha: why is `slash` using UtilsLib.toInt256 and not here?

certora/specs/BalanceEffects.spec
QGarchery: We can tackle that separately, but maybe just adding comment is enough, as I think the spec would become quite messy if we tried to prove it with callbacks

test/LiquidationTest.sol
```diff
@@ -270,7 +271,50 @@ contract LiquidationTest is BaseTest {
...
+            ? previousLossIndex
+            : type(uint128).max
+                - (type(uint128).max - previousLossIndex).mulDivDown(oldTotalUnits - expectedBadDebt, oldTotalUnits);
+
+        vm.expectEmit(true, true, true, true);
+        emit EventsLib.Liquidate(address(this), id, 0, 0, 0, borrower, expectedBadDebt, expectedLossIndex);
+        midnight.liquidate(obligation, 0, 0, 0, borrower, "");
+    }
+
+    function testSlashEvent(uint256 units) public {
```
peyha, line 307: could be moved to `OtherFunctionsTest.sol` but no big deal

test/MaxAmountsTest.sol
```diff
@@ -33,6 +33,9 @@ contract MaxAmountsTest is BaseTest {
         obligation.rcfThreshold = 0;
 
         id = toId(obligation);
+
+        vm.prank(borrower);
+        midnight.setIsAuthorized(borrower, address(this), true);
```
peyha:
```suggestion
        authorize(borrower, address(this))
```
nitpick

src/Midnight.sol
peyha: oh yes you're right I mixed it up

test/LiquidationTest.sol
MathisGD: as it's a bad debt related thing, it can be justified here

certora/specs/BalanceEffects.spec
```diff
@@ -0,0 +1,170 @@
...
+
+    int256 balanceBefore = balanceOf(id, user);
+    int256 otherBalanceBefore = balanceOf(anyId, anyUser);
+
+    slash(e, id, user);
+
+    int256 balanceAfter = balanceOf(id, user);
+    int256 otherBalanceAfter = balanceOf(anyId, anyUser);
+
+    assert balanceAfter <= balanceBefore;
```
QGarchery:
```suggestion
    assert balanceAfter <= balanceBefore;
    assert balanceAfter == balanceAfterSlashingBefore;
```
it is more about `balanceAfterSlashing` than about `slash`, but still interesting.

We could also show: 
- that `balanceAfterSlashing` is either always to the balance of the interacting user (after the interaction)
- that starting to a `balanceAfterSlashing` that is equal to the `balanceOf`, the only way that it differs is because of an interaction

-> feel free to put that as an issue in the wish list

test/LiquidationTest.sol
peyha: yes it's fine

certora/specs/SlashBeforeBalance.spec
```diff
@@ -0,0 +1,66 @@
...
+        && f.selector != sig:debtOf(bytes32, address).selector
+        && f.selector != sig:balanceOfAfterSlashing(bytes32, address).selector
+        && f.selector != sig:isHealthy(Midnight.Obligation, bytes32, address).selector
+} {
+    f(e, args);
+    assert !balanceReadWithoutSlash;
+}
+
+rule balanceWrittenAfterSlash(method f, env e, calldataarg args) {
+    f(e, args);
```
QGarchery, line 56:
```suggestion
    require !balanceWrittenWithoutSlash, "initialize the ghost variable";
    f(e, args);
```
to make the spec go through

certora/specs/SlashBeforeBalance.spec
```diff
@@ -0,0 +1,66 @@
...
+
+// View functions that read balanceOf don't call slash (they can't mutate state).
+rule balanceReadAfterSlash(method f, env e, calldataarg args)
+filtered {
+    f -> f.selector != sig:balanceOf(bytes32, address).selector
+        && f.selector != sig:debtOf(bytes32, address).selector
+        && f.selector != sig:balanceOfAfterSlashing(bytes32, address).selector
+        && f.selector != sig:isHealthy(Midnight.Obligation, bytes32, address).selector
+} {
+    f(e, args);
```
QGarchery, line 50:
```suggestion
    require !balanceReadWithoutSlash, "initialize the ghost variable";
    f(e, args);
```

certora/specs/SlashBeforeBalance.spec
QGarchery: I think this was not changed

src/periphery/TakeBundler.sol
```diff
@@ -113,15 +51,11 @@ contract TakeBundler {
...
             ) returns (
-                uint256 filledBuyerAssets,
-                uint256 filledSellerAssets,
-                uint256 filledObligationUnits,
-                uint256 filledObligationShares
+                uint256 filledBuyerAssets, uint256 filledSellerAssets, uint256 filledObligationUnits
             ) {
                 totalFilledUnits += filledObligationUnits;
                 totalBuyerAssets += filledBuyerAssets;
                 totalSellerAssets += filledSellerAssets;
```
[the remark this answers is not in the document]
QGarchery: if this is documented, then it means that the issue is acknowledged

peyha (approved):
LGTM

test/BundlerTest.sol
```diff
@@ -212,54 +135,44 @@ contract BundlerTest is BaseTest {
...
+            obligationUnits: offerUnits1,
             sig: sig([offers[1]]),
             root: root([offers[1]]),
             proof: proof([offers[1]])
         });
 
         _authorizeBundler();
 
-        if (offerShares1 >= units - fromOffer0) {
+        if (offerUnits1 >= units - fromOffer0) {
```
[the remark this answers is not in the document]
MathisGD: very useful!

certora/specs/BalanceEffects.spec
```diff
@@ -0,0 +1,174 @@
...
+    assert anyUser != borrower || anyId != id => otherBalanceAfter == otherBalanceBefore;
+}
+
+/// SLASH ///
+
+/// slash can only decrease balances (or keep them unchanged), preserves non-positive balances,
+/// leaves the balance non-negative, and only changes position[id][user].balance.
+/// Requires the system invariant that the obligation's lossIndex >= the user's lossIndex.
+rule slashEffects(env e, bytes32 id, address user, bytes32 anyId, address anyUser) {
+    require userLossIndex(id, user) <= currentContract.obligationState[id].lossIndex, "TODO prove this";
```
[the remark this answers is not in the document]
MathisGD: it's verified in Midnight.spec

MathisGD:
@codex review

src/periphery/TakeBundler.sol
```diff
@@ -145,23 +77,20 @@ contract TakeBundler {
         address receiverIfTakerIsSeller,
         Take[] calldata takes,
         uint256 minObligationUnits,
-        uint256 maxObligationUnits,
-        uint256 minObligationShares,
-        uint256 maxObligationShares
+        uint256 maxObligationUnits
     ) external {
         require(taker == msg.sender || midnight.isAuthorized(taker, msg.sender), "unauthorized");
         bytes32 id = midnight.touchObligation(takes[0].offer.obligation); // to have the correct trading fees.
```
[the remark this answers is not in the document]
MathisGD: it's ok

src/Midnight.sol
```diff
@@ -574,22 +563,28 @@ contract Midnight is IMidnight {
         return abi.decode(create2Address.code, (Obligation));
     }
 
-    function debtOf(bytes32 id, address user) external view returns (uint256) {
-        return borrowerState[id][user].debt;
+    function balanceOfAfterSlashing(bytes32 id, address user) public view returns (int256) {
+        Position storage _position = position[id][user];
+        int256 balance = _position.balance;
+        uint128 _userLossIndex = _position.lossIndex;
```
adhusson, line 568: what about calling `lossIndex` `lossFactor` so we don't call 2 different things LI, but instead we call that one LF?

src/Midnight.sol
```diff
@@ -440,7 +387,13 @@ contract Midnight is IMidnight {
         require(block.timestamp > obligation.maturity || originalDebt > maxDebt, "position is not liquidatable");
 
         if (badDebt > 0) {
-            _state.debt -= UtilsLib.toUint128(badDebt);
+            _position.balance += UtilsLib.toInt256(badDebt);
+            uint256 oldTotalUnits = _obligationState.totalUnits;
+            _obligationState.lossIndex = UtilsLib.toUint128(
+                type(uint128).max
+                    - (type(uint128).max - _obligationState.lossIndex)
+                    .mulDivDown(oldTotalUnits - badDebt, oldTotalUnits)
```
QGarchery, line 403: We have to round up the loss index, as we do here, otherwise you don't have the invariant that the sum of positive balances after slashing is less than total units. But then it can effectively mean that the bad debt is rounded up for lenders. So multiple slashing could be an issue, if you can make so there are multiple bad debts

src/Midnight.sol
MathisGD: I added a comment about roundings

yum0e:
👀

adhusson (approved):

PR #424, opened 2026-02-27
midnight

MathisGD:
https://morpholabs.slack.com/archives/C08T3G97VJS/p1772215894369679

### DIFF
--- a/certora/specs/Liquidate.spec
+++ b/certora/specs/Liquidate.spec
@@ -3,10 +3,10 @@
 methods {
     function multicall(bytes[]) external => HAVOC_ALL DELETE;
 
-    function isHealthy(MorphoV2.Obligation obligation, bytes20 id, address borrower) external returns (bool) envfree;
+    function isHealthy(Midnight.Obligation obligation, bytes20 id, address borrower) external returns (bool) envfree;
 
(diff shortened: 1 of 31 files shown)


qdrn (approved):
Couldn't find any mention of v2. LGTM.

certora/specs/Liquidate.spec
adhusson:
```suggestion
bool isHealthyBefore = isHealthy(obligation, id, borrower);
liquidate(e, obligation, collateralIndex, seizedAssets, repaidUnits, borrower, data);

// it's okay to check only after the call that the prover chose the correct id.
require id == lastId, "id should be derived from obligation";

assert !isHealthyBefore || e.block.timestamp > obligation.maturity, "liquidate can only be called on unhealthy obligations";
```

adhusson (approved):

certora/specs/Liquidate.spec
MathisGD: I don't understand the change

adhusson: `isHealty` -> `isHealthy`, fixed by hand to ensure no mistake

QGarchery (commented):
LGTM

PR #430, opened 2026-03-02
lif formula

MathisGD:
first step of lif modularization

### DIFF
--- a/src/Midnight.sol
+++ b/src/Midnight.sol
@@ -10,7 +10,7 @@ import {
     WAD,
     ORACLE_PRICE_SCALE,
     FEE_STEP,
-    MAX_LIF,
+    LIQUIDATION_CURSOR,
     TIME_TO_MAX_LIF,
     MAX_COLLATERALS,
     MAX_COLLATERALS_PER_BORROWER,
@@ -425,7 +425,7 @@ contract Midnight is IMidnight {
     /// the liquidation could leave a collateral with a value that would not be enough to repay rcfThreshold units.
     /// @dev Recovery close factor means that debtOf - repaidUnits >= maxDebt - repaidUnits*LIF*LLTV, which is
     /// equivalent to repaidUnits <= (debtOf-maxDebt) / (1 - LIF*LLTV).
-    /// @dev If an account is healthy, the LIF grows linearly from 1 at maturity to MAX_LIF at maturity +
+    /// @dev If an account is healthy, the LIF grows linearly from 1 at maturity to maxLif(lltv) at maturity +
     /// TIME_TO_MAX_LIF.
     /// @dev Returns the seized assets and the repaid units.
     function liquidate(
@@ -453,7 +453,7 @@ contract Midnight is IMidnight {
             if (i == collateralIndex) liquidatedCollatPrice = price;
             uint256 collateralQuoted = collateralOf[id][borrower][i].mulDivDown(price, ORACLE_PRICE_SCALE);
             maxDebt += collateralQuoted.mulDivDown(_collateral.lltv, WAD);
-            badDebt = badDebt.zeroFloorSub(collateralQuoted.mulDivDown(WAD, MAX_LIF));
+            badDebt = badDebt.zeroFloorSub(collateralQuoted.mulDivDown(WAD, maxLif(_collateral.lltv)));
             bitmap ^= (1 << i);
         }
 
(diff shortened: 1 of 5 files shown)


src/Midnight.sol
```diff
@@ -567,6 +568,10 @@ contract Midnight is IMidnight {
 
     /// VIEW FUNCTIONS ///
 
+    function maxLif(uint256 lltv) public pure returns (uint256) {
+        return WAD.mulDivDown(WAD, WAD - LIQUIDATION_CURSOR.mulDivDown(WAD - lltv, WAD));
```
peyha, line 573: Don't we also want to cap it at 115% just like we did on Blue ?

MathisGD: for the moment no, let's see when we settle on the final formula

src/Midnight.sol
```diff
@@ -567,6 +568,10 @@ contract Midnight is IMidnight {
 
     /// VIEW FUNCTIONS ///
 
+    function maxLif(uint256 lltv) public pure returns (uint256) {
```
peyha, line 572:
```suggestion
    function preMaturityLif(uint256 lltv) public pure returns (uint256) {
```
prop (probably more relevant with LI tiers than in this design)

MathisGD: but post maturity the lif comes back to this after 15 min?

peyha: yes, though the post maturity lif can be seen as linearly increasing between 0 and pre maturity lif. My point is that max might not be the max for a given lltv (but at the same time we could also add another parameter to the function to specify high or low)

MathisGD: yes, will do that in an other PR

peyha (approved):

src/Midnight.sol
```diff
@@ -453,7 +453,7 @@ contract Midnight is IMidnight {
             if (i == collateralIndex) liquidatedCollatPrice = price;
             uint256 collateralQuoted = collateralOf[id][borrower][i].mulDivDown(price, ORACLE_PRICE_SCALE);
             maxDebt += collateralQuoted.mulDivDown(_collateral.lltv, WAD);
-            badDebt = badDebt.zeroFloorSub(collateralQuoted.mulDivDown(WAD, MAX_LIF));
+            badDebt = badDebt.zeroFloorSub(collateralQuoted.mulDivDown(WAD, maxLif(_collateral.lltv)));
```
adhusson, line 457: related to maxlif capping, this divides by zero when `lltv=wad`

MathisGD: wait it doesn't divide by zero right?

src/Midnight.sol
adhusson, line 487: related to maxlif capping, this divides by zero when `lltv=wad`

src/Midnight.sol
adhusson: clicked the wrong line, it's below (will add a comment there)

src/Midnight.sol
adhusson: It is probably fine anyway because in this situation you can't liquidate anything, just realize bad debt. And if you don't liquidate this `maxRepaid` is never computed.

MathisGD: documented as edge-case

src/Midnight.sol
```diff
@@ -552,7 +554,7 @@ contract Midnight is IMidnight {
             for (uint256 i = 0; i < obligation.collaterals.length; i++) {
                 address collateralToken = obligation.collaterals[i].token;
                 require(collateralToken > previousCollateralToken, "collaterals not sorted");
-                require(obligation.collaterals[i].lltv < WAD.mulDivDown(WAD, MAX_LIF), "lltv too high or LIF too high"); // temporary.
+                require(obligation.collaterals[i].lltv <= WAD, "lltv too high");
```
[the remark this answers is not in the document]
MathisGD: this is documented

QGarchery (approved):
See #442

src/Midnight.sol
```diff
@@ -425,8 +425,9 @@ contract Midnight is IMidnight {
     /// the liquidation could leave a collateral with a value that would not be enough to repay rcfThreshold units.
     /// @dev Recovery close factor means that debtOf - repaidUnits >= maxDebt - repaidUnits*LIF*LLTV, which is
     /// equivalent to repaidUnits <= (debtOf-maxDebt) / (1 - LIF*LLTV).
-    /// @dev If an account is healthy, the LIF grows linearly from 1 at maturity to MAX_LIF at maturity +
+    /// @dev If an account is healthy, the LIF grows linearly from 1 at maturity to maxLif(lltv) at maturity +
     /// TIME_TO_MAX_LIF.
+    /// @dev Liquidations non zero amounts revert if LLTV = 1.
```
qdrn, line 430:
```suggestion
    /// @dev Liquidating non zero amounts revert if LLTV = 1.
```

PR #436, opened 2026-03-02
certora: use foundry solc

MathisGD:


MathisGD:
actually this doesn't work, mb

PR #439, opened 2026-03-03
Use forge solc in certora CI

QGarchery:
inspiration from #436

### DIFF
--- a/.github/workflows/certora.yml
+++ b/.github/workflows/certora.yml
@@ -54,7 +54,7 @@ jobs:
 
       - name: Copy solc
         run: |
-          forge install
+          forge build
           ls -la ~/.local/share/svm
           sudo cp ~/.local/share/svm/solc-0.8.31/solc-0.8.31 /usr/local/bin/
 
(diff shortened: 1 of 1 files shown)


.github/workflows/certora.yml
```diff
@@ -49,11 +49,14 @@ jobs:
           echo "Fix lint errors by applying the following diff to $file:"
           diff -u "$file" <(certoraCVLFormatter "$file")
 
+      - name: Install Foundry
+        uses: foundry-rs/foundry-toolchain@v1
+        run: forge build
+
       - name: Install solc
```
MathisGD:
```suggestion
      - name: Copy solc
```

.github/workflows/certora.yml
```diff
@@ -49,11 +49,14 @@ jobs:
           echo "Fix lint errors by applying the following diff to $file:"
           diff -u "$file" <(certoraCVLFormatter "$file")
 
+      - name: Install Foundry
+        uses: foundry-rs/foundry-toolchain@v1
+        run: forge build
```
MathisGD: why?

QGarchery: I need to run forge so that it installs the binary, but maybe I can just do forge install

MathisGD: oh ok, makes sense

MathisGD (approved):

PR #442, opened 2026-03-03
[Certora] lif and lltv bounds

QGarchery:


MathisGD (approved):

lilCertora (approved):

jhoenicke (commented):
Looks good to me.  The property `lltv*lif <= WAD()*WAD()`  is exactly what I need to prove that liquidated users stay healthy if they were healthy before.

PR #451, opened 2026-03-04
ids as bytes32

QGarchery:
context in [this thread](https://morpholabs.slack.com/archives/C08A40KKN9W/p1771925393982049), inspired from @jhoenicke's idea [here](https://morpholabs.slack.com/archives/C0ACSA8TNMA/p1772564829296709?thread_ts=1772441125.820019&cid=C0ACSA8TNMA)

Why it works:
- assume that there cannot be any bytes32 collisions (but bytes20 collisions are still possible)
- now the `toId` function returns a bytes32, so because of the assumption above, it cannot be misused
- `toObligation` uses a bytes20 internally, but it checks that the obligation is created. So the passed id corresponds necessarily to the `toId` of an obligation, and because of the above assumption that obligation is unique

### DIFF
--- a/certora/specs/CreatedObligations.spec
+++ b/certora/specs/CreatedObligations.spec
@@ -7,11 +7,11 @@ methods {
     function multicall(bytes[]) external => HAVOC_ALL DELETE;
     function _.price() external => NONDET;
 
-    function Midnight.totalUnits(bytes20) external returns (uint256) envfree;
-    function Midnight.totalShares(bytes20) external returns (uint256) envfree;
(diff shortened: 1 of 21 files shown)


peyha (commented):
The added cost (both in gas and code complexity) is very low and it completely solves the issue so I think we should go for it

adhusson (approved):
I don't have the added cost of created in mind but the complexity of this vs. adding the id prefix to the bytecode of the sstore2 contract is convincing.

MathisGD (approved):

PR #453, opened 2026-03-04
[rounding 3] round in favor of the fee recipient

QGarchery:
Note that this is on top of #270, so it rounds in favor of the passive lenders, in favor of the maker, in favor of the fee recipient. The taker is taking all the rounding losses
[Motivation](https://morpholabs.slack.com/archives/C08A40KKN9W/p1772789576617189?thread_ts=1772731881.347929&cid=C08A40KKN9W) for merging this PR

### DIFF
--- a/src/Midnight.sol
+++ b/src/Midnight.sol
@@ -201,12 +201,6 @@ contract Midnight is IMidnight {
                 offer.receiverIfMakerIsSeller
             );
 
-        uint256 offerPrice = TickLib.tickToPrice(offer.tick);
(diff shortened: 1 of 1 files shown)


MathisGD (commented):
not completely convinced. notably, it makes the computation for your price as a taker much more complex (before it was offer.price +- tradingFee), for an not super tangible benefit

QGarchery:
Closing because the computation are not "free" after merging #270. Having a clearer computation, to be able to have your exact price, feels more important than to be able to compute the fees without having to do the full computation of the buyer and seller assets

PR #455, opened 2026-03-04
fix: consume attack

MathisGD:
if you are trying to put consumed to type(uint256).max, someone can front-run and take 1 of one of your offers to make you revert. one idea could be to target type(uint128).max, but i thought that it was simpler to do it this way. credits to @spennyp for the [finding](https://morpholabs.slack.com/archives/C08UY4ZS5PH/p1772582360067159)

### DIFF
--- a/src/Midnight.sol
+++ b/src/Midnight.sol
@@ -490,8 +490,11 @@ contract Midnight is IMidnight {
         return (seizedAssets, repaidUnits);
     }
 
+    /// @dev Pass type(uint256).max to cancel all offers in the group. Never reverts.
     function consume(bytes32 group, uint256 amount) external {
-        consumed[msg.sender][group] += amount;
+        require(amount >= consumed[msg.sender][group], "consumed");
+
+        consumed[msg.sender][group] = amount;
 
         emit EventsLib.Consume(msg.sender, group, amount);
     }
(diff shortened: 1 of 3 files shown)


src/Midnight.sol
```diff
@@ -490,8 +490,11 @@ contract Midnight is IMidnight {
         return (seizedAssets, repaidUnits);
     }
 
+    /// @dev Pass type(uint256).max to cancel all offers in the group. Never reverts.
     function consume(bytes32 group, uint256 amount) external {
```
peyha: Now that this function is not incremental anymore, should we change its name ?  something like `setConsumed` would work

src/Midnight.sol
```diff
@@ -490,8 +490,11 @@ contract Midnight is IMidnight {
         return (seizedAssets, repaidUnits);
     }
 
+    /// @dev Pass type(uint256).max to cancel all offers in the group. Never reverts.
```
```suggestion
    /// @dev Pass type(uint256).max to cancel all offers in the group. Cancelling in this manner never reverts.
```
otherwise one could think that the function never reverts

src/Midnight.sol
QGarchery, line 1: Looks good modulo [the renaming](https://github.com/morpho-org/midnight/pull/455/changes#r2888643605). I thought that it was fine (and I think it's not a big issue because of the workaround), and that it allowed to easily have offers increment the `consumed` variable by a given `delta`. But it's still possible onchain: just read the `oldConsumed` and set the new amount to `oldConsumed + delta`

QGarchery (approved):

PR #457, opened 2026-03-04
bad debt underestimation

MathisGD:
https://morpholabs.slack.com/archives/C08A40KKN9W/p1772643325804879

### DIFF
--- a/src/Midnight.sol
+++ b/src/Midnight.sol
@@ -425,10 +425,10 @@ contract Midnight is IMidnight {
             Collateral memory _collateral = obligation.collaterals[i];
             uint256 price = IOracle(_collateral.oracle).price();
             if (i == collateralIndex) liquidatedCollatPrice = price;
-            maxDebt += collateralOf[id][borrower][i].mulDivDown(price, ORACLE_PRICE_SCALE)
-                .mulDivDown(_collateral.lltv, WAD);
+            uint256 _collateralOf = collateralOf[id][borrower][i];
+            maxDebt += _collateralOf.mulDivDown(price, ORACLE_PRICE_SCALE).mulDivDown(_collateral.lltv, WAD);
             badDebt = badDebt.zeroFloorSub(
-                collateralOf[id][borrower][i].mulDivUp(price, ORACLE_PRICE_SCALE).mulDivUp(WAD, _collateral.maxLif)
+                _collateralOf.mulDivUp(price, ORACLE_PRICE_SCALE).mulDivUp(WAD, _collateral.maxLif)
             );
             bitmap ^= (1 << i);
         }
@@ -458,7 +458,7 @@ contract Midnight is IMidnight {
                 uint256 lltv = obligation.collaterals[collateralIndex].lltv;
                 // Rounded up to avoid consecutive max liquidations.
                 // Acknowledged that the position could be slightly healthy after a liquidation.
-                // Note that debt >= maxDebt at this point.
+                // Note that debt >= maxDebt after the bad debt realization.
                 uint256 maxRepaid = (_state.debt - maxDebt).mulDivUp(WAD, WAD - lif.mulDivUp(lltv, WAD));
                 require(
                     repaidUnits <= maxRepaid
--- a/src/Midnight.sol
+++ b/src/Midnight.sol
@@ -427,7 +427,7 @@ contract Midnight is IMidnight {
             if (i == collateralIndex) liquidatedCollatPrice = price;
             uint256 collateralQuoted = collateralOf[id][borrower][i].mulDivDown(price, ORACLE_PRICE_SCALE);
             maxDebt += collateralQuoted.mulDivDown(_collateral.lltv, WAD);
-            badDebt = badDebt.zeroFloorSub(collateralQuoted.mulDivDown(WAD, _collateral.maxLif));
+            badDebt = badDebt.zeroFloorSub(collateralQuoted.mulDivUp(WAD, _collateral.maxLif));
             bitmap ^= (1 << i);
         }
 
--- a/src/Midnight.sol
+++ b/src/Midnight.sol
@@ -458,7 +458,7 @@ contract Midnight is IMidnight {
                 uint256 lltv = obligation.collaterals[collateralIndex].lltv;
                 // Rounded up to avoid consecutive max liquidations.
                 // Acknowledged that the position could be slightly healthy after a liquidation.
-                // Note that debt >= Σ collateralQuoted * 1 / lif >= Σ collateralQuoted * lltv = maxDebt.
+                // Note that debt >= maxDebt at this point.
                 uint256 maxRepaid = (_state.debt - maxDebt).mulDivUp(WAD, WAD - lif.mulDivUp(lltv, WAD));
                 require(
                     repaidUnits <= maxRepaid
--- a/src/Midnight.sol
+++ b/src/Midnight.sol
@@ -458,7 +458,7 @@ contract Midnight is IMidnight {
                 uint256 lltv = obligation.collaterals[collateralIndex].lltv;
                 // Rounded up to avoid consecutive max liquidations.
                 // Acknowledged that the position could be slightly healthy after a liquidation.
-                // Note that debt >= maxDebt after the bad debt realization.
+                // Note that debt >= maxDebt in this branch.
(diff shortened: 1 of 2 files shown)


src/Midnight.sol
```diff
@@ -427,7 +427,7 @@ contract Midnight is IMidnight {
             if (i == collateralIndex) liquidatedCollatPrice = price;
             uint256 collateralQuoted = collateralOf[id][borrower][i].mulDivDown(price, ORACLE_PRICE_SCALE);
             maxDebt += collateralQuoted.mulDivDown(_collateral.lltv, WAD);
-            badDebt = badDebt.zeroFloorSub(collateralQuoted.mulDivDown(WAD, _collateral.maxLif));
+            badDebt = badDebt.zeroFloorSub(collateralQuoted.mulDivUp(WAD, _collateral.maxLif));
```
```suggestion
            badDebt = badDebt.zeroFloorSub(collateralOf[id][borrower][i].mulDivUp(price, ORACLE_PRICE_SCALE).mulDivUp(WAD, _collateral.maxLif));
```


collateralQuoted is rounded in the wrong direction, original version was this

src/Midnight.sol
QGarchery: This is another mention of `collateralQuoted` that should be removed.

This comment is wrong, because in case there is no bad debt, then we can have debt < Σ collateralQuoted * 1 / lif. But in the path it's still true (both before and after this PR) that debt >= maxDebt, so we could just rephrase as

```suggestion
                // Note that debt >= maxDebt at this point.
```

test/LiquidationTest.sol
```diff
@@ -292,24 +296,28 @@ contract LiquidationTest is BaseTest {
...
         liquidationOraclePrice = bound(liquidationOraclePrice, 1, badDebtPriceDown());
         collateralize(obligation, borrower, units);
         setupObligation(obligation, units);
         Oracle(obligation.collaterals[0].oracle).setPrice(liquidationOraclePrice);
         uint256 debtAfterBadDebt = units - _badDebt();
+        uint256 lif0 = obligation.collaterals[0].maxLif;
         uint256 maxRepaid = _maxRepaid(units, debtAfterBadDebt, liquidationOraclePrice);
+        uint256 maxSeizedFromRepaid = UtilsLib.min(maxRepaid, debtAfterBadDebt).mulDivDown(lif0, WAD)
+            .mulDivDown(ORACLE_PRICE_SCALE, liquidationOraclePrice);
+        uint256 maxSeized = UtilsLib.min(midnight.collateralOf(id, borrower, 0), maxSeizedFromRepaid);
```
QGarchery:
```suggestion
        uint256 maxSeized = midnight.collateralOf(id, borrower, 0);
```
this should work, and indeed it works if we add `vm.assume(_badDebt() > 0)`, so I think that `badDebtPriceDown` is broken (the comment above this test assumes that there is bad debt)

test/LiquidationTest.sol
```diff
@@ -264,8 +264,9 @@ contract LiquidationTest is BaseTest {
         uint256 debtAfterBadDebt = units - _badDebt();
         uint256 maxRepaid = _maxRepaid(units, debtAfterBadDebt, liquidationOraclePrice);
         uint256 lif0 = obligation.collaterals[0].maxLif;
-        uint256 maxSeized = UtilsLib.min(maxRepaid, debtAfterBadDebt).mulDivDown(lif0, WAD)
+        uint256 maxSeizedFromRepaid = UtilsLib.min(maxRepaid, debtAfterBadDebt).mulDivDown(lif0, WAD)
             .mulDivDown(ORACLE_PRICE_SCALE, liquidationOraclePrice);
+        uint256 maxSeized = UtilsLib.min(midnight.collateralOf(id, borrower, 0), maxSeizedFromRepaid);
```
QGarchery: Same thing to simplify here as well

test/LiquidationTest.sol
peyha, line 667: We could remove `badDebtPriceDown` and `badDebtPriceUp` and keep only `badDebtPrice` cf [this PR](https://github.com/morpho-org/midnight/pull/462)

QGarchery: Then it's not clear that you necessarily get bad debt when you have a price lower than `badDebtPrice`, see [this](https://github.com/morpho-org/midnight/pull/457#discussion_r2890039030)

test/LiquidationTest.sol
```diff
@@ -292,24 +296,28 @@ contract LiquidationTest is BaseTest {
...
         Oracle(obligation.collaterals[0].oracle).setPrice(liquidationOraclePrice);
         uint256 debtAfterBadDebt = units - _badDebt();
+        uint256 lif0 = obligation.collaterals[0].maxLif;
         uint256 maxRepaid = _maxRepaid(units, debtAfterBadDebt, liquidationOraclePrice);
+        uint256 maxSeizedFromRepaid = UtilsLib.min(maxRepaid, debtAfterBadDebt).mulDivDown(lif0, WAD)
+            .mulDivDown(ORACLE_PRICE_SCALE, liquidationOraclePrice);
+        uint256 maxSeized = UtilsLib.min(midnight.collateralOf(id, borrower, 0), maxSeizedFromRepaid);
 
-        midnight.liquidate(obligation, 0, 0, UtilsLib.min(maxRepaid, debtAfterBadDebt), borrower, "");
+        midnight.liquidate(obligation, 0, maxSeized, 0, borrower, "");
```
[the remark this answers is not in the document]
QGarchery: it is expected, now you can't easily repay max but you can easily seize max

test/LiquidationTest.sol
peyha: yes, it only works in one direction. Higher price have no bad debt. Lower prices probably have but not sure what the exact threshold is

src/Midnight.sol
```diff
@@ -456,7 +458,7 @@ contract Midnight is IMidnight {
                 uint256 lltv = obligation.collaterals[collateralIndex].lltv;
                 // Rounded up to avoid consecutive max liquidations.
                 // Acknowledged that the position could be slightly healthy after a liquidation.
-                // Note that debt >= Σ collateralQuoted * 1 / lif >= Σ collateralQuoted * lltv = maxDebt.
+                // Note that debt >= maxDebt at this point.
```
```suggestion
                // Note that debt >= maxDebt through bad debt realization
```

test/LiquidationTest.sol
```diff
@@ -238,11 +238,29 @@ contract LiquidationTest is BaseTest {
         midnight.liquidate(obligation, 0, seized, 0, borrower, "");
     }
 
+    function testBadDebtPriceDownGivesBadDebt(uint256 units) public {
```
peyha, line 241: we could add `testBadDebtPriceDownIsMaximal` to verify that the bound is optimal

peyha (approved):

test/LiquidationTest.sol
MathisGD: but it's not true I think

peyha: pretty sure it is, the following test pass
```
    function testBadDebtPriceDownIsMaximal(uint256 units) public {
        units = bound(units, 10, MAX_UNITS);
        Oracle(obligation.collaterals[0].oracle).setPrice(badDebtPriceDown(units) + 1);

        assertEq(_badDebt(), 0, "should have no bad debt at badDebtPriceDown");
    }
```
it also means that badDebtPriceUp could just be badDebtPriceDown + 1

MathisGD: I'm not sure to understand why this is true, but cool ^^

src/Midnight.sol
```diff
@@ -456,7 +458,7 @@ contract Midnight is IMidnight {
                 uint256 lltv = obligation.collaterals[collateralIndex].lltv;
                 // Rounded up to avoid consecutive max liquidations.
                 // Acknowledged that the position could be slightly healthy after a liquidation.
-                // Note that debt >= Σ collateralQuoted * 1 / lif >= Σ collateralQuoted * lltv = maxDebt.
+                // Note that debt >= maxDebt after the bad debt realization.
```
```suggestion
                // Note that debt >= maxDebt in this branch.
```
or add all the reasons:

```suggestion
                // Note that debt >= maxDebt after the bad debt realization and when the liquidation happens before maturity.
```
because after maturity you can have an arbitrarily low non-zero debt

QGarchery (approved):

PR #461, opened 2026-03-05
[Certora] Consume

bhargavbh:
- [x] only `setCconsumed` and `take` can modify consumed
- [x] calling `setConsumed` only affects onBehalf's consumed
- [x] calling `take` only affects maker's consumed
- [x] consumed is non-decreasing
- [x] after a successful take, consumed does not exceed offer's max amount
- [x] after a successful take, consume is changed exactly as expected
- [x] if the maker's consumed is already >= offer's max amount, then it remains unchanged
- [x] a fully consumed offer always reverts when take input is non-zero in the offer's consumption (thanks to @lilCertora)

### DIFF
--- a/certora/specs/Consume.spec
+++ b/certora/specs/Consume.spec
@@ -13,7 +13,6 @@ methods {
 
 ///  Only `setConsumed` and `take` can modify the consumed mapping.
 rule onlySetConsumedAndTakeChangeConsumed(env e, method f, calldataarg args, address user, bytes32 group) filtered { f -> f.selector != sig:setConsumed(bytes32, uint256, address).selector && f.selector != sig:take(uint256, address, address, bytes, address, Midnight.Offer, Midnight.Signature, bytes32, bytes32[]).selector } {
-
     uint256 consumedBefore = consumed(user, group);
 
     f(e, args);
--- a/certora/specs/Consume.spec
+++ b/certora/specs/Consume.spec
@@ -6,12 +6,11 @@ methods {
     function consumed(address user, bytes32 group) external returns (uint256) envfree;
     function totalUnits(bytes32 id) external returns (uint256) envfree;
     function totalShares(bytes32 id) external returns (uint256) envfree;
-    function toId(Midnight.Obligation obligation) external returns (bytes32);
 
     function _.price() external => NONDET;
 }
 
-///  Only `setConsumed` and `take` can modify the consumed mapping.
+///  Only `setConsumed` and `take` can modify the `consumed` mapping.
 rule onlySetConsumedAndTakeChangeConsumed(env e, method f, calldataarg args, address user, bytes32 group) filtered { f -> f.selector != sig:setConsumed(bytes32, uint256, address).selector && f.selector != sig:take(uint256, address, address, bytes, address, Midnight.Offer, Midnight.Signature, bytes32, bytes32[]).selector } {
     uint256 consumedBefore = consumed(user, group);
 
@@ -20,7 +19,7 @@ rule onlySetConsumedAndTakeChangeConsumed(env e, method f, calldataarg args, add
     assert consumed(user, group) == consumedBefore;
 }
 
-/// Calling `setConsumed` only affects msg.sender's consumed value for the given group.
+/// Calling `setConsumed` only affects onBehalf's consumed value for the given group.
 /// No other (user, group) pair is modified.
 rule setConsumedOnlyAffectsOnBehalf(env e, bytes32 group, uint256 amount, address onBehalf, address otherUser, bytes32 otherGroup) {
     uint256 otherConsumedBefore = consumed(otherUser, otherGroup);
@@ -74,7 +73,7 @@ rule takeConsumedAtMaxUnchanged(env e, uint256 obligationShares, address taker, 
 }
 
 /// A fully-consumed offer always reverts when the take input is non-zero in the offer's consumption dimension.
-rule fullyConsumedOfferRevertsOnNonTrivialTake(env e, uint256 obligationShares, address taker, address takerCallback, bytes takerCallbackData, address receiverIfTakerIsSeller, Midnight.Offer offer, Midnight.Signature signature, bytes32 root, bytes32[] proof) {
+rule fullyConsumedOfferRevertsOnNonTrivialTake(env e, uint256 obligationShares, address taker, address takerCallback, bytes takerCallbackData, address receiver, Midnight.Offer offer, Midnight.Signature signature, bytes32 root, bytes32[] proof) {
     uint256 consumedBefore = consumed(offer.maker, offer.group);
 
     bytes32 id = toId(e, offer.obligation);
@@ -86,7 +85,7 @@ rule fullyConsumedOfferRevertsOnNonTrivialTake(env e, uint256 obligationShares, 
     // When consumption is units-based, prevent rounding down to 0
     require offer.obligationUnits > 0 => to_mathint(obligationShares) * (to_mathint(_totalUnits) + 1) >= to_mathint(_totalShares) + 1;
 
-    take@withrevert(e, obligationShares, taker, takerCallback, takerCallbackData, receiverIfTakerIsSeller, offer, signature, root, proof);
+    take@withrevert(e, obligationShares, taker, takerCallback, takerCallbackData, receiver, offer, signature, root, proof);
 
     assert lastReverted;
 }
--- a/certora/specs/Consume.spec
+++ b/certora/specs/Consume.spec
@@ -4,6 +4,9 @@ methods {
     function multicall(bytes[]) external => HAVOC_ALL DELETE;
 
     function consumed(address user, bytes32 group) external returns (uint256) envfree;
+    function totalUnits(bytes32 id) external returns (uint256) envfree;
+    function totalShares(bytes32 id) external returns (uint256) envfree;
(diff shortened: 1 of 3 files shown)


certora/specs/Consume.spec
```diff
@@ -0,0 +1,60 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+methods {
+    function multicall(bytes[]) external => HAVOC_ALL DELETE;
+
+    function consumed(address user, bytes32 group) external returns (uint256) envfree;
+
+    // assume price does not revert and does not modify storage.
```
QGarchery:
```suggestion
```
`price` is view so it does not modify storage, and we already implicitly only consider non-reverting paths (except in liveness rules, with the `@withrevert` suffix)

certora/specs/Consume.spec
```diff
@@ -0,0 +1,91 @@
...
+    assert consumed(offer.maker, offer.group) <= maxAmount;
+}
+
+/// If consumed[offer.maker][offer.group] is already at or above the offer's max amount before a `take`,
+/// it remains unchanged.
+rule takeConsumedAtMaxUnchanged(env e, uint256 obligationShares, address taker, address takerCallback, bytes takerCallbackData, address receiver, Midnight.Offer offer, Midnight.Signature signature, bytes32 root, bytes32[] proof) {
+    uint256 consumedBefore = consumed(offer.maker, offer.group);
+    uint256 maxAmount = offer.obligationUnits > 0 ? offer.obligationUnits : offer.obligationShares;
+
+    require consumedBefore >= maxAmount;
```
QGarchery: This is missing a reason string. We want to add it just to avoid the warning, since this is just an assumption of the rule. Other option, that would make sense here, is to remove this require, and add that assumption in the final `assert`:
```
    assert consumedBefore >= maxAmount => consumed(offer.maker, offer.group) == consumedBefore;
```

certora/specs/Consume.spec
```diff
@@ -0,0 +1,91 @@
...
+
+/// A fully-consumed offer always reverts when the take input is non-zero in the offer's consumption dimension.
+rule fullyConsumedOfferRevertsOnNonTrivialTake(env e, uint256 obligationShares, address taker, address takerCallback, bytes takerCallbackData, address receiver, Midnight.Offer offer, Midnight.Signature signature, bytes32 root, bytes32[] proof) {
+    uint256 consumedBefore = consumed(offer.maker, offer.group);
+
+    bytes32 id = toId(e, offer.obligation);
+    uint256 _totalUnits = totalUnits(id);
+    uint256 _totalShares = totalShares(id);
+
+    require (offer.obligationUnits > 0 && consumedBefore >= offer.obligationUnits && obligationShares > 0) || (offer.obligationShares > 0 && consumedBefore >= offer.obligationShares && obligationShares > 0);
```
QGarchery: We are requiring `offer.obligationShares > 0` in both paths. Since this is an assumption about the fact that the "input is non-zero in the offer's consumption dimension", it would make more sense to extract it out of this `require`

QGarchery:
```suggestion
    require (offer.obligationUnits > 0 && consumedBefore >= offer.obligationUnits && obligationShares > 0) || (offer.obligationShares > 0 && consumedBefore >= offer.obligationShares && obligationShares > 0), "assume that the offer is fully consumed";
```
this reason would make even more sense once the above is resolved

certora/specs/Consume.spec
```diff
@@ -0,0 +1,91 @@
...
+
+    // When consumption is units-based, prevent rounding down to 0
+    require offer.obligationUnits > 0 => to_mathint(obligationShares) * (to_mathint(_totalUnits) + 1) >= to_mathint(_totalShares) + 1;
```
QGarchery: let's add a reason string here too

certora/specs/Consume.spec
```diff
@@ -0,0 +1,91 @@
...
+    assert consumed(offer.maker, offer.group) == consumedBefore;
+}
+
+/// A fully-consumed offer always reverts when the take input is non-zero in the offer's consumption dimension.
+rule fullyConsumedOfferRevertsOnNonTrivialTake(env e, uint256 obligationShares, address taker, address takerCallback, bytes takerCallbackData, address receiver, Midnight.Offer offer, Midnight.Signature signature, bytes32 root, bytes32[] proof) {
+    uint256 consumedBefore = consumed(offer.maker, offer.group);
+
+    bytes32 id = toId(e, offer.obligation);
+    uint256 _totalUnits = totalUnits(id);
+    uint256 _totalShares = totalShares(id);
```
QGarchery: We can inline those calls

lilCertora (approved):

certora/specs/Consume.spec
```diff
@@ -0,0 +1,99 @@
...
+
+    assert consumed(user, group) >= consumedBefore;
+}
+
+/// After a successful `take`, consumed[offer.maker][offer.group] does not exceed the offer's max amount
+/// (offer.obligationUnits if units-based, offer.obligationShares if shares-based).
+rule takeConsumedBoundedByMax(env e, uint256 obligationShares, address taker, address takerCallback, bytes takerCallbackData, address receiver, Midnight.Offer offer, Midnight.Signature signature, bytes32 root, bytes32[] proof) {
+    uint256 consumedBefore = consumed(offer.maker, offer.group);
+
+    uint256 returnedUnits;
```
QGarchery:
```suggestion
```

QGarchery (approved):

certora/specs/Consume.spec
```diff
@@ -0,0 +1,96 @@
...
+    assert consumed(user, group) >= consumedBefore;
+}
+
+/// After a successful `take`, consumed[offer.maker][offer.group] does not exceed the offer's max amount
+/// (offer.obligationUnits if units-based, offer.obligationShares if shares-based).
+rule takeConsumedBoundedByMax(env e, uint256 obligationShares, address taker, address takerCallback, bytes takerCallbackData, address receiver, Midnight.Offer offer, Midnight.Signature signature, bytes32 root, bytes32[] proof) {
+    take(e, obligationShares, taker, takerCallback, takerCallbackData, receiver, offer, signature, root, proof);
+
+    uint256 maxAmount = offer.obligationUnits > 0 ? offer.obligationUnits : offer.obligationShares;
+    assert consumed(offer.maker, offer.group) <= maxAmount;
```
MathisGD, line 58: if you no-op take, this could not be true right?

certora/specs/Consume.spec
MathisGD, line 1: why didn't we add a proof that delta <= maxAmount - consumed? maybe it's implied by the current properties but it's not super direct. while it's the most interesting proof

certora/specs/Consume.spec
QGarchery: a no-op take on an offer that has < consumed actually reverts !

certora/specs/Consume.spec
QGarchery: yes it's exactly the combination of `takeConsumedBoundedByMax` and `takeConsumedDelta`. To me it's not clear which proof is the most interesting, because if one of the 2 properties fail then the whole mechanism makes no sense. For example you could have `delta <= maxAmount - consumed` but `consumed` is badly updated and so future checks on `consumed` don't have the intended meaning

The way it is currently written is nice because in one of the 2 rules you don't need do compare with `consumedBefore`

certora/specs/Consume.spec
MathisGD: indeed!

bhargavbh: correct, the no-op take on an offer would revert [here](https://github.com/morpho-org/morpho-v2/blob/c91d0dd88ee95a2e7cd9437856bc4460c6621c12/src/Midnight.sol#L228) if consumed[offer.maker][offer.group] > offer.obligationShares. But isn't this intended behaviour?

certora/specs/Consume.spec
bhargavbh: we can show that the two are equivalent for the current implementation atleast: [prover link](https://prover.certora.com/output/1438622/f7b4fe79627b409ea06fe6e3a81aff52?anonymousKey=31b12d2ffc3d72c0c89c23360dc9834f45afa4bb&params=%7B%7D&generalState=%7B%22fileViewOpen%22%3Atrue%2C%22fileViewCollapsed%22%3Atrue%2C%22mainTreeViewCollapsed%22%3Atrue%2C%22callTraceClosed%22%3Atrue%2C%22mainSideNavItem%22%3A%22rules%22%2C%22globalResSelected%22%3Afalse%2C%22isSideBarCollapsed%22%3Afalse%2C%22isRightSideBarCollapsed%22%3Atrue%2C%22selectedFile%22%3A%7B%22uiID%22%3A%225e5dae%22%2C%22output%22%3A%22.certora_sources%2Fcertora%2Fspecs%2FConsume.spec%22%2C%22name%22%3A%22Consume.spec%22%7D%2C%22fileViewFilter%22%3A%22%22%2C%22mainTreeViewFilter%22%3A%22%22%2C%22contractsFilter%22%3A%22%22%2C%22globalCallResolutionFilter%22%3A%22%22%2C%22currentRuleUiId%22%3Anull%2C%22counterExamplePos%22%3A1%2C%22expandedKeysState%22%3A%22%22%2C%22expandedFilesState%22%3A%5B%5D%2C%22outlinedfilterShared%22%3A%22000000000%22%7D). See the rule `takeConsumed`

PR #462, opened 2026-03-05
test: single bad debt price

peyha:


peyha:
can be more precise on bad debt price

PR #463, opened 2026-03-05
continuous fee (borrower side)

MathisGD:


src/Midnight.sol
```diff
@@ -498,13 +541,40 @@ contract Midnight is IMidnight {
...
             IdLib.storeInCode(obligation);
 
             emit EventsLib.ObligationCreated(id, obligation);
         }
         return id;
     }
 
+    function accrueFees(bytes32 id, address user) public {
+        ObligationState storage _obligationState = obligationState[id];
+        uint256 elapsed = block.timestamp - userLastUpdate[id][user];
```
peyha, line 556:
```suggestion
        uint256 elapsed = UtilsLib.min(block.timestamp, obligation.maturity) - userLastUpdate[id][user];
```

src/Midnight.sol
```diff
@@ -361,6 +405,8 @@ contract Midnight is IMidnight {
     ) external returns (uint256, uint256) {
         require(UtilsLib.atMostOneNonZero(repaidUnits, seizedAssets), "inconsistent input");
         bytes32 id = touchObligation(obligation);
+        accrueFees(id, borrower);
+        slash(id, borrower);
```
MathisGD, line 409: the slash here is useless

src/Midnight.sol
```diff
@@ -498,13 +541,40 @@ contract Midnight is IMidnight {
...
+            // forge-lint: disable-next-line(unsafe-typecast)
+            userLastUpdate[id][user] = uint56(block.timestamp);
+
+            uint256 fee = _obligationState.continuousFee;
+            if (fee > 0) {
+                int256 balance = balanceOf[id][user];
+                if (balance < 0) {
+                    uint256 feeFraction = elapsed * fee;
+                    // forge-lint: disable-next-line(unsafe-typecast)
+                    uint256 feeAmount = uint256(-balance).mulDivUp(feeFraction, WAD);
```
adhusson, line 567: defining feeAmount as a uint128 means fewer `UtilsLib.toUint128` later

src/Midnight.sol
```diff
@@ -498,13 +541,40 @@ contract Midnight is IMidnight {
...
         }
         return id;
     }
 
+    function accrueFees(bytes32 id, address user) public {
+        ObligationState storage _obligationState = obligationState[id];
+        uint256 elapsed = block.timestamp - userLastUpdate[id][user];
+        if (elapsed > 0) {
+            // forge-lint: disable-next-line(unsafe-typecast)
+            userLastUpdate[id][user] = uint56(block.timestamp);
```
adhusson, line 559: this can go into borrower state

src/Midnight.sol
```diff
@@ -59,11 +60,20 @@ contract Midnight is IMidnight {
...
 
+    /// @dev Default continuous fee per loan token. Set when the obligation is created. Can be later decreased by the
+    /// feeSetter.
+    mapping(address loanToken => uint64) public defaultContinuousFees;
+
+    /// @dev Trading fee recipient.
     address public tradingFeeRecipient;
 
+    /// @dev Continuous fee recipient.
+    address public continuousFeeRecipient;
```
adhusson, line 75: is it worth having 2 separate recipients

MathisGD: No need, I agree. Especially as they are not claimed in the same way, it's still easy to separate them

PR #465, opened 2026-03-05
feat: check post-callback execution state

peyha:
Checking that the computed amounts have not been modified through the callback by checking the consistency of the obligation totalUnits, totalShares and trading fee.

hopefully solving #496


src/Midnight.sol
```diff
@@ -253,6 +254,13 @@ contract Midnight is IMidnight {
                     obligationShares,
                     buyerCallbackData
                 );
+
+            uint256 newTradingFee = tradingFee(id, timeToMaturity);
+            require(
+                _totalUnits == _obligationState.totalUnits && _totalShares == _obligationState.totalShares
+                    && _tradingFee == newTradingFee,
+                "inconsistent obligation state"
```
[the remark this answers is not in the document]
peyha: we don't care about reentrant tokens tbh but good catch

src/Midnight.sol
```diff
@@ -271,6 +272,29 @@ contract Midnight is IMidnight {
             require(newConsumed <= offer.maxUnits, "consumed");
         }
 
+        if (buyerCallback != address(0)) {
+            ICallbacks(buyerCallback)
+                .onBuy(id, offer.obligation, buyer, buyerAssets, sellerAssets, units, buyerCallbackData);
+
+            uint256 currentTradingFee = tradingFee(id, timeToMaturity);
```
adhusson, line 279: why check that trading fee has not changed?

src/Midnight.sol
```diff
@@ -271,6 +272,29 @@ contract Midnight is IMidnight {
...
+            uint256 currentLossIndex = _obligationState.lossIndex;
```
adhusson, line 280: why not take `currentLossIndex` before both callbacks?

src/Midnight.sol
```diff
@@ -271,6 +272,29 @@ contract Midnight is IMidnight {
             require(newConsumed <= offer.maxUnits, "consumed");
         }
 
+        if (buyerCallback != address(0)) {
```
adhusson, line 275: I'm not clear on
* Do we want callbacks to happen in the pre-trade state?
* Was it unsafe for callbacks to happen in the post-trade state?

src/Midnight.sol
```diff
@@ -271,6 +272,29 @@ contract Midnight is IMidnight {
...
+        SafeTransferLib.safeTransferFrom(offer.obligation.loanToken, buyer, receiver, sellerAssets);
+
+        if (sellerCallback != address(0)) {
+            ICallbacks(sellerCallback)
+                .onSell(id, offer.obligation, seller, buyerAssets, sellerAssets, units, sellerCallbackData);
+
+            uint256 currentTradingFee = tradingFee(id, timeToMaturity);
+            uint256 currentLossIndex = _obligationState.lossIndex;
+
+            require(currentTradingFee == _tradingFee && currentLossIndex == _lossIndex, "callback changed the state");
```
MathisGD, line 295: we could do only one at the end, no?

src/Midnight.sol
MathisGD: trading fee is used to compute assets, so it's much easier to check that we are safe if we know that it didn't change

src/Midnight.sol
MathisGD: you mean after?

src/Midnight.sol
MathisGD, line 279: too bad to redo the computation

src/Midnight.sol
```diff
@@ -254,6 +254,7 @@ contract Midnight is IMidnight {
         uint256 offerPrice = TickLib.tickToPrice(offer.tick);
         uint256 timeToMaturity = UtilsLib.zeroFloorSub(offer.obligation.maturity, block.timestamp);
         uint256 _tradingFee = tradingFee(id, timeToMaturity);
+        uint256 _lossIndex = _obligationState.lossIndex;
```
MathisGD, line 257: why is it here?

src/Midnight.sol
MathisGD, line 223: we should do that after the callbacks

src/Midnight.sol
adhusson: you start by caching the current one before doing the callbacks

src/Midnight.sol
adhusson: safe vs someone increasing/decreasing the trading fee in a callback & reusing the old one? so indexers know that after a trading fee change event all trades use the new trading fee?

peyha:
Broken as seen [here ](https://morpholabs.slack.com/archives/C08A40KKN9W/p1775126535121419?thread_ts=1775125827.035119&cid=C08A40KKN9W)

PR #466, opened 2026-03-05
[no shares] sliding window user loss

adhusson:


test/LossTest.sol
```diff
@@ -0,0 +1,265 @@
...
+        vm.warp(obligation.maturity + TIME_TO_MAX_LIF);
+        midnight.liquidate(obligation, 0, 0, 0, _borrower, "");
+        oracle1.setPrice(ORACLE_PRICE_SCALE);
+    }
+
+    function _expectedBalance(bytes32 _id, address user, uint256 balance) internal view returns (int256) {
+        (uint128 userScale, uint32 userShifts) = midnight.loss(_id, user);
+        (,, uint128 oblScale, uint32 oblShifts,) = midnight.obligationState(_id);
+        uint32 shifts = oblShifts - userShifts;
+        return int256(balance.mulDivDown(userScale, oblScale) >> (64 * shifts));
```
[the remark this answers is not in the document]
adhusson: fixed just after

src/Midnight.sol
```diff
@@ -385,8 +385,13 @@ contract Midnight is IMidnight {
             // forge-lint: disable-next-line(unsafe-typecast)
             balanceOf[id][borrower] += int256(badDebt);
             uint256 oldTotalUnits = _obligationState.totalUnits;
-            _obligationState.lossIndex =
-                WAD - (WAD - _obligationState.lossIndex).mulDivDown(oldTotalUnits - badDebt, oldTotalUnits);
+            uint256 rawScale = _obligationState.lossScale.mulDivUp(oldTotalUnits + 1, oldTotalUnits - badDebt + 1);
+            // 64 bit shifts needed to get a 128 bit loss scale
+            uint32 newShifts =
+                rawScale > (uint256(type(uint128).max) << 64) ? 2 : (rawScale > type(uint128).max ? 1 : 0);
```
QGarchery, line 391: Why isn't it this ?
```suggestion
                rawScale > (uint256(type(uint192).max)) ? 2 : (rawScale > type(uint128).max ? 1 : 0);
```

src/Midnight.sol
```diff
@@ -385,8 +385,13 @@ contract Midnight is IMidnight {
...
+            // forge-lint: disable-next-line(unsafe-typecast)
+            _obligationState.lossScale = uint128(((rawScale - 1) >> (64 * newShifts)) + 1);
```
QGarchery, line 393: Why do we subtract 1 ? At least if there is no shift ever it seems like it changes the expected computation
```suggestion
            _obligationState.lossScale = uint128((rawScale >> (64 * newShifts)) + 1);
```

peyha: If we remove the subscraft 1 we can also remove the add 1

src/Midnight.sol
adhusson: Because the goal is to end up `<= type(uint128).max` and later we round up the scale (to avoid creating unclaimable units). So for instance `(type(uint192).max - 1 >> 64) + 1 > type(uint128).max` (it's 1 more).

src/Midnight.sol
adhusson: It's a ceil (`((a-1) >> N) + 1 = ceil(a/2^N)`). If there is no shift it's `rawScale - 1 + 1 = rawScale`. And rawScale can't be zero so no underflow.

src/Midnight.sol
adhusson: btw there is a variation of this where we get the exact amount to shift so it gets `<= uint128.max` using `clz` but probably not useful

PR #468, opened 2026-03-06
Consume overflow fix with substraction

peyha:


adhusson (approved):
yes although maybe the full arithmetic [fix](https://morpholabs.slack.com/files/U036QFAHCTF/F0AJQJUQTGS/error.diff) of @QGarchery is better, not sure.

QGarchery (changes requested):
I think it is too complicated for the gains

MathisGD:
https://morpholabs.slack.com/archives/C08A40KKN9W/p1773066424455479?thread_ts=1772650459.620499&cid=C08A40KKN9W

PR #470, opened 2026-03-06
[continuous fee] per borrower, remaining

adhusson:
continuous fee implementation
- per borrower
- predictable face value
- no fee after maturity

specs done
- [x] debt are never read before accrual
- [x] lastAccrual = 0 => pending=0

rest of verif wish list moved to https://github.com/morpho-org/midnight/issues/109

good summary of the tradeoffs [here](https://morpholabs.slack.com/archives/C08A40KKN9W/p1773931754654349)


src/Midnight.sol
```diff
@@ -256,22 +256,22 @@ contract Midnight is IMidnight {
             sharesOf[id][buyer] += obligationShares;
         } else {
             // Borrower exits.
-            accrueContinuousFee(id, buyer);
-            borrowerState[id][buyer].debt -= UtilsLib.toUint128(obligationUnits);
+            accrueContinuousFee(id, buyer, offer.obligation.maturity);
+            BorrowerState storage _state = borrowerState[id][buyer];
+            _state.remainingContinuousFee -= uint128(
+                _state.remainingContinuousFee.mulDivDown(obligationUnits, _state.debt)
+            );
```
MathisGD: isn't it supposed to be remainingContinuousFee -= obligationUnit * fee

src/Midnight.sol
```diff
@@ -522,6 +528,12 @@ contract Midnight is IMidnight {
             _state.debt -= UtilsLib.toUint128(repaidUnits);
         }
 
+        if (originalDebt > 0) {
+            _state.remainingContinuousFee -= uint128(
+                _state.remainingContinuousFee.mulDivDown(badDebt + repaidUnits, originalDebt)
+            );
+        }
```
MathisGD: alternatively, there could be one update in the bad debt block and one in `if (repaidUnits > 0 || seizedAssets > 0) {`

src/Midnight.sol
MathisGD, line 256: I would prefer to keep the same 4 branch structure as on main

src/Midnight.sol
adhusson: The current fee should not matter. It's a borrower buying back debt so he should keep his average fee per debt. 

For example imagine ttm is `1 year`, the borrower has 1000 debt taken at `1%/yr` and 1000 debt taken at `2%/yr` so now his remaining fee is `30` (avg 1.5%/yr). He buys back 50 of his debt. His remaining fee should go down proportionally to 22.5 so it's still 1.5%/yr.

src/Midnight.sol
adhusson: agree it mirrors the debt updates so it should be clearer but then in-context the 2nd update is like this:

```
            if (originalDebt > 0) {
                _state.remainingContinuousFee -= uint128(
                    _state.remainingContinuousFee.mulDivDown(repaidUnits, _state.debt)
                );
            }
            _state.debt -= UtilsLib.toUint128(repaidUnits);
```

with denominator `_state.debt = originalDebt - badDebt`, and the order matters since then it gets updated again. So slightly harder to think about, wdyt?

src/Midnight.sol
adhusson: <details>
<summary>to compare it's version A (i removed all guards related to not touching state for noops):</summary>

```solidity
if (buyerIsLender) {
    // Lender enters.
    sharesOf[id][buyer] += obligationShares;
} else {
    // Borrower exits.
    accrueContinuousFee(id, buyer, offer.obligation.maturity);
    BorrowerState storage _state = borrowerState[id][buyer];
    _state.remainingContinuousFee -= uint128(
        _state.remainingContinuousFee.mulDivDown(obligationUnits, _state.debt)
    );
    _state.debt -= UtilsLib.toUint128(obligationUnits);
}

if (sellerIsBorrower) {
    // Borrower enters.
    accrueContinuousFee(id, seller, offer.obligation.maturity);
    borrowerState[id][seller].remainingContinuousFee += uint128(
        _obligationState.continuousFee.mulDivDown(obligationUnits * timeToMaturity, WAD)
    );
    borrowerState[id][seller].debt += UtilsLib.toUint128(obligationUnits);
} else {
    // Lender exits.
    sharesOf[id][seller] -= obligationShares;
}

if (buyerIsLender && sellerIsBorrower) {
    _obligationState.totalShares += UtilsLib.toUint128(obligationShares);
    _obligationState.totalUnits += UtilsLib.toUint128(obligationUnits);
} else if (!buyerIsLender && !sellerIsBorrower) {
    _obligationState.totalShares -= UtilsLib.toUint128(obligationShares);
    _obligationState.totalUnits -= UtilsLib.toUint128(obligationUnits);
}
```

</details>



<details>
<summary>vs version B:
</summary>

```solidity
accrueContinuousFee(id, buyer, offer.obligation.maturity);
accrueContinuousFee(id, seller, offer.obligation.maturity);

if (buyerIsLender && sellerIsBorrower) {
    sharesOf[id][buyer] += obligationShares;
    borrowerState[id][seller].debt += UtilsLib.toUint128(obligationUnits);
    _obligationState.totalShares += UtilsLib.toUint128(obligationShares);
    _obligationState.totalUnits += UtilsLib.toUint128(obligationUnits);
} else if (buyerIsLender && !sellerIsBorrower) {
    sharesOf[id][buyer] += obligationShares;
    sharesOf[id][seller] -= obligationShares;
} else if (!buyerIsLender && sellerIsBorrower) {
    borrowerState[id][buyer].debt -= UtilsLib.toUint128(obligationUnits);
    borrowerState[id][seller].debt += UtilsLib.toUint128(obligationUnits);
} else {
    borrowerState[id][buyer].debt -= UtilsLib.toUint128(obligationUnits);
    sharesOf[id][seller] -= obligationShares;
    _obligationState.totalShares -= UtilsLib.toUint128(obligationShares);
    _obligationState.totalUnits -= UtilsLib.toUint128(obligationUnits);
}

if (!buyerIsLender) {
    BorrowerState storage _buyerState = borrowerState[id][buyer];
    _buyerState.remainingContinuousFee -= uint128(
        _buyerState.remainingContinuousFee.mulDivDown(obligationUnits, _buyerState.debt + obligationUnits)
    );
}

if (sellerIsBorrower) {
    borrowerState[id][seller].remainingContinuousFee += uint128(
        _obligationState.continuousFee.mulDivDown(obligationUnits * timeToMaturity, WAD)
    );
}
```
</details>



<details>
<summary>
btw we could also do a version B' with the 2 final ifs removed:
</summary>

```diff
         accrueContinuousFee(id, buyer, offer.obligation.maturity);
         accrueContinuousFee(id, seller, offer.obligation.maturity);

         if (buyerIsLender && sellerIsBorrower) {
             sharesOf[id][buyer] += obligationShares;
+            borrowerState[id][seller].remainingContinuousFee += uint128(
+                _obligationState.continuousFee.mulDivDown(obligationUnits * timeToMaturity, WAD)
+            );
             borrowerState[id][seller].debt += UtilsLib.toUint128(obligationUnits);
             _obligationState.totalShares += UtilsLib.toUint128(obligationShares);
             _obligationState.totalUnits += UtilsLib.toUint128(obligationUnits);
         } else if (buyerIsLender && !sellerIsBorrower) {
             sharesOf[id][buyer] += obligationShares;
             sharesOf[id][seller] -= obligationShares;
         } else if (!buyerIsLender && sellerIsBorrower) {
-            borrowerState[id][buyer].debt -= UtilsLib.toUint128(obligationUnits);
+            BorrowerState storage _buyerState = borrowerState[id][buyer];
+            _buyerState.remainingContinuousFee -= uint128(
+                _buyerState.remainingContinuousFee.mulDivDown(obligationUnits, _buyerState.debt)
+            );
+            _buyerState.debt -= UtilsLib.toUint128(obligationUnits);
+            borrowerState[id][seller].remainingContinuousFee += uint128(
+                _obligationState.continuousFee.mulDivDown(obligationUnits * timeToMaturity, WAD)
+            );
             borrowerState[id][seller].debt += UtilsLib.toUint128(obligationUnits);
         } else {
-            borrowerState[id][buyer].debt -= UtilsLib.toUint128(obligationUnits);
+            BorrowerState storage _buyerState = borrowerState[id][buyer];
+            _buyerState.remainingContinuousFee -= uint128(
+                _buyerState.remainingContinuousFee.mulDivDown(obligationUnits, _buyerState.debt)
+            );
+            _buyerState.debt -= UtilsLib.toUint128(obligationUnits);
             sharesOf[id][seller] -= obligationShares;
             _obligationState.totalShares -= UtilsLib.toUint128(obligationShares);
             _obligationState.totalUnits -= UtilsLib.toUint128(obligationUnits);
         }
-
-        if (!buyerIsLender) {
-            BorrowerState storage _buyerState = borrowerState[id][buyer];
-            _buyerState.remainingContinuousFee -= uint128(
-                _buyerState.remainingContinuousFee.mulDivDown(obligationUnits, _buyerState.debt + obligationUnits)
-            );
-        }
-
-        if (sellerIsBorrower) {
-            borrowerState[id][seller].remainingContinuousFee += uint128(
-                _obligationState.continuousFee.mulDivDown(obligationUnits * timeToMaturity, WAD)
-            );
-        }
```
</details>

<details>
<summary>
and a B'' with the 2 additional ifs earlier & accrual inside (edited)
</summary>

```diff
-        accrueContinuousFee(id, buyer, offer.obligation.maturity);
-        accrueContinuousFee(id, seller, offer.obligation.maturity);
-
+        if (!buyerIsLender) {
+            accrueContinuousFee(id, buyer, offer.obligation.maturity);
+            BorrowerState storage _buyerState = borrowerState[id][buyer];
+            _buyerState.remainingContinuousFee -= uint128(
+                _buyerState.remainingContinuousFee.mulDivDown(obligationUnits, _buyerState.debt)
+            );
+        }
+
+        if (sellerIsBorrower) {
+            accrueContinuousFee(id, seller, offer.obligation.maturity);
+            borrowerState[id][seller].remainingContinuousFee += uint128(
+                _obligationState.continuousFee.mulDivDown(obligationUnits * timeToMaturity, WAD)
+            );
+        }
+
         if (buyerIsLender && sellerIsBorrower) {
             sharesOf[id][buyer] += obligationShares;
             borrowerState[id][seller].debt += UtilsLib.toUint128(obligationUnits);
             _obligationState.totalShares += UtilsLib.toUint128(obligationShares);
             _obligationState.totalUnits += UtilsLib.toUint128(obligationUnits);
         } else if (buyerIsLender && !sellerIsBorrower) {
             sharesOf[id][buyer] += obligationShares;
             sharesOf[id][seller] -= obligationShares;
         } else if (!buyerIsLender && sellerIsBorrower) {
             borrowerState[id][buyer].debt -= UtilsLib.toUint128(obligationUnits);
             borrowerState[id][seller].debt += UtilsLib.toUint128(obligationUnits);
         } else {
             borrowerState[id][buyer].debt -= UtilsLib.toUint128(obligationUnits);
             sharesOf[id][seller] -= obligationShares;
             _obligationState.totalShares -= UtilsLib.toUint128(obligationShares);
             _obligationState.totalUnits -= UtilsLib.toUint128(obligationUnits);
         }
-
-        if (!buyerIsLender) {
-            BorrowerState storage _buyerState = borrowerState[id][buyer];
-            _buyerState.remainingContinuousFee -= uint128(
-                _buyerState.remainingContinuousFee.mulDivDown(obligationUnits, _buyerState.debt + obligationUnits)
-            );
-        }
-
-        if (sellerIsBorrower) {
-            borrowerState[id][seller].remainingContinuousFee += uint128(
-                _obligationState.continuousFee.mulDivDown(obligationUnits * timeToMaturity, WAD)
-            );
-        }
```
</details>
My preference is B'' > A > B > B', wdyt?

src/Midnight.sol
```diff
@@ -598,6 +651,18 @@ contract Midnight is IMidnight {
...
+        return borrowerState[id][user].remainingContinuousFee;
+    }
+
+    function lastContinuousFeeAccrual(bytes32 id, address user) external view returns (uint48) {
+        return borrowerState[id][user].lastContinuousFeeAccrual;
+    }
+
     /// @dev This function should be called with the id corresponding to the obligation.
     /// @dev This function does not call any oracle if debt is 0.
     function isHealthy(Obligation memory obligation, bytes32 id, address borrower) public view returns (bool) {
```
adhusson: should this always accrue or should we add a comment

src/Midnight.sol
```diff
@@ -628,6 +693,36 @@ contract Midnight is IMidnight {
...
+            if (block.timestamp >= maturity) {
+                feeUnits = remaining;
+            } else {
+                uint256 elapsed = block.timestamp - _state.lastContinuousFeeAccrual;
+                feeUnits = uint128(remaining.mulDivDown(elapsed, maturity - _state.lastContinuousFeeAccrual));
+            }
+
+            if (feeUnits > 0) {
+                ObligationState storage _obligationState = obligationState[id];
+                uint256 feeShares =
```
adhusson: - as long as we have shares the invariant (debt == 0 || shares ==0) can break here
- if shares are gone fee recipient can reduce debt here

src/Midnight.sol
adhusson: i'm doing B'' for now in the interest of speed

MathisGD: why would we factorize these but not the other state updates?

adhusson: - it avoids reading some cold state in accrual
- it's easier to read: first you take care of fee-related updates, then you take care of take-related updates

It's not completely consistent but it happens to be easier to read. The repetition in the 4-way `if` is worth it because you read "what does the take do", but it's not worth it for fee updates.

src/Midnight.sol
```diff
@@ -628,6 +695,36 @@ contract Midnight is IMidnight {
...
+        uint128 remaining = _state.remainingContinuousFee;
+
+        if (remaining > 0 && _state.lastContinuousFeeAccrual > 0) {
+            uint128 feeUnits;
+            if (block.timestamp >= maturity) {
+                feeUnits = remaining;
+            } else {
+                uint256 elapsed = block.timestamp - _state.lastContinuousFeeAccrual;
+                feeUnits = uint128(remaining.mulDivDown(elapsed, maturity - _state.lastContinuousFeeAccrual));
+            }
```
MathisGD: now if elapsed=0 but timestamp==maturity, feeUnit=remaining. would be great to have elapsed=0 => feeUnit=0. I'm fixing this in #476

src/Midnight.sol
adhusson: Problems
* We can't ask that the fee recipient has no debt when it's set because we don't track users' sum of all debts.
* We can't revert if the recipient has some debt it would break liveness.

Other options but they are a little weird:
* We can send fees to an arbitrary address like 0x1.
* We can not accrue fees if the recipient has debt.

MathisGD: I fixed it in #476, if you want

adhusson: you picked 

> We can not accrue fees if the recipient has debt.

right?

src/Midnight.sol
```diff
@@ -228,6 +251,21 @@ contract Midnight is IMidnight {
             require(newConsumed <= offer.obligationShares, "consumed");
         }
 
+        if (!buyerIsLender) {
+            accrueContinuousFee(id, buyer, offer.obligation.maturity);
```
MathisGD: see [this](https://github.com/morpho-org/midnight/pull/476/#discussion_r2904391633), there might be an issue

src/Midnight.sol
adhusson: added a comment for now

src/Midnight.sol
MathisGD: >it avoids reading some cold state in accrual

indeed, but actually [this](https://github.com/morpho-org/midnight/pull/470#discussion_r2904455708) might be a problem

src/Midnight.sol
adhusson: actually if elapsed=0 and timestamp=maturity it means the entire fee was taken at the previous accrual no?

src/Midnight.sol
adhusson: another idea: shares go to passive recipient account, fee recipient can only withdraw assets

see ~https://github.com/morpho-org/midnight/pull/483~ (merged here)

src/Midnight.sol
adhusson: It is very error-prone to wrongly say healthy when in fact the user might be liquidatable. Also it's as if `preview*` funtions in vault v2 did not accrue before doing conversions.

I'm changing to accruing for now to check it doesn't affect other things & we can discuss later.

peyha (commented):
The weird thing with this approach is that if a borrower knows that a trading fee update is coming, he'll fully repay his debt just before the update (to make his remaining fee at 0) and re-open his borrow position after the update to benefit from the lower fee

src/Midnight.sol
```diff
@@ -338,7 +375,15 @@ contract Midnight is IMidnight {
...
         bytes32 id = touchObligation(obligation);
 
-        borrowerState[id][onBehalf].debt -= UtilsLib.toUint128(obligationUnits);
+        accrueContinuousFee(id, onBehalf, obligation.maturity);
+
+        BorrowerState storage _repayState = borrowerState[id][onBehalf];
+        if (_repayState.debt > 0) {
+            _repayState.remainingContinuousFee -= uint128(
+                _repayState.remainingContinuousFee.mulDivDown(obligationUnits, _repayState.debt)
+            );
```
MathisGD: aren't you afraid of the rounding introduced by this? for example now it might be possible to arrive to debt=0 with remaining>0. But it could also hide a bunch of things

src/Midnight.sol
adhusson: i rechecked and indeed if elapsed=0 and last call was accrue it means at the last accrue we had `timestamp == maturity` so we returned `_state.pendingFee` and then either `feeUnits` was 0, or we updated `_sate.pendingFee -= _state.pendingFee` so the next `remaining` is 0.

if last call was a take then `timeToMaturity` was 0 so no fee accrued.

src/Midnight.sol
adhusson: currently fixed by merging https://github.com/morpho-org/midnight/pull/483

src/Midnight.sol
adhusson: currently fixed by merging https://github.com/morpho-org/midnight/pull/483

src/Midnight.sol
```diff
@@ -338,7 +378,13 @@ contract Midnight is IMidnight {
         require(onBehalf == msg.sender || isAuthorized[onBehalf][msg.sender], "unauthorized");
         bytes32 id = touchObligation(obligation);
 
-        borrowerState[id][onBehalf].debt -= UtilsLib.toUint128(obligationUnits);
+        accrueContinuousFee(id, onBehalf, obligation.maturity);
+
+        BorrowerState storage _state = borrowerState[id][onBehalf];
+        if (_state.debt > 0) {
+            _state.pendingFee -= uint128(_state.pendingFee.mulDivUp(obligationUnits, _state.debt));
```
adhusson: would this be simpler (and same for other exits)?

```solidity
_buyerState.pendingFee = uint128(_state.pendingFee.mulDivDown(_state.debt - obligationUnits, _state.debt));
```

src/Midnight.sol
adhusson: you couldn't, in this case units = debt so pendingFee -= pendingFee * debt / debt = 0

src/Midnight.sol
```diff
@@ -228,6 +253,19 @@ contract Midnight is IMidnight {
             require(newConsumed <= offer.obligationShares, "consumed");
         }
 
+        if (!buyerIsLender) {
+            accrueContinuousFee(id, buyer, offer.obligation.maturity);
+            BorrowerState storage _buyerState = borrowerState[id][buyer];
+            _buyerState.pendingFee -= uint128(_buyerState.pendingFee.mulDivUp(obligationUnits, _buyerState.debt));
```
MathisGD: why lint doesn't flag this as an unsafe cast

src/Midnight.sol
MathisGD: is it safe to accrueContinuousFee after querying this? It's not hardcoded that debt=0 couldn't become nonzero in accrual. Or the seller could be the fee recipient

adhusson: yes because it uses a passive fee recipient

src/Midnight.sol
MathisGD: this would never underflow because obligationUnits <= debt?

src/Midnight.sol
MathisGD: >It's not hardcoded that debt=0 couldn't become nonzero in accrual.

and this?

MathisGD: still not fully fully sold. then I would consider rewriting the 4 branches completely

src/Midnight.sol
adhusson: its more if it underflows now it would have underflowed later anyway

src/Midnight.sol
```diff
@@ -228,6 +253,19 @@ contract Midnight is IMidnight {
...
+        }
+
+        if (sellerIsBorrower) {
+            accrueContinuousFee(id, seller, offer.obligation.maturity);
+            borrowerState[id][seller].pendingFee += uint128(
+                _obligationState.continuousFee.mulDivDown(obligationUnits * timeToMaturity, WAD)
+            );
```
MathisGD: (minor) probably better to do a safe cast here

src/Midnight.sol
```diff
@@ -311,7 +347,11 @@ contract Midnight is IMidnight {
         address onBehalf,
         address receiver
     ) external returns (uint256, uint256) {
-        require(onBehalf == msg.sender || isAuthorized[onBehalf][msg.sender], "unauthorized");
+        require(
+            onBehalf == msg.sender || isAuthorized[onBehalf][msg.sender]
+                || (onBehalf == PASSIVE_FEE_RECIPIENT && msg.sender == feeRecipient),
```
MathisGD, line 309: not a huge fan of this. the fee recipient will be read for the tradingFee anyway, so it's not so much of a gas opti. is it only to make sure that the fee recipient cannot take?

certora/specs/Midnight.spec
```diff
@@ -36,10 +40,14 @@ hook Sstore borrowerState[KEY bytes32 id][KEY address owner].debt uint128 newDeb
 
 function summaryMulDiv(uint256 x, uint256 y, uint256 d) returns uint256 {
     if (x == 0 || y == 0) return 0;
+    if (d > 0 && y == d) return x;
+    if (d > 0 && x == d) return y;
```
MathisGD, line 47: curious why is this needed?

MathisGD: oh for noRemainingContinuousFeeWithoutDebt?

src/Midnight.sol
MathisGD: I don't really see the point in not updating accrueContinuousFee for both users at the beginning. It's at most one useless cold SLOAD, half of the time.

src/Midnight.sol
MathisGD: I'm speaking about the -=

adhusson: yes the -= can underflow if obligationUnbits > _buyerState.debt, but if that is the case, later debt -= obligationUnits will underflow too

src/Midnight.sol
```diff
@@ -67,10 +70,15 @@ contract Midnight is IMidnight {
     /// @dev Whether an address is authorized to act on behalf of another address.
     mapping(address authorizer => mapping(address authorized => bool)) public isAuthorized;
 
-    /// @dev Default fees per loan token. Set when the obligation is created. Can be later overridden by the feeSetter.
+    /// @dev Default trading fees per loan token. Set when the obligation is created. Can be later overriden by the
+    /// feeSetter.
     mapping(address loanToken => uint16[7]) public defaultFees;
```
```suggestion
    mapping(address loanToken => uint16[7]) public defaultTradingFees;
```

certora/specs/Midnight.spec
adhusson: yes, at least repay was failing, IIUC because a full repay that brings debt to zero, with `repaid=debt`, must remove exactly pendingFee from pendingFee, so it needs `pendingFee.mulDivDown(repaidWhichIsEqualToDebt,debt) == pendingFee`

src/Midnight.sol
adhusson: Yes, the alternative was to do `isAuthorized[PASSIVE_FEE_RECIPIENT][feeRecipient] = true` (and undo the previous one) when setting a fee recipient, but there is no granularity so then the `PASSIVE_FEE_RECIPIENT` could take too and we would be back to square 1.

We could also check in take that the fee recipient cannot take

src/Midnight.sol
adhusson: good point, no idea

src/Midnight.sol
MathisGD: if you do that then the fee recipient can authorize other people (that you won't undo when removing them)

src/Midnight.sol
MathisGD, line 1: Properties to add before merging:
- [ ] debt at maturity prediction is accurate
- [ ] effective fee is less than the max fee (/max fee on the period)
- [ ] roundings are in the user's favor (the ratio pending/debt decreases)
- [ ] equivalence reverts (continuous fee activation never make you revert in new ways)
- [x] lastAccrual = 0 => pending=0

(feel free to edit)

adhusson: > It's not hardcoded that debt=0 couldn't become nonzero in accrual.

Sorry missed this. I think we could show `debtOfZeroCanOnlyIncreaseViaTake`, which would also mean a public accrual alone cannot make a zero debt nonzero. But it's still subtle, i'll push accrual up.

adhusson: btw i'm making accrueContinuousFee public, but maybe not a great idea?

MathisGD: no it's great I think

peyha: I don't think it changes much you can always accrue the fee with a no-op

src/Midnight.sol
peyha: isn't similar to blue's healthy function which didn't accrue interest ? (though the function was internal while its public here)
Not sure it caused any issues in the past

src/Midnight.sol
MathisGD, line 1: we probably want some events updates, because right now it's pretty difficult to index pending

src/Midnight.sol
MathisGD: why "if originalDebt > 0"?

src/Midnight.sol
MathisGD: to me it's fine to not accrue and warn

src/Midnight.sol
MathisGD, line 1: the takeAmountsLib and/or bundler needs to be updated

src/Midnight.sol
```diff
@@ -632,6 +715,26 @@ contract Midnight is IMidnight {
         return tentativeSigner;
     }
 
+    function accrueContinuousFee(bytes32 id, address borrower, uint256 maturity) public {
```
```suggestion
    function accrueContinuousFee(bytes32 id, address borrower, uint256 maturity) public {
    require(obligationState[id].created, "obligation not created");
```

src/interfaces/IMidnight.sol
```diff
@@ -44,6 +44,8 @@ struct Signature {
 struct BorrowerState {
     uint128 debt;
     uint128 activatedCollaterals;
+    uint128 pendingFee;
+    uint48 lastContinuousFeeAccrual;
```
MathisGD: the second slot is unzeroed for all lenders...

src/Midnight.sol
MathisGD: https://github.com/morpho-org/midnight/pull/505#discussion_r2920151776

we cannot trust the maturity if the function is public

src/Midnight.sol
peyha: I would guess the ir equivalent of the current version requires one more dup but it might be equivalent after optimization but both formula are equivalent so I would go for your proposal

peyha: After a quick check, your proposal saves ~20 gas

src/interfaces/IMidnight.sol
adhusson: i don't understand

test/ContinuousFeeTest.sol
```diff
@@ -0,0 +1,526 @@
...
+        uint256 remaining = midnight.pendingFee(id, borrower);
+
+        vm.warp(block.timestamp + elapsed);
+        uint256 expectedFee = remaining.mulDivDown(elapsed, ttm);
+
+        // Via repay
+        uint256 snap = vm.snapshotState();
+        midnight.repay(obligation, 0, borrower);
+        assertEq(midnight.debtOf(id, borrower), debt + expectedFee, "debt after repay");
+        assertEq(midnight.pendingFee(id, borrower), remaining - expectedFee, "remaining after repay");
```
peyha, line 81: why don't you also check the pending fee after withdraw collateral & take ?

src/Midnight.sol
adhusson: and wdyt in terms of human understanding of the line? i'm not sure

src/Midnight.sol
adhusson: working on it

adhusson: working on it

test/ContinuousFeeTest.sol
```diff
@@ -0,0 +1,575 @@
...
+
+    function testExitViaLiquidation(uint256 debt, uint256 feeRate, uint256 ttm, uint256 elapsed) public {
+        debt = bound(debt, 1e18, MAX_DEBT);
+        feeRate = bound(feeRate, 0, MAX_CONTINUOUS_FEE);
+        ttm = bound(ttm, 10, 360 days);
+        elapsed = bound(elapsed, 1, ttm - 1);
+
+        setupBorrower(debt, feeRate, ttm);
+
+        // Make liquidatabl
```
peyha:
```suggestion
        // Make liquidatable
```

test/ContinuousFeeTest.sol
```diff
@@ -0,0 +1,575 @@
...
+                .mulDivUp(WAD, obligation.collaterals[0].maxLif)
+        );
+        uint256 expectedRemaining = remainingAfterAccrual - remainingAfterAccrual.mulDivUp(badDebt, debtAfterAccrual);
+
+        deal(address(loanToken), address(this), debtAfterAccrual);
+        vm.expectEmit();
+        emit EventsLib.AccrueContinuousFee(id, borrower, feeUnits, feeShares, remainingAfterAccrual);
+        vm.expectEmit();
+        emit EventsLib.SetPendingFee(id, borrower, expectedRemaining);
+        midnight.liquidate(obligation, 0, 0, 0, borrower, "");
```
peyha, line 370: this is a no-op, is it normal ?

src/Midnight.sol
adhusson: accrual effect in all functions: (wip)

- [ ] * withdraw: `debtOf(id, onBehalf) == debtBefore + accruedFeeBefore`
- [ ] * repay: `debtOf(id, onBehalf) + obligationUnits == debtBefore + accruedFeeBefore`
- [ ] * liquidate: `debtOf(id, borrower) <= debtBefore + accruedFeeBefore`

test/ContinuousFeeTest.sol
adhusson: there is some bad debt that gets written off, but yes its not liquidating a >0 amount yet

src/Midnight.sol
adhusson: back to internal so no need?

src/libraries/EventsLib.sol
```diff
@@ -12,7 +12,14 @@ library EventsLib {
...
     event SetDefaultTradingFee(address indexed loanToken, uint256 indexed index, uint256 newTradingFee);
-    event SetTradingFeeRecipient(address indexed feeRecipient);
+    event SetFeeRecipient(address indexed feeRecipient);
+    event SetObligationContinuousFee(bytes32 indexed id_, uint256 newContinuousFee);
+    event SetDefaultContinuousFee(address indexed loanToken, uint256 newContinuousFee);
+    event AccrueContinuousFee(
+        bytes32 indexed id_, address indexed borrower, uint256 accruedFee, uint256 feeShares, uint256 newPendingFee
+    );
+    /// @dev SetPendingFee should always happen on a freshly accrued borrower state.
+    event SetPendingFee(bytes32 indexed id_, address indexed borrower, uint256 pendingFee);
```
```suggestion
    event UpdatePendingFee(bytes32 indexed id_, address indexed borrower, uint256 newPendingFee);
```
not a fan of the current naming because it's not settable (compared to other set events)

src/interfaces/IMidnight.sol
MathisGD: sorry. this slot is opened (20k gas) for lenders while it's useless. your initial implementation didn't have this problem but I found it much harder to verify

src/libraries/EventsLib.sol
peyha: we could also store the pending fee delta (because that's what we do for debt/shares events) but it means that it should be converted to int256

src/Midnight.sol
peyha: the current version looks slightly nicer so not sure

src/Midnight.sol
adhusson: if originalDebt==0 this divides by zero

src/Midnight.sol
adhusson: @peyha every single `_isHealthy` call in blue is preceded by an accrueInterest. 

@MathisGD tbh i have a very hard time seeing why it's not net positive to give a correct answer right away. We would at least need to rename so the number of people who wrongly (and reasonably) assume the function call takes into account the current onchain state is reduced.

src/libraries/EventsLib.sol
adhusson: 👍 on rename

For what to log, final value is easier on indexers and easy to do here.

In vaultV2 where we log deltas the reasons seems to be:
- for `Allocate`, `Deallocate`, `ForceDeallocate`: there would be as many final values to log as there are ids, not worth it, so we log the common delta
- for `Deposit` `Withdraw`, `Transfer`: it follows an EIP
- for `AccrueInterest`: I think logging only new performance and management fee shares is for code simplicity (the accrual returns the new shares).

src/interfaces/IMidnight.sol
adhusson: ok we can only accrue under the ifs like before, or early return in accrue when `debt==0` (since we have the invariant no real need to check pendingFee=0). Or we could include the uint48 in the first slot and reduce the size of `activateCollaterals`, and always update. Maybe cleaner overall.

src/libraries/EventsLib.sol
peyha: yes in the end I think both would work well and the current version is the simplest.  there are now two user state variables that cannot be re-computed by adding event data: `consumed` and `pendingFee` but it's fine because it's not the most important state data

adhusson: not sure i get it. I would say soemthing like: they can be computed by reading the last event, so it's strictly simpler than needing all the past events, wdyt?

peyha: The simplest way for indexers to process data is to perform sql queries. To compute current assets/shares they usually do `SELECT SUM(assets) OVER (PARTITION BY obligationId, borrower ORDER BY block_number), obligationId, borrower, block_number FROM event_data GROUP BY obligationId, borrower`. Here the indexer will have to implement both SUM and LAST_VALUE operators and LAST_VALUE is slightly less optimized than SUM but it's not a big deal imo because I'm not even sure that indexers will try to index `consumed` and `pendingFee`

src/Midnight.sol
MathisGD: if _state.debt=0, no?

src/periphery/TakeAmountsLib.sol
```diff
@@ -10,56 +10,84 @@ import {WAD} from "../libraries/ConstantsLib.sol";
...
+        returns (uint256 expectedTotalUnits, uint256 expectedTotalShares, bool buyerIsLender)
     {
+        expectedTotalUnits = midnight.totalUnits(id);
+        expectedTotalShares = midnight.totalShares(id);
+
         address buyer = offer.buy ? offer.maker : taker;
-        bool buyerIsLender = midnight.debtOf(id, buyer) == 0;
+
+        uint256 buyerAccruedFee = midnight.pendingContinuousFee(id, buyer, offer.obligation.maturity);
+        expectedTotalShares += buyerAccruedFee.mulDivDown(expectedTotalShares + 1, expectedTotalUnits + 1);
```
MathisGD: uses an outdated share price. will open a PR on top to fix

MathisGD: actually no

src/Midnight.sol
adhusson: will be resolved by making internal in #515

src/Midnight.sol
```diff
@@ -624,6 +711,24 @@ contract Midnight is IMidnight {
         return tentativeSigner;
     }
 
+    function accrueContinuousFee(bytes32 id, address borrower, uint256 maturity) internal {
```
MathisGD: Should revert if the obligation is not created

src/Midnight.sol
```diff
@@ -231,6 +260,19 @@ contract Midnight is IMidnight {
             _obligationState.totalUnits - oldSellerDebt - oldBuyerDebt + sellerPos.debt + buyerPos.debt
         );
 
+        if (!buyerIsLender) {
+            // forge-lint: disable-next-item(unsafe-typecast) as if obligationUnits > debt, an underflow occurs later.
+            buyerPos.pendingFee -= uint128(uint256(buyerPos.pendingFee).mulDivUp(obligationUnits, oldBuyerDebt));
```
```suggestion
            buyerPos.pendingFee -= uint128(uint256(buyerPos.pendingFee).mulDivUp(buyerDebtReduction, oldBuyerDebt));
```

src/Midnight.sol
```diff
@@ -231,6 +260,19 @@ contract Midnight is IMidnight {
...
+            emit EventsLib.UpdatePendingFee(id, buyer, buyerPos.pendingFee);
+        }
+
+        if (sellerIsBorrower) {
+            sellerPos.pendingFee += UtilsLib.toUint128(
+                uint256(_obligationState.continuousFee).mulDivDown(obligationUnits * timeToMaturity, WAD)
+            );
```
```suggestion
        uint sellerDebtIncrease = obligationUnits - UtilsLib.min(sellerPos.credit, obligationUnits);
        if (sellerDebtIncrease > 0) {
            sellerPos.pendingFee += UtilsLib.toUint128(
                uint256(_obligationState.continuousFee).mulDivDown(sellerDebtIncrease * timeToMaturity, WAD)
            );
```

src/Midnight.sol
```diff
@@ -595,11 +659,23 @@ contract Midnight is IMidnight {
...
+    function lastContinuousFeeAccrual(bytes32 id, address user) external view returns (uint48) {
+        return position[id][user].lastContinuousFeeAccrual;
+    }
+
     /// @dev This function should be called with the id corresponding to the obligation.
     /// @dev This function does not call any oracle if debt is 0.
     function isHealthy(Obligation memory obligation, bytes32 id, address borrower) public view returns (bool) {
         Position storage _position = position[id][borrower];
-        uint256 debt = debtOf(id, borrower);
+        uint256 debt = debtOf(id, borrower) + pendingContinuousFee(id, borrower, obligation.maturity);
```
adhusson: elsewhere we use `_position.debt` rather than the function

src/Midnight.sol
adhusson: maybe there is a globally simpler way since we need this var anyway, and don't need `sellerIsBorrower`

src/Midnight.sol
adhusson: yes my bad

certora/specs/OnlyAuthorizedCanChange.spec
```diff
@@ -56,3 +72,62 @@ rule onlyAuthorizedCanChangeCreditAndDebtExceptLiquidateAndSlash(env e, method f
...
+}
+
+rule repayDebtMatchesAccrualAndRepayment(env e, Midnight.Obligation obligation, uint256 obligationUnits, address onBehalf) {
+    bytes32 id = toId(obligation);
+    mathint debtBefore = debtOf(id, onBehalf);
+    mathint accruedFeeBefore = accruedContinuousFeeBefore(id, onBehalf, e.block.timestamp, obligation.maturity);
+
+    repay(e, obligation, obligationUnits, onBehalf);
+    assert debtOf(id, onBehalf) + obligationUnits == debtBefore + accruedFeeBefore;
+}
```
MathisGD:
```suggestion
```

src/interfaces/IMidnight.sol
MathisGD: see #536

certora/specs/BalanceEffects.spec
```diff
@@ -162,7 +196,9 @@ filtered {
         && f.selector != sig:repay(Midnight.Obligation, uint256, address).selector
         && f.selector != sig:liquidate(Midnight.Obligation, uint256, uint256, uint256, address, bytes).selector
         && f.selector != sig:slash(bytes32, address).selector
+        && f.selector != sig:withdrawCollateral(Midnight.Obligation, uint256, uint256, address, address).selector
 } {
+    require noAccrual(e, id, user);
```
MathisGD: not needed

src/interfaces/IMidnight.sol
```diff
@@ -47,13 +47,16 @@ struct ObligationState {
...
     uint16[7] fees;
+    uint64 continuousFee;
 }
 
 struct Position {
     uint128 credit;
     uint128 lossIndex;
     uint128 debt;
     uint128 activatedCollaterals;
+    uint128 pendingFee;
```
peyha, line 58:
```suggestion
uint128 debt;
uint128 pendingFee;
uint128 activatedCollaterals;
```
pendingFee and debt are updated together so it could make sense to pack it together but I don't think solc is efficient enough to remove a sstore

src/Midnight.sol
MathisGD: moved to the description

certora/specs/BalanceEffects.spec
```diff
@@ -49,12 +47,19 @@ function summaryMulDiv(uint256 x, uint256 y, uint256 d) returns uint256 {
     return ghostMulDiv(x, y, d);
 }
 
+// All rules in this file assume no accrual.
+definition noAccrual(env e, bytes32 id, address borrower) returns bool = currentContract.position[id][borrower].pendingFee == 0 || e.block.timestamp == currentContract.position[id][borrower].lastContinuousFeeAccrual;
+
 /// REPAY ///
 
 /// repay decreases onBehalf's debt by exactly obligationUnits and only changes position[id][onBehalf].debt.
+/// When no fee accrual occurs during repay.
```
adhusson:
```suggestion
/// When no fee accrual occurs, repay decreases onBehalf's debt by exactly obligationUnits and only changes position[id][onBehalf].debt
```

certora/specs/BalanceEffects.spec
```diff
@@ -114,9 +125,13 @@ rule takeEffects(env e, uint256 obligationUnits, address taker, address takerCal
 
 /// liquidate decreases the borrower's debt by at least repaidUnits,
 /// and only changes position[id][borrower].debt.
+/// When no fee accrual occurs during liquidate.
```
adhusson:
```suggestion
/// When no fee accrual occurs, liquidate decreases the borrower's debt by at least repaidUnits,
/// and only changes position[id][borrower].debt.
```

certora/specs/BalanceEffects.spec
```diff
@@ -151,9 +166,28 @@ rule slashEffects(env e, bytes32 id, address user, bytes32 anyId, address anyUse
     assert anyUser != user || anyId != id => creditOf(anyId, anyUser) == otherCreditBefore;
 }
 
+/// WITHDRAW COLLATERAL ///
+
+/// withdrawCollateral does not change any user's credit or debt.
+/// When no fee accrual occurs during withdrawCollateral.
```
adhusson:
```suggestion
/// When no fee accrual occurs, withdrawCollateral does not change any user's credit or debt.
```

src/Midnight.sol
```diff
@@ -222,15 +249,28 @@ contract Midnight is IMidnight {
...
 
+        if (buyerDebtReduction > 0) {
+            // forge-lint: disable-next-item(unsafe-typecast) as pendingFee reduction <= pendingFee
+            buyerPos.pendingFee -= uint128(uint256(buyerPos.pendingFee).mulDivUp(buyerDebtReduction, oldBuyerDebt));
+            emit EventsLib.UpdatePendingFee(id, buyer, buyerPos.pendingFee);
+        }
+
+        if (sellerDebtIncrease > 0) {
+            sellerPos.pendingFee += UtilsLib.toUint128(
+                _obligationState.continuousFee.mulDivDown(sellerDebtIncrease * timeToMaturity, WAD)
```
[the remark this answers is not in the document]
adhusson: to deal with later

src/Midnight.sol
```diff
@@ -290,7 +332,15 @@ contract Midnight is IMidnight {
         require(onBehalf == msg.sender || isAuthorized[onBehalf][msg.sender], "unauthorized");
         bytes32 id = touchObligation(obligation);
 
-        position[id][onBehalf].debt -= UtilsLib.toUint128(obligationUnits);
+        accrueContinuousFee(obligation, id, onBehalf);
+
+        Position storage _position = position[id][onBehalf];
+        if (_position.debt > 0) {
+            // forge-lint: disable-next-item(unsafe-typecast) as if obligationUnits > debt, an underflow occurs later.
+            _position.pendingFee -= uint128(uint256(_position.pendingFee).mulDivUp(obligationUnits, _position.debt));
```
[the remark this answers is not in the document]
adhusson: to deal with later

src/interfaces/IMidnight.sol
QGarchery:
```suggestion
uint256 withdrawable;
uint128 totalUnits;
```
it's one less slot, and in a typical interaction everything is touched (except maybe `withdrawable`)

src/Midnight.sol
```suggestion
                sellerDebtIncrease.mulDivDown(_obligationState.continuousFee * timeToMaturity, WAD)
```
same computation but clearer, because the continuous fee is a percentage (in WAD) divided by the time elapsed, which is clear in the definition of `MAX_CONTINUOUS_FEE`

src/libraries/ConstantsLib.sol
```diff
@@ -5,10 +5,12 @@ pragma solidity ^0.8.0;
 uint256 constant WAD = 1e18;
 uint256 constant ORACLE_PRICE_SCALE = 1e36;
 uint256 constant FEE_STEP = 1e12;
+uint256 constant MAX_CONTINUOUS_FEE = uint256(0.01e18) / uint256(365 days);
```
QGarchery: somewhat weird that we have both 360 days and 365 days in the repo, but I guess this one is not so critical so it's ok

src/interfaces/IMidnight.sol
adhusson: Even if the compiler does not remove the sstore, a sstore to a slot that already had its value changed in the same tx only costs 100 gas... but i thought it would help and whne i tried it increased cost a tiny bit for repay. The layout is `debt; activated; pending; lastAccrual` so I think it's because repay has an accrual first, which touches `pending` and `lastAccrual` anyway. So if we do `debt; pending; activated; lastAccrual` now 2 slots are touched.

src/Midnight.sol
```diff
@@ -612,6 +687,17 @@ contract Midnight is IMidnight {
         return maxDebt >= debt;
     }
 
+    function pendingContinuousFee(bytes32 id, address borrower, uint256 maturity) public view returns (uint256) {
```
QGarchery: let's inline this function, as it's used only once (and the naming pending continuous fee is overloaded because of the event)

src/Midnight.sol
```diff
@@ -624,6 +710,34 @@ contract Midnight is IMidnight {
...
+        view
+        returns (uint128)
+    {
+        // forge-lint: disable-next-item(unsafe-typecast) as accrued fee is <= pendingFee
+        return uint128(pendingContinuousFee(id, borrower, obligation.maturity));
+    }
+
+    /// @dev Expects the obligation to be touched.
+    /// @dev Expects the id to correspond to the obligation's id.
+    function accrueContinuousFee(Obligation memory obligation, bytes32 id, address borrower) internal {
```
QGarchery, line 729: Should we have it public ? It's not so different from `slash`, and the benefit is the same: have the possibility to have a fully updated state (no need for virtually accrued variables)

src/interfaces/IMidnight.sol
adhusson: And in take it looks worse because there are 2 accruals, so 2 more slots used with the order switch (but you only gain on 1 debt/pending update).

src/Midnight.sol
adhusson: wdyt
```
            buyerPos.pendingFee -= uint128(uint256(buyerPos.pendingFee).mulDivUp(buyerDebtReduction, oldBuyerDebt));
```       
            ->
            
``` 
            buyerPos.pendingFee -= uint128(buyerDebtReduction.mulDivUp(buyerPos.pendingFee, oldBuyerDebt));
```

by itself not more or less sense but harmonizes with below

src/Midnight.sol
adhusson: but it's public, you mean remove it from external API?

src/Midnight.sol
adhusson: slash is public too, you mean make both internal?

src/Midnight.sol
```diff
@@ -134,14 +142,31 @@ contract Midnight is IMidnight {
...
         emit EventsLib.SetDefaultTradingFee(loanToken, index, newTradingFee);
     }
 
-    function setTradingFeeRecipient(address feeRecipient) external {
+    function setFeeRecipient(address newFeeRecipient) external {
         require(msg.sender == owner, "only owner");
-        tradingFeeRecipient = feeRecipient;
-        emit EventsLib.SetTradingFeeRecipient(feeRecipient);
+        feeRecipient = newFeeRecipient;
+        emit EventsLib.SetFeeRecipient(newFeeRecipient);
```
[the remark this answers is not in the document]
adhusson: it's fine

src/Midnight.sol
```diff
@@ -624,6 +710,34 @@ contract Midnight is IMidnight {
...
+        if (accruedFee > 0) {
+            _position.pendingFee -= accruedFee;
+            _position.debt += accruedFee;
+            _obligationState.totalUnits += accruedFee;
+            slash(id, PASSIVE_FEE_RECIPIENT);
+            position[id][PASSIVE_FEE_RECIPIENT].credit += accruedFee;
+        }
+        _position.lastContinuousFeeAccrual = uint48(block.timestamp);
+
+        emit EventsLib.AccrueContinuousFee(id, borrower, accruedFee, accruedFee, _position.pendingFee);
```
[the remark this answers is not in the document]
adhusson: not worth updating for me but not sure

src/Midnight.sol
```diff
@@ -441,6 +497,12 @@ contract Midnight is IMidnight {
             }
             _obligationState.withdrawable += repaidUnits;
             _position.debt -= UtilsLib.toUint128(repaidUnits);
+            // forge-lint: disable-next-item(unsafe-typecast) as repaidUnits <= originalDebt - badDebt
+            _position.pendingFee -= uint128(uint256(_position.pendingFee).mulDivUp(repaidUnits, originalDebt - badDebt));
```
```suggestion
            // forge-lint: disable-next-item(unsafe-typecast)
            _position.pendingFee -= uint128(uint256(_position.pendingFee).mulDivUp(repaidUnits, _position.debt));
_position.debt -= UtilsLib.toUint128(repaidUnits);
```
because then you don't have to check that the debt was updated to `orinalDebt - badDebt`, and this is in the order of the other updates (pending fee updated first then debt is updated)

src/Midnight.sol
```diff
@@ -624,6 +710,34 @@ contract Midnight is IMidnight {
...
+    /// @dev Expects the obligation to be touched.
+    /// @dev Expects the id to correspond to the obligation's id.
+    function accrueContinuousFee(Obligation memory obligation, bytes32 id, address borrower) internal {
+        Position storage _position = position[id][borrower];
+        ObligationState storage _obligationState = obligationState[id];
+        uint128 accruedFee = accrueContinuousFeeView(obligation, id, borrower);
+        if (accruedFee > 0) {
+            _position.pendingFee -= accruedFee;
+            _position.debt += accruedFee;
+            _obligationState.totalUnits += accruedFee;
```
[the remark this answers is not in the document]
adhusson: fixing

src/Midnight.sol
MathisGD: we cannot because it requires obligation and id

src/Midnight.sol
```diff
@@ -624,6 +710,34 @@ contract Midnight is IMidnight {
         return tentativeSigner;
     }
 
+    /// @dev Returns the accrued fee.
```
QGarchery, line 703:
```suggestion
    /// @dev Returns the accrued fee.
    /// @dev Expects the id to correspond to the obligation's id.
```

src/Midnight.sol
MathisGD: yes, imo it's not needed in the API

src/Midnight.sol
adhusson: what about adding `toId(obligation)`?

src/Midnight.sol
QGarchery: > but it's public, you mean remove it from external API?

`accrueContinuousFeeView` returns the same thing, and it's public

src/Midnight.sol
MathisGD: >slash is public too, you mean make both internal?

he meant make both public

src/Midnight.sol
adhusson: * it's debt then pending fee just above l452
* just above it uses `originalDebt`, not `_position.debt` even though it could use `originalDebt`
* I think it's good and clearer to reuse the cached vars rather than the mutable versions.

MathisGD: in take, the debt is updated before the pendingFee

src/Midnight.sol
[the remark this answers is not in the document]
MathisGD: nb: I added two simple tests for this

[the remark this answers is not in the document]
MathisGD: > fixing

how do you want to fix it?

src/Midnight.sol
MathisGD: anyway I'll try to uniformize

src/Midnight.sol
```diff
@@ -134,14 +142,31 @@ contract Midnight is IMidnight {
...
-        emit EventsLib.SetTradingFeeRecipient(feeRecipient);
+        feeRecipient = newFeeRecipient;
+        emit EventsLib.SetFeeRecipient(newFeeRecipient);
+    }
+
+    function setObligationContinuousFee(bytes32 id, uint256 newContinuousFee) external {
+        require(msg.sender == feeSetter, "only fee setter");
+        require(newContinuousFee <= MAX_CONTINUOUS_FEE, "continuous fee too high");
+        require(obligationState[id].created, "obligation not created");
+        // forge-lint: disable-next-line(unsafe-typecast) as newContinuousFee <= 317097919 < type(uint64).max
```
```suggestion
        // forge-lint: disable-next-line(unsafe-typecast) as newContinuousFee <= MAX_CONTINUOUS_FEE < type(uint64).max
```
more future proof/clean

certora/specs/OnlyAuthorizedCanChange.spec
```diff
@@ -45,17 +41,19 @@ function CVL_signer() returns address {
...
+    bool isPassiveFeeRecipient = user == Utils.passiveFeeRecipient();
 
     uint256 creditBefore = creditOf(id, user);
     uint256 debtBefore = debtOf(id, user);
     f(e, args);
     uint256 creditAfter = creditOf(id, user);
     uint256 debtAfter = debtOf(id, user);
 
-    assert (creditAfter == creditBefore && debtAfter == debtBefore) || userIsAuthorized || signed[user];
+    assert (creditAfter == creditBefore && debtAfter == debtBefore) || userIsAuthorized || signed[user] || isPassiveFeeRecipient;
```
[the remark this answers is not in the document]
MathisGD: added to wish list

certora/specs/BalanceEffects.spec
```diff
@@ -50,12 +48,18 @@ function summaryMulDiv(uint256 x, uint256 y, uint256 d) returns uint256 {
     return ghostMulDiv(x, y, d);
 }
 
+// All rules in this file assume no accrual.
+definition noAccrual(env e, bytes32 id, address borrower) returns bool = currentContract.position[id][borrower].pendingFee == 0 || e.block.timestamp == currentContract.position[id][borrower].lastContinuousFeeAccrual;
```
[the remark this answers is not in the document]
MathisGD: added to wish list

src/Midnight.sol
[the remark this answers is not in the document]
MathisGD: see #536

src/Midnight.sol
```diff
@@ -245,17 +281,17 @@ contract Midnight is IMidnight {
             receiver,
             offer.group,
             newConsumed,
-            _obligationState.totalUnits
+            _obligationState.totalUnits,
+            buyerPos.pendingFee,
+            sellerPos.pendingFee
```
peyha, line 284:
```suggestion
            sellerPos.pendingFee,
            buyerDebtIncrease,
            sellerDebtIncrease
```
buyerPos.debt and sellerPos.debt might be difficult to re-compute without this. It is updated by `buyerPos.debt -= UtilsLib.toUint128(buyerDebtDecrease)` where `uint256 buyerDebtDecrease = UtilsLib.min(oldBuyerDebt, obligationUnits)` which require the debt from the previous interaction (difficult to get)

but at the same time, `Take` already logs a lot of information

adhusson: Could be worth it but at the same time those are not hard updates (still, need to recompute increaes & decrease first).

certora/confs/Midnight.conf
```diff
@@ -9,7 +10,7 @@
   "optimistic_loop": true,
   "loop_iter": 2,
   "optimistic_hashing": true,
-  "hashing_length_bound": 2048,
+  "hashing_length_bound": 1024,
```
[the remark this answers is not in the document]
adhusson: i think its fine, @QGarchery @MathisGD ?

src/Midnight.sol
adhusson: made an issue

certora/specs/DebtAccrual.spec
```diff
@@ -0,0 +1,75 @@
...
+
+    function IdLib.toId(Midnight.Obligation memory, uint256, address) internal returns (bytes32) => NONDET;
+
+    function accrueContinuousFee(Midnight.Obligation memory, bytes32 id, address borrower) internal => summaryAccrueContinuousFee(id, borrower);
+}
+
+/// GHOSTS ///
+
+/// Whether accrueContinuousFee was called for (id, user) in this transaction.
+persistent ghost mapping(bytes32 => mapping(address => bool)) accrued;
```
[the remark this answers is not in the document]
adhusson: once accrued it stays accrued, expected

src/Midnight.sol
```diff
@@ -539,43 +603,43 @@ contract Midnight is IMidnight {
...
         address create2Address = address(uint160(uint256(id)));
         return abi.decode(create2Address.code, (Obligation));
     }
 
-    function creditAfterSlashing(bytes32 id, address user) public view returns (uint256) {
+    function creditAfterSlashing(bytes32 id, address user) external view returns (uint256) {
         Position storage _position = position[id][user];
         uint128 _userLossIndex = _position.lossIndex;
         uint128 lossIndex = obligationState[id].lossIndex;
         if (_userLossIndex == lossIndex) return _position.credit;
```
```suggestion

```
would work similarly except if `_userlossIndex=type(uint128.max)` so it would feel cleaner to not do this check or to check ``_userlossIndex==type(uint128.max)` instead

certora/confs/Midnight.conf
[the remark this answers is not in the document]
QGarchery: Yes it's fine (note that the worst case is not on `Obligation` but on `Offer`), because it still leaves plenty of room to have non-trivial obligations. Also the prover outputs an error if this leads to vacuous rules

src/Midnight.sol
QGarchery: We had a problem in the spec when the implementation of `creditAfterSlashing` differs from `slash`: if the loss index and the user loss index are both at max, then `slash` reverts but `creditAfterSlashing` doesn't

src/Midnight.sol
```diff
@@ -620,6 +700,52 @@ contract Midnight is IMidnight {
         return tentativeSigner;
     }
 
+    /// @dev Returns the accrued fee.
+    function accrueContinuousFeeView(Obligation memory obligation, address borrower) public view returns (uint256) {
+        return _accrueContinuousFeeView(obligation, IdLib.toId(obligation, block.chainid, address(this)), borrower);
+    }
+
+    /// @dev Expects the id to correspond to the obligation's id.
+    function _accrueContinuousFeeView(Obligation memory obligation, bytes32 id, address borrower)
```
peyha: why not pass only the maturity instead of the entire obligation ?

certora/specs/Midnight.spec
```diff
@@ -8,10 +10,23 @@ methods {
...
     function userLossIndex(bytes32 id, address user) external returns (uint128) envfree;
+    function Utils.passiveFeeRecipient() external returns (address) envfree;
 
     function _.price() external => NONDET;
     function IdLib.toId(Midnight.Obligation memory, uint256, address) internal returns (bytes32) => NONDET;
+
+    function tradingFee(bytes32, uint256) internal returns (uint256) => NONDET;
+    function isHealthy(Midnight.Obligation memory, bytes32, address) internal returns (bool) => NONDET;
+    function SafeTransferLib.safeTransferFrom(address, address, address, uint256) internal => NONDET;
+    function SafeTransferLib.safeTransfer(address, address, uint256) internal => NONDET;
```
QGarchery, line 23: Let's add a comment to say that the tokens are assumed to not reenter

certora/specs/OnlyAuthorizedCanChange.spec
```diff
@@ -32,6 +24,10 @@ methods {
 
 /// HELPERS ///
 
+definition noAccrual(env e, bytes32 id, address borrower) returns bool = currentContract.position[id][borrower].pendingFee == 0 || e.block.timestamp == currentContract.position[id][borrower].lastContinuousFeeAccrual;
+
+use invariant noRemainingContinuousFeeWithoutDebt;
+
```
QGarchery, line 30:
```suggestion
```
unused here

adhusson:
closing in favor of #547

certora/specs/Midnight.spec
QGarchery: will move it to #547

certora/specs/OnlyAuthorizedCanChange.spec
QGarchery: will move it to #547

PR #472, opened 2026-03-06
sig based authorize

MathisGD:
signature based setIsAuthorized

### DIFF
--- a/src/Midnight.sol
+++ b/src/Midnight.sol
@@ -516,7 +516,7 @@ contract Midnight is IMidnight {
 
     function setIsAuthorized(address authorized, bool newIsAuthorized) external {
         isAuthorized[msg.sender][authorized] = newIsAuthorized;
-        emit EventsLib.SetIsAuthorized(msg.sender, authorized, newIsAuthorized);
+        emit EventsLib.SetIsAuthorized(msg.sender, msg.sender, authorized, newIsAuthorized);
     }
 
     function setIsAuthorizedWithSig(Authorization memory authorization, Signature memory signature) external {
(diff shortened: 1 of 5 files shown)


src/Midnight.sol
```diff
@@ -513,7 +516,22 @@ contract Midnight is IMidnight {
...
+        require(block.timestamp <= authorization.deadline, "signature expired");
+        require(authorization.nonce == nonce[authorization.authorizer]++, "invalid nonce");
+
+        bytes32 structHash = keccak256(abi.encode(AUTHORIZATION_TYPEHASH, authorization));
+        bytes32 digest = keccak256(bytes.concat("\x19\x01", domainSeparator(), structHash));
+        address signatory = ecrecover(digest, signature.v, signature.r, signature.s);
+        require(signatory != address(0) && signatory == authorization.authorizer, "invalid signature");
+
+        isAuthorized[authorization.authorizer][authorization.authorized] = authorization.isAuthorized;
+        emit EventsLib.SetIsAuthorized(
```
QGarchery, line 532:
```suggestion
        emit EventsLib.SetIsAuthorizedWithSig(
```
otherwise, when reconstructing the state from events, you don't know whether you should increase the `nonce` or not (you could have `authorization.authorizer == msg.sender`)

src/libraries/EventsLib.sol
```diff
@@ -67,5 +67,7 @@ library EventsLib {
     event ShuffleSession(address indexed user, bytes32 session);
     event FlashLoan(address indexed caller, address indexed token, uint256 assets);
 
-    event SetIsAuthorized(address indexed authorizer, address indexed authorized, bool newIsAuthorized);
+    event SetIsAuthorized(
+        address indexed sender, address indexed authorizer, address indexed authorized, bool newIsAuthorized
```
QGarchery, line 71:
```suggestion
        address indexed caller, address indexed authorizer, address indexed authorized, bool newIsAuthorized
```

src/Midnight.sol
QGarchery, line 1: Far fetched idea, but interesting to think about:
remove the `authorizer` from the `Authorization` struct, this way you can "create" an account that is controlled by the `authorizer` (similar idea to Nick's method). We didn't do this in Blue, but it becomes even more powerful with the ability to [authorize everything](https://github.com/morpho-org/midnight/pull/473). Because now someone could create a new account, fill it, and transfer the authorization to another address by removing itself as authorized at the end

MathisGD:
see #25

src/Midnight.sol
MathisGD: #474

PR #482, opened 2026-03-09
[Certora] share price temp fix2

lilCertora:
created due to linter issue on 481


lilCertora:
closed because #481 is better

PR #488, opened 2026-03-09
Update copyright year in Midnight.sol

fdemiramon:


MathisGD (changes requested):
not sure what the copyright year is useful for. we can update it if you feel. but on all files then

fdemiramon:
just a feeling of having something outdated / forgot.
but I dunno if this has to be the year of publication or first line written...

MathisGD:
apparently it's the year of first publication, but what is publication exactly... again, feel free to push the full change but otherwise I'll close

PR #491, opened 2026-03-10
feat: separate fee and debt tracking

peyha:


src/interfaces/IMidnight.sol
```diff
@@ -42,7 +42,8 @@ struct Signature {
 }
 
 struct BorrowerState {
-    uint128 debt;
+    uint128 pureDebt;
+    uint128 feeDebt;
```
MathisGD: have you thought about storing totalDebt and feeDebt? maybe it simplifies things bit in the contract

peyha: nope, will try it

peyha: It works well if repay decrease the pure debt first. Otherwise it's pretty similar to using pure debt.

src/interfaces/IMidnight.sol
```diff
@@ -42,7 +42,8 @@ struct Signature {
 }
 
 struct BorrowerState {
-    uint128 debt;
+    uint128 totalDebt;
+    uint128 feeDebt;
```
MathisGD: purely a naming change
```suggestion
    uint128 debt;
    uint128 accruedFees;
```

src/Midnight.sol
```diff
@@ -267,11 +267,16 @@ contract Midnight is IMidnight {
             sharesOf[id][seller] -= obligationShares;
         } else if (!buyerIsLender && sellerIsBorrower) {
             // Borrower exits + borrower enters.
-            borrowerState[id][buyer].debt -= UtilsLib.toUint128(obligationUnits);
-            borrowerState[id][seller].debt += UtilsLib.toUint128(obligationUnits);
+            borrowerState[id][buyer].totalDebt -= UtilsLib.toUint128(obligationUnits);
+            borrowerState[id][buyer].feeDebt =
+                UtilsLib.toUint128(UtilsLib.min(borrowerState[id][buyer].feeDebt, borrowerState[id][buyer].totalDebt));
```
MathisGD: do we need to do this at all? I'm thinking that in Accrue we could do zeroFloorSub and never decrease feeDebt at all

peyha: not a huge fan of this abstraction intuitively because borrowers could repay their fee debt early on and the state would not update it accordingly

MathisGD: but they would save on fee only on what they paid. which doesn't seem too bad nor gameable

peyha: it's a "small" tradeoff between favoring the borrower vs the protocol. the only thing that I don't really like is that the "currently owed continuous fee" cannot be computed onchain anymore

MathisGD:
close as per this: https://morpholabs.slack.com/archives/C02N7CZ088N/p1773237914237309

PR #492, opened 2026-03-10
[Certora] onlyAuthorizedCanChange liquidate

QGarchery:
add liquidate to onlyAuthorizedCanChange


certora/specs/OnlyAuthorizedCanChange.spec
```diff
@@ -62,20 +121,25 @@ function CVL_signer() returns address {
...
 
     uint256 creditBefore = creditOf(id, user);
     uint256 debtBefore = debtOf(id, user);
     f(e, args);
     uint256 creditAfter = creditOf(id, user);
     uint256 debtAfter = debtOf(id, user);
 
-    assert (creditAfter == creditBefore && debtAfter == debtBefore) || userIsAuthorized || signed[user] || isPassiveFeeRecipient;
+    assert creditAfter == creditBefore || userIsAuthorized || signed[user] || isPassiveFeeRecipient;
+    assert debtAfter == debtBefore || userIsAuthorized || signed[user] || isPassiveFeeRecipient || !isHealthyBefore || obligation.maturity < e.block.timestamp;
```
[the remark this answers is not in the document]
bhargavbh:
```suggestion

    // case: f is not liquidate
    assert debtAfter == debtBefore || userIsAuthorized || signed[user] || isPassiveFeeRecipient || f.selector == sig:liquidate(Midnight.Obligation, uint256, uint256, uint256, address, bytes).selector;
    
    //case: f is liquidate
    assert debtAfter == debtBefore || userIsAuthorized || signed[user] || isPassiveFeeRecipient || !isHealthyBefore || obligation.maturity < e.block.timestamp;
```
^ we can make a case distinction. If this looks ugly, we can pull out the case of liquidate in another rule.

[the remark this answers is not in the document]
lilCertora: don't we risk to miss a bug in a non liquidate function this way ?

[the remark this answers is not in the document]
bhargavbh: for any non liquidate function, the assert in line-144 boils down to `assert debtAfter == debtBefore || userIsAuthorized || signed[user] || isPassiveFeeRecipient`, which is what we want. 
We could also add an antecedent `f!=liquidate => ....` in line-147 assert but that is not strictly necessary as the earlier asser already covers it.

certora/specs/OnlyAuthorizedCanChange.spec
```diff
@@ -48,6 +51,62 @@ methods {
...
+
+ghost CVL_price(address) returns uint256;
+
+ghost CVL_msb(uint128) returns uint256;
+
+ghost CVL_mulDivDown(uint256, uint256, uint256) returns uint256;
+
+ghost CVL_mulDivUp(uint256, uint256, uint256) returns uint256;
+
 definition noAccrual(env e, bytes32 id, address borrower) returns bool = currentContract.position[id][borrower].pendingFee == 0 || e.block.timestamp == currentContract.position[id][borrower].lastAccrual;
```
lilCertora, line 59: is it still used ?

certora/specs/OnlyAuthorizedCanChange.spec
lilCertora, line 88: maybe we could apply the same patter as function above by having `assert collateralAfter == collateralBefore || userIsAuthorized || !isHealthyBefore || obligation.maturity < e.block.timestamp;` and include liquidate into this rule ?

certora/specs/OnlyAuthorizedCanChange.spec
bhargavbh: good catch. no, we can discard it.

certora/specs/OnlyAuthorizedCanChange.spec
bhargavbh: makes sense

certora/specs/OnlyAuthorizedCanChange.spec
```diff
@@ -92,6 +173,19 @@ rule onlyAuthorizedCanChangeCollateralExceptLiquidate(env e, method f, calldataa
...
+/// Assumes no reentrancy: callbacks and token transfers are not modeled as re-entering Midnight, so re-entrant collateral changes are not covered.
+rule onlyAuthorizedCanChangeCollateralForLiquidate(env e, Midnight.Obligation obligation, bytes32 id, address user, uint256 collateralIndex, uint256 seizedAssets, uint256 repaidUnits, address borrower, bytes data) {
+    bool userIsAuthorized = user == e.msg.sender || isAuthorized(user, e.msg.sender);
+    bool isHealthyBefore = isHealthy(e, obligation, id, user);
+
+    uint256 collateralBefore = collateralOf(id, user, collateralIndex);
+    liquidate(e, obligation, collateralIndex, seizedAssets, repaidUnits, borrower, data);
+    uint256 collateralAfter = collateralOf(id, user, collateralIndex);
+
+    assert collateralAfter == collateralBefore || userIsAuthorized || !isHealthyBefore || obligation.maturity < e.block.timestamp;
```
MathisGD:
```suggestion
    assert collateralAfter == collateralBefore || !isHealthyBefore || obligation.maturity < e.block.timestamp;
```

certora/specs/OnlyAuthorizedCanChange.spec
```diff
@@ -75,7 +132,31 @@ rule onlyAuthorizedCanChangeCreditAndDebtExceptLiquidateAndSlash(env e, method f
...
+    uint256 debtBefore = debtOf(id, user);
+    uint256 collateralIndex;
+
+    liquidate(e, obligation, collateralIndex, seizedAssets, repaidUnits, borrower, data);
+
+    uint256 creditAfter = creditOf(id, user);
+    uint256 debtAfter = debtOf(id, user);
+
+    assert creditAfter == creditBefore || userIsAuthorized || signed[user] || isPassiveFeeRecipient;
+    assert debtAfter == debtBefore || userIsAuthorized || signed[user] || !isHealthyBefore || obligation.maturity < e.block.timestamp;
```
MathisGD:
```suggestion
    assert creditAfter == creditBefore;
    assert debtAfter == debtBefore || !isHealthyBefore || obligation.maturity < e.block.timestamp;
```

certora/specs/OnlyAuthorizedCanChange.spec
```diff
@@ -75,7 +132,31 @@ rule onlyAuthorizedCanChangeCreditAndDebtExceptLiquidateAndSlash(env e, method f
...
+    Midnight.Obligation obligation = getGlobalObligation();
+    bytes32 id = globalId;
+
+    bool userIsAuthorized = user == e.msg.sender || isAuthorized(user, e.msg.sender);
+    bool isPassiveFeeRecipient = user == Utils.passiveFeeRecipient();
+    bool isHealthyBefore = isHealthy(e, obligation, id, user);
+
+    uint256 creditBefore = creditOf(id, user);
+    uint256 debtBefore = debtOf(id, user);
+    uint256 collateralIndex;
```
MathisGD: this should be an "argument" of the rule

certora/specs/OnlyAuthorizedCanChange.spec
MathisGD, line 1: no need to do it in this PR, but we should rename this file and the rules. they are more about some kind of "anti-liveness" (if none of this is met, reverts), than purely about authorized

bhargavbh: on second thoughts, feels like the right place for the property on liquidate we are proving here is Liquidate.spec. We also do not need the whole globalID persistent ghost setup if we use the trick of equating id==lastId after the liquidate call (similar to the rule liquidateRequireUnhealthy). These newly added properties are not really access control. We can make this change in the current PR itself. @MathisGD

certora/specs/Liquidate.spec
```diff
@@ -37,6 +45,36 @@ ghost CVL_price(address) returns uint256;
...
+    // it's okay to check only after the call that the prover chose the correct id.
+    require id == lastId, "id should be derived from obligation";
+
+    uint256 creditAfter = creditOf(id, user);
+    uint256 debtAfter = debtOf(id, user);
+    uint256 collateralAfter = collateralOf(id, user, collateralIndex);
+    uint256 borrowerDebtAfter = debtOf(id, borrower);
+    uint256 borrowerCollateralAfter = collateralOf(id, borrower, collateralIndex);
+
+    assert creditAfter == creditBefore || user == Utils.passiveFeeRecipient();
```
jhoenicke: Why is this exception even necessary?   Only updatePosition adds fees and this is not called from liquidate. 

If it's needed, can we show that credit is increasing?

certora/specs/Liquidate.spec
```diff
@@ -37,6 +45,36 @@ ghost CVL_price(address) returns uint256;
...
+    bytes32 id;
+    bool liquidatable = !isHealthy(e, obligation, id, borrower) || obligation.maturity < e.block.timestamp;
+
+    uint256 creditBefore = creditOf(id, user);
+    uint256 debtBefore = debtOf(id, user);
+    uint256 collateralBefore = collateralOf(id, user, collateralIndex);
+    uint256 borrowerDebtBefore = debtOf(id, borrower);
+    uint256 borrowerCollateralBefore = collateralOf(id, borrower, collateralIndex);
+
+    liquidate(e, obligation, collateralIndex, seizedAssets, repaidUnits, borrower, data);
```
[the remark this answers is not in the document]
jhoenicke: The default semantics for an external call is HAVOC_ECF, which basically assumes that no-reentrancy happens: it assumes that no variable changes during the callback.  This is why the verification didn't fail.

In some sense, this is not sound, as onLiquidate could trigger a withdraw() or similar.  But since we are only interested in the changes that were caused by the liquidation itself, this is good enough for this property.

certora/specs/Liquidate.spec
bhargavbh: you are right, the assert can be unconditional. thanks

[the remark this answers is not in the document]
bhargavbh: that assert is now unconditional

certora/specs/Liquidate.spec
```diff
@@ -8,6 +10,12 @@ methods {
     function UtilsLib.msb(uint128 bitmap) internal returns (uint256) => CVL_msb(bitmap);
     function UtilsLib.mulDivDown(uint256 a, uint256 b, uint256 denominator) internal returns (uint256) => CVL_mulDivDown(a, b, denominator);
     function UtilsLib.mulDivUp(uint256 a, uint256 b, uint256 denominator) internal returns (uint256) => CVL_mulDivUp(a, b, denominator);
+
+    function creditOf(bytes32 id, address user) external returns (uint256) envfree;
+    function debtOf(bytes32 id, address user) external returns (uint256) envfree;
+    function collateralOf(bytes32 id, address user, uint256 index) external returns (uint128) envfree;
+    function Utils.passiveFeeRecipient() external returns (address) envfree;
```
MathisGD:
```suggestion
```

certora/specs/Liquidate.spec
```diff
@@ -37,6 +45,36 @@ ghost CVL_price(address) returns uint256;
...
+    uint256 borrowerCollateralAfter = collateralOf(id, borrower, collateralIndex);
+
+    assert creditAfter == creditBefore;
+    assert debtAfter == debtBefore || (user == borrower && liquidatable);
+    assert collateralAfter == collateralBefore || (user == borrower && liquidatable);
+    assert borrowerDebtAfter <= borrowerDebtBefore;
+    assert borrowerCollateralAfter <= borrowerCollateralBefore;
+}
+
 rule liquidateRequireUnhealthy(env e, Midnight.Obligation obligation, uint256 collateralIndex, uint256 seizedAssets, uint256 repaidUnits, address borrower, bytes data) {
```
MathisGD:
```suggestion
rule liquidateRequireLiquidatable(env e, Midnight.Obligation obligation, uint256 collateralIndex, uint256 seizedAssets, uint256 repaidUnits, address borrower, bytes data) {
```

certora/specs/Liquidate.spec
```diff
@@ -37,6 +45,36 @@ ghost CVL_price(address) returns uint256;
...
     bytes32 id;
     bool isHealthyBefore = isHealthy(e, obligation, id, borrower);
```
MathisGD:
```suggestion
    bool isLiquidatable = isLiquidatable(e, obligation, id, borrower);
```

certora/specs/Liquidate.spec
```diff
@@ -37,17 +44,36 @@ ghost CVL_price(address) returns uint256;
 
 // RULES ///
 
-rule liquidateRequireUnhealthy(env e, Midnight.Obligation obligation, uint256 collateralIndex, uint256 seizedAssets, uint256 repaidUnits, address borrower, bytes data) {
```
jhoenicke, line 40: This rule is no longer valid, right?  It's still in the CI?

certora/specs/Liquidate.spec
```diff
@@ -37,17 +44,36 @@ ghost CVL_price(address) returns uint256;
...
+    uint256 creditAfter = creditOf(id, user);
+    uint256 debtAfter = debtOf(id, user);
+    uint256 collateralAfter = collateral(id, user, collateralIndex);
+    uint256 borrowerDebtAfter = debtOf(id, borrower);
+    uint256 borrowerCollateralAfter = collateral(id, borrower, collateralIndex);
+
+    assert creditAfter == creditBefore;
+    assert debtAfter == debtBefore || (user == borrower && wasLiquidatable);
+    assert collateralAfter == collateralBefore || (user == borrower && wasLiquidatable);
+    assert borrowerDebtAfter <= borrowerDebtBefore;
```
jhoenicke: You could use the debt/collateralAfter/Before instead of borrower* here; if user == borrower, it's the same and otherwise, it still holds, because of equality.  Then you don't need borrower* at all.

I'm not sure what looks nicer, though.

jhoenicke (approved):
Looks good to me.  I had an optional suggestion to simplify the rule.

certora/specs/Liquidate.spec
bhargavbh: went with your suggestion, fairly readable after the change as well.

certora/confs/Liquidate.conf
```diff
@@ -1,5 +1,6 @@
 {
   "files": [
+    "certora/helpers/Utils.sol",
     "src/Midnight.sol"
   ],
   "verify": "Midnight:certora/specs/Liquidate.spec",
```
lilCertora, line 5: can we add sanity check here ?

certora/specs/Liquidate.spec
```diff
@@ -37,17 +44,32 @@ ghost CVL_price(address) returns uint256;
 
 // RULES ///
 
-rule liquidateRequireUnhealthy(env e, Midnight.Obligation obligation, uint256 collateralIndex, uint256 seizedAssets, uint256 repaidUnits, address borrower, bytes data) {
+/// Credit, debt, and collateral of a user can only change via liquidate if the position is liquidatable and user is borrower.
```
lilCertora:
```suggestion
/// Credit never change, debt, and collateral of a user can only change via liquidate if the position is liquidatable and user is borrower.
```

lilCertora (approved):

certora/specs/Liquidate.spec
QGarchery: it's still valid but less powerful than the next one, because `isLiquidatable` requires more than being unhealthy or being past maturity (it also requires the liquidation to not be locked)

certora/specs/Liquidate.spec
```diff
@@ -37,17 +44,32 @@ ghost CVL_price(address) returns uint256;
...
     require id == lastId, "id should be derived from obligation";
 
-    assert !isHealthyBefore || e.block.timestamp > obligation.maturity, "liquidate can only be called on unhealthy obligations";
+    uint256 creditAfter = creditOf(id, user);
+    uint256 debtAfter = debtOf(id, user);
+    uint256 collateralAfter = collateral(id, user, collateralIndex);
+
+    assert creditAfter == creditBefore;
+    assert debtAfter == debtBefore || (user == borrower && wasLiquidatable);
+    assert collateralAfter == collateralBefore || (user == borrower && wasLiquidatable);
```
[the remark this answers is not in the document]
QGarchery: fixed in latest commit

certora/specs/Liquidate.spec
```diff
@@ -3,58 +3,66 @@
...
-
-    // it's okay to check only after the call that the prover chose the correct id.
-    require id == lastId, "id should be derived from obligation";
+    bool wasLiquidatable = isLiquidatable(e, obligation, id, borrower);
 
-    assert !isHealthyBefore || e.block.timestamp > obligation.maturity, "liquidate can only be called on unhealthy obligations";
-}
+    uint256 creditBefore = creditOf(id, user);
+    uint256 debtBefore = debtOf(id, user);
+    uint256 collateralBefore = collateral(id, user, collateralIndex);
```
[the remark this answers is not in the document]
QGarchery: it's fine because if it reverts the rule passes

bhargavbh (approved):
`assert id == liqId => wasLiquidatable` nicely subsumes the "liquidateRevertsWhenNotLiquidatable" rule. 
LGTM

MathisGD (approved):

PR #493, opened 2026-03-10
[Certora] Specification for collateral bitmap

jhoenicke:
Show that the correct bits are set in the bitmap.
Show that the isHealthy function is equal to one running over all collaterals.

### DIFF
--- a/certora/confs/CollateralBitmap.conf
+++ b/certora/confs/CollateralBitmap.conf
@@ -1,0 +1,20 @@
+{
+  "files": [
+    "certora/helpers/MidnightWrapper.sol"
+  ],
+  "verify": "MidnightWrapper:certora/specs/CollateralBitmap.spec",
+  "solc": "solc-0.8.31",
+  "solc_via_ir": true,
+  "solc_evm_version": "osaka",
+  "optimistic_loop": true,
+  "loop_iter": 3,
+  "optimistic_hashing": true,
+  "hashing_length_bound": 2048,
+  "precise_bitwise_ops": true,
+  "prover_args": [
+    "-depth 5",
+    "-mediumTimeout 20",
+    "-timeout 3600"
+  ],
+  "msg": "Midnight: collateral bitmap"
+}
--- a/certora/helpers/MidnightWrapper.sol
+++ b/certora/helpers/MidnightWrapper.sol
@@ -12,7 +12,7 @@ contract MidnightWrapper is Midnight {
     using UtilsLib for uint256;
     using UtilsLib for uint128;
     
-    /* This isHealthy function iterates over all collaterals and also works if the collateral bitmap is broken. */
+    /* This isHealthy function iterates over all collaterals, it doesn't use the collateral bitmap. */
 
     function isHealthyNoBitmap(Obligation memory obligation, bytes32 id, address borrower) public view returns (bool) {
         BorrowerState storage _borrowerState = borrowerState[id][borrower];
--- a/certora/helpers/MidnightWrapper.sol
+++ b/certora/helpers/MidnightWrapper.sol
@@ -19,7 +19,8 @@ contract MidnightWrapper is Midnight {
         uint256 debt = _borrowerState.debt;
         uint256 maxDebt;
         uint256 len = obligation.collaterals.length;
-        for (uint256 i = 0; i < len && maxDebt < debt; i++) {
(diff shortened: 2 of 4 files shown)


certora/helpers/MidnightWrapper.sol
```diff
@@ -19,7 +19,8 @@ contract MidnightWrapper is Midnight {
         uint256 debt = _borrowerState.debt;
         uint256 maxDebt;
         uint256 len = obligation.collaterals.length;
-        for (uint256 i = 0; i < len && maxDebt < debt; i++) {
+        for (uint256 i = len; i > 0 && maxDebt < debt; ) {
```
jhoenicke, line 22: This needs to run in reverse because isHealthy does it reverse and if we do it differently we can have different overflow behavior.

certora/confs/CollateralBitmap.conf
QGarchery, line 1: let's add it to the `certora.yml`

certora/helpers/MidnightWrapper.sol
QGarchery:
```suggestion
    /* This isHealthy function iterates over all collaterals, it doesn't use the collateral bitmap. */
```
not sure what is the purpose of this comment otherwise

certora/specs/CollateralBitmap.spec
```diff
@@ -0,0 +1,53 @@
...
+// that does not use collateral bitmap returns true.  We also check that the latter function
+// does not revert if isHealthy does not revert.
+rule isHealthyEquivalant(Midnight.Obligation obligation, bytes32 id, address borrower) {
+    // We restrict to at most three collaterals
+    require obligation.collaterals.length <= 3, "restrict to three collaterals";
+    requireInvariant bitsetIffCollateral(id, borrower, 0);
+    requireInvariant bitsetIffCollateral(id, borrower, 1);
+    requireInvariant bitsetIffCollateral(id, borrower, 2);
+
+    // We make no claim about isHealthyNoBitmap() if isHealthy() reverts.
```
QGarchery, line 46: IIUC, we are not sure that calling `isHealthyNoBitmap` will not reduce the search space. For example, in the `stayHealthy` rule, some functions don't call at all `isHealthy`. But we do `callIsHealthy` before calling the function, so it means that it doesn't consider the cases where `isHealthyNoBitmap` reverts

certora/specs/CollateralBitmap.spec
```diff
@@ -0,0 +1,53 @@
...
+}
+
+// Check that a collateral bit is set exactly when there is collateral for that index.
+invariant bitsetIffCollateral(bytes32 id, address borrower, uint256 idx)
+    idx < 128 => (collateralBitSet(id, borrower, idx) <=> collateralOf(id, borrower, idx) != 0);
+
+// This shows that the real isHealthy returns true if and only if the isHealthy function
+// that does not use collateral bitmap returns true.  We also check that the latter function
+// does not revert if isHealthy does not revert.
+rule isHealthyEquivalant(Midnight.Obligation obligation, bytes32 id, address borrower) {
```
QGarchery:
```suggestion
rule isHealthyEquivalent(Midnight.Obligation obligation, bytes32 id, address borrower) {
```

QGarchery (approved):

certora/specs/CollateralBitmap.spec
jhoenicke: There are cases where both functions revert because of overflows.  We would need assumptions about maximum price the oracle can return, maximum collateral a user can have, etc.

I also tried to prove equal reverting behavior, but that is tricky:  We need another pre-condition that obligation and id match and prove an invariant that the bitmap never contains indices larger than the number of collaterals.  Otherwise, I get a counterexample where the bitmap has a high bit set and then isHealthy() reverts, but isHealthyNoBitmap() is fine.

lilCertora (approved):

MathisGD (commented):
LGTM

certora/helpers/MidnightWrapper.sol
MathisGD: how could one sum overflow and not the other? this is weird no,

jhoenicke: Because of the `maxDebt < debt` early exit.  One sum exits early before overflowing, the other comes from the other side and overflows.

PR #497, opened 2026-03-10
[Certora] liquidation is profitable and bounded by LIF

lilCertora:
1. Liquidation is profitable: the value of paid out collateral is larger than the value of paid loan tokens. 
2. The profit of liquidation is bounded by LIF 

[corresponding thread](https://morpholabs.slack.com/archives/C0ACSA8TNMA/p1773655802547359)


.github/workflows/certora.yml
```diff
@@ -24,7 +24,18 @@ jobs:
...
+          - CreatedObligations
+          - ExactMath
+          - Liquidate
+          - LiquidationBoundedByLIF
+          - LiquidationProfitability
+          - Midnight
+          - OnlyAuthorizedCanChange
+          - SharePrice
+          - Solvency
+          - WithdrawableMonotonicity
```
QGarchery: wrong merge here

certora/specs/LiquidationBoundedByLIF.spec
```diff
@@ -0,0 +1,67 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+methods {
+    function multicall(bytes[]) external => HAVOC_ALL DELETE;
+
+    // Assume price doesn't change during the execution of a transaction.
```
QGarchery: We don't really assume that, since the price is used only once per Collateral. To me this is more of a trick to be able to talk about the price in the spec

certora/specs/LiquidationBoundedByLIF.spec
```diff
@@ -0,0 +1,67 @@
...
+
+definition ORACLE_PRICE_SCALE() returns uint256 = 10 ^ 36;
+
+persistent ghost summaryPrice(address) returns uint256;
+
+persistent ghost summaryObligationId(address, uint256) returns bytes32;
+
+function summaryToId(Midnight.Obligation obligation) returns bytes32 {
+    return summaryObligationId(obligation.loanToken, obligation.maturity);
+}
```
QGarchery, line 44: It would be best to use the same method as in `CreatedObligations.spec` so that obligations with different Collateral array are not conflated. Would the spec go through with that change ?

certora/specs/LiquidationBoundedByLIF.spec
```diff
@@ -0,0 +1,67 @@
...
+
+/// LIF BOUNDARIES ///
+
+/// Liquidation profit is bounded by maxLif (repaidUnits input)
+rule liquidationProfitBounded_repaidUnits(env e, Midnight.Obligation obligation, uint256 collateralIndex, uint256 repaidUnits, address borrower, bytes data) {
```
QGarchery:
```suggestion
rule liquidationProfitBoundedInputRepaidUnits(env e, Midnight.Obligation obligation, uint256 collateralIndex, uint256 repaidUnits, address borrower, bytes data) {
```
to use camelCase consistently
-> more occurrences below

certora/specs/LiquidationBoundedByLIF.spec
```diff
@@ -0,0 +1,67 @@
...
+    mathint maxLif = obligation.collaterals[collateralIndex].maxLif;
+    require maxLif >= WAD();
+    require repaidUnits > 0;
```
QGarchery: let's put reason strings everywhere. In this last case, could it be put in the final assert ?

certora/specs/LiquidationBoundedByLIF.spec
```diff
@@ -0,0 +1,67 @@
...
+    mathint maxLif = obligation.collaterals[collateralIndex].maxLif;
+    require maxLif >= WAD();
+    require seizedAssets > 0;
+
+    // Safe: seized/repaid amounts are computed before the callback; data.length == 0 skips it for prover performance.
+    require data.length == 0;
+
+    // Restrict to 1 active collateral so the loop processes collateralIndex and sets liquidatedCollatPrice.
+    bytes32 id0 = summaryObligationId(obligation.loanToken, obligation.maturity);
+    require to_mathint(currentContract.borrowerState[id0][borrower].activatedCollaterals) == 2 ^ to_mathint(collateralIndex);
```
QGarchery: it's unclear to me why we need this assumption here but not in the rule with repaid units as input

lilCertora: This comes from the calculation difference : 
```
if (seizedAssets > 0) {
       repaidUnits = seizedAssets.mulDivUp(liquidatedCollatPrice, ORACLE_PRICE_SCALE).mulDivUp(WAD, lif);
} else {
}
```
seized path needs the constraint where the `liquidatedCollatPrice != 0` because if the collateral is not set for it , the liquidate function returns `(seizedAssets, 0)` and assertion fails so we force the loop to visit `collateralIndex`.
On the repaid path, if  `liquidatedCollatPrice == 0`, the calculation revert due to divison by 0.

certora/specs/LiquidationBoundedByLIF.spec
lilCertora: done, the spec go through : https://prover.certora.com/output/7508195/7a3d204935c24106b41c5083a5766aa2/?anonymousKey=f081c2d2e393dbde2bd20da25b2c5f8aac134001
thanks !

.github/workflows/certora.yml
QGarchery, line 1: I'm confused, why does it show changes to this file ?

certora/specs/LiquidationBoundedByLIF.spec
QGarchery: Ok I see. I think it would be best to not make this assumption, and instead rely on the fact that the bitmap is correct. This is another property that we could show (added it to the wish list): a user does not have collateral if the corresponding bit is not set. It means that if the `seized` assets are non-zero, then we know that the bit is set and we don't have this counter-example

certora/specs/LiquidationProfitability.spec
```diff
@@ -0,0 +1,77 @@
...
+
+persistent ghost summaryPrice(address) returns uint256;
+
+persistent ghost summaryMulDivDownM(mathint, mathint, mathint) returns mathint {
+    axiom forall mathint a. forall mathint b. forall mathint d. a >= 0 && b >= 0 && d > 0 => summaryMulDivDownM(a, b, d) >= 0;
+    axiom forall mathint a. forall mathint b. forall mathint d. a >= 0 && b >= 0 && d > 0 => (summaryMulDivDownM(a, b, d) + 1) * d > a * b;
+    axiom forall mathint a. forall mathint b. forall mathint d. a >= 0 && b >= 0 && d > 0 => summaryMulDivDownM(a, b, d) * d <= a * b;
+}
+
+persistent ghost summaryMulDivUpM(mathint, mathint, mathint) returns mathint {
```
QGarchery: If those ghost instead take `uint256` as input and return `uint256`, then we can avoid an axiom like `axiom forall mathint a. forall mathint b. forall mathint d. a >= 0 && b >= 0 && d > 0 => summaryMulDivUpM(a, b, d) >= 0;` and we can avoid having premises of the form `a >= 0 && b >= 0` in the corresponding axioms

certora/specs/LiquidationProfitability.spec
```diff
@@ -0,0 +1,77 @@
...
+definition WAD() returns uint256 = 10 ^ 18;
+
+definition ORACLE_PRICE_SCALE() returns uint256 = 10 ^ 36;
+
+persistent ghost summaryPrice(address) returns uint256;
+
+persistent ghost summaryMulDivDownM(mathint, mathint, mathint) returns mathint {
+    axiom forall mathint a. forall mathint b. forall mathint d. a >= 0 && b >= 0 && d > 0 => summaryMulDivDownM(a, b, d) >= 0;
+    axiom forall mathint a. forall mathint b. forall mathint d. a >= 0 && b >= 0 && d > 0 => (summaryMulDivDownM(a, b, d) + 1) * d > a * b;
+    axiom forall mathint a. forall mathint b. forall mathint d. a >= 0 && b >= 0 && d > 0 => summaryMulDivDownM(a, b, d) * d <= a * b;
```
QGarchery: It would be good to prove those axioms separately, ideally in the same place/file as in #388

certora/specs/LiquidationProfitability.spec
```diff
@@ -0,0 +1,77 @@
...
+    if (overflow || d == 0) {
+        revert();
+    }
+    return require_uint256(summaryMulDivUpM(a, b, d));
+}
+
+/// LIQUIDATION PROFITABILITY ///
+
+/// Liquidation is profitable up 1 collateral token unit in price-scaled terms due to floor rounding
+rule liquidationIsProfitable_repaidUnits(env e, Midnight.Obligation obligation, uint256 collateralIndex, uint256 repaidUnits, address borrower, bytes data) {
```
QGarchery: let's use camelCase here as well

certora/specs/LiquidationBoundedByLIF.spec
```diff
@@ -0,0 +1,63 @@
...
+    require repaidUnits > 0, "repaidUnits must be positive";
+    require data.length == 0, "no callback for prover performance";
+
+    uint256 seizedResult;
+    uint256 repaidResult;
+    seizedResult, repaidResult = liquidate(e, obligation, collateralIndex, 0, repaidUnits, borrower, data);
+
+    mathint price = summaryPrice(obligation.collaterals[collateralIndex].oracle);
+
+    assert maxLif >= WAD() => to_mathint(seizedResult) * price * WAD() <= to_mathint(repaidResult) * ORACLE_PRICE_SCALE() * maxLif;
```
QGarchery:
```suggestion
    assert maxLif >= WAD() => seizedResult * price * WAD() <= repaidResult * ORACLE_PRICE_SCALE() * maxLif;
```
we can remove `to_mathint` everywhere

certora/specs/LiquidationProfitability.spec
```diff
@@ -0,0 +1,77 @@
...
+rule liquidationIsProfitable_seizedAssets(env e, Midnight.Obligation obligation, uint256 collateralIndex, uint256 seizedAssets, address borrower, bytes data) {
+    require seizedAssets > 0, "seizedAssets must be positive";
+
+    uint256 seizedResult;
+    uint256 repaidResult;
+    seizedResult, repaidResult = liquidate(e, obligation, collateralIndex, seizedAssets, 0, borrower, data);
+
+    mathint price = summaryPrice(obligation.collaterals[collateralIndex].oracle);
+
+    assert obligation.collaterals[collateralIndex].maxLif >= WAD() => to_mathint(seizedResult) * price + ORACLE_PRICE_SCALE() > to_mathint(repaidResult) * ORACLE_PRICE_SCALE();
```
QGarchery:
```suggestion
    assert obligation.collaterals[collateralIndex].maxLif >= WAD() => seizedResult * price > (repaidResult - 1) * ORACLE_PRICE_SCALE();
```
this looks more like the description of the rule

certora/specs/LiquidationProfitability.spec
```diff
@@ -0,0 +1,77 @@
...
+rule liquidationIsProfitable_repaidUnits(env e, Midnight.Obligation obligation, uint256 collateralIndex, uint256 repaidUnits, address borrower, bytes data) {
+    require repaidUnits > 0, "repaidUnits must be positive";
+
+    uint256 seizedResult;
+    uint256 repaidResult;
+    seizedResult, repaidResult = liquidate(e, obligation, collateralIndex, 0, repaidUnits, borrower, data);
+
+    mathint price = summaryPrice(obligation.collaterals[collateralIndex].oracle);
+
+    assert obligation.collaterals[collateralIndex].maxLif >= WAD() => (to_mathint(seizedResult) + 1) * price > to_mathint(repaidResult) * ORACLE_PRICE_SCALE();
```
QGarchery: I'm skeptical about this spec overall, because being barely profitable (or not even profitable if we take into account the rounding) is not really a useful property, as liquidators need to be truly incentivized.

So could we also add the lif in there ?
```suggestion
    assert obligation.collaterals[collateralIndex].maxLif >= WAD() => (to_mathint(seizedResult) + 1) * price * WAD() > to_mathint(repaidResult) * ORACLE_PRICE_SCALE() * computedLif;
```

certora/specs/LiquidationProfitability.spec
lilCertora: i think the best would be to wait for 388 to be merged and then replace them no ?

certora/specs/LiquidationProfitability.spec
lilCertora: i get your point , this is true, i will work on that, thank you

.github/workflows/certora.yml
lilCertora: this is good now

certora/specs/LiquidationProfitability.spec
lilCertora: done

certora/specs/LiquidationProfitability.spec
lilCertora: yes , good idea , thanks

certora/specs/LiquidationProfitability.spec
QGarchery: Ok good point, let's merge this after 388

certora/specs/LiquidationBoundedByLIF.spec
```diff
@@ -0,0 +1,60 @@
...
+    require data.length == 0, "no callback for prover performance";
+    require maxLif >= WAD(), "maxLif must be at least 1x for profit boundedness";
+
+    uint256 seizedResult;
+    uint256 repaidResult;
+    seizedResult, repaidResult = liquidate(e, obligation, collateralIndex, 0, repaidUnits, borrower, data);
+
+    mathint price = summaryPrice(obligation.collaterals[collateralIndex].oracle);
+
+    assert repaidUnits > 0 => seizedResult * price * WAD() <= repaidResult * ORACLE_PRICE_SCALE() * maxLif;
```
bhargavbh: if repaidUnits = 0, then liquidate anyways reverts. 
```suggestion
    assert seizedResult * price * WAD() <= repaidResult * ORACLE_PRICE_SCALE() * maxLif;
```

certora/specs/LiquidationBoundedByLIF.spec
```diff
@@ -0,0 +1,60 @@
...
+    bytes32 id0 = summaryObligationId(obligation.loanToken, obligation.maturity);
+    require to_mathint(currentContract.position[id0][borrower].activatedCollaterals) == 2 ^ to_mathint(collateralIndex), "exactly 1 active collateral at collateralIndex so the loop sets liquidatedCollatPrice";
+
+    uint256 seizedResult;
+    uint256 repaidResult;
+    seizedResult, repaidResult = liquidate(e, obligation, collateralIndex, seizedAssets, 0, borrower, data);
+
+    mathint price = summaryPrice(obligation.collaterals[collateralIndex].oracle);
+
+    assert seizedAssets > 0 => seizedResult * price * WAD() <= repaidResult * ORACLE_PRICE_SCALE() * maxLif;
```
bhargavbh: similar, if seizedAssets = 0, then liquidate reverts
```suggestion
    assert seizedResult * price * WAD() <= repaidResult * ORACLE_PRICE_SCALE() * maxLif;
```

certora/specs/LiquidationBoundedByLIF.spec
```diff
@@ -0,0 +1,60 @@
...
+    assert repaidUnits > 0 => seizedResult * price * WAD() <= repaidResult * ORACLE_PRICE_SCALE() * maxLif;
+}
+
+/// Liquidation profit is bounded by maxLif (seizedAssets input)
+rule liquidationProfitBoundedSeizedAssets(env e, Midnight.Obligation obligation, uint256 collateralIndex, uint256 seizedAssets, address borrower, bytes data) {
+    mathint maxLif = obligation.collaterals[collateralIndex].maxLif;
+    require data.length == 0, "no callback for prover performance";
+    require maxLif >= WAD(), "maxLif must be at least 1x for profit boundedness";
+
+    bytes32 id0 = summaryObligationId(obligation.loanToken, obligation.maturity);
```
bhargavbh: looking at the asymmetry here, feels like we need to add a `require(liquidatedCollatPrice > 0)` in liquidate().

bhargavbh (commented):
if we agree on https://github.com/morpho-org/midnight/pull/535, then the rule `liquidationProfitBoundedSeizedAssets` can be simplified

certora/specs/LiquidationBoundedByLIF.spec
lilCertora: true, applying

lilCertora: actually it does not revert it's ok with 0 amount, reverting back

certora/specs/LiquidationProfitability.spec
```diff
@@ -0,0 +1,105 @@
...
+    assert seizedResult * price > (repaidResult - 1) * ORACLE_PRICE_SCALE();
+}
+
+
+/// Liquidation profit is at least maxLif-proportional when lif = maxLif (repaidUnits input).
+/// Pre-maturity forces the borrower to be unhealthy, which forces lif = maxLif.
+rule liquidationLIFProfitability_RepaidUnits(env e, Midnight.Obligation obligation, uint256 collateralIndex, uint256 repaidUnits, address borrower, bytes data) {
+    mathint maxLif = obligation.collaterals[collateralIndex].maxLif;
+    require maxLif >= WAD();
+    require e.block.timestamp <= obligation.maturity || e.block.timestamp >= require_uint256(obligation.maturity + 900);
```
jhoenicke: Wouldn't it be better to use `!isHealthy() || e.block.timestamp >= require_uint256(obligation.maturity + 900`?  This way you also cover the case where the user is unhealthy and during the 15 minutes.

The tricky part is the obligation id you need for isHealthy, see Liquidate.spec.

lilCertora: yes really good idea

certora/specs/LiquidationProfitability.spec
```diff
@@ -0,0 +1,105 @@
...
+rule liquidationIsProfitableRepaidUnits(env e, Midnight.Obligation obligation, uint256 collateralIndex, uint256 repaidUnits, address borrower, bytes data) {
+    require obligation.collaterals[collateralIndex].maxLif >= WAD(), "maxLif must be at least 1x for profitability";
+
+    uint256 seizedResult;
+    uint256 repaidResult;
+    seizedResult, repaidResult = liquidate(e, obligation, collateralIndex, 0, repaidUnits, borrower, data);
+
+    mathint price = summaryPrice(obligation.collaterals[collateralIndex].oracle);
+
+    assert seizedResult * price > (repaidResult - 1) * ORACLE_PRICE_SCALE();
```
jhoenicke: Copy-Paste error?  This should be `seizedResult + 1` and `repaidResult`, right?

certora/specs/LiquidationBoundedByLIF.spec
bhargavbh: actually such a require is not needed. resolved in slack convo.

certora/specs/LiquidationProfitability.spec
lilCertora: yes this was an error, thanks

certora/specs/LiquidationBoundedByLIF.spec
```diff
@@ -0,0 +1,60 @@
...
+    require to_mathint(currentContract.position[id0][borrower].activatedCollaterals) == 2 ^ to_mathint(collateralIndex), "exactly 1 active collateral at collateralIndex so the loop sets liquidatedCollatPrice";
```
jhoenicke: This looks dangerous... I see why it's needed, though.

We have proved an invariant that a bit in activatedCollaterals is set, if and only if the collateralOf is non-zero.  However this invariant is not enough here, unless we also do bit-precise reasoning, which probably times out.  Without bit-precise reasoning the solver cannot prove that it will iterate over the collateral.

lilCertora: should be better now by using your bitmap summaries, it's easier to reason about collateral bitmap

certora/specs/LiquidationBoundedByLIF.spec
```diff
@@ -0,0 +1,112 @@
...
+
+definition WAD() returns uint256 = 10 ^ 18;
+
+definition ORACLE_PRICE_SCALE() returns uint256 = 10 ^ 36;
+
+persistent ghost summaryPrice(address) returns uint256;
+
+persistent ghost summaryObligationId(address, uint256) returns bytes32;
+
+// Bitmap abstraction (see BitmapSummaries.spec for the proven invariant).
```
jhoenicke: You could just `import "BitmapSummaries.spec";` at the beginning of the file instead.

certora/specs/LiquidationBoundedByLIF.spec
```diff
@@ -0,0 +1,112 @@
...
+    require bit < 128, "see Bitmap.spec";
+    require summaryGetBit(bitmap, bit), "see Bitmap.spec";
+    require forall uint256 otherBit. summaryGetBit(bitmap, otherBit) => otherBit <= bit, "see Bitmap.spec";
+    return bit;
+}
+
+/// INVARIANTS ///
+
+/// Proven in BitmapSummaries.spec; assumed here via requireInvariant (not re-proven in this spec).
+strong invariant nonZeroCollateralsAreActivated(bytes32 id, address user, uint256 idx)
```
jhoenicke: This will now move to CollateralBitmap.spec in PR #388

It's impossible to import that, because that file defines to many things you don't want to import.

To leave out the proof of this invariant you had to list all the other rules explicitly.   I'm wondering if we can put the invariants into an Invariants.spec (without proving the spec file), import the Invariants.spec when necessary and then have one of  the spec file use the invariant, so it is proved there.

certora/specs/LiquidationBoundedByLIF.spec
```diff
@@ -0,0 +1,112 @@
...
+    assert seizedResult * price * WAD() <= repaidResult * ORACLE_PRICE_SCALE() * maxLif;
+}
+
+/// Liquidation profit is bounded by maxLif (seizedAssets input)
+rule liquidationProfitBoundedSeizedAssets(env e, Midnight.Obligation obligation, uint256 collateralIndex, uint256 seizedAssets, address borrower, bytes data) {
+    mathint maxLif = obligation.collaterals[collateralIndex].maxLif;
+    require data.length == 0, "no callback for prover performance";
+    require maxLif >= WAD(), "maxLif must be at least 1x for profit boundedness (see touchObligation validation and ExactMath.spec)";
+
+    require collateralIndex < 128, "collateralIndex must be less than 128";
```
jhoenicke: Is this necessary?  liquidate should revert if `collateralIndex >= 128`, right?

jhoenicke (changes requested):
I suggest to try importing BitmapSummary.spec and not duplicating the summary here.  The other part with the invariant, we can solve later.   However you may need to wait for #388 to import the summary (or merge it locally).

certora/confs/LiquidationBoundedByLIF.conf
```diff
@@ -0,0 +1,25 @@
...
+  ],
+  "verify": "Midnight:certora/specs/LiquidationBoundedByLIF.spec",
+  "solc": "solc-0.8.31",
+  "solc_via_ir": true,
+  "solc_evm_version": "osaka",
+  "optimistic_loop": true,
+  "loop_iter": 2,
+  "optimistic_hashing": true,
+  "hashing_length_bound": 2048,
+  "smt_timeout": 7200,
```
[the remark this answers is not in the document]
lilCertora: thanks

certora/confs/LiquidationProfitability.conf
```diff
@@ -0,0 +1,23 @@
...
+  "optimistic_hashing": true,
+  "hashing_length_bound": 2048,
+  "smt_timeout": 7200,
+  "rule_sanity": "basic",
+  "prover_args": [
+    "-depth 5",
+    "-mediumTimeout 20",
+    "-timeout 3600",
+    "-split false",
+    " -destructiveOptimizations twostage -backendStrategy singleRace -smt_useLIA false -smt_useNIA true -s [z3:def{randomSeed=11},z3:def{randomSeed=12},z3:def{randomSeed=13},z3:def{randomSeed=14},z3:def{randomSeed=15},z3:def{randomSeed=16},z3:def{randomSeed=17},z3:def{randomSeed=18},z3:def{randomSeed=19},z3:def{randomSeed=20}]"
```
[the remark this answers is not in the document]
lilCertora: that's ok

certora/specs/LiquidationBoundedByLIF.spec
lilCertora: what do you think @MathisGD ? is it ok if we create an invariants.spec and gather the main invariants there so that it's easier to import those needed on new specs without having to re-prove them and just use `requireInvariants` ?

MathisGD: sure, we can do that. ideally in a separated PR. Midnight.spec was weird anyway

certora/specs/LiquidationProfitability.spec
```diff
@@ -0,0 +1,110 @@
...
+    require maxLif >= WAD(), "maxLif must be at least 1x for profitability";
+    bytes32 id = toId(e, obligation);
+    require !isHealthy(obligation, id, borrower) || e.block.timestamp >= require_uint256(obligation.maturity + TIME_TO_MAX_LIF()), "lif = maxLif when borrower is unhealthy or >= 15 min post-maturity";
+
+    uint256 seizedResult;
+    uint256 repaidResult;
+    seizedResult, repaidResult = liquidate(e, obligation, collateralIndex, 0, repaidUnits, borrower, data);
+    mathint price = summaryPrice(obligation.collaterals[collateralIndex].oracle);
+
+    assert (seizedResult + 1) * price * WAD() + WAD() * ORACLE_PRICE_SCALE() > repaidResult * ORACLE_PRICE_SCALE() * maxLif;
```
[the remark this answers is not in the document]
lilCertora: it's already handled

certora/confs/LiquidationProfitability.conf
```diff
@@ -0,0 +1,23 @@
+{
+  "files": [
+    "src/Midnight.sol"
+  ],
+  "verify": "Midnight:certora/specs/LiquidationProfitability.spec",
+  "solc": "solc-0.8.31",
```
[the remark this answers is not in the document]
lilCertora: done

jhoenicke (approved):
Looks good now.  The rules make sense.

certora/specs/LiquidationBoundedByLIF.spec
lilCertora: I will do it as a separated PR in a future work

certora/confs/LiquidationBoundedByLIF.conf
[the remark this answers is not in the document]
QGarchery: It seems to be the default behavior [according to the docs](https://docs.certora.com/en/latest/docs/prover/checking/sanity.html#:~:text=When%20not%20using%20the%20rule_sanity%20option%20at%20all%2C%20or%20when%20using%20it%20with%20the%20value%20basic%2C%20the%20vacuity%20and%20trivial%20invariant%20checks%20are%20performed.) -> let's remove it if not necessary

certora/specs/LiquidationProfitability.spec
```diff
@@ -4,7 +4,12 @@ methods {
     function multicall(bytes[]) external => HAVOC_ALL DELETE;
 
     function isHealthy(Midnight.Obligation obligation, bytes32 id, address borrower) external returns (bool) envfree;
-    function toId(Midnight.Obligation) external returns (bytes32);
+
+    // Deterministic toId summary using a ghost that takes simple types (no struct).
+    function IdLib.toId(Midnight.Obligation memory obligation, uint256 chainId, address midnight) internal returns (bytes32) => summaryObligationId(obligation.loanToken, obligation.maturity);
+
+    // Skip obligation creation logic: removes the collateral-validation loop.
+    function touchObligation(Midnight.Obligation memory obligation) internal returns (bytes32) => summaryObligationId(obligation.loanToken, obligation.maturity);
```
QGarchery: Why do we need that now ?

certora/specs/LiquidationBoundedByLIF.spec
QGarchery: I don't think this was applied, it's still conflating obligations with different Collateral array

certora/specs/LiquidationProfitability.spec
```diff
@@ -0,0 +1,117 @@
...
+
+    assert seizedAssets > 0 => seizedResult * price > (repaidResult - 1) * ORACLE_PRICE_SCALE();
+}
+
+/// When lif = maxLif (borrower unhealthy or >= 15 min post-maturity), the liquidator receives collateral worth at least maxLif/WAD * repaid debt, up to rounding.
+rule liquidationIsProfitableRepaidUnits(env e, Midnight.Obligation obligation, uint256 collateralIndex, uint256 repaidUnits, address borrower, bytes data) {
+    uint256 maxLif = obligation.collateralParams[collateralIndex].maxLif;
+    require maxLif >= WAD(), "maxLif must be at least 1x for profitability";
+    bytes32 id = toId(e, obligation);
+    require !isHealthy(obligation, id, borrower) || e.block.timestamp >= require_uint256(obligation.maturity + TIME_TO_MAX_LIF()), "lif = maxLif when borrower is unhealthy or >= 15 min post-maturity";
```
QGarchery: Could we instead recompute the lif and show:
- that it is greater or equal to 1
- that in case the borrower is unhealthy, or 15min after maturity, it is equal to the maxLif

This way we wouldn't need the first 2 rules (solvent)

lilCertora: yes makes sense, let me try

certora/specs/LiquidationBoundedByLIF.spec
```diff
@@ -0,0 +1,109 @@
...
+    function Utils.hashObligation(Midnight.Obligation) external returns (bytes32) envfree;
+
+    // Summary to capture the oracle price so the spec can reference it in assertions.
+    function _.price() external => summaryPrice(calledContract) expect(uint256);
+
+    // Deterministic toId summary using a wrapper that extracts all scalar Obligation fields.
+    function IdLib.toId(Midnight.Obligation memory obligation, uint256 chainId, address midnight) internal returns (bytes32) => summaryToId(obligation);
+
+    // Skip obligation creation logic: removes the collateral-validation loop.
+    function touchObligation(Midnight.Obligation memory obligation) internal returns (bytes32) => summaryToId(obligation);
```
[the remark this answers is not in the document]
lilCertora: Codex misunderstands CVL semantics

certora/specs/LiquidationProfitability.spec
```diff
@@ -0,0 +1,95 @@
...
+
+/// LIF CHARACTERIZATION ///
+
+/// For repaidUnits input: lif >= WAD (solvency), and lif == maxLif when borrower is unhealthy or >= 15 min post-maturity (profitability).
+rule liquidationLifRepaidUnits(env e, Midnight.Obligation obligation, uint256 collateralIndex, uint256 repaidUnits, address borrower, bytes data) {
+    uint256 maxLif = obligation.collateralParams[collateralIndex].maxLif;
+    require maxLif >= WAD(), "maxLif must be at least 1x";
+
+    bytes32 id = toId(e, obligation);
+    bool maxLifReached = !isHealthy(obligation, id, borrower) || e.block.timestamp >= require_uint256(obligation.maturity + TIME_TO_MAX_LIF());
```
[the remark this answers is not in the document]
lilCertora: liquidate itself iterates over the bitmap in the same way when computing maxDebt so if bitmap is inconsisten isHealthy would revert the same way

certora/specs/LiquidationProfitability.spec
```diff
@@ -0,0 +1,112 @@
...
+    uint256 repaidResult;
+    seizedResult, repaidResult = liquidate(e, obligation, collateralIndex, seizedAssets, 0, borrower, data);
+
+    mathint price = summaryPrice(obligation.collateralParams[collateralIndex].oracle);
+
+    // lif >= WAD: liquidator receives collateral worth at least the repaid debt (up to 1 unit ceil rounding on repaidUnits).
+    assert seizedAssets > 0 => seizedResult * price > (repaidResult - 1) * ORACLE_PRICE_SCALE();
+
+    // lif == maxLif when borrower is unhealthy or >= 15 min post-maturity: full liquidation incentive factor applies.
+    assert seizedAssets > 0 && maxLifReached => seizedResult * price * WAD() + ORACLE_PRICE_SCALE() * WAD() > (repaidResult - 1) * maxLif * ORACLE_PRICE_SCALE();
```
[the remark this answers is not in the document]
lilCertora: not needed

certora/specs/LiquidationProfitability.spec
lilCertora: for prover performance mainly, I'm hitting timeout if i remove it and as we're testing liquidation on valid obligations it looks ok

certora/specs/LiquidationProfitability.spec
```diff
@@ -0,0 +1,112 @@
...
+
+/// LIF CHARACTERIZATION ///
+
+/// For repaidUnits input: lif >= WAD (solvency), and lif == maxLif when borrower is unhealthy or >= 15 min post-maturity (profitability).
+rule liquidationLifRepaidUnits(env e, Midnight.Obligation obligation, uint256 collateralIndex, uint256 repaidUnits, address borrower, bytes data) {
+    uint256 maxLif = obligation.collateralParams[collateralIndex].maxLif;
+    require maxLif >= WAD(), "maxLif must be at least 1x";
+
+    bytes32 id = toId(e, obligation);
+    bool maxLifReached = !isHealthy(obligation, id, borrower) || e.block.timestamp >= require_uint256(obligation.maturity + TIME_TO_MAX_LIF());
```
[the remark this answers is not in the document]
lilCertora: obligation.maturity is a uint256 in Solidity

certora/specs/LiquidationProfitability.spec
```diff
@@ -0,0 +1,112 @@
...
+    }
+    return summaryMulDivUpM(a, b, d);
+}
+
+/// LIF CHARACTERIZATION ///
+
+/// For repaidUnits input: lif >= WAD (solvency), and lif == maxLif when borrower is unhealthy or >= 15 min post-maturity (profitability).
+rule liquidationLifRepaidUnits(env e, Midnight.Obligation obligation, uint256 collateralIndex, uint256 repaidUnits, address borrower, bytes data) {
+    uint256 maxLif = obligation.collateralParams[collateralIndex].maxLif;
+    require maxLif >= WAD(), "maxLif must be at least 1x";
```
QGarchery:
```suggestion
    require maxLif >= WAD(), "see the rule maxLifIsAtLeastWad";
```

certora/specs/LiquidationProfitability.spec
```diff
@@ -0,0 +1,112 @@
...
+    assert repaidUnits > 0 => (seizedResult + 1) * price > repaidResult * ORACLE_PRICE_SCALE();
+
+    // lif == maxLif when borrower is unhealthy or >= 15 min post-maturity: full liquidation incentive factor applies.
+    assert repaidUnits > 0 && maxLifReached => (seizedResult + 1) * price * WAD() + WAD() * ORACLE_PRICE_SCALE() > repaidResult * ORACLE_PRICE_SCALE() * maxLif;
+}
+
+/// For seizedAssets input: lif >= WAD (solvency), and lif == maxLif when borrower is unhealthy or >= 15 min post-maturity (profitability).
+rule liquidationLifSeizedAssets(env e, Midnight.Obligation obligation, uint256 collateralIndex, uint256 seizedAssets, address borrower, bytes data) {
+    uint256 maxLif = obligation.collateralParams[collateralIndex].maxLif;
+    require maxLif >= WAD(), "maxLif must be at least 1x";
```
QGarchery:
```suggestion
    require maxLif >= WAD(), "see the rule maxLifIsAtLeastWad";
```

certora/specs/LiquidationProfitability.spec
QGarchery:
```suggestion
    assert maxLifReached => seizedResult * price * WAD() + ORACLE_PRICE_SCALE() * WAD() > (repaidResult - 1) * maxLif * ORACLE_PRICE_SCALE();
```
wondering why `seizedAssets > 0` was added (there was no need for it before, see [here](https://github.com/morpho-org/midnight/blob/ba56e031c01870e398d0fbed2eba78f94bb1a29f/certora/specs/LiquidationProfitability.spec#L104-L117))
(similarly for all the other asserts)

certora/specs/LiquidationProfitability.spec
```diff
@@ -0,0 +1,112 @@
...
+
+persistent ghost summaryMulDivDownM(uint256, uint256, uint256) returns uint256 {
+    axiom forall uint256 a. forall uint256 b. forall uint256 d. d > 0 => (summaryMulDivDownM(a, b, d) + 1) * d > a * b;
+    axiom forall uint256 a. forall uint256 b. forall uint256 d. d > 0 => summaryMulDivDownM(a, b, d) * d <= a * b;
+}
+
+persistent ghost summaryMulDivUpM(uint256, uint256, uint256) returns uint256 {
+    axiom forall uint256 a. forall uint256 b. forall uint256 d. d > 0 => summaryMulDivUpM(a, b, d) * d < a * b + d;
+    axiom forall uint256 a. forall uint256 b. forall uint256 d. d > 0 => summaryMulDivUpM(a, b, d) * d >= a * b;
+}
```
QGarchery: let's show the axioms in MulDiv.spec and reference those proofs here

certora/specs/LiquidationProfitability.spec
```diff
@@ -0,0 +1,112 @@
...
+function summaryToId(Midnight.Obligation obligation) returns bytes32 {
+    return Utils.hashObligation(obligation);
+}
+
+// Non-deterministic overflow models potential revert on x * y overflow.
+function summaryMulDivDown(uint256 a, uint256 b, uint256 d) returns uint256 {
+    bool overflow;
+    if (overflow || d == 0) {
+        revert();
+    }
```
QGarchery: maybe not necessary to add the revert behavior, since we are not looking at reverts in this spec. Note that we already make this kind of assumption by summarizing transfers with NONDET

certora/specs/LiquidationProfitability.spec
```diff
@@ -0,0 +1,112 @@
...
+    uint256 repaidResult;
+    seizedResult, repaidResult = liquidate(e, obligation, collateralIndex, 0, repaidUnits, borrower, data);
+
+    mathint price = summaryPrice(obligation.collateralParams[collateralIndex].oracle);
+
+    // lif >= WAD: liquidator receives collateral worth at least the repaid debt (up to 1 unit floor rounding on seizedAssets).
+    assert repaidUnits > 0 => (seizedResult + 1) * price > repaidResult * ORACLE_PRICE_SCALE();
+
+    // lif == maxLif when borrower is unhealthy or >= 15 min post-maturity: full liquidation incentive factor applies.
+    assert repaidUnits > 0 && maxLifReached => (seizedResult + 1) * price * WAD() + WAD() * ORACLE_PRICE_SCALE() > repaidResult * ORACLE_PRICE_SCALE() * maxLif;
```
QGarchery:
```suggestion
    assert repaidUnits > 0 && maxLifReached => (seizedResult + 1) * price * WAD() + ORACLE_PRICE_SCALE() * WAD() > repaidResult * maxLif * ORACLE_PRICE_SCALE();
```
(nitpick) to have the same order as for the seized assets case

certora/specs/LiquidationProfitability.spec
lilCertora: yes, i removed it for `liquidationLifSeizedAssets` , we need it however in `liquidationLifRepaidUnits`

certora/confs/LiquidationBoundedByLIF.conf
```diff
@@ -0,0 +1,27 @@
...
+  "rule": [
+    "liquidationProfitBoundedInputRepaidUnits",
+    "liquidationProfitBoundedSeizedAssets"
+  ],
```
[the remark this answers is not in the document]
lilCertora: invariant is proved elsewhere

certora/specs/LiquidationProfitability.spec
lilCertora: yes that's true too, thanks

certora/specs/LiquidationProfitability.spec
```diff
@@ -0,0 +1,112 @@
...
+    bool maxLifReached = !isHealthy(obligation, id, borrower) || e.block.timestamp >= require_uint256(obligation.maturity + TIME_TO_MAX_LIF());
+
+    uint256 seizedResult;
+    uint256 repaidResult;
+    seizedResult, repaidResult = liquidate(e, obligation, collateralIndex, 0, repaidUnits, borrower, data);
+
+    mathint price = summaryPrice(obligation.collateralParams[collateralIndex].oracle);
+
+    // lif >= WAD: liquidator receives collateral worth at least the repaid debt (up to 1 unit floor rounding on seizedAssets).
+    assert (seizedResult + 1) * price > repaidResult * ORACLE_PRICE_SCALE();
```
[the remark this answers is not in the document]
lilCertora: true

certora/specs/LiquidationProfitability.spec
```diff
@@ -0,0 +1,112 @@
...
+    bool maxLifReached = !isHealthy(obligation, id, borrower) || e.block.timestamp >= require_uint256(obligation.maturity + TIME_TO_MAX_LIF());
+
+    uint256 seizedResult;
+    uint256 repaidResult;
+    seizedResult, repaidResult = liquidate(e, obligation, collateralIndex, seizedAssets, 0, borrower, data);
+
+    mathint price = summaryPrice(obligation.collateralParams[collateralIndex].oracle);
+
+    // lif >= WAD: liquidator receives collateral worth at least the repaid debt (up to 1 unit ceil rounding on repaidUnits).
+    assert seizedResult * price > (repaidResult - 1) * ORACLE_PRICE_SCALE();
```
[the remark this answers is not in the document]
lilCertora: same

certora/specs/LiquidationBoundedByLIF.spec
```diff
@@ -0,0 +1,109 @@
...
+    collateralIndex < 128 => (collateral(id, user, collateralIndex) != 0 <=> summaryGetBit(currentContract.position[id][user].activatedCollaterals, collateralIndex));
+
+/// LIF BOUNDARIES ///
+
+/// Liquidation profit is bounded by maxLif (repaidUnits input).
+/// Unlike the seizedAssets rule, no requireInvariant is needed here: if collateralIndex is not in the bitmap because mulDivDown(..., 0) reverts.
+rule liquidationProfitBoundedInputRepaidUnits(env e, Midnight.Obligation obligation, uint256 collateralIndex, uint256 repaidUnits, address borrower, bytes data) {
+    mathint maxLif = obligation.collateralParams[collateralIndex].maxLif;
+    require data.length == 0, "no callback for prover performance";
+    require maxLif >= WAD(), "maxLif must be at least 1x for profit boundedness (see touchObligation validation and ExactMath.spec)";
```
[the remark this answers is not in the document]
lilCertora: Same reasoning

certora/specs/LiquidationBoundedByLIF.spec
```diff
@@ -0,0 +1,109 @@
...
+    mathint price = summaryPrice(obligation.collateralParams[collateralIndex].oracle);
+
+    assert seizedResult * price * WAD() <= repaidResult * ORACLE_PRICE_SCALE() * maxLif;
+}
+
+/// Liquidation profit is bounded by maxLif (seizedAssets input)
+rule liquidationProfitBoundedSeizedAssets(env e, Midnight.Obligation obligation, uint256 collateralIndex, uint256 seizedAssets, address borrower, bytes data) {
+    mathint maxLif = obligation.collateralParams[collateralIndex].maxLif;
+    require data.length == 0, "no callback for prover performance";
+    require maxLif >= WAD(), "maxLif must be at least 1x for profit boundedness (see touchObligation validation and ExactMath.spec)";
```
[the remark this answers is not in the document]
lilCertora: again

certora/specs/LiquidationProfitability.spec
QGarchery: Weird, because in the rule `liquidationIsProfitableRepaidUnits` we didn't need it

QGarchery (approved):

certora/specs/LiquidationProfitability.spec
```diff
@@ -0,0 +1,111 @@
...
+
+/// LIF CHARACTERIZATION ///
+
+/// For repaidUnits input: lif >= WAD (solvency), and lif == maxLif when borrower is unhealthy or >= 15 min post-maturity (profitability).
+rule liquidationLifRepaidUnits(env e, Midnight.Obligation obligation, uint256 collateralIndex, uint256 repaidUnits, address borrower, bytes data) {
+    uint256 maxLif = obligation.collateralParams[collateralIndex].maxLif;
+    require maxLif >= WAD(), "see the rule maxLifIsAtLeastWad";
+
+    bytes32 id = toId(e, obligation);
+    bool maxLifReached = !isHealthy(obligation, id, borrower) || e.block.timestamp >= require_uint256(obligation.maturity + TIME_TO_MAX_LIF());
```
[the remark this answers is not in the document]
lilCertora: already discussed

QGarchery (approved):

certora/confs/LiquidationBoundedByLIF.conf
```diff
@@ -0,0 +1,28 @@
...
```
MathisGD, line 18: why don't we verify all the rules?

edit: there are only these 2 rules, but then why do we have this?

lilCertora: we have also the strong invariant `nonZeroCollateralsAreActivated` already proven in CollateralBitmap.spec but we don't want to import all the summaries used by CollateralBitmap.spec

MathisGD: >/// Proven in CollateralBitmap.spec; assumed here via requireInvariant (not re-proven in this spec).

certora/specs/LiquidationProfitability.spec
```diff
@@ -0,0 +1,111 @@
...
+
+    uint256 seizedResult;
+    uint256 repaidResult;
+    seizedResult, repaidResult = liquidate(e, obligation, collateralIndex, 0, repaidUnits, borrower, receiver, callback, data);
+
+    mathint price = summaryPrice(obligation.collateralParams[collateralIndex].oracle);
+
+    // lif >= WAD: liquidator receives collateral worth at least the repaid debt (up to 1 unit floor rounding on seizedAssets).
```
MathisGD: we should precise "at the oracle price". otherwise could be a bit misleading

certora/specs/LiquidationProfitability.spec
```diff
@@ -0,0 +1,111 @@
...
+    assert repaidUnits > 0 => (seizedResult + 1) * price > repaidResult * ORACLE_PRICE_SCALE();
```
MathisGD: should be true too, no?
```suggestion
    assert (seizedResult + 1) * price > repaidResult * ORACLE_PRICE_SCALE();
```

lilCertora: when `repaidUnits == 0` , `seizedAssets = 0` [here](https://github.com/morpho-org/midnight/blob/76ab6c1115579514b6cdbcebb77eab2ddea5c359/src/Midnight.sol#L575) it causes `(seizedResult + 1) * price > repaidResult * ORACLE_PRICE_SCALE();` evaluating to `0 > 0` which is false, aditionaly the profitability statement is not meaningful if both `seizedResult` and `repaidResult` are 0

certora/specs/LiquidationProfitability.spec
```diff
@@ -0,0 +1,111 @@
...
+definition ORACLE_PRICE_SCALE() returns uint256 = 10 ^ 36;
+
+definition TIME_TO_MAX_LIF() returns uint256 = 900; // 15 min
+
+persistent ghost summaryPrice(address) returns uint256;
+
+// Axioms proven in MulDiv.spec (mulDivDownTightBound, mulDivDownRoundsDown).
+persistent ghost summaryMulDivDownM(uint256, uint256, uint256) returns uint256 {
+    axiom forall uint256 a. forall uint256 b. forall uint256 d. d > 0 => (summaryMulDivDownM(a, b, d) + 1) * d > a * b;
+    axiom forall uint256 a. forall uint256 b. forall uint256 d. d > 0 => summaryMulDivDownM(a, b, d) * d <= a * b;
```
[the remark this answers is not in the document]
lilCertora: true

certora/specs/LiquidationBoundedByLIF.spec
```diff
@@ -0,0 +1,109 @@
...
+persistent ghost summaryPrice(address) returns uint256;
+
+function summaryToId(Midnight.Obligation obligation) returns bytes32 {
+    return Utils.hashObligation(obligation);
+}
+
+// Axioms proven in MulDiv.spec (mulDivDownRoundsDown, mulDivDownTightBound).
+persistent ghost ghostMulDivDown(uint256, uint256, uint256) returns uint256 {
+    axiom forall uint256 a. forall uint256 b. forall uint256 d. d > 0 => ghostMulDivDown(a, b, d) * d <= a * b;
+    axiom forall uint256 a. forall uint256 b. forall uint256 d. d > 0 => (ghostMulDivDown(a, b, d) + 1) * d > a * b;
```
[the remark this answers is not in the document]
lilCertora: true

certora/specs/LiquidationProfitability.spec
MathisGD: we could use >= then, no?

not a big deal anyway

lilCertora: yes not a big deal, applied

certora/specs/LiquidationBoundedByLIF.spec
```diff
@@ -0,0 +1,113 @@
...
+
+/// Liquidation profit is bounded by maxLif (seizedAssets input)
+rule liquidationProfitBoundedSeizedAssets(env e, Midnight.Obligation obligation, uint256 collateralIndex, uint256 seizedAssets, address borrower, address receiver, address callback, bytes data) {
+    mathint maxLif = obligation.collateralParams[collateralIndex].maxLif;
+    require data.length == 0, "no callback for prover performance";
+    require maxLif >= WAD(), "maxLif must be at least 1x for profit boundedness (see touchObligation validation and ExactMath.spec)";
+
+    // Soundness: nonZeroCollateralsAreActivated is proven in CollateralBitmap.spec,
+    bytes32 id0 = summaryToId(obligation);
+    requireInvariant nonZeroCollateralsAreActivated(id0, borrower, collateralIndex);
```
MathisGD, line 104: what's the intuition for why this is needed?

if we remove it, we can merge both rules

certora/specs/LiquidationProfitability.spec
```diff
@@ -0,0 +1,115 @@
...
+    uint256 repaidResult;
+    seizedResult, repaidResult = liquidate(e, obligation, collateralIndex, seizedAssets, 0, borrower, receiver, callback, data);
+
+    mathint price = summaryPrice(obligation.collateralParams[collateralIndex].oracle);
+
+    // lif >= WAD: liquidator receives collateral worth at least the repaid debt (up to 1 unit ceil rounding on repaidUnits) at the oracle price.
+    assert seizedResult * price > (repaidResult - 1) * ORACLE_PRICE_SCALE();
+
+    // lif == maxLif when borrower is unhealthy or >= 15 min post-maturity: full liquidation incentive factor applies.
+    assert maxLifReached => seizedResult * price * WAD() + ORACLE_PRICE_SCALE() * WAD() > (repaidResult - 1) * maxLif * ORACLE_PRICE_SCALE();
```
MathisGD, line 114: have you considered merging both rules and proving only the worst case? at the end that's almost the only interesting thing

MathisGD (approved):

certora/specs/LiquidationBoundedByLIF.spec
lilCertora: we need it for the case where a borrower has collateral at `collateralIndex` but the bit is not set, the loop never visits that index, so `liquidatedCollatPrice` stays 0 : [prover link](https://prover.certora.com/output/7508195/cc0b2858188d47b28f734a4704232c69/?anonymousKey=b22c3ed3185f77132581dac86aff6f08a50c523e).

certora/specs/LiquidationProfitability.spec
lilCertora: yes but there are timeouts issues : https://prover.certora.com/output/7508195/45fbe0335ab54d37bc8a9c36ea1f3174/?anonymousKey=99c75b208e9388e0311ac159546f489bc96097cf

PR #498, opened 2026-03-10
[Certora] more robust consume spec

QGarchery:
Motivation is to not depend on the conversion from shares to units. Notably it fixes the issues that it breaks after #238, see [this run](https://prover.certora.com/output/66603/b3abc13c77f34201b9b33be1ad2c5d55?anonymousKey=d228c2943911ee17004620c5047df2af89762aff)

### DIFF
--- a/certora/specs/Consume.spec
+++ b/certora/specs/Consume.spec
@@ -81,7 +81,7 @@ rule takeConsumedAtMaxUnchanged(env e, uint256 obligationShares, address taker, 
     assert consumedBefore >= maxAmount => consumed(offer.maker, offer.group) == consumedBefore;
 }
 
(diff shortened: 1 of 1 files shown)


MathisGD (approved):

certora/specs/Consume.spec
bhargavbh:
```suggestion
/// For a fully consumed offer, take does not revert implies that the input to the take was zero. 
```

bhargavbh (approved):
LGTM. changed the description to better reflect what is being proved.

certora/specs/Consume.spec
MathisGD:
```suggestion
/// A fully-consumed offer in shares (rest. in units) only allows no-ops in shares (rest. in units).
```

PR #500, opened 2026-03-10
[no shares] loss index goes up

adhusson:


adhusson:
Closing as an index overflow blocks liquidation.

PR #507, opened 2026-03-12
[Certora] refactor fee boundaries

QGarchery:
- remove ghost & hooks
- remove mathint
- remove forall
- remove `maxFeeUnits` and use `maxTradingFee` directly instead
- add more requireInvariant
- add rule `tradingFeeAtBreakpoint`


lilCertora (approved):
looks good, this is better this way

PR #505, opened 2026-03-11
470 certora

MathisGD:
add rules for #470 and fix failing ones

PR #509, opened 2026-03-12
verif: mulDiv summary

MathisGD:
useful in #505

### DIFF
--- a/certora/helpers/MulDivHelper.sol
+++ b/certora/helpers/MulDivHelper.sol
@@ -1,0 +1,14 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity ^0.8.0;
+
+import {UtilsLib} from "../../src/libraries/UtilsLib.sol";
+
+contract MulDivHelper {
+    function mulDivDown(uint256 x, uint256 y, uint256 d) external pure returns (uint256) {
+        return UtilsLib.mulDivDown(x, y, d);
+    }
+
+    function mulDivUp(uint256 x, uint256 y, uint256 d) external pure returns (uint256) {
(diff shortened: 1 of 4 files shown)


certora/specs/MulDivSummary.spec
```diff
@@ -0,0 +1,45 @@
...
+
+function summaryMulDivUp(uint256 x, uint256 y, uint256 d) returns uint256 {
+    if (x == 0 || y == 0) return 0;
+    if (d > 0 && y == d) return x;
+    if (d > 0 && x == d) return y;
+    uint256 res;
+
+    // Exact ceil: res = ceil(x*y/d).
+    require to_mathint(res) * to_mathint(d) >= to_mathint(x) * to_mathint(y);
+    require res == 0 || (to_mathint(res) - 1) * to_mathint(d) < to_mathint(x) * to_mathint(y);
```
QGarchery, line 28: Tricky to see why adding `res == 0 ||` is safe. Also I don't see why it would be needed

certora/specs/MulDivSummary.spec
QGarchery, line 1: I think it makes more sense to base this PR on top of one that depends on it, otherwise we might be adding things that we don't need (it's unclear where it would be used by looking at this PR alone)

certora/helpers/MulDivHelper.sol
QGarchery, line 1: https://github.com/morpho-org/midnight/pull/388 has a very complete summarization of mulDiv. Would that be enough for the use case that you have in mind ? Also let's try to not duplicate the functionalities (#388 also defines a MulDiv contract that exposes the mulDiv functions)

PR #510, opened 2026-03-12
verif: accrue before touch debt

MathisGD:


certora/specs/DebtAccrual.spec
```diff
@@ -0,0 +1,94 @@
...
+    // Summary: replace accrueContinuousFee with a ghost flag update.
+    // The original function body is replaced, so its internal debt reads/writes do not fire hooks.
+    function accrueContinuousFee(bytes32 id, address borrower, uint256 maturity) internal => summaryAccrueContinuousFee(id, borrower);
+}
+
+/// GHOSTS ///
+
+/// Whether accrueContinuousFee was called for (id, user) in this transaction.
+persistent ghost mapping(bytes32 => mapping(address => bool)) accrued {
+    init_state axiom (forall bytes32 id. forall address user. accrued[id][user] == false);
```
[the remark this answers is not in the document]
MathisGD: not relevant in CVL

certora/specs/DebtAccrual.spec
```diff
@@ -0,0 +1,90 @@
...
+
+    function TickLib.tickToPrice(uint256) internal returns (uint256) => NONDET;
+    function TickLib.wExp(int256) internal returns (uint256) => NONDET;
+    function UtilsLib.isLeaf(bytes32, bytes32, bytes32[] memory) internal returns (bool) => NONDET;
+    function UtilsLib.msb(uint256) internal returns (uint256) => NONDET;
+    function UtilsLib.countBits(uint128) internal returns (uint256) => NONDET;
+
+    function IdLib.toId(Midnight.Obligation memory, uint256, address) internal returns (bytes32) => NONDET;
+
+    function isHealthy(Midnight.Obligation memory, bytes32, address) internal returns (bool) => NONDET;
```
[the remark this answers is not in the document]
MathisGD: was actually wondering how this could pass... nice finding!

certora/specs/DebtAccrual.spec
```diff
@@ -0,0 +1,88 @@
...
+
+    f(e, args);
+
+    assert !debtStoredBeforeAccrual[id][user], "debt was stored before accrueContinuousFee was called";
+}
+
+/// @title Debt is never loaded before accrueContinuousFee is called.
+/// Since accrueContinuousFee is summarized (its body does not execute), any SLOAD of debt
+/// comes from the calling function. If the flag is not set at that point, accrue was not called first.
+rule debtNotLoadedBeforeAccrual(env e, method f, calldataarg args, bytes32 id, address user) filtered { f -> f.selector != sig:isHealthy(Midnight.Obligation memory, bytes32, address).selector && f.selector != sig:debtOf(bytes32, address).selector && f.selector != sig:borrowerState(bytes32, address).selector } {
```
QGarchery: if `isHealthy` is internal as discussed here, then we can remove it from here. But I just checked and it seems like it will return an error if we forget to remove it from here ✔️

certora/specs/DebtAccrual.spec
```diff
@@ -0,0 +1,88 @@
...
+    // Summary: replace accrueContinuousFee with a ghost flag update.
+    // The original function body is replaced, so its internal debt reads/writes do not fire hooks.
+    function accrueContinuousFee(bytes32 id, address borrower, uint256 maturity) internal => summaryAccrueContinuousFee(id, borrower);
+}
+
+/// GHOSTS ///
+
+/// Whether accrueContinuousFee was called for (id, user) in this transaction.
+persistent ghost mapping(bytes32 => mapping(address => bool)) accrued {
+    init_state axiom (forall bytes32 id. forall address user. accrued[id][user] == false);
```
QGarchery:
```suggestion
```
this is only useful for invariants, because the init state is about the state after the deployment, not the arbitrary state at the start of the rules
-> same for the other ghosts

certora/specs/DebtAccrual.spec
```diff
@@ -0,0 +1,88 @@
...
+    if (!accrued[id][user]) {
+        debtLoadedBeforeAccrual[id][user] = true;
+    }
+}
+
+/// RULES ///
+
+/// @title Debt is never stored before accrueContinuousFee is called.
+/// Since accrueContinuousFee is summarized (its body does not execute), any SSTORE to debt
+/// comes from the calling function. If the flag is not set at that point, accrue was not called first.
```
QGarchery:
```suggestion
/// comes from the calling function.
```
(optional) it's a common pattern so I think we don't have to mention it
-> same below

certora/specs/DebtAccrual.spec
```diff
@@ -0,0 +1,88 @@
...
+
+    function TickLib.tickToPrice(uint256) internal returns (uint256) => NONDET;
+    function TickLib.wExp(int256) internal returns (uint256) => NONDET;
+    function UtilsLib.isLeaf(bytes32, bytes32, bytes32[] memory) internal returns (bool) => NONDET;
+    function UtilsLib.msb(uint256) internal returns (uint256) => NONDET;
+    function UtilsLib.countBits(uint128) internal returns (uint256) => NONDET;
+
+    function IdLib.toId(Midnight.Obligation memory, uint256, address) internal returns (bytes32) => NONDET;
+
+    // Summary: replace accrueContinuousFee with a ghost flag update.
```
QGarchery:
```suggestion
```

certora/specs/DebtAccrual.spec
```diff
@@ -0,0 +1,88 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+methods {
+    function multicall(bytes[]) external => HAVOC_ALL DELETE;
+
+    function _.price() external => NONDET;
+
+    function tradingFee(bytes32, uint256) internal returns (uint256) => NONDET;
+    function SafeTransferLib.safeTransferFrom(address, address, address, uint256) internal => NONDET;
+    function SafeTransferLib.safeTransfer(address, address, uint256) internal => NONDET;
```
QGarchery:
```suggestion
```
this should work because we have persistent ghosts. Otherwise we would need to explain why we make those assumptions (they could reenter)

certora/specs/DebtAccrual.spec
```diff
@@ -0,0 +1,88 @@
...
+
+hook Sload uint128 val borrowerState[KEY bytes32 id][KEY address user].debt {
+    if (!accrued[id][user]) {
+        debtLoadedBeforeAccrual[id][user] = true;
+    }
+}
+
+/// RULES ///
+
+/// @title Debt is never stored before accrueContinuousFee is called.
```
QGarchery:
```suggestion
/// Check that debt is never stored before accrueContinuousFee is called.
```
there is no natspec integrations for spec files (that I am aware of)
-> same for the other rule below

QGarchery (approved):

MathisGD:
we should add totalShares and totalUnits (otherwise you could be using an outdated share price)

PR #513, opened 2026-03-12
verif: faster created spec

MathisGD:
https://docs.certora.com/en/latest/docs/user-guide/out-of-resources/timeout.html#:~:text=This%20option%20enables%20some%20aggressive%20simplifications%20that%20speed%20up%20the%20Prover%20in%20many%20cases%2C%20but%20breaks%20call%20trace%20generation%20in%20case%20a%20rule%20is%20violated


MathisGD:
not worth for 5 min

PR #517, opened 2026-03-13
[bad debt refactor] events (on top of #423)

adhusson:


src/Midnight.sol
MathisGD, line 1: should we log the new index in liquidate too

adhusson: good point much more convenient

src/libraries/EventsLib.sol
```diff
@@ -52,8 +52,10 @@ library EventsLib {
         uint256 seizedAssets,
         uint256 repaidUnits,
         address indexed borrower,
-        uint256 badDebt
+        uint256 badDebt,
+        uint256 lossIndex
```
```suggestion
        uint256 newLossIndex
```

src/Midnight.sol
```diff
@@ -516,11 +518,14 @@ contract Midnight is IMidnight {
         uint256 lossIndex = obligationState[id].lossIndex;
         if (_userLossIndex != lossIndex) {
             int256 balance = balanceOf[id][user];
+            int256 newBalance = balance;
```
peyha: we could keep a single `balance` var no ?

src/libraries/EventsLib.sol
adhusson: it may not have changed (if bad debt is 0)

adhusson: but still you're right it's not clear, renaming and also in slash event

peyha (approved):

MathisGD:
@codex review

MathisGD:
I'm on the failing tests

MathisGD:
@codex review

PR #521, opened 2026-03-16
fix omit first line in Certora CI

QGarchery:
otherwise `jq` starts by reading the first line and consumes it from the input stream, so it would skip the first spec file


adhusson (approved):

MathisGD:
@codex review

MathisGD (approved):

PR #524, opened 2026-03-17
1-index

MathisGD:


src/Midnight.sol
```diff
@@ -387,12 +387,10 @@ contract Midnight is IMidnight {
...
 
         if (badDebt > 0) {
-            _position.balance += UtilsLib.toInt256(badDebt);
+            _position.balance += UtilsLib.toInt128(badDebt);
             uint256 oldTotalUnits = _obligationState.totalUnits;
             _obligationState.lossIndex = UtilsLib.toUint128(
-                type(uint128).max
-                    - (type(uint128).max - _obligationState.lossIndex)
-                    .mulDivDown(oldTotalUnits - badDebt, oldTotalUnits)
+                uint256(_obligationState.lossIndex).mulDivDown(oldTotalUnits - badDebt, oldTotalUnits)
```
MathisGD: fixes the rounding at the same time

src/interfaces/IMidnight.sol
```diff
@@ -50,7 +50,7 @@ struct ObligationState {
 }
 
 struct Position {
-    int256 balance;
+    int128 balance;
     uint128 lossIndex;
```
MathisGD: if we don't want to decrease the maxAmount
```suggestion
    int136 balance;
    uint120 lossIndex;
```

src/Midnight.sol
adhusson: isn't it exactly the same thing?

src/Midnight.sol
adhusson, line 35: should be updated

adhusson: not if https://github.com/morpho-org/midnight/pull/524#discussion_r2945329810

src/Midnight.sol
MathisGD: yes mb. and actually there is no pb

src/Midnight.sol
```diff
@@ -517,14 +516,16 @@ contract Midnight is IMidnight {
...
+            int136 balance = _position.balance;
             if (balance > 0) {
-                balance = UtilsLib.toInt256(
-                    // forge-lint: disable-next-line(unsafe-typecast) as balance > 0
-                    uint256(balance).mulDivDown(type(uint128).max - lossIndex, type(uint128).max - _userLossIndex)
+                balance = UtilsLib.toInt136(
+                    UtilsLib.toInt256(
+                        // forge-lint: disable-next-line(unsafe-typecast) as balance > 0
+                        uint256(uint136(balance)).mulDivDown(lossIndex, _userLossIndex)
+                    )
```
peyha, line 528:
```suggestion
                balance = UtilsLib.toInt136(
                        // forge-lint: disable-next-line(unsafe-typecast) as balance > 0
                        uint256(uint136(balance)).mulDivDown(lossIndex, _userLossIndex)
```
there should be no need to cast to int256

MathisGD:
will reopen after #527

PR #525, opened 2026-03-17
[Certora] debtOf doesn't read positive balances

QGarchery:
https://morpholabs.slack.com/archives/C08A40KKN9W/p1773685307524299


QGarchery:
closing as this is not needed after #527

PR #530, opened 2026-03-17
1-index

MathisGD:
only downside is that borrowers have to open one more slot..


src/Midnight.sol
adhusson: About useless slot opening see [532](https://github.com/morpho-org/midnight/pull/532/changes) that just avoids opening the slot  for nothing on take. It doesn't try to prevent wasting gas on slash, but it's ok.

MathisGD:
replaced by #600

PR #531, opened 2026-03-17
add testing to split balance change

adhusson:


test/LiquidationTest.sol
```diff
@@ -700,6 +700,84 @@ contract LiquidationTest is BaseTest {
...
+
+        midnight.slash(id, lender);
+
+        assertEq(midnight.creditOf(id, lender), creditAfterFirstSlash, "credit unchanged");
+        assertEq(midnight.userLossIndex(id, lender), lossIndexAfterFirstSlash, "loss index unchanged");
+    }
+
+    // full bad debt test.
+
+    function testFullBadDebtWithdrawCollateral(uint256 units) public {
```
MathisGD, line 759: we are missing slash non-full test, right?

adhusson: it was already there before the pr but badly named (testSlashEvent), will rename

MathisGD (approved):

PR #532, opened 2026-03-18
[1-loss-index] save slot on no credit take

adhusson:
invariant for this is `user credit > 0 -> user index >= global index`

### DIFF
--- a/src/Midnight.sol
+++ b/src/Midnight.sol
@@ -172,8 +172,6 @@ contract Midnight is IMidnight {
         require(UtilsLib.isLeaf(root, keccak256(abi.encode(offer)), proof), "invalid proof");
         require(offer.session == session[offer.maker], "invalid session");
         bytes32 id = touchObligation(offer.obligation);
-        slash(id, offer.maker);
-        slash(id, taker);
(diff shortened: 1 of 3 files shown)


src/Midnight.sol
```diff
@@ -221,10 +219,12 @@ contract Midnight is IMidnight {
         Position storage sellerPos = position[id][seller];
         uint256 oldBuyerDebt = buyerPos.debt;
         uint256 oldSellerDebt = sellerPos.debt;
-        uint256 buyerDebtReduction = UtilsLib.min(oldBuyerDebt, obligationUnits);
+        if (sellerPos.credit > 0) slash(id, seller);
```
MathisGD, line 222: A simpler approach could be
- always slash the buyer 
- slash the seller if they have non null credit

adhusson: slash buyer if its a borrower doing an exit means opening their slot for nothing

MathisGD: yes, it's the tradeoff (but simpler to reason about)

MathisGD:
replaced by #550

PR #533, opened 2026-03-18
[certora] RCF Properties (WIP)

bhargavbh:
- [x] RecoveryCloseFactor.spec: structural properties hence sufficient to summarise muldivs using ghost functions
-  RCF condition holds for pre-maturity liquidations. special case of rcfThreshold==0 also covered
-  if rcfThreshold is max_uint256, then RCF is deactivated. Uses CVL's `satisfy` for specifying.


certora/specs/RecoveryCloseFactor.spec
```diff
@@ -0,0 +1,156 @@
...
+    function activatedCollaterals(bytes32, address) external returns (uint128) envfree;
+    function isHealthy(Midnight.Obligation, bytes32, address) external returns (bool) envfree;
+    function _.price() external => CVL_price(calledContract) expect(uint256);
+
+    function IdLib.toId(Midnight.Obligation memory obligation, uint256 chainId, address midnight) internal returns (bytes32) => CVL_toId(obligation, chainId, midnight);
+    function UtilsLib.msb(uint256 bitmap) internal returns (uint256) => CVL_msb(bitmap);
+    function UtilsLib.mulDivDown(uint256 a, uint256 b, uint256 d) internal returns (uint256) => CVL_mulDivDown(a, b, d);
+    function UtilsLib.mulDivUp(uint256 a, uint256 b, uint256 d) internal returns (uint256) => CVL_mulDivUp(a, b, d);
+
+    // Summaries for external calls.
```
QGarchery, line 16: This comment is not really helpful, why can we use those summaries ?

certora/specs/RecoveryCloseFactor.spec
```diff
@@ -0,0 +1,156 @@
...
+
+definition ORACLE_PRICE_SCALE() returns uint256 = 10 ^ 36;
+
+// For any rcfThreshold, the RCF condition holds on all pre-maturity liquidations:
+// repaidUnits <= maxRepaid || collateralValue.zeroFloorSub(maxRepaid) < rcfThreshold.
+// As a special, shows that RCF is always active pre-maturity if rcfThreshold is zero.
+// Proven for obligations with a single collateral.
+rule liquidationRespectsRcfBound(env e, Midnight.Obligation obligation, uint256 seizedAssets, uint256 repaidUnits, address borrower, bytes data) {
+    require obligation.collaterals.length == 1, "we assume single collateral obligations";
+    uint256 collateralIndex = 0;
```
QGarchery, line 56:
```suggestion
```
we use `0` almost everywhere anyway

certora/specs/RecoveryCloseFactor.spec
```diff
@@ -0,0 +1,156 @@
...
+
+    uint256 lltv = obligation.collaterals[0].lltv;
+    uint256 maxLif = obligation.collaterals[0].maxLif;
+    uint256 lifTimesLltv = CVL_mulDivUp(maxLif, lltv, WAD());
+    require lifTimesLltv < WAD(), "See lifTimesLltvIsLessThanOrEqualToOne in ExactMath.spec";
```
QGarchery, line 61: In the rule it is proven that `lif * lltv <= 1`, and in fact we have `lif * lltv = 1` iff `lltv = 1`. Let's explain that liquidations don't make sense for `lltv = 1`

QGarchery: Actually, since we are calling `liquidate` pre-maturity, we can use probably and non-zero assets, we can remove this assumption I think

certora/specs/RecoveryCloseFactor.spec
```diff
@@ -0,0 +1,156 @@
...
+
+    // Read pre-call state via getters
+    bytes32 id;
+    uint256 collatBefore = collateralOf(id, borrower, 0);
+    uint256 debtBefore = debtOf(id, borrower);
+
+    // call setup
+    require collatBefore > 0, "call setup";
+    require activatedCollaterals(id, borrower) == 1, "call setup";
+    require CVL_msb(1) == 0, "call setup";
```
QGarchery, line 73: I don't get why those are "call setup". The "call setup" reason was originally there to describe the fact that in a liveness rule, if we want to show that it is possible to call a function, then we are free to choose the arguments of that function. But this is not a liveness rule here, and I feel like writing "call setup" makes us miss the reason why we make those assumptions.
The last 2 in the highlight correspond more to a simplification, and probably belong close to the first `require` about the fact that we consider only one collateral.
The first one I'm not sure why we need it or why it's safe

certora/specs/RecoveryCloseFactor.spec
```diff
@@ -0,0 +1,156 @@
...
+    uint256 collateralIndex = 0;
+
+    uint256 lltv = obligation.collaterals[0].lltv;
+    uint256 maxLif = obligation.collaterals[0].maxLif;
+    uint256 lifTimesLltv = CVL_mulDivUp(maxLif, lltv, WAD());
+    require lifTimesLltv < WAD(), "See lifTimesLltvIsLessThanOrEqualToOne in ExactMath.spec";
+
+    uint256 price = CVL_price(obligation.collaterals[0].oracle);
+
+    // Read pre-call state via getters
```
QGarchery, line 65:
```suggestion
```

certora/specs/RecoveryCloseFactor.spec
```diff
@@ -0,0 +1,156 @@
...
+
+    // call setup
+    require collatBefore > 0, "call setup";
+    require activatedCollaterals(id, borrower) == 1, "call setup";
+    require CVL_msb(1) == 0, "call setup";
+    require seizedAssets > 0 || repaidUnits > 0, "call setup";
+
+    require e.block.timestamp <= obligation.maturity, "assume liquidation pre-maturity";
+
+    // liquidate called.
```
QGarchery, line 78:
```suggestion
```

certora/specs/RecoveryCloseFactor.spec
```diff
@@ -0,0 +1,156 @@
...
+    uint256 effectiveDebt = assert_uint256(debtBefore - badDebt);
+
+    // Mirror maxRepaid: (effectiveDebt - maxDebt).mulDivUp(WAD, WAD - lif.mulDivUp(lltv, WAD))
+    uint256 _maxRepaid = CVL_mulDivUp(assert_uint256(effectiveDebt - _maxDebt), WAD(), assert_uint256(WAD() - lifTimesLltv));
+
+    // Mirror remaining collat computed to compare against rcfThreshold.
+    uint256 collatValueForRcf = CVL_mulDivDown(CVL_mulDivDown(collatBefore, price, ORACLE_PRICE_SCALE()), WAD(), maxLif);
+    uint256 remainingCollat = collatValueForRcf > _maxRepaid ? assert_uint256(collatValueForRcf - _maxRepaid) : 0;
+
+    assert actualRepaid <= _maxRepaid || remainingCollat < obligation.rcfThreshold, "RCF conditions must hold on all pre-maturity liquidations";
```
QGarchery, line 102: I'm not sold on why this rule is useful, as it mirrors very closely the code. It means we could do such a rule for every `require` we have. 
What would be very nice though, is a rule more high-level, that states "one cannot seize more than what is needed to put the position back to healthy". I expect it to be a bit difficult to formalize, but we can make some assumptions: for example only consider the fact that the threshold is 0

certora/specs/RecoveryCloseFactor.spec
```diff
@@ -0,0 +1,156 @@
...
+
+    //special case: if rcfThreshold is zero then RCF is always activated.
+    assert obligation.rcfThreshold != 0 || actualRepaid <= _maxRepaid, "rcfThreshold=0 must enforce repaidUnits <= maxRepaid on all pre-maturity liquidations";
```
QGarchery, line 105:
```suggestion
```
we can remove since it can easily be seen as a corollary of the assert above

certora/specs/RecoveryCloseFactor.spec
```diff
@@ -0,0 +1,156 @@
...
+    // Mirror bad-debt deduction
+    uint256 collatValueUp = CVL_mulDivUp(collatBefore, price, ORACLE_PRICE_SCALE());
+    uint256 maxSupportedDebt = CVL_mulDivUp(collatValueUp, WAD(), maxLif);
+    uint256 badDebt = debtBefore > maxSupportedDebt ? assert_uint256(debtBefore - maxSupportedDebt) : 0;
+    uint256 effectiveDebt = assert_uint256(debtBefore - badDebt);
+
+    // Mirror maxRepaid: (effectiveDebt - maxDebt).mulDivUp(WAD, WAD - lif.mulDivUp(lltv, WAD))
+    uint256 _maxRepaid = CVL_mulDivUp(assert_uint256(effectiveDebt - _maxDebt), WAD(), assert_uint256(WAD() - lifTimesLltv));
+
+    satisfy actualRepaid > _maxRepaid, "if rcfThreshold is max_uint256, RCF is not enforced";
```
QGarchery, line 155: The problem here is that it just finds a setting where the RCF threshold where the RCF is not enforced. So the naming "maxRcfThresholdNeverEnforcesRcf" is somewhat misleading.
In particular, the property also holds for any non-zero threshold. So with the following diff it also passes in [this run](https://prover.certora.com/output/66603/665a8dd847724e0f9cc874253e3bd893?anonymousKey=7389a66b0ce216f4dfae4705e587e62bc545bc65):
```diff
diff --git a/certora/specs/RecoveryCloseFactor.spec b/certora/specs/RecoveryCloseFactor.spec
index 4cf0680c..d708f67c 100644
--- a/certora/specs/RecoveryCloseFactor.spec
+++ b/certora/specs/RecoveryCloseFactor.spec
@@ -129,7 +129,7 @@ rule maxRcfThresholdNeverEnforcesRcf(env e, Midnight.Obligation obligation, uint
     require CVL_msb(1) == 0, "call setup";
     require seizedAssets > 0 || repaidUnits > 0, "call setup";
 
-    require obligation.rcfThreshold == max_uint256, "assume rcfThreshold is MAX UINT256";
+    require obligation.rcfThreshold == 1, "assume rcfThreshold is MAX UINT256";
     require e.block.timestamp <= obligation.maturity, "assume liquidation pre-maturity";
 
     // Liquidate called.
```

bhargavbh:
decided to prove properties on the exact bound of repaidUnits.

PR #534, opened 2026-03-18
[Certora] Spec for balance and lossIndex

jhoenicke:
Work in progress: Verification currently times out.

This adds a specification for the lossIndex.  The goal is to prove that the sum of all positive balanceOfAfterSlashing is less equal to totalUnits.

We prove as invariant that `(sumof balance/user.lossIndex*PRECISION) * obligation.lossIndex <= totalUnits*PRECISION` where the sum only ranges over the positive balances.  We keep the values `balance/user.lossIndex` as a fixed point number in the spec with a very large PRECISION value that avoids rounding.   The value is 0 if the balance is negative or the user.lossIndex is 0.  Think of PRECISION as being 2^128!, a number that is divisible by any positive uint128.   We leave PRECISION unspecified and only require it to be positive and divisible by the accessed user.lossIndex.  So the solver can choose any positive value that is divisible by any user lossIndex it sees.

There are currently timeouts related to static analysis; this is why the pull request is not ready yet.

Everything except withdraw, liquidate, and take is already verified, but these missing functions are the most important ones.


QGarchery:
closed in favor #1063

PR #546, opened 2026-03-19
[Certora] Formal verification of bitmap operations.

jhoenicke:
Note that this changes the implementation to encapsulate all bitmap functions in the UtilsLib library.  I hope this is an acceptable change.

The idea is that we can later use the summaries in `BitmapSummaries.spec` to verify properties that would otherwise require bit-precise reasoning.  To test if the summaries are powerful enough, I rechecked the collateral bitmap invariant using only the summary.

`Bitmap.spec` defines basic rules about setBit, clearBit, toggleBit and msb.
`BitmapSummaries.spec` summarizes these by non-deterministic functions and requires that they fulfill the properties proved in `Bitmap.spec`.

To discuss: 
* It may be cleaner to make these operations to work only on uint128 bitmaps.  The summary currently is written in a way that it reports errors if the bitmap exceeds uint128, but we may also want to write Bitmap.spec in this way.
* The `msb` function returns 2^256-1 if bitmap is zero, which is kind of odd behavior.  In the summary, I report an error if it is ever called with 0.  The current implementation never triggers this error.
* We may include the summaries later in other files.  But then it should only summarize the functions, not prove an invariant and require non-det summaries for other functions.
* `getBit` is currently unused in the implementation (it's needed in the verification, but we could just move it to `certora/helper/Utils.sol`)
* The `toggleBit` function is currently only used to clear bits.  It may be slightly more gas efficient.
* I haven't checked the gas overhead, but I believe it should be very small to none, depending on how good solidity is at inlining the functions.

[corresponding thread](https://morpholabs.slack.com/archives/C0ACSA8TNMA/p1774363431599649)


certora/specs/BitmapSummaries.spec
```diff
@@ -0,0 +1,71 @@
...
+    return result;
+}
+
+function summaryMsb(uint256 bitmap) returns (uint256) {
+    uint256 bit;
+    assert bitmap < 2 ^ 128 && bitmap != 0;
+
+    require bit < 128, "see Bitmap.spec";
+    require summaryGetBit(bitmap, bit), "see Bitmap.spec";
+    require forall uint256 otherbit. otherbit < 256 && summaryGetBit(bitmap, otherbit) => otherbit <= bit, "see Bitmap.spec";
```
lilCertora: what is the reason behind not using `otherbit < 128` here ?

src/Midnight.sol
```diff
@@ -384,7 +384,7 @@ contract Midnight is IMidnight {
             badDebt = badDebt.zeroFloorSub(
                 _collateralOf.mulDivUp(price, ORACLE_PRICE_SCALE).mulDivUp(WAD, _collateral.maxLif)
             );
-            bitmap ^= (1 << i);
+            bitmap = bitmap.toggleBit(i);
```
lilCertora: toggle is more efficient than clearBit and we're sure that this bit is set right ?

certora/specs/BitmapSummaries.spec
jhoenicke: It somehow happened, because I proved it for 256 bit first.  Yes, maybe we should make it 128 here also to make everything 128 bit.  Maybe we should also only prove the whole Bitmap.spec only for the 128 bit case.

src/Midnight.sol
jhoenicke: I think this is the idea.  I didn't want to change the code more than necessary.  We are talking about 3 gas I think, so it's not that important.

certora/specs/BitmapSummaries.spec
jhoenicke: In principle you could even leave out the `otherbit < 256` completely, but we are then making use of implementation details, like that shifting 1 by more than 256 bit to the left returns 0 and therefore getBit(bitmap, otherbit) always returns false if otherbit is out of range.

lilCertora: oh yes ok i see, thanks for clarification

certora/specs/BitmapSummaries.spec
QGarchery, line 1: > It may be cleaner to make these operations to work only on uint128 bitmaps. The summary currently is written in a way that it reports errors if the bitmap exceeds uint128, but we may also want to write Bitmap.spec in this way.

Yes let's do that !

src/libraries/UtilsLib.sol
QGarchery, line 1: > The msb function returns 2^256-1 if bitmap is zero, which is kind of odd behavior. In the summary, I report an error if it is ever called with 0. The current implementation never triggers this error.

We should document it, adding a comment on top of `msb`

QGarchery, line 1: > getBit is currently unused in the implementation (it's needed in the verification, but we could just move it to certora/helper/Utils.sol)

Yes let's move it if not necessary in the code

src/Midnight.sol
QGarchery: Let's remove the `toggleBit` function as well !

lilCertora (approved):

src/libraries/UtilsLib.sol
QGarchery: done in #560

certora/specs/BitmapSummaries.spec
QGarchery, line 1: The naming of the files is a bit confusing, shouldn't it be the other way around ?

QGarchery, line 1: Added my review as a PR: #560

jhoenicke: I originally intended BitmapSummaries to just contain the summaries.  But then added the property to check if the summaries are working.   Should we split this up into BitmapSummaries containing just part of the method block and the summaries and CollateralBitmap importing BitmapSummaries and containing the invariant and the nondet summaries?

MathisGD:
@codex review

certora/helpers/Utils.sol
```diff
@@ -4,14 +4,32 @@ pragma solidity ^0.8.0;
...
+        return UtilsLib.clearBit(bitmap, bit);
+    }
+
+    function msb(uint256 bitmap) external pure returns (uint256) {
+        return UtilsLib.msb(bitmap);
+    }
+
     function passiveFeeRecipient() external pure returns (address) {
         return PASSIVE_FEE_RECIPIENT;
+
```
MathisGD: (minor) we could setup forge fmt to also format here
```suggestion
```

certora/helpers/Utils.sol
```diff
@@ -4,14 +4,32 @@ pragma solidity ^0.8.0;
...
     }
```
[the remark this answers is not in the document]
MathisGD: duplicate

certora/specs/Bitmap.spec
```diff
@@ -0,0 +1,62 @@
...
+
+    bool otherBefore = getBit(bitmap, otherBit);
+
+    uint256 bitmapAfter = clearBit(bitmap, bit);
+    bool otherAfter = getBit(bitmapAfter, otherBit);
+    bool bitAfter = getBit(bitmapAfter, bit);
+
+    assert !bitAfter, "clearBit clears the bit";
+    assert otherBit != bit => otherBefore == otherAfter, "clearBit doesn't change other bits";
+    assert bitmapAfter < 2 ^ 128, "result fits in 128 bit";
```
MathisGD: (minor) stronger
```suggestion
    assert bitmapAfter < bitmap, "result fits in 128 bit";
```

certora/specs/Bitmap.spec
MathisGD, line 1: nb: we don't really verify that getBit works as expected, but we have `getBit(setBit())` == 1 and `getBit(clearBit())` == 0 which is enough

certora/specs/BitmapSummaries.spec
```diff
@@ -0,0 +1,59 @@
...
+    assert bitmap < 2 ^ 128 && bitmap != 0;
+
+    require bit < 128, "see Bitmap.spec";
+    require summaryGetBit(bitmap, bit), "see Bitmap.spec";
+    require forall uint256 otherbit. summaryGetBit(bitmap, otherbit) => otherbit <= bit, "see Bitmap.spec";
+    return bit;
+}
+
+strong invariant nonZeroCollateralsAreActivated(bytes32 id, address user, uint256 collateralIndex)
+    collateralIndex < 128 => (collateralOf(id, user, collateralIndex) != 0 <=> summaryGetBit(currentContract.position[id][user].activatedCollaterals, collateralIndex));
```
MathisGD, line 55: as collateralOf reverts for an index too big (it's an array, not a mapping)
```suggestion
    collateralOf(id, user, collateralIndex) != 0 <=> summaryGetBit(currentContract.position[id][user].activatedCollaterals, collateralIndex);
```

MathisGD (approved):

certora/specs/Bitmap.spec
QGarchery: Note also that the spec depends on `getBit` (we say that getBit is true iff collateral is non-zero), so it's not "just" a ghost variable

certora/specs/Bitmap.spec
QGarchery: We want 2^128 in this PR, because we use precisely this property in BitmapSummaries. Note that it will be removed with #560, but we can always add your spec (as another property that isn't used, but that's okay)

certora/specs/BitmapSummaries.spec
QGarchery: It's a bit tricky to define what an invariant means when the formula can revert. If you say that the property is false when it reverts, then if it goes from reverting to not reverting after a function call, then you are trivially proving the invariant (because the heredity condition is `require false`) when in fact it could not hold.
If you say that the property is true, then it's unexpected (and this wasn't your intuition here), and I think there are other issues.

Not sure what Certora is doing here actually, @jhoenicke knows more about it

certora/specs/BitmapSummaries.spec
QGarchery: Lets do that in another PR if it is needed, so we merge this one more quickly

certora/specs/Bitmap.spec
QGarchery: Actually it would be `bitmapAfter <= bitmap`, because you can clear a bit that was already unset. For symmetry we could add that `setBit` returns a bigger bitmap as well

QGarchery (approved):
Couple of (very) minor comments open, but it looks good !

certora/specs/BitmapSummaries.spec
jhoenicke: I think it is better to formulate the invariant in a way that it doesn't revert.

I'm not sure if we have fixed the meaning of a reverting invariant yet. But either it should have a fixed boolean value like false if it reverts, or it should be undefined, meaning the prover can chose the most evil value and report an error.  In both cases you need the implication for the invariant to be correct.

certora/specs/Bitmap.spec
jhoenicke: This is intended, as we don't want the solver to reason about bit-vectors.  We basically abstract from uint128 from being a bit-vector.  The prover can choose arbitrary abstract values, e.g., setBit(0, 1) can return 25.   But it is consistent in the way that then `getBit(25,1)` is true and for all other bits `getBit(25,otherBit)` is false.

certora/specs/Bitmap.spec
MathisGD: let's not add it then, agree

certora/specs/BitmapSummaries.spec
MathisGD: let's keep it like that then

PR #528, opened 2026-03-17
Scale obligation units by 1e6

peyha:

PR #547, opened 2026-03-19
lender side continuous fee

MathisGD:
#470 but lender side


src/Midnight.sol
```diff
@@ -558,6 +548,7 @@ contract Midnight is IMidnight {
         if (!obligationState[id].created) {
             require(obligation.collaterals.length > 0, "no collaterals");
             require(obligation.collaterals.length <= MAX_COLLATERALS, "too many collaterals");
+            require(obligation.maturity <= block.timestamp + 100 * 365 days, "ttm too high");
```
QGarchery: Not that it matters much, but curious why we need this `require`

src/Midnight.sol
```diff
@@ -247,20 +246,20 @@ contract Midnight is IMidnight {
...
-            buyerPos.pendingFee -= UtilsLib.toUint128(
-                uint256(buyerPos.pendingFee).mulDivUp(buyerDebtDecrease, oldBuyerDebt)
+        uint256 buyerCreditIncrease = UtilsLib.zeroFloorSub(units, buyerPos.debt);
+        uint256 sellerCreditDecrease = UtilsLib.min(units, sellerPos.credit);
+        buyerPos.debt -= UtilsLib.toUint128(units - buyerCreditIncrease);
+        buyerPos.pendingFee += UtilsLib.toUint128(
+            buyerCreditIncrease.mulDivDown(_obligationState.continuousFee * timeToMaturity, WAD)
+        );
+        buyerPos.credit += UtilsLib.toUint128(buyerCreditIncrease);
+        if (sellerPos.credit > 0 && sellerCreditDecrease > 0) {
```
```suggestion
        if (sellerCreditDecrease > 0) {
```
if `sellerCreditDecrease > 0` then `sellerPos.credit > 0`

src/Midnight.sol
```diff
@@ -588,9 +579,16 @@ contract Midnight is IMidnight {
         uint128 _userLossIndex = _position.lossIndex;
         uint128 lossIndex = obligationState[id].lossIndex;
         if (_userLossIndex != lossIndex) {
+            uint256 creditBefore = _position.credit;
             uint256 newCredit =
-                _position.credit.mulDivDown(type(uint128).max - lossIndex, type(uint128).max - _userLossIndex);
-            // forge-lint: disable-next-item(unsafe-typecast) as newCredit <= credits.
+                creditBefore.mulDivDown(type(uint128).max - lossIndex, type(uint128).max - _userLossIndex);
+            uint256 creditDecrease = creditBefore - newCredit;
+            if (creditBefore > 0 && creditDecrease > 0) {
```
```suggestion
            if (creditDecrease > 0) {
```
the following works too, it depends on the rounding direction when computing `newCredit`, but it could be nice to remove nesting and have the above condition in the `if` be `_userLossIndex != lossIndex && creditBefore > 0`

```suggestion
            if (creditBefore > 0) {
```

src/Midnight.sol
MathisGD: so that we can prove pending <= credit

src/Midnight.sol
```suggestion
    function _accrueContinuousFeeView(Obligation memory obligation, bytes32 id, address user)
```

src/Midnight.sol
MathisGD: indeed but I removed sellerCreditDecrease > 0 because we want to make it clear that we avoid the div by zero

src/Midnight.sol
MathisGD: if we remove the nesting, we have a problem with the event

src/Midnight.sol
MathisGD: what is your suggestions? (GitHub bug?)

src/Midnight.sol
```diff
@@ -588,9 +579,16 @@ contract Midnight is IMidnight {
...
+                _position.pendingFee -= UtilsLib.toUint128(
+                    uint256(_position.pendingFee).mulDivUp(creditDecrease, creditBefore)
+                );
```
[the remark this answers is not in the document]
MathisGD: this is expected

certora/specs/CreditAccrual.spec
```diff
@@ -0,0 +1,84 @@
...
+    function UtilsLib.msb(uint256) internal returns (uint256) => NONDET;
+    function UtilsLib.countBits(uint128) internal returns (uint256) => NONDET;
+
+    function IdLib.toId(Midnight.Obligation memory, uint256, address) internal returns (bytes32) => NONDET;
+
+    function accrueContinuousFee(Midnight.Obligation memory, bytes32 id, address user) internal => summaryAccrueContinuousFee(id, user);
+
+    // Summarize slash so that its credit writes do not fire the credit hooks below.
+    // Slash-before-credit correctness is verified separately in SlashBeforeBalance.spec.
+    function slash(bytes32 id, address user) internal => summarySlash(id, user);
```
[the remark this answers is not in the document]
MathisGD: this is expected

src/Midnight.sol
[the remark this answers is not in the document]
QGarchery: Fixing this would require to store the all the timestamps of the bad debt events...

We could say that it is not a problem, because the approximation is done in favor of the user, but this creates a weird incentive to not touch the position

src/Midnight.sol
QGarchery: yes github bug I think, I was suggesting to rename `borrower` into `user` in the `_accrueInterestView` function

certora/specs/CreatedObligations.spec
```diff
@@ -103,11 +103,42 @@ rule obligationIsCreatedAfterLiquidate(env e, Midnight.Obligation obligation, ui
...
+    !Midnight.obligationCreated(id) => userHasNoActivatedCollaterals(id, user);
+
+strong invariant obligationPendingFeeIsEmptyIfNotCreated(bytes32 id, address user)
+    !Midnight.obligationCreated(id) => userHasNoRemainingContinuousFee(id, user);
+
+strong invariant obligationLastContinuousFeeAccrualIsEmptyIfNotCreated(bytes32 id, address user)
+    !Midnight.obligationCreated(id) => userHasNoLastContinuousFeeAccrual(id, user);
+
+strong invariant obligationCollateralIsEmptyIfNotCreated(bytes32 id, address user)
+    !Midnight.obligationCreated(id) => userHasNoCollateral(id, user);
```
QGarchery: Now we can remove the `forall` in `userHasNoCollateral`
```suggestion
strong invariant obligationCollateralIsEmptyIfNotCreated(bytes32 id, address user, uint256 collateralIndex)
    !Midnight.obligationCreated(id) => userHasNoCollateral(id, user, collateralIndex);
```

src/Midnight.sol
```diff
@@ -724,26 +720,25 @@ contract Midnight is IMidnight {
     function accrueContinuousFee(Obligation memory obligation, address borrower) external {
         bytes32 id = IdLib.toId(obligation, block.chainid, address(this));
         require(obligationState[id].created, "not created");
+        slash(id, borrower);
         accrueContinuousFee(obligation, id, borrower);
```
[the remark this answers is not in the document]
MathisGD: indeed, will merge the view functions into a `slashAndAccrueView`

test/ContinuousFeeTest.sol
```diff
@@ -52,414 +52,455 @@ contract ContinuousFeeTest is BaseTest {
...
-        uint256 expectedRemaining =
-            pendingAfterBadDebt - pendingAfterBadDebt.mulDivUp(repaidUnits, debtAfterAccrual - badDebt);
 
         deal(address(loanToken), address(this), repaidUnits);
-        vm.expectEmit();
-        emit EventsLib.AccrueContinuousFee(id, borrower, feeUnits, remainingAfterAccrual);
         midnight.liquidate(obligation, 0, 0, repaidUnits, borrower, "");
 
-        assertApproxEqAbs(midnight.pendingFee(id, borrower), expectedRemaining, 1, "remaining after liquidation");
+        assertEq(midnight.pendingFee(id, borrower), 0, "borrower never has pending fee");
```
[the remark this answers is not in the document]
MathisGD: I'll remove these tests, they are useless

certora/specs/Midnight.spec
```diff
@@ -117,8 +161,33 @@ rule userLossIndexMonotonicallyIncreases(bytes32 id, address user, method f, env
...
+        preserved take(uint256 units, address taker, address takerCallback, bytes takerCallbackData, address receiverIfTakerIsSeller, Midnight.Offer offer, Midnight.Signature signature, bytes32 root, bytes32[] proof) with (env e) {
+            requireInvariant createdObligationsHaveBoundedTtm(e, offer.obligation);
+            requireInvariant pendingContinuousFeeBoundedByCredit(summaryToId(offer.obligation), offer.maker);
+            requireInvariant pendingContinuousFeeBoundedByCredit(summaryToId(offer.obligation), taker);
+            require currentContract.defaultContinuousFee[offer.obligation.loanToken] * MAX_TTM() <= 10 ^ 18;
+            require currentContract.obligationState[summaryToId(offer.obligation)].continuousFee * MAX_TTM() <= 10 ^ 18;
+        }
+    }
+
+strong invariant noRemainingContinuousFeeWithoutCredit(bytes32 id, address user)
```
QGarchery: this is a corollary of the invariant above, so it’s best to only use a rule that does `requireInvariant` and then asserts the statement, rather than to go through every function

certora/specs/Midnight.spec
```diff
@@ -117,8 +161,33 @@ rule userLossIndexMonotonicallyIncreases(bytes32 id, address user, method f, env
...
+
+strong invariant pendingContinuousFeeBoundedByCredit(bytes32 id, address user)
+    pendingFee(id, user) <= creditOf(id, user)
+    {
+        preserved take(uint256 units, address taker, address takerCallback, bytes takerCallbackData, address receiverIfTakerIsSeller, Midnight.Offer offer, Midnight.Signature signature, bytes32 root, bytes32[] proof) with (env e) {
+            requireInvariant createdObligationsHaveBoundedTtm(e, offer.obligation);
+            requireInvariant pendingContinuousFeeBoundedByCredit(summaryToId(offer.obligation), offer.maker);
+            requireInvariant pendingContinuousFeeBoundedByCredit(summaryToId(offer.obligation), taker);
+            require currentContract.defaultContinuousFee[offer.obligation.loanToken] * MAX_TTM() <= 10 ^ 18;
+            require currentContract.obligationState[summaryToId(offer.obligation)].continuousFee * MAX_TTM() <= 10 ^ 18;
```
QGarchery: I think we can prove those 2 `require`

src/Midnight.sol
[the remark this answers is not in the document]
MathisGD: not sure how to do actually

src/Midnight.sol
```diff
@@ -666,14 +662,14 @@ contract Midnight is IMidnight {
...
-    function lastContinuousFeeAccrual(bytes32 id, address user) external view returns (uint48) {
+    function lastContinuousFeeAccrual(bytes32 id, address user) external view returns (uint128) {
         return position[id][user].lastContinuousFeeAccrual;
     }
 
     /// @dev This function should be called with the id corresponding to the obligation.
     /// @dev This function does not call any oracle if debt is 0.
     /// @dev Expects the id to correspond to the obligation's id.
-    function isHealthy(Obligation memory obligation, bytes32 id, address borrower) internal view returns (bool) {
+    function isHealthy(Obligation memory obligation, bytes32 id, address borrower) public view returns (bool) {
```
adhusson, line 712: back to public because no accrual question I guess, but it takes both obligation and id input. So either we do an internal version of `isHealthy` or we undo the change to `_accrueContinuousFeeView` that removed the redundant argument API. Having both doesn't make sense.

certora/specs/Midnight.spec
```diff
@@ -117,8 +161,33 @@ rule userLossIndexMonotonicallyIncreases(bytes32 id, address user, method f, env
...
 
-strong invariant noRemainingContinuousFeeWithoutDebt(bytes32 id, address user)
-    debtOf(id, user) == 0 => pendingFee(id, user) == 0;
+strong invariant createdObligationsHaveBoundedTtm(env e, Midnight.Obligation obligation)
+    obligationIsCreated(obligation) => obligation.maturity <= e.block.timestamp + MAX_TTM()
+    {
+        preserved with (env e_f) {
+            require e.block.timestamp >= e_f.block.timestamp;
+        }
+    }
```
QGarchery: I don't understand what this invariant is proving, together with the `require` in the `preserved` block. It's clearly unsafe to `requireInvariant` it with any value, because you can chose 0 for the timestamp (and prove that you only have obligations that end in the 1970's 😅), so that makes me have some doubts.
Note that invariants are not checked against time passing, only against function calls

src/Midnight.sol
```diff
@@ -588,9 +579,14 @@ contract Midnight is IMidnight {
...
         uint128 lossIndex = obligationState[id].lossIndex;
         if (_userLossIndex != lossIndex) {
+            uint256 creditBefore = _position.credit;
             uint256 newCredit =
-                _position.credit.mulDivDown(type(uint128).max - lossIndex, type(uint128).max - _userLossIndex);
-            // forge-lint: disable-next-item(unsafe-typecast) as newCredit <= credits.
+                creditBefore.mulDivDown(type(uint128).max - lossIndex, type(uint128).max - _userLossIndex);
+            uint256 creditDecrease = creditBefore - newCredit;
+            if (creditBefore > 0) {
+                _position.pendingFee -= UtilsLib.toUint128(_position.pendingFee.mulDivUp(creditDecrease, creditBefore));
```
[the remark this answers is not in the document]
MathisGD: same as [here](https://github.com/morpho-org/midnight/pull/547#discussion_r2964831403)

src/Midnight.sol
```diff
@@ -247,20 +246,20 @@ contract Midnight is IMidnight {
...
-        sellerPos.pendingFee += UtilsLib.toUint128(
-            sellerDebtIncrease.mulDivDown(_obligationState.continuousFee * timeToMaturity, WAD)
-        );
-        sellerPos.credit -= UtilsLib.toUint128(units - sellerDebtIncrease);
-        sellerPos.debt += UtilsLib.toUint128(sellerDebtIncrease);
+        sellerPos.credit -= UtilsLib.toUint128(sellerCreditDecrease);
+        sellerPos.debt += UtilsLib.toUint128(units - sellerCreditDecrease);
         _obligationState.totalUnits = UtilsLib.toUint128(
             _obligationState.totalUnits - oldSellerDebt - oldBuyerDebt + sellerPos.debt + buyerPos.debt
         );
```
adhusson: `oldSellerDebt` and `oldBuyerDebt` are now only used here, it would remove those 2 local variables to just do

```
  _obligationState.totalUnits = UtilsLib.toUint128(
      _obligationState.totalUnits + buyerCreditIncrease - sellerCreditDecrease
  );
```

certora/specs/Midnight.spec
```diff
@@ -41,12 +56,41 @@ hook Sstore position[KEY bytes32 id][KEY address owner].debt uint128 newDebt (ui
...
+ghost ghostMulDivUp(uint256, uint256, uint256) returns uint256 {
+    // When the multiplier is at most the denominator, ceil(x * y / d) cannot exceed x.
+    axiom forall uint256 x. forall uint256 y. forall uint256 d. d > 0 && y <= d => ghostMulDivUp(x, y, d) <= x;
+
+    // x * d / d == x.
+    axiom forall uint256 x. forall uint256 d. d > 0 => ghostMulDivUp(x, d, d) == x;
+
+    // For x <= d and y <= d, ceil(x * y / d) is large enough that the remainder
+    // after subtracting it from x fits within the denominator remainder d - y.
+    axiom forall uint256 x. forall uint256 y. forall uint256 d. d > 0 && x <= d && y <= d => x - ghostMulDivUp(x, y, d) <= d - y;
```
QGarchery: The proof is interesting, it goes like this: `(d - x) (d - y)` is non-negative. Expand it and dividing by `d` gives the result. It's unclear why we need it though

src/Midnight.sol
```diff
@@ -312,8 +311,13 @@ contract Midnight is IMidnight {
         bytes32 id = touchObligation(obligation);
         ObligationState storage _obligationState = obligationState[id];
         slash(id, onBehalf);
+        accrueContinuousFee(obligation, id, onBehalf);
 
-        position[id][onBehalf].credit -= UtilsLib.toUint128(units);
+        Position storage _position = position[id][onBehalf];
+        if (_position.credit > 0 && units > 0) {
+            _position.pendingFee -= UtilsLib.toUint128(uint256(_position.pendingFee).mulDivUp(units, _position.credit));
```
```suggestion
            _position.pendingFee -= UtilsLib.toUint128(_position.pendingFee.mulDivUp(units, _position.credit));
```

src/Midnight.sol
```diff
@@ -312,8 +311,13 @@ contract Midnight is IMidnight {
         bytes32 id = touchObligation(obligation);
         ObligationState storage _obligationState = obligationState[id];
         slash(id, onBehalf);
+        accrueContinuousFee(obligation, id, onBehalf);
 
-        position[id][onBehalf].credit -= UtilsLib.toUint128(units);
+        Position storage _position = position[id][onBehalf];
+        if (_position.credit > 0 && units > 0) {
```
```suggestion
        if (_position.credit > 0) {
```

src/Midnight.sol
adhusson: why not just check `buyer.pendingFee <= buyer.credit` in take

MathisGD: that would be better for you?

certora/specs/Midnight.spec
MathisGD: I think that the LLM got lost a bit in all his assumptions. will clean everything and push

certora/specs/Midnight.spec
MathisGD: will be fixed by [this](https://github.com/morpho-org/midnight/pull/547#discussion_r2965112789)

src/Midnight.sol
MathisGD: convinced, will do

certora/specs/Midnight.spec
MathisGD: will be fixed by [this](https://github.com/morpho-org/midnight/pull/547#discussion_r2965112789)

src/Midnight.sol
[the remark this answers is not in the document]
adhusson: see thread https://morpholabs.slack.com/archives/C08A40KKN9W/p1774011741294809?thread_ts=1774011577.480279&cid=C08A40KKN9W not a real finding at least in the case where accrue is immdiately before slash

test/OtherFunctionsTest.sol
```diff
@@ -367,6 +367,21 @@ contract OtherFunctionsTest is BaseTest {
         midnight.touchObligation(_obligation);
     }
 
+    function testTtmTooHigh() public {
```
peyha:
```suggestion
    function testTtmTooHigh(uint256 ttm) public {
         ttm = bound(ttm, block.timestamp + 100 * 365 days + 1, type(uint256).max)

```

certora/specs/BalanceEffects.spec
```diff
@@ -73,11 +73,14 @@ rule repayEffects(env e, Midnight.Obligation obligation, uint256 units, address
 
 /// WITHDRAW ///
 
-/// withdraw decreases onBehalf's post-slash credit by exactly units, and only changes position[id][onBehalf].credit.
+/// When no fee accrual occurs, withdraw decreases onBehalf's post-slash credit by exactly units, and only changes position[id][onBehalf].credit.
 rule withdrawEffects(env e, Midnight.Obligation obligation, uint256 units, address onBehalf, address receiver, bytes32 anyId, address anyUser) {
     bytes32 id = toId(e, obligation);
     require userLossIndex(id, onBehalf) <= currentContract.obligationState[id].lossIndex, "see Midnight.spec";
 
+    // Exclude fee accrual effects: withdraw now calls accrueContinuousFee which decreases credit.
```
QGarchery: We can remove the require on `repay` now

QGarchery:
```suggestion
    // Exclude fee accrual effects.
```

src/Midnight.sol
adhusson: inconsistency still here

src/Midnight.sol
```diff
@@ -700,15 +705,34 @@ contract Midnight is IMidnight {
...
+        }
+        return accrueView(id, user, obligation.maturity);
     }
 
-    /// @dev Expects the id to correspond to the obligation's id.
-    function _accrueContinuousFeeView(bytes32 id, uint256 maturity, address borrower) internal view returns (uint256) {
-        Position storage _position = position[id][borrower];
+    function slashView(bytes32 id, address user) internal view returns (uint256) {
+        Position storage _position = position[id][user];
+        uint128 _userLossIndex = _position.lossIndex;
```
```suggestion
        uint128 userLossIndex = _position.lossIndex;
```

src/Midnight.sol
QGarchery, line 1: I don't care much that functions are categorized by view functions or not, but having the slash and accrue mechanism separated is quite annoying in this case

MathisGD: agree, will change that

src/Midnight.sol
```diff
@@ -582,21 +570,49 @@ contract Midnight is IMidnight {
         return id;
     }
 
-    function slash(bytes32 id, address user) public {
+    function slashAndAccrue(Obligation memory obligation, address user) external {
```
```suggestion
    function updatePosition(Obligation memory obligation, address user) external {
```
not a huge fan of this naming but no big deal

src/Midnight.sol
MathisGD: not sure what you mean, accrueContinuousFeeView is gone

src/Midnight.sol
```diff
@@ -696,49 +705,41 @@ contract Midnight is IMidnight {
...
-        bytes32 id = IdLib.toId(obligation, block.chainid, address(this));
-        require(obligationState[id].created, "not created");
-        slash(id, borrower);
-        accrueContinuousFee(obligation, id, borrower);
-    }
-
-    /// @dev Expects the obligation to be touched.
-    /// @dev Expects the id to correspond to the obligation's id.
-    function accrueContinuousFee(Obligation memory obligation, bytes32 id, address user) internal {
+    function accrueView(bytes32 id, address user, uint256 maturity) internal view returns (uint256) {
```
adhusson: returns uint128? (avoids casting after, but can also be implicitly cast to uint256 by callers if they do `uint256 x = accrueView(...)`)

src/Midnight.sol
```diff
@@ -696,49 +705,41 @@ contract Midnight is IMidnight {
...
+        Position storage _position = position[id][user];
+        uint256 creditBefore = _position.credit;
+        if (creditBefore > 0) {
+            uint256 creditAfter = slashView(id, user);
+            if (creditAfter < creditBefore) {
+                uint256 fee = accrueView(id, user, obligation.maturity);
+                return fee - fee.mulDivUp(creditBefore - creditAfter, creditBefore);
+            }
+        }
+        return accrueView(id, user, obligation.maturity);
```
adhusson: Actually not equivalent to slash then accrue, it calls accrue on the current state so rounding diff. Pushing a failing test for it.

src/Midnight.sol
```diff
@@ -700,15 +705,34 @@ contract Midnight is IMidnight {
         return tentativeSigner;
     }
 
-    /// @dev Returns the accrued fee.
-    function accrueContinuousFeeView(Obligation memory obligation, address borrower) public view returns (uint256) {
+    /// @dev Returns the accrued fee after simulating slash.
+    function slashAndAccrueView(Obligation memory obligation, address user) public view returns (uint256) {
```
QGarchery: Confused by this function:
- in case credit is 0, it returns `accrueView(id, user, obligation. maturity)`, but this will always return 0
- more importantly, the fee returned first computes the fee without slashing, and then applies slashing. So I'm guessing that it will not be precise up to roundings

src/Midnight.sol
MathisGD: just spotted too. I'm refactoring so that it's much more clearly correct

src/Midnight.sol
MathisGD: https://github.com/morpho-org/midnight/pull/547#discussion_r2966301883

src/Midnight.sol
```diff
@@ -582,21 +570,49 @@ contract Midnight is IMidnight {
...
-    function slash(bytes32 id, address user) public {
+    function slashAndAccrue(Obligation memory obligation, address user) external {
+        bytes32 id = IdLib.toId(obligation, block.chainid, address(this));
+        require(obligationState[id].created, "not created");
+        slash(id, user);
+        accrueContinuousFee(obligation, id, user);
+    }
+
+    function slash(bytes32 id, address user) internal {
         require(obligationState[id].created, "not created");
```
QGarchery, line 626:
```suggestion
```
since it's internal now

src/Midnight.sol
adhusson: ok will let you do the test

adhusson: note i had something like 

```
        uint256 creditBefore = midnight.creditOf(id, lender);
        uint128 userLossIndex = midnight.userLossIndex(id, lender);
        (, uint128 lossIndex,,,) = midnight.obligationState(id);
        uint256 creditAfterSlash = creditBefore.mulDivDown(type(uint128).max - lossIndex, type(uint128).max - userLossIndex);
        uint256 accruedFeeView = midnight.slashAndAccrueView(obligation, lender);
```

which shows it wasn't easy to use anyway, the function should return something like `expectedCredit` (credit after next slash;accrue) or `expectedCreditReduction` (total reduction to expect)

MathisGD: yes, did exactly this

src/Midnight.sol
```diff
@@ -245,26 +244,24 @@ contract Midnight is IMidnight {
...
+        uint256 buyerCreditIncrease = UtilsLib.zeroFloorSub(units, buyerPos.debt);
+        uint256 sellerCreditDecrease = UtilsLib.min(units, sellerPos.credit);
+        buyerPos.debt -= UtilsLib.toUint128(units - buyerCreditIncrease);
+        buyerPos.pendingFee += UtilsLib.toUint128(
+            buyerCreditIncrease.mulDivDown(_obligationState.continuousFee * timeToMaturity, WAD)
+        );
+        buyerPos.credit += UtilsLib.toUint128(buyerCreditIncrease);
+        if (sellerPos.credit > 0) {
+            sellerPos.pendingFee -= UtilsLib.toUint128(
+                uint256(sellerPos.pendingFee).mulDivUp(sellerCreditDecrease, sellerPos.credit)
```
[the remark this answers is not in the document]
MathisGD: this is solved by the precision #528

src/Midnight.sol
MathisGD: viable only if we do 551. let's see in there

src/Midnight.sol
```diff
@@ -700,15 +705,34 @@ contract Midnight is IMidnight {
...
+    function slashAndAccrueView(Obligation memory obligation, address user) public view returns (uint256) {
         bytes32 id = IdLib.toId(obligation, block.chainid, address(this));
-        return _accrueContinuousFeeView(id, obligation.maturity, borrower);
+        Position storage _position = position[id][user];
+        uint256 creditBefore = _position.credit;
+        if (creditBefore > 0) {
+            uint256 creditAfter = slashView(id, user);
+            if (creditAfter < creditBefore) {
+                uint256 fee = accrueView(id, user, obligation.maturity);
+                return fee - fee.mulDivUp(creditBefore - creditAfter, creditBefore);
```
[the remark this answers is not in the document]
MathisGD: solved

src/Midnight.sol
```diff
@@ -245,26 +244,24 @@ contract Midnight is IMidnight {
...
-        uint256 sellerDebtIncrease = units.zeroFloorSub(sellerPos.credit);
-        if (oldBuyerDebt > 0) {
-            buyerPos.pendingFee -= UtilsLib.toUint128(
-                uint256(buyerPos.pendingFee).mulDivUp(buyerDebtDecrease, oldBuyerDebt)
+        uint256 buyerCreditIncrease = UtilsLib.zeroFloorSub(units, buyerPos.debt);
+        uint256 sellerCreditDecrease = UtilsLib.min(units, sellerPos.credit);
+        buyerPos.debt -= UtilsLib.toUint128(units - buyerCreditIncrease);
+        buyerPos.pendingFee += UtilsLib.toUint128(
+            buyerCreditIncrease.mulDivDown(_obligationState.continuousFee * timeToMaturity, WAD)
+        );
```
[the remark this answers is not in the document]
MathisGD: this is ok

src/Midnight.sol
```suggestion
        emit EventsLib.Repay(msg.sender, id, units, onBehalf);
```

```suggestion
        emit EventsLib.Withdraw(msg.sender, id, units, onBehalf, receiver, _position.pendingFee);
```

```suggestion
```

src/Midnight.sol
```diff
@@ -582,22 +559,60 @@ contract Midnight is IMidnight {
...
+    function _updatePosition(Obligation memory obligation, bytes32 id, address user) internal {
+        Position storage _position = position[id][user];
+        uint128 lossIndex = obligationState[id].lossIndex;
+        (uint128 newCredit, uint128 newPending, uint128 accruedFee) = updatePositionView(obligation, user);
+        if (_position.lossIndex != lossIndex) {
             _position.lossIndex = lossIndex;
-            emit EventsLib.Slash(msg.sender, id, user, newCredit, lossIndex);
         }
+        if (accruedFee > 0) {
+            position[id][PASSIVE_FEE_RECIPIENT].credit += accruedFee;
```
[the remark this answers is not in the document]
MathisGD: this is on purpose. but nice catch. 

will document

test/ContinuousFeeTest.sol
```diff
@@ -39,588 +38,319 @@ contract ContinuousFeeTest is BaseTest {
...
+        uint256 accruedFee = pendingAfterSlash.mulDivDown(elapsed2, ttm - elapsed1);
 
-        vm.warp(block.timestamp + elapsed);
+        midnight.updatePosition(obligation, lender);
 
-        vm.expectRevert();
-        midnight.repay(obligation, 0, borrower);
+        assertEq(midnight.creditOf(id, lender), creditAfterSlash - accruedFee, "credit after slash and accrual");
+        assertApproxEqAbs(
+            midnight.pendingFee(id, lender), pendingAfterSlash - accruedFee, 1, "remaining after slash and accrual"
```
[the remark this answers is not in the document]
MathisGD: same

test/BaseTest.sol
```diff
@@ -234,6 +234,7 @@ abstract contract BaseTest is Test {
         }
         collaterals = sortCollaterals(collaterals);
         obligation.collaterals = collaterals;
+        obligation.maturity = min(obligation.maturity, block.timestamp + 100 * 365 days);
```
peyha:
```suggestion
```

src/Midnight.sol
adhusson: when i wrote this i'm pretty sure there was 1 viewer that took `obligation,id` and one that did not.

Now both `isHealthy` and `updatePositionView` take both an obligation and an id, which does not go in my preferred direction obviously but it's consistent at least

certora/specs/Midnight.spec
```diff
@@ -8,16 +11,41 @@ methods {
...
+    function Utils.hashObligation(Midnight.Obligation) external returns (bytes32) envfree;
 
     function _.price() external => NONDET;
-    function IdLib.toId(Midnight.Obligation memory, uint256, address) internal returns (bytes32) => NONDET;
+    function IdLib.toId(Midnight.Obligation memory obligation, uint256, address) internal returns (bytes32) => summaryToId(obligation);
+
+    function tradingFee(bytes32, uint256) internal returns (uint256) => NONDET;
+    function isHealthy(Midnight.Obligation memory, bytes32, address) internal returns (bool) => NONDET;
+    function SafeTransferLib.safeTransferFrom(address, address, address, uint256) internal => NONDET;
+    function SafeTransferLib.safeTransfer(address, address, uint256) internal => NONDET;
```
QGarchery, line 30: Let's add a comment to say that the tokens are assumed to not reenter

certora/specs/OnlyAuthorizedCanChange.spec
```diff
@@ -32,6 +24,10 @@ methods {
 
 /// HELPERS ///
 
+definition noAccrual(env e, bytes32 id, address borrower) returns bool = currentContract.position[id][borrower].pendingFee == 0 || e.block.timestamp == currentContract.position[id][borrower].lastContinuousFeeAccrual;
+
+use invariant noRemainingContinuousFeeWithoutCredit;
+
```
QGarchery:
```suggestion
```
unused here

src/Midnight.sol
```diff
@@ -288,16 +318,24 @@ contract Midnight is IMidnight {
...
+        );
         bytes32 id = touchObligation(obligation);
         ObligationState storage _obligationState = obligationState[id];
-        slash(id, onBehalf);
+        _updatePosition(obligation, id, onBehalf);
 
-        position[id][onBehalf].credit -= UtilsLib.toUint128(units);
+        Position storage _position = position[id][onBehalf];
+        if (_position.credit > 0) {
+            _position.pendingFee -= UtilsLib.toUint128(_position.pendingFee.mulDivUp(units, _position.credit));
```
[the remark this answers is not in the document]
MathisGD: this will be fixed by #528

src/Midnight.sol
```suggestion
```
not used anymore ? weird that the linter is not catching this

peyha, line 50: should we write some comment on the continuous fee ?

src/Midnight.sol
```diff
@@ -306,7 +344,8 @@ contract Midnight is IMidnight {
         require(onBehalf == msg.sender || isAuthorized[onBehalf][msg.sender], "unauthorized");
         bytes32 id = touchObligation(obligation);
 
-        position[id][onBehalf].debt -= UtilsLib.toUint128(units);
+        Position storage _position = position[id][onBehalf];
+        _position.debt -= UtilsLib.toUint128(units);
```
```suggestion
position[id][onBehalf].debt -= UtilsLib.toUint128(units);
```
probably due to fork but previous version was cleaner

src/Midnight.sol
```diff
@@ -535,68 +574,98 @@ contract Midnight is IMidnight {
...
+    /// SLASHING AND CONTINUOUS FEE ACCRUAL ///
+
+    /// @dev Expects the id to correspond to the obligation's id.
+    /// @dev Returns the new credit, new pending fee, and accrued fee after simulating updatePosition.
+    function updatePositionView(Obligation memory obligation, bytes32 id, address user)
+        public
+        view
+        returns (uint128, uint128, uint128)
+    {
+        Position memory _position = position[id][user];
```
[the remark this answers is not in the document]
QGarchery: `memory` -> `storage`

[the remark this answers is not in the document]
MathisGD: but then we read from storage the values multiples times?

src/Midnight.sol
```diff
@@ -535,68 +574,98 @@ contract Midnight is IMidnight {
...
+        uint256 postSlashCredit = _position.lossIndex < type(uint128).max
+            ? _position.credit
+                .mulDivDown(type(uint128).max - obligationState[id].lossIndex, type(uint128).max - _position.lossIndex)
+            : 0;
```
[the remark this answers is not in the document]
QGarchery: Actually the new behavior is great, and now we always have that the sum of credit after update is lower than the total units 💯

peyha (approved):
lgtm

src/Midnight.sol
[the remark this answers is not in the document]
MathisGD: because we didn't have it before?

src/Midnight.sol
```diff
@@ -535,68 +584,98 @@ contract Midnight is IMidnight {
...
-            _position.credit = uint128(newCredit);
-            _position.lossIndex = lossIndex;
-            emit EventsLib.Slash(msg.sender, id, user, newCredit, lossIndex);
-        }
+        (uint128 newCredit, uint128 newPending, uint128 accruedFee) = updatePositionView(obligation, id, user);
+
+        _position.credit = newCredit;
+        _position.lossIndex = lossIndex;
+        _position.pendingFee = newPending;
+        _position.lastContinuousFeeAccrual = uint128(block.timestamp);
```
```suggestion
        _position.pendingFee = newPending;
        
        _position.lossIndex = obligationState[id].lossIndex;
        _position.lastContinuousFeeAccrual = uint128(block.timestamp);
```
(style) and remove the definition of `lossIndex` above

src/Midnight.sol
```diff
@@ -535,68 +584,98 @@ contract Midnight is IMidnight {
...
             emit EventsLib.ObligationCreated(id, obligation);
         }
         return id;
     }
 
-    function slash(bytes32 id, address user) public {
+    /// SLASHING AND CONTINUOUS FEE ACCRUAL ///
+
+    /// @dev Expects the id to correspond to the obligation's id.
+    /// @dev Returns the new credit, new pending fee, and accrued fee after simulating updatePosition.
```
```suggestion
    /// @dev Returns the new credit, new pending fee, and accrued fee after having updated the position.
```
because `updatePosition` uses `updatePositionView`, so it's a bit unclear what it would mean for `updatePositionView` to simulate `updatePosition`

src/Midnight.sol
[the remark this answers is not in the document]
MathisGD: oh but now we read the 128 slots all the time, crazy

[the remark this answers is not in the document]
QGarchery: Maybe not a huge deal since it will be warm. I guess we can cache those variables. If we want to go further there is also `obligationState[id].lossIndex` that is read twice between `_updatePositionView` and `updatePosition`

src/Midnight.sol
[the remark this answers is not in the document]
QGarchery: No, before I think we were bypassing the slashing if the loss index was already maxed out

src/Midnight.sol
```diff
@@ -535,68 +584,97 @@ contract Midnight is IMidnight {
...
+        returns (uint128, uint128, uint128)
+    {
+        Position storage _position = position[id][user];
+        uint128 credit = _position.credit;
+        uint128 lossIndex = _position.lossIndex;
+        uint256 postSlashCredit = lossIndex < type(uint128).max
+            ? credit.mulDivDown(type(uint128).max - obligationState[id].lossIndex, type(uint128).max - lossIndex)
+            : 0;
+        uint128 _pendingFee = _position.pendingFee;
+        uint256 postSlashPending = credit > 0 ? _pendingFee - _pendingFee.mulDivUp(credit - postSlashCredit, credit) : 0;
```
QGarchery, line 612:
```suggestion
        uint256 postSlashPending = credit > 0 ? _pendingFee.mulDivDown(postSlashCredit, credit) : 0;
```
this gives the same result. Not sure which is best between the current code and the one where we do this change + rewrite the other decrease of the pending fee with this idea

certora/specs/BalanceEffects.spec
```diff
@@ -35,24 +32,14 @@ methods {
...
-function summaryMulDiv(uint256 x, uint256 y, uint256 d) returns uint256 {
-    if (x == 0 || y == 0) return 0;
-    return ghostMulDiv(x, y, d);
-}
+definition noSlash(bytes32 id, address user) returns bool = currentContract.position[id][user].lossIndex == currentContract.obligationState[id].lossIndex;
 
 /// REPAY ///
 
-/// repay decreases onBehalf's debt by exactly units and only changes position[id][onBehalf].debt.
+/// When no fee accrual occurs, repay decreases onBehalf's debt by exactly units and only changes position[id][onBehalf].debt
```
QGarchery:
```suggestion
/// repay decreases onBehalf's debt by exactly units and only changes position[id][onBehalf].debt
```

certora/specs/Midnight.spec
```diff
@@ -28,8 +58,11 @@ hook Sstore position[KEY bytes32 id][KEY address owner].debt uint128 newDebt (ui
 
 function summaryMulDiv(uint256 x, uint256 y, uint256 d) returns uint256 {
     if (x == 0 || y == 0) return 0;
+    if (d > 0 && y == d) return x;
+    if (d > 0 && x == d) return y;
     uint256 res;
     require y > d || res <= x;
```
QGarchery:
```suggestion
    require y <= d => res <= x;
```
(style)

certora/specs/Midnight.spec
```diff
@@ -28,8 +58,11 @@ hook Sstore position[KEY bytes32 id][KEY address owner].debt uint128 newDebt (ui
...
+    require y > d || x > d || x - res <= d - y;
```
QGarchery: still unclear why we need this assumption, but it's true (I noted to verify it later)

certora/specs/UpdateBeforeCredit.spec
```diff
@@ -0,0 +1,76 @@
...
+
+    f(e, args);
+
+    assert !creditStoredBeforeUpdate[id][user], "credit was stored before _updatePosition was called";
+}
+
+/// Check that credit is never loaded before _updatePosition is called.
+/// The SLOADs of _updatePosition are ignored (see summary above).
+rule creditNotLoadedBeforeUpdate(env e, method f, calldataarg args, bytes32 id, address user) filtered { f -> f.selector != sig:creditOf(bytes32, address).selector && f.selector != sig:updatePositionView(Midnight.Obligation, bytes32, address).selector } {
+    require !updated[id][user], "initialize the ghost variable";
```
QGarchery:
```suggestion
```
we could remove it (same for the rule above), but maybe it's less clear

src/Midnight.sol
MathisGD: nb: The other change will not be as simple

QGarchery (approved):

certora/specs/BalanceEffects.spec
```diff
@@ -17,10 +16,8 @@ methods {
     function UtilsLib.isLeaf(bytes32, bytes32, bytes32[] memory) internal returns (bool) => NONDET;
     function UtilsLib.msb(uint256) internal returns (uint256) => NONDET;
     function TickLib.tickToPrice(uint256) internal returns (uint256) => NONDET;
-    function isHealthy(Midnight.Obligation memory, bytes32, address) internal returns (bool) => NONDET;
-    function tradingFee(bytes32, uint256) internal returns (uint256) => NONDET;
-    function UtilsLib.mulDivDown(uint256 x, uint256 y, uint256 d) internal returns (uint256) => summaryMulDiv(x, y, d);
-    function UtilsLib.mulDivUp(uint256 x, uint256 y, uint256 d) internal returns (uint256) => summaryMulDiv(x, y, d);
+    function UtilsLib.mulDivDown(uint256 x, uint256 y, uint256 d) internal returns (uint256) => NONDET;
+    function UtilsLib.mulDivUp(uint256 x, uint256 y, uint256 d) internal returns (uint256) => NONDET;
```
QGarchery: You can't have something non-deterministic in this rule, because you want to be able to predict what will be the value after the call

certora/specs/Midnight.spec
QGarchery:
```suggestion
    require y <= d && x <= d => x - res <= d - y;
```

MathisGD: i removed it

src/Midnight.sol
```diff
@@ -180,8 +216,8 @@ contract Midnight is IMidnight {
         require(UtilsLib.isLeaf(root, keccak256(abi.encode(offer)), proof), "invalid proof");
         require(offer.session == session[offer.maker], "invalid session");
         bytes32 id = touchObligation(offer.obligation);
-        slash(id, offer.maker);
-        slash(id, taker);
+        _updatePosition(offer.obligation, id, offer.maker);
+        _updatePosition(offer.obligation, id, taker);
```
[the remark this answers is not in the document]
MathisGD: indeed, but wont fix

certora/specs/BalanceEffects.spec
MathisGD: I think that it's not needed anymore (I assume noSlash noAccrue now)

src/periphery/TakeBundler.sol
```diff
@@ -70,6 +70,7 @@ contract TakeBundler {
     /// @dev Not usable if buyerPrice > WAD, because not all buyerAssets are reachable then.
     /// @dev buyerAssetsToUnits is evaluated before midnight.take, so reverts there (e.g. underflow when offerPrice <
     /// tradingFee) are not caught by the try/catch and will abort the bundle.
+    /// @dev Requires a non-empty takes array.
     function bundleTakeBuyerAssets(
```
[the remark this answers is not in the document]
MathisGD: it's documented, no need to do more

src/Midnight.sol
adhusson: prefer to keep the "fee -=" style everywhere at the cost of more complex here but not a strong opinion at all

MathisGD: slightly like @adhusson, let's keep like that then

certora/specs/Midnight.spec
```diff
@@ -59,12 +59,11 @@ hook Sstore position[KEY bytes32 id][KEY address owner].debt uint128 newDebt (ui
...
-    if (x == 0 || y == 0) return 0;
-    if (d > 0 && y == d) return x;
-    if (d > 0 && x == d) return y;
-    uint256 res;
-    require y <= d => res <= x;
-    return res;
+    uint256 r;
+    require x == 0 => r == 0;
+    require d > 0 && y <= d => r <= x;
+    require d > 0 && x <= d && y <= d => x - r <= d - y;
```
adhusson, line 65: Issue was in `postSlashPending = pending - pending.mulDivUp(credit - postSlashCredit, credit)`.

So in mulDivUp we need to know that when `pending <= credit` and  `credit - postSlashCredit <= credit` then `postSlashPending = pending - pending.mulDivUp(...) <= credit - (credit - postSlashCredit) = postSlashCredit` otherwise the invariant is not maintained. The line is also true for muldivdown so a single summary is ok.

adhusson (approved):

PR #548, opened 2026-03-19
certora: no div by zero

MathisGD:
thought that was interesting. courtesy of Opus 4.6


certora/specs/NoDivisionByZero.spec
```diff
@@ -0,0 +1,149 @@
...
+//   cursor >= WAD could cause a division by zero, but this would simply revert (Solidity 0.8
+//   checked arithmetic). No state change occurs.
+//
+// - liquidate: Excluded because it requires a combination of precise summaries that NONDET cannot
+//   provide. Specifically:
+//   (a) UtilsLib.msb is NONDET, so the prover can construct paths where the liquidated collateral
+//       index is not visited in the bitmap loop, leaving liquidatedCollatPrice = 0. In reality,
+//       msb returns the MSB of the bitmap, ensuring all activated collateral indices are visited.
+//       Calling liquidate on an inactive collateral index is an invalid input that reverts via
+//       Solidity's checked arithmetic (division by zero or underflow).
```
MathisGD: @jhoenicke I'm thinking that we could be able to fix this

certora/specs/NoDivisionByZero.spec
```diff
@@ -0,0 +1,149 @@
...
+//   (b) IdLib.toId is NONDET, so the prover cannot infer that calldata obligation fields match the
+//       stored obligation. In reality, toId is a deterministic hash, so different obligation content
+//       produces different ids.
```
MathisGD: here we should copy what was done in #388

certora/specs/NoDivisionByZero.spec
```diff
@@ -0,0 +1,149 @@
...
+//   (c) The recovery close factor divisor WAD - ceil(lif*lltv/WAD) requires proving that
+//       lif*lltv < WAD^2, which depends on the maxLif formula and involves nonlinear arithmetic.
```
MathisGD: we should be able to prove that too

jhoenicke:
I think there is one scenario where liquidate() divides by zero in practice:  if you liquidate a non-activated collateral with repaidUnits > 0, the function divides by liquidatedCollatPrice, which is never set.

The liquidate function never explicitly checks that the collateralIndex is activated.  Of course, if you liquidate a non-zero amount, the subtraction `uint128 newCollateralOf = _position.collateral[collateralIndex] - UtilsLib.toUint128(seizedAssets);` would underflow.

MathisGD:
then it would be a reverting run, so it wouldn't be considered by the prover, no?

certora/specs/NoDivisionByZero.spec
jhoenicke: @QGarchery already did this in lifTimesLltvIsLessThanOrEqualToOne in ExactMath (with `<=`).

jhoenicke:
> then it would be a reverting run, so it wouldn't be considered by the prover, no?

yes, since you check the div by zero flag at the end, you should not see this error in the prover.  If you would assert directly in the mulDiv function the error would be reported.

certora/specs/NoDivisionByZero.spec
MathisGD: added a comment

certora/specs/NoDivisionByZero.spec
```diff
@@ -0,0 +1,170 @@
...
+
+    // Single collateral (SMT solver can't handle multi-iteration bitmap loops).
+    require obligation.collaterals.length == 1;
+    require collateralIndex == 0;
+
+    // Sound: touchObligation enforces maxLif >= WAD and maxLif*lltv <= WAD^2 (ExactMath.spec).
+    require obligation.collaterals[0].maxLif >= WAD();
+
+    // Ensures recovery close factor divisor WAD - ceil(lif*lltv/WAD) > 0; tight at lltv = WAD.
+    require to_mathint(obligation.collaterals[0].maxLif) * to_mathint(obligation.collaterals[0].lltv) <= to_mathint(WAD()) * (to_mathint(WAD()) - 1);
```
jhoenicke: ExactMath can only prove <= WAD()*WAD().  You explicitly allow lltv = LLTV_8 (WAD) and in this case maxLif = WAD().

So liquidate() always reverts if lltv = LLTV_8 and maturity not yet reached?
Is this intentional?

certora/specs/NoDivisionByZero.spec
```diff
@@ -0,0 +1,170 @@
...
+//
+// The liquidate function is verified in a separate rule (noDivisionByZeroLiquidate), restricted
+// to a single collateral due to SMT solver limitations with multi-iteration bitmap loops.
+// The toId summary follows the approach from PR #388: a ghost-backed deterministic function.
+// The msb summary uses a bitwise constraint (bitmap >> bit == 1).
+
+methods {
+    function multicall(bytes[]) external => HAVOC_ALL DELETE;
+
+    function _.price() external => ghostPrice() expect(uint256);
```
jhoenicke: price depends on oracle, so this should be
```suggestion
    function _.price() external => ghostPrice(calledContract) expect(uint256);
```

certora/specs/NoDivisionByZero.spec
```diff
@@ -0,0 +1,170 @@
...
+}
+
+// Follows from totalUnitsEqualsSumNegativeDebtPlusWithdrawable in Midnight.spec.
+hook Sload uint128 value obligationState[KEY bytes32 id].totalUnits {
+    require value > 0;
+}
+
+/// SUMMARIES ///
+
+ghost ghostPrice() returns uint256;
```
jhoenicke:
```suggestion
ghost ghostPrice(address) returns uint256;
```

certora/specs/NoDivisionByZero.spec
```diff
@@ -0,0 +1,170 @@
...
+    require collateralIndex == 0;
+
+    // Sound: touchObligation enforces maxLif >= WAD and maxLif*lltv <= WAD^2 (ExactMath.spec).
+    require obligation.collaterals[0].maxLif >= WAD();
+
+    // Ensures recovery close factor divisor WAD - ceil(lif*lltv/WAD) > 0; tight at lltv = WAD.
+    require to_mathint(obligation.collaterals[0].maxLif) * to_mathint(obligation.collaterals[0].lltv) <= to_mathint(WAD()) * (to_mathint(WAD()) - 1);
+
+    // Sound: inactive collateral will repaid input actually revert with div by zero.
+    require (collateralOf(globalId, borrower, 0) > 0 && ghostPrice() > 0) || repaidUnits == 0;
```
jhoenicke: This assumption seems to be there just to exclude the division by zero for liquidatedCollatPrice.  It's not even clear if this is an assumption on the oracle or on the collateralBitmap.

For example, if the user has no collateral, and the division by zero returns zero the prover finds a counter example where liquidate() would only revert because of the division by zero.

You can say it's kind of unimportant edge case, because a user with no collateral cannot be liquidated anyway.  Similar you can say that the case where the oracle returns zero is unimportant, as the collateral would not be worth anything (in that case liquidate() would work if called with `seizedAssets = collateralOf(id, borrower)` but not with `seizedAssets == 0 && repaidUnits == 1 wei`).

I think we should just say something like:
```
    // Assume that the collateral price is non-zero and the collateral is active.  Otherwise, liquidate may revert with div by zero.
    require ghostPrice(obligation.collaterals[collateralIndex].oracle) > 0, "Assumption: the collateral price is not zero";
    require summaryGetBit(currentContract.position[globalId][borrower].activatedCollaterals, collateralIndex), "Assumption: liquidated collateral was activated";
```

certora/specs/NoDivisionByZero.spec
```diff
@@ -0,0 +1,170 @@
...
+// Therefore, only mulDivDown and mulDivUp can have variable denominators.
+//
+// maxLif(uint256, uint256) is excluded: it is a pure function callable with arbitrary inputs.
+// A standalone call with cursor >= WAD causes a safe revert (Solidity checked arithmetic).
+//
+// The liquidate function is verified in a separate rule (noDivisionByZeroLiquidate), restricted
+// to a single collateral due to SMT solver limitations with multi-iteration bitmap loops.
+// The toId summary follows the approach from PR #388: a ghost-backed deterministic function.
+// The msb summary uses a bitwise constraint (bitmap >> bit == 1).
+
```
jhoenicke, line 18: We could use the BitmapSummaries here:
```suggestion

import "BitmapSummaries.spec";

```

certora/specs/NoDivisionByZero.spec
```diff
@@ -0,0 +1,170 @@
...
+    function multicall(bytes[]) external => HAVOC_ALL DELETE;
+
+    function _.price() external => ghostPrice() expect(uint256);
+
+    function collateralOf(bytes32 id, address user, uint256 index) external returns (uint128) envfree;
+
+    function IdLib.toId(Midnight.Obligation memory obligation, uint256 chainId, address midnight) internal returns (bytes32) => summaryToId(obligation, chainId, midnight);
+
+    function UtilsLib.isLeaf(bytes32, bytes32, bytes32[] memory) internal returns (bool) => NONDET;
+    function UtilsLib.msb(uint128 bitmap) internal returns (uint256) => summaryMsbFn(bitmap);
```
jhoenicke: remove this if you use BitmapSummaries, see above.
```suggestion
```

certora/specs/NoDivisionByZero.spec
```diff
@@ -0,0 +1,170 @@
...
+rule noDivisionByZero(method f, env e, calldataarg args) filtered { f -> f.selector != sig:maxLif(uint256, uint256).selector && f.selector != sig:liquidate(Midnight.Obligation, uint256, uint256, uint256, address, bytes).selector } {
+    require !divisionByZero;
+    f(e, args);
+    assert !divisionByZero, "division by zero detected in mulDivDown or mulDivUp";
+}
+
+rule noDivisionByZeroLiquidate(env e, Midnight.Obligation obligation, uint256 collateralIndex, uint256 seizedAssets, uint256 repaidUnits, address borrower, bytes data) {
+    require equalsGlobalObligation(obligation);
+
+    // Single collateral (SMT solver can't handle multi-iteration bitmap loops).
```
jhoenicke: These two assumptions are not needed with BitmapSummaries.

But you need to replace 0 with collateralIndex in the require statements for maxLif below.

certora/confs/NoDivisionByZero.conf
```diff
@@ -0,0 +1,21 @@
...
+  "optimistic_hashing": true,
+  "hashing_length_bound": 2048,
+  "smt_timeout": 7200,
+  "prover_args": [
+    "-depth 5",
+    "-mediumTimeout 60",
+    "-s [z3:def{randomSeed=1},z3:def{randomSeed=2},z3:def{randomSeed=3},z3:def{randomSeed=4},z3:def{randomSeed=5}]"
+  ],
+  "msg": "No Division By Zero",
+  "wait_for_results": "all"
```
jhoenicke: Do we want to explicitly wait for the result?   The other confs don't do that.

jhoenicke (changes requested):
Some suggested changes.   With the BitmapSummaries, you can have more than one collateral.

certora/specs/NoDivisionByZero.spec
MathisGD: LLTV8 was added to experiment with "undercollateralized" where the collateral itself is not seizable but is a representation of the borrowing power. In this context it doesn't make sense to liquidate so we thought that this was fine. It is documented btw.

Though fixing it would be really simple, so I think that it's worth doing at the end.

MathisGD: will do in a separate PR: #594

certora/specs/NoDivisionByZero.spec
```diff
@@ -0,0 +1,161 @@
...
+
+definition WAD() returns uint256 = 1000000000000000000;
+
+definition collateralMatches(Midnight.Obligation obligation, uint256 index) returns bool = (index < globalObligationCollateralLength => obligation.collaterals[index].oracle == globalObligationCollateralOracle[index] && obligation.collaterals[index].token == globalObligationCollateralToken[index] && obligation.collaterals[index].lltv == globalObligationCollateralLLTV[index] && obligation.collaterals[index].maxLif == globalObligationCollateralMaxLif[index]);
+
+function equalsGlobalObligation(Midnight.Obligation obligation) returns (bool) {
+    return obligation.loanToken == globalObligationLoanToken && obligation.collaterals.length == globalObligationCollateralLength && collateralMatches(obligation, 0) && collateralMatches(obligation, 1) && collateralMatches(obligation, 2) && obligation.maturity == globalObligationMaturity && obligation.rcfThreshold == globalObligationRcfThreshold && obligation.enterGate == globalObligationEnterGate && obligation.liquidatorGate == globalObligationLiquidatorGate;
+}
+
+function summaryToId(Midnight.Obligation obligation, uint256 chainId, address morpho) returns (bytes32) {
```
lilCertora, line 90: nice!

certora/specs/NoDivisionByZero.spec
```diff
@@ -0,0 +1,161 @@
...
+// The toId summary follows the approach from PR #388: a ghost-backed deterministic function.
+
+import "BitmapSummaries.spec";
+
+methods {
+    function multicall(bytes[]) external => HAVOC_ALL DELETE;
+
+    function _.price() external => ghostPrice(calledContract) expect(uint256);
+
+    function collateralOf(bytes32 id, address user, uint256 index) external returns (uint128) envfree;
```
lilCertora: do we need this summary ? seems unused

certora/specs/NoDivisionByZero.spec
```diff
@@ -0,0 +1,161 @@
...
+hook Sload uint128 value obligationState[KEY bytes32 id].lossIndex {
+    require value < max_uint128;
+}
+
+// Follows from userLossIndexLeqObligationLossIndex in Midnight.spec and the hook above.
+hook Sload uint128 value position[KEY bytes32 id][KEY address user].lossIndex {
+    require value < max_uint128;
+}
+
+// Follows from totalUnitsEqualsSumNegativeDebtPlusWithdrawable in Midnight.spec.
```
lilCertora: is it true ? `totalUnitsEqualsSumNegativeDebtPlusWithdrawable` proves `totalUnits == sumDebt + withdrawable` which does not guarantee > 0 no ? this is ok here because when using this hook in liquidate as a divisor we know debt > 0 because badDebt > 0 thus totalUnits > 0

certora/confs/NoDivisionByZero.conf
```diff
@@ -0,0 +1,20 @@
...
+  "files": [
+    "src/Midnight.sol"
+  ],
+  "verify": "Midnight:certora/specs/NoDivisionByZero.spec",
+  "solc": "solc-0.8.31",
+  "solc_via_ir": true,
+  "solc_evm_version": "osaka",
+  "optimistic_loop": true,
+  "loop_iter": 2,
+  "optimistic_hashing": true,
```
lilCertora, line 11: is there a reason for not using "rule_sanity": "basic", ?

certora/specs/NoDivisionByZero.spec
```diff
@@ -0,0 +1,161 @@
...
+    // Needed for the bitmap loop which calls mulDivUp(WAD, maxLif) for every activated collateral.
+    require forall uint256 i. i < obligation.collaterals.length => obligation.collaterals[i].maxLif >= WAD();
+
+    // Ensures recovery close factor divisor WAD - ceil(lif*lltv/WAD) > 0; tight at lltv = WAD.
+    // Not true for LLTV=1.
+    require to_mathint(obligation.collaterals[collateralIndex].maxLif) * to_mathint(obligation.collaterals[collateralIndex].lltv) <= to_mathint(WAD()) * (to_mathint(WAD()) - 1);
+
+    // Assume that the collateral price is non-zero and the collateral is active. Otherwise, liquidate may revert with div by zero.
+    require ghostPrice(obligation.collaterals[collateralIndex].oracle) > 0, "Assumption: the collateral price is not zero";
+    require summaryGetBit(currentContract.position[globalId][borrower].activatedCollaterals, collateralIndex), "Assumption: liquidated collateral was activated";
```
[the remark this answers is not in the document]
lilCertora: reverting path correctly excluded

certora/specs/NoDivisionByZero.spec
```diff
@@ -0,0 +1,154 @@
...
+persistent ghost address globalObligationEnterGate;
+
+persistent ghost address globalObligationLiquidatorGate;
+
+persistent ghost bytes32 globalId;
+
+/// HOOKS ///
+
+// lossIndex < max: obligation can't lose 100% to bad debt.
+hook Sload uint128 value obligationState[KEY bytes32 id].lossIndex {
```
bhargavbh, line 69: is this a protocol assumption or does this follow from other lower-level assumptions? Not clear from the comment

lilCertora (approved):

certora/specs/NoDivisionByZero.spec
```diff
@@ -0,0 +1,154 @@
...
+rule noDivisionByZeroLiquidate(env e, Midnight.Obligation obligation, uint256 collateralIndex, uint256 seizedAssets, uint256 repaidUnits, address borrower, bytes data) {
+    require equalsGlobalObligation(obligation);
+
+    // Sound: touchObligation enforces maxLif >= WAD for all collaterals (ExactMath.spec).
+    // Needed for the bitmap loop which calls mulDivUp(WAD, maxLif) for every activated collateral.
+    require forall uint256 i. i < obligation.collaterals.length => obligation.collaterals[i].maxLif >= WAD();
+
+    // Ensures recovery close factor divisor WAD - ceil(lif*lltv/WAD) > 0; tight at lltv = WAD.
+    // Not true for LLTV=1.
+    require to_mathint(obligation.collaterals[collateralIndex].maxLif) * to_mathint(obligation.collaterals[collateralIndex].lltv) <= to_mathint(WAD()) * (to_mathint(WAD()) - 1);
```
bhargavbh, line 145: this assumption seems stronger than the rule lifTimesLltvIsLessThanOrEqualToOne in ExactMath.spec. In ExactMath.spec, you can actually prove the stronger inequality. The equality holds only for two cases: 
1. lltv=WAD() 
2. cursor=0.

```
rule maxLifIsAtLeastWadStrict(uint256 lltv, uint256 cursor) {
    require lltv < WAD(), "see rule createdObligationsHaveLltvLessThanOrEqualToOne";
    require cursor > 0, "see the definition of LIQUIDATION_CURSOR_LOW and LIQUIDATION_CURSOR_HIGH";
    require cursor < WAD(), "see the definition of LIQUIDATION_CURSOR_LOW and LIQUIDATION_CURSOR_HIGH";
    assert lltv * maxLif(lltv, cursor) <= WAD() * (WAD() - 1);
}
```
Here is the passing prover [link](https://prover.certora.com/output/1438622/af5006752f3f4c2d9dc63c3e2904eb4a)

bhargavbh (approved):
lgtm. suggested adding a rule in ExactMath.spec to support the assumption (maxLif*lltv <= wad * (wad-1)) in rule noDivisionByZeroLiquidate

jhoenicke (approved):
The new changes looked fine.

certora/specs/NoDivisionByZero.spec
jhoenicke: We cannot exclude WAD(), because of LLTV_8, but we can prove:

```
assert lltv=WAD() || lltv * maxLif(lltv, cursor) <= WAD() * (WAD() - 1);
```

certora/specs/NoDivisionByZero.spec
MathisGD: Right, I'll clarity the comment.

There is nothing preventing the bad debt to go to 100%. But the protocol documents that it might not behave as expected then (which is fine, because it should never happen).

certora/specs/NoDivisionByZero.spec
```diff
@@ -0,0 +1,154 @@
...
+
+persistent ghost address globalObligationEnterGate;
+
+persistent ghost address globalObligationLiquidatorGate;
+
+persistent ghost bytes32 globalId;
+
+/// HOOKS ///
+
+// lossIndex < max: obligation can't lose 100% to bad debt.
```
MathisGD:
```suggestion
// lossIndex < max: the protocol stop behaving correctly if this happens (documented).
```

certora/specs/NoDivisionByZero.spec
MathisGD: let's fix once we merged #594

PR #549, opened 2026-03-20
[ratification] buy/sell callbacks

adhusson:


src/Midnight.sol
```diff
@@ -239,16 +269,30 @@ contract Midnight is IMidnight {
...
+                    .onBuy(
+                        offer,
+                        offer.buy ? offerSigner : address(0),
+                        buyer,
+                        buyerAssets,
+                        sellerAssets,
+                        obligationUnits,
+                        obligationShares,
+                        buyerCallbackData
+                    ) == CALLBACK_SUCCESS,
```
adhusson: This is because `isAuthorized` can include EOAs that will never revert.

Also several 7702-style wallets don't revert on unknown signatures (see [safe](https://github.com/safe-fndn/safe-smart-account/blob/main/contracts/base/FallbackManager.sol#L66-L68), [argent](https://github.com/argentlabs/argent-contracts/blob/develop/contracts/wallet/BaseWallet.sol?utm_source=chatgpt.com#L145), [solady](https://github.com/Vectorized/solady/blob/main/src/accounts/Receiver.sol#L36), [ambire](https://github.com/AmbireTech/wallet/blob/main/contracts/AmbireAccount.sol#L57)). Some return 0, others return the passed signature. It seems safer to check for a special value (`CALLBACK_SUCCESS`) than to check for a common nonzero value like `1`.

src/Midnight.sol
```diff
@@ -239,16 +267,30 @@ contract Midnight is IMidnight {
...
         );
 
         if (buyerCallback != address(0)) {
-            ICallbacks(buyerCallback)
-                .onBuy(offer.obligation, buyer, buyerAssets, sellerAssets, units, buyerCallbackData);
+            require(
+                ICallbacks(buyerCallback)
+                    .onBuy(
+                        offer,
+                        (offer.buy && !offerPreValidated) ? offerSigner : address(1),
```
adhusson: `address(1)` means: no need to ratify the offer. 

The callback must know if it needs to ratify the offer or not. Otherwise imagine there are 2 callbacks authorized by the user, one does other things like withdraw collateral and the other does ratification. Without this argument the callback doesn't know if it has to check the offer or not.

Another solution is to have a new mapping `isRatifier`, separate from `isAuthorized`, and non-ratifying callbacks are only added to `isRatifier`.

src/Midnight.sol
```diff
@@ -247,8 +256,19 @@ contract Midnight is IMidnight {
         );
 
         if (buyerCallback != address(0)) {
```
adhusson, line 325: below, `address(1)` means: no need to ratify the offer. 

The callback must know if it needs to ratify the offer or not. Otherwise imagine there are 2 callbacks authorized by the user, one does other things like withdraw collateral and the other does ratification. Without this argument the callback doesn't know if it has to check the offer or not.

Another solution is to have a new mapping `isRatifier`, separate from `isAuthorized`, and non-ratifying callbacks are only added to `isRatifier`.

adhusson:
closing in favor of https://github.com/morpho-org/midnight/pull/588

PR #551, opened 2026-03-20
[lender side fee] merge `slash` and `accrue`

QGarchery:


MathisGD:
great! I'm implementing directly in #547

PR #555, opened 2026-03-23
[certora] account for update in effects spec

adhusson:


certora/specs/BalanceEffectsWithUpdate.spec
```diff
@@ -0,0 +1,106 @@
...
+
+    // Summarize internals irrelevant to credit and debt tracking.
+    function IdLib.storeInCode(Midnight.Obligation memory) internal returns (address) => NONDET;
+    function SafeTransferLib.safeTransfer(address, address, uint256) internal => NONDET;
+    function SafeTransferLib.safeTransferFrom(address, address, address, uint256) internal => NONDET;
+    function UtilsLib.isLeaf(bytes32, bytes32, bytes32[] memory) internal returns (bool) => NONDET;
+    function UtilsLib.msb(uint256) internal returns (uint256) => NONDET;
+    function TickLib.tickToPrice(uint256) internal returns (uint256) => NONDET;
+
+    // Assume no reentrancy: callbacks and token transfers do not re-enter Midnight.
```
QGarchery:
```suggestion
    // Assume no reentrancy: callbacks and token transfers do not re-enter Midnight.
    // This is justified because the properties we verify are about the effect of each function's own
    // body on credit and debt, not the effect of the full transaction including callbacks.
```

certora/specs/BalanceEffectsWithUpdate.spec
```diff
@@ -0,0 +1,106 @@
...
+    function _.onBuy(Midnight.Obligation, address, uint256, uint256, uint256, bytes) external => NONDET;
+    function _.onSell(Midnight.Obligation, address, uint256, uint256, uint256, bytes) external => NONDET;
+    function _.transfer(address, uint256) external => NONDET;
```
QGarchery:
```suggestion
```
unused because `SafeTransferLib` is summarized. 
Let's also put the `SafeTransferLib` summaries here (the comment says "callbacks and token transfers do not re-enter Midnight.")

certora/specs/BalanceEffectsWithUpdate.spec
```diff
@@ -0,0 +1,106 @@
...
+
+    function creditOf(bytes32 id, address user) external returns (uint256) envfree;
+    function debtOf(bytes32 id, address user) external returns (uint256) envfree;
+    function Utils.passiveFeeRecipient() external returns (address) envfree;
+
+    function updatePositionView(Midnight.Obligation memory, bytes32, address) external returns (uint128, uint128, uint128);
+
+    function _.price() external => NONDET;
+
+    // Summarize internals irrelevant to credit and debt tracking.
```
QGarchery:
```suggestion
```
those are sound summaries, so no need to say anything (same for the summaries above)

certora/specs/BalanceEffectsWithUpdate.spec
```diff
@@ -0,0 +1,130 @@
...
+    function _.onBuy(Midnight.Obligation, address, uint256, uint256, uint256, bytes) external => NONDET;
+    function _.onSell(Midnight.Obligation, address, uint256, uint256, uint256, bytes) external => NONDET;
+    function signer(bytes32, Midnight.Signature memory) internal returns (address) => NONDET;
+}
+
+/// The passive fee recipient has no pending fee, because they only receive credit via fee accrual
+/// and never participate in take.
+strong invariant feeRecipientHasNoPendingFee(bytes32 id)
+    pendingFee(id, Utils.passiveFeeRecipient()) == 0
+    filtered { f -> f.selector != sig:take(uint256, address, address, bytes, address, Midnight.Offer, Midnight.Signature, bytes32, bytes32[]).selector }
```
QGarchery: It would be best to have a `preserved` block for `take` that assumes exactly that the fee recipient cannot participate in `take`: can't be the sender, can't authorize another account (this is another invariant), can't sign. This way we would catch a potential bug where in `take` the pending fee of the passive fee recipient is changed

certora/specs/BalanceEffectsWithUpdate.spec
```diff
@@ -0,0 +1,130 @@
...
+
+    uint256 anyCredit = creditOf(anyId, anyUser);
+    uint256 anyDebt = debtOf(anyId, anyUser);
+    uint256 feeRecipientCredit = creditOf(id, passiveFeeRecipient);
+
+    updatePosition(e, obligation, user);
+
+    assert debtOf(anyId, anyUser) == anyDebt;
+    assert (anyUser != passiveFeeRecipient && anyUser != user) => creditOf(anyId, anyUser) == anyCredit;
+    assert (anyId != id) => creditOf(anyId, anyUser) == anyCredit;
```
QGarchery:
```suggestion
    assert (anyId != id) || (anyUser != passiveFeeRecipient && anyUser != user) => creditOf(anyId, anyUser) == anyCredit;
```
for consistency with other specs, and to make sure we have one assert per variable change to check
-> same for `withdrawEffects` below

certora/specs/BalanceEffectsWithUpdate.spec
```diff
@@ -0,0 +1,130 @@
...
+    assert creditOf(id, user) == updatedUserCredit;
+
+    // Depends on antecedent because fee recipient is not slashed in other user updates.
```
QGarchery:
```suggestion
    // Premise is needed because fee recipient is not slashed in other user updates.
```

certora/specs/BalanceEffectsWithUpdate.spec
```diff
@@ -0,0 +1,130 @@
...
+    assert user != passiveFeeRecipient => creditOf(id, passiveFeeRecipient) == feeRecipientCredit + userFee;
```
QGarchery:
```suggestion
    assert user != passiveFeeRecipient => creditOf(id, passiveFeeRecipient) == feeRecipientCredit + userFee;
    assert user == passiveFeeRecipient => userFee == 0;
```
not completely in line with the spec name, but worth checking I think

certora/specs/BalanceEffects.spec
```diff
@@ -130,25 +76,6 @@ rule liquidateEffects(env e, Midnight.Obligation obligation, uint256 collateralI
...
-    uint256 otherDebtBefore = debtOf(anyId, anyUser);
-
-    updatePosition(e, obligation, user);
-
-    assert creditOf(id, user) == creditBefore;
-    assert debtOf(anyId, anyUser) == otherDebtBefore;
-    assert anyUser != user || anyId != id => creditOf(anyId, anyUser) == otherCreditBefore;
-}
-
 /// ALL OTHER FUNCTIONS ///
```
QGarchery, line 203: Let's write in this file that `take` and `withdraw` are checked separately. The reader could be confused why they don't appear here

certora/specs/BalanceEffectsWithUpdate.spec
```diff
@@ -0,0 +1,130 @@
...
+methods {
+    function multicall(bytes[]) external => HAVOC_ALL DELETE;
+
+    function creditOf(bytes32 id, address user) external returns (uint256) envfree;
+    function debtOf(bytes32 id, address user) external returns (uint256) envfree;
+    function pendingFee(bytes32 id, address user) external returns (uint128) envfree;
+    function Utils.passiveFeeRecipient() external returns (address) envfree;
+
+    function updatePositionView(Midnight.Obligation memory, bytes32, address) external returns (uint128, uint128, uint128);
+
```
QGarchery:
```suggestion
```
this does nothing

certora/specs/BalanceEffectsWithUpdate.spec
```diff
@@ -0,0 +1,130 @@
...
+    function _.price() external => NONDET;
+
+    function IdLib.storeInCode(Midnight.Obligation memory) internal returns (address) => NONDET;
```
QGarchery:
```suggestion
    function IdLib.storeInCode(Midnight.Obligation memory) internal returns (address) => HAVOC_ECF;
```
this is a sound summary for this non-view function, because it does not reenter

certora/specs/BalanceEffectsWithUpdate.spec
```diff
@@ -0,0 +1,130 @@
...
+/// TAKE ///
+
+/// take changes maker's and taker's net credit-debt by +/- units relative to their post-update values
+/// and only changes credit of maker, taker, and passive fee recipient and debt of maker and taker at the obligation id.
+/// Assumes the passive fee recipient cannot participate in a take.
+rule takeEffects(env e, uint256 units, address taker, address takerCallback, bytes takerCallbackData, address receiver, Midnight.Offer offer, Midnight.Signature signature, bytes32 root, bytes32[] proof, bytes32 anyId, address anyUser) {
+    bytes32 id = toId(e, offer.obligation);
+    address passiveFeeRecipient = Utils.passiveFeeRecipient();
+
+    require offer.maker != passiveFeeRecipient;
```
bhargavbh: add a reason string to suppress warnings in Rule Notifcation
```suggestion
    require offer.maker != passiveFeeRecipient, "Assume passiveFeeRecipient can't participate in take()";
```

certora/specs/BalanceEffectsWithUpdate.spec
```diff
@@ -0,0 +1,130 @@
...
+    require taker != passiveFeeRecipient;
```
bhargavbh: same as above
```suggestion
    require taker != passiveFeeRecipient, "Assume passiveFeeRecipient can't participate in take()";
```

certora/specs/BalanceEffectsWithUpdate.spec
```diff
@@ -0,0 +1,130 @@
...
+
+    // Depends on antecedent because fee recipient is not slashed in other user updates.
+    assert onBehalf != passiveFeeRecipient => creditOf(id, passiveFeeRecipient) == feeRecipientCredit + userFee;
+}
+
+/// TAKE ///
+
+/// take changes maker's and taker's net credit-debt by +/- units relative to their post-update values
+/// and only changes credit of maker, taker, and passive fee recipient and debt of maker and taker at the obligation id.
+/// Assumes the passive fee recipient cannot participate in a take.
```
bhargavbh: added clarity
```suggestion
/// Assumes the passive fee recipient cannot participate in a take since it is an internal account.
```

bhargavbh (approved):
some minor comments; lgtm

certora/specs/BalanceEffectsWithUpdate.spec
adhusson: indeed but we summarize it as nondet elsewhere, should i change it everywhere?

QGarchery: Yes if it goes through that's great

certora/specs/BalanceEffectsWithUpdate.spec
```diff
@@ -0,0 +1,133 @@
...
+
+/// The passive fee recipient has no pending fee, because they only receive credit via fee accrual
+/// and never participate in take.
+strong invariant feeRecipientHasNoPendingFee(bytes32 id)
+    pendingFee(id, Utils.passiveFeeRecipient()) == 0
+    {
+        preserved take(uint256 units, address taker, address takerCallback, bytes takerCallbackData, address receiver, Midnight.Offer offer, Midnight.Signature signature, bytes32 root, bytes32[] proof) with (env e) {
+            require e.msg.sender != Utils.passiveFeeRecipient();
+            require offer.maker != Utils.passiveFeeRecipient();
+            require taker != Utils.passiveFeeRecipient();
```
QGarchery:
```suggestion
            require e.msg.sender != Utils.passiveFeeRecipient(), "passive fee recipient can't sign";
            require offer.maker != Utils.passiveFeeRecipient(), "passive fee recipient can't sign";
            require taker != Utils.passiveFeeRecipient(), "passive fee recipient can't sign, and can't authorize another account";
```
that's why those requires are ok (not being able to participate in take is a consequence)

certora/specs/BalanceEffectsWithUpdate.spec
```diff
@@ -0,0 +1,133 @@
...
+
+    // Premise is needed because fee recipient is not slashed in other user updates.
+    assert onBehalf != passiveFeeRecipient => creditOf(id, passiveFeeRecipient) == feeRecipientCredit + userFee;
+}
+
+/// TAKE ///
+
+/// take changes maker's and taker's net credit-debt by +/- units relative to their post-update values
+/// and only changes credit of maker, taker, and passive fee recipient and debt of maker and taker at the obligation id.
+/// Assumes the passive fee recipient cannot participate in take since its address derives from the hash of a human readable string.
```
QGarchery:
```suggestion
/// Assumes the passive fee recipient can't sign since its address derives from the hash of a human readable string.
```

certora/specs/BalanceEffectsWithUpdate.spec
```diff
@@ -0,0 +1,133 @@
...
+rule takeEffects(env e, uint256 units, address taker, address takerCallback, bytes takerCallbackData, address receiver, Midnight.Offer offer, Midnight.Signature signature, bytes32 root, bytes32[] proof, bytes32 anyId, address anyUser) {
+    bytes32 id = toId(e, offer.obligation);
+    address passiveFeeRecipient = Utils.passiveFeeRecipient();
+
+    require offer.maker != passiveFeeRecipient, "passive fee recipient can't participate in take()";
+    require taker != passiveFeeRecipient, "passive fee recipient can't participate in take()";
```
QGarchery:
```suggestion
    require offer.maker != passiveFeeRecipient, "passive fee recipient can't sign";
    require taker != passiveFeeRecipient, "passive fee recipient can't sign";
```

certora/specs/BalanceEffectsWithUpdate.spec
QGarchery: It would be great to write an invariant where we show that the passive fee recipient can't authorize another account

QGarchery (commented):
LGTM, a few minor things and an idea for an invariant

certora/specs/BalanceEffectsWithUpdate.spec
adhusson: will do separately

certora/specs/BalanceEffectsWithUpdate.spec
adhusson: I tried it, only requires left are:
* it's not `msg.sender`
* `signer` never returns it

certora/specs/BalanceEffectsWithUpdate.spec
adhusson: Actually it's not enough to not be able to sign, a contract could authorize / call. It's really that it derives from a particular string.

certora/specs/BalanceEffectsWithUpdate.spec
```diff
@@ -0,0 +1,150 @@
...
+    function SafeTransferLib.safeTransfer(address, address, uint256) internal => NONDET;
+    function SafeTransferLib.safeTransferFrom(address, address, address, uint256) internal => NONDET;
+    function _.onBuy(Midnight.Obligation, address, uint256, uint256, uint256, bytes) external => NONDET;
+    function _.onSell(Midnight.Obligation, address, uint256, uint256, uint256, bytes) external => NONDET;
+    function signer(bytes32, Midnight.Signature memory) internal returns (address) => signerSummary();
+}
+
+function signerSummary() returns address {
+    address returnedSigner;
+    require returnedSigner != Utils.passiveFeeRecipient(), "passive fee recipient can't sign";
```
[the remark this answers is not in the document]
adhusson: that is the point, to try to encode the fact that we don't expect the passive fee recipient to ever sign

certora/specs/BalanceEffectsWithUpdate.spec
```diff
@@ -0,0 +1,150 @@
...
+}
+
+/// The passive fee recipient can't authorize another account, because it can't sign
+/// and setIsAuthorized requires msg.sender == onBehalf || isAuthorized[onBehalf][msg.sender].
+strong invariant feeRecipientCantAuthorize(address authorized)
+    !isAuthorized(Utils.passiveFeeRecipient(), authorized)
+    {
+        preserved with (env e) {
+            require e.msg.sender != Utils.passiveFeeRecipient(), "passive fee recipient can't sign or call";
+            requireInvariant feeRecipientCantAuthorize(e.msg.sender);
```
bhargavbh: the self-referential requireInvariant feeRecipientCantAuthorize within preserved block puzzled me a bit, but should be fine.

bhargavbh (approved):

certora/specs/BalanceEffectsWithUpdate.spec
MathisGD, line 1: this doesn't make more assumptions than BalanceEffects right? then we could merge both specs, that would make things easier to read

certora/specs/BalanceEffectsWithUpdate.spec
```diff
@@ -0,0 +1,150 @@
...
+    function signer(bytes32, Midnight.Signature memory) internal returns (address) => signerSummary();
+}
+
+function signerSummary() returns address {
+    address returnedSigner;
+    require returnedSigner != Utils.passiveFeeRecipient(), "passive fee recipient can't sign";
+    return returnedSigner;
+}
+
+/// The passive fee recipient can't authorize another account, because it can't sign
```
lilCertora: would it be interesting to also prove that `PASSIVE_FEE_RECIPIENT` has no debt and zero lossIndex maybe ?

certora/specs/BalanceEffectsWithUpdate.spec
```diff
@@ -0,0 +1,150 @@
...
+
+    // Premise is needed because fee recipient is not slashed in other user updates.
+    assert onBehalf != passiveFeeRecipient => creditOf(id, passiveFeeRecipient) == feeRecipientCredit + userFee;
+}
+
+/// TAKE ///
+
+/// take changes maker's and taker's net credit-debt by +/- units relative to their post-update values
+/// and only changes credit of maker, taker, and passive fee recipient and debt of maker and taker at the obligation id.
+/// Assumes the passive fee recipient can't sign or call since its address derives from the hash of a human readable string.
```
lilCertora: worth verifying that passiveFeeRecipient receive the fees here maybe ?

certora/specs/BalanceEffectsWithUpdate.spec
```diff
@@ -0,0 +1,150 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+// The file is separate from BalanceEffects.spec because it cannot use a mulDiv summary.
```
lilCertora: that would be nice to explain briefly why it would not be good to use the summary here

lilCertora:
LGTM, I added some comments

certora/specs/BalanceEffectsWithUpdate.spec
adhusson: added the debt invariant! The loss index of the passive recipient could be updated though.

certora/specs/BalanceEffectsWithUpdate.spec
adhusson: I thought no summary in balanceeffect rules would be too long but it's ok, merged & removed summary.

certora/specs/BalanceEffectsWithUpdate.spec
adhusson: In the end i merged back into `BalanceEffects.sol` without a summary at all and it all goes through in ~7mn so I'm thinking no comment at all

certora/specs/BalanceEffectsWithUpdate.spec
adhusson: will do in a future PR to get this one through

certora/specs/BalanceEffectsWithUpdate.spec
lilCertora: yes that's nice if no summary is needed at the end, perfect

lilCertora (approved):

certora/specs/BalanceEffects.spec
```diff
@@ -107,46 +159,45 @@ rule takeEffects(env e, uint256 units, address taker, address takerCallback, byt
     assert makerNetAfter == makerNetBefore + makerDelta;
     mathint takerDelta = offer.buy ? -units : units;
     assert takerNetAfter == takerNetBefore + takerDelta;
-    assert anyId != id || (anyUser != offer.maker && anyUser != taker) => creditOf(anyId, anyUser) == otherCreditBefore && debtOf(anyId, anyUser) == otherDebtBefore;
+    assert anyId != id || (anyUser != offer.maker && anyUser != taker) => debtOf(anyId, anyUser) == otherDebtBefore;
+    assert anyId != id || (anyUser != offer.maker && anyUser != taker && anyUser != passiveFeeRecipient) => creditOf(anyId, anyUser) == otherCreditBefore;
```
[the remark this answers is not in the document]
adhusson: will address later

certora/specs/BalanceEffects.spec
```diff
@@ -27,74 +30,123 @@ methods {
...
+    bytes32 id = toId(e, obligation);
+    address passiveFeeRecipient = Utils.passiveFeeRecipient();
 
-definition noSlash(bytes32 id, address user) returns bool = currentContract.position[id][user].lossIndex == currentContract.obligationState[id].lossIndex && currentContract.position[id][user].lossIndex < max_uint128;
+    requireInvariant feeRecipientHasNoPendingFee(id);
 
-/// REPAY ///
+    uint128 updatedUserCredit;
+    uint128 userFee;
+    updatedUserCredit, _, userFee = updatePositionView(e, obligation, id, user);
```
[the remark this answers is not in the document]
adhusson: I think it's clear enough but not sure. We could add this as a kind of documentation but the rule passes with or without it so it's also misleading.

```
require (e.block.timestamp >= currentContract.position[id][user].lastAccrual, "lastAccrual is always set to block.timestamp"");
```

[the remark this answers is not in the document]
adhusson: will check liveness later

certora/specs/BalanceEffects.spec
```diff
@@ -27,74 +30,123 @@ methods {
...
-// Under noSlash, all mulDiv calls in _updatePosition hit y == d (identity) or y == 0.
-function summaryMulDiv(uint256 x, uint256 y, uint256 d) returns uint256 {
-    if (x == 0 || y == 0) return 0;
-    if (d > 0 && y == d) return x;
-    uint256 res;
-    return res;
+function signerSummary() returns address {
+    address returnedSigner;
+    require returnedSigner != Utils.passiveFeeRecipient(), "passive fee recipient can't sign";
+    return returnedSigner;
```
MathisGD, line 39: elegant!

certora/specs/BalanceEffects.spec
```diff
@@ -27,74 +30,123 @@ methods {
...
-// All rules in this file assume no accrual and no slash.
-definition noAccrual(env e, bytes32 id, address user) returns bool = currentContract.position[id][user].pendingFee == 0 || e.block.timestamp == currentContract.position[id][user].lastAccrual;
+/// The passive fee recipient can't authorize another account, because it can't sign
+/// and setIsAuthorized requires msg.sender == onBehalf || isAuthorized[onBehalf][msg.sender].
+strong invariant feeRecipientCantAuthorize(address authorized)
+    !isAuthorized(Utils.passiveFeeRecipient(), authorized)
+    {
+        preserved with (env e) {
+            require e.msg.sender != Utils.passiveFeeRecipient(), "passive fee recipient can't sign or call";
+            requireInvariant feeRecipientCantAuthorize(e.msg.sender);
```
MathisGD, line 49: why is that ok?

adhusson: Before the call we implicitly know `forall x. !authorized(recipient, x)` but the only fact added to the prover is `!authorized(recipient, authorized)`, so we can soundly assume `!authorized(recipient, e.msg.sender)` to get that `!authorized(recipient, authorized)` is true after the call.

adhusson: (based on the end of [this section](https://docs.certora.com/en/latest/docs/cvl/invariants.html#invariants-and-induction))

certora/specs/BalanceEffects.spec
```diff
@@ -27,74 +30,123 @@ methods {
...
     withdraw(e, obligation, units, onBehalf, receiver);
 
-    assert creditOf(id, onBehalf) == creditBefore - units;
-    assert debtOf(anyId, anyUser) == otherDebtBefore;
-    assert anyUser != onBehalf || anyId != id => creditOf(anyId, anyUser) == otherCreditBefore;
+    assert creditOf(id, onBehalf) == updatedUserCredit - units;
+    assert debtOf(anyId, anyUser) == anyDebt;
+    assert (anyId != id) || (anyUser != passiveFeeRecipient && anyUser != onBehalf) => creditOf(anyId, anyUser) == anyCredit;
+
+    // Premise is needed because fee recipient is not slashed in other user updates.
```
MathisGD: this is not the reason right? it's just that if they withdraw, they will get less credit at the end 
```suggestion
```

adhusson: Yes! That's the only reason in `updatePositionEffects`, here it would be slash+deduct withdrawn, fixing.

MathisGD (approved):

certora/specs/BalanceEffects.spec
jhoenicke: Requiring an invariant is sound, even the invariant that we just prove.  The reason is that we only require the invariant before the call.   There we can assume it, just like we already assume the invariant with the same parameters.

The reason we need this additional invariant is the recursive authorisation, right?

certora/specs/BalanceEffects.spec
```diff
@@ -27,74 +30,123 @@ methods {
...
 
     withdraw(e, obligation, units, onBehalf, receiver);
 
-    assert creditOf(id, onBehalf) == creditBefore - units;
-    assert debtOf(anyId, anyUser) == otherDebtBefore;
-    assert anyUser != onBehalf || anyId != id => creditOf(anyId, anyUser) == otherCreditBefore;
+    assert creditOf(id, onBehalf) == updatedUserCredit - units;
+    assert debtOf(anyId, anyUser) == anyDebt;
+    assert (anyId != id) || (anyUser != passiveFeeRecipient && anyUser != onBehalf) => creditOf(anyId, anyUser) == anyCredit;
+
```
jhoenicke, line 127: Do we also want to say that updatePositionView returns the same as creditOf after the function?  Currently we compare creditOf just with updatePositionView before the function, but maybe we want to say the updatePositionView after the function returns the same as updatePositionView before the function minus the withdrawn units.

MathisGD: good idea! we could add to the wish list if we want to move fast (break nothing)

certora/specs/BalanceEffects.spec
```diff
@@ -16,8 +21,6 @@ methods {
     function UtilsLib.isLeaf(bytes32, bytes32, bytes32[] memory) internal returns (bool) => NONDET;
     function UtilsLib.msb(uint128) internal returns (uint256) => NONDET;
     function TickLib.tickToPrice(uint256) internal returns (uint256) => NONDET;
-    function UtilsLib.mulDivDown(uint256 x, uint256 y, uint256 d) internal returns (uint256) => summaryMulDiv(x, y, d);
```
jhoenicke, line 19: If the solver runs into timeout problems here, you may want to consider a ghost summary, similar to what we have in Healthiness.

You may not even need the axioms, just that it's deterministic.

jhoenicke (approved):
Look good. 

I'll add some entries to the wishlist, like updatePositionView(anyUser) is unchanged before and after.

certora/specs/BalanceEffects.spec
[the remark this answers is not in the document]
jhoenicke: I would not add any require, as long as you can show the correctness without it.  Every require weakens the property.

PR #560, opened 2026-03-24
Use uint128 bitmaps

QGarchery:


src/Midnight.sol
```diff
@@ -507,7 +507,7 @@ contract Midnight is IMidnight {
             _position.collateral[collateralIndex] = newCollateralOf;
             if (newCollateralOf == 0 && seizedAssets > 0) {
                 // forge-lint: disable-next-item(unsafe-typecast) as collateralIndex < MAX_COLLATERALS (128)
```
jhoenicke:
```suggestion
```

test/LiquidationTest.sol
```diff
@@ -794,7 +794,8 @@ contract LiquidationTest is BaseTest {
                 midnight.collateralOf(id, borrower, i).mulDivUp(price, ORACLE_PRICE_SCALE)
                     .mulDivUp(WAD, _collateral.maxLif)
             );
-            bitmap ^= (1 << i);
+            require(i < 128, "i is too large");
```
jhoenicke, line 797: is this needed?  msb(bitmap) should already ensure this is the case.

src/Midnight.sol
```diff
@@ -405,7 +405,7 @@ contract Midnight is IMidnight {
 
         if (newCollateralOf == 0 && assets > 0) {
             // forge-lint: disable-next-item(unsafe-typecast) as collateralIndex < MAX_COLLATERALS (128)
```
jhoenicke:
```suggestion
```

src/Midnight.sol
```diff
@@ -377,7 +377,7 @@ contract Midnight is IMidnight {
 
         if (oldCollateralOf == 0 && assets > 0) {
             // forge-lint: disable-next-item(unsafe-typecast) as collateralIndex < MAX_COLLATERALS (128)
```
jhoenicke:
```suggestion
```

certora/helpers/Utils.sol
```diff
@@ -12,19 +12,19 @@ contract Utils {
         return keccak256(abi.encode(obligation));
     }
 
-    function getBit(uint256 bitmap, uint256 bit) external pure returns (bool) {
-        return bitmap & (1 << bit) != 0;
+    function getBit(uint128 bitmap, uint256 bit) external pure returns (bool) {
+        return uint256(bitmap) & (1 << bit) != 0;
```
jhoenicke:
```suggestion
        return bitmap & (1 << bit) != 0;
```

I think the upcast does nothing here.

src/Midnight.sol
```diff
@@ -713,13 +713,13 @@ contract Midnight is IMidnight {
...
         uint256 maxDebt;
-        uint256 bitmap = _position.activatedCollaterals;
+        uint128 bitmap = _position.activatedCollaterals;
         while (maxDebt < debt && bitmap != 0) {
             uint256 i = UtilsLib.msb(bitmap);
             Collateral memory collateral = obligation.collaterals[i];
             uint256 price = IOracle(collateral.oracle).price();
             maxDebt += _position.collateral[i].mulDivDown(price, ORACLE_PRICE_SCALE).mulDivDown(collateral.lltv, WAD);
-            bitmap = bitmap.clearBit(i);
+            bitmap = UtilsLib.clearBit(bitmap, i);
```
jhoenicke: Why change this but not the other clearBit/setBit in Midnight.sol?

src/libraries/UtilsLib.sol
```diff
@@ -68,17 +68,20 @@ library UtilsLib {
...
         assembly {
             res := sub(255, clz(bitmap))
         }
     }
 
-    function setBit(uint256 bitmap, uint256 bit) internal pure returns (uint256) {
-        return bitmap | (1 << bit);
+    /// @dev Assumes `bit` is less than 128.
+    function setBit(uint128 bitmap, uint256 bit) internal pure returns (uint128) {
+        return uint128(bitmap | (1 << bit));
```
jhoenicke, line 81: is a lint comment needed?
```
// forge-lint: disable-next-item(unsafe-typecast) as bit < 128
```

Or should one add a safe cast to be sure?

src/libraries/UtilsLib.sol
```diff
@@ -68,17 +68,20 @@ library UtilsLib {
...
     }
 
-    function clearBit(uint256 bitmap, uint256 bit) internal pure returns (uint256) {
-        return bitmap & ~(1 << bit);
+    /// @dev Assumes `bit` is less than 128.
+    function clearBit(uint128 bitmap, uint256 bit) internal pure returns (uint128) {
+        return uint128(bitmap & ~(1 << bit));
```
jhoenicke, line 87: lint comment for unsafe case?   In this case it's even safe fore bit >=128

jhoenicke (changes requested):
Looks good. I added a few suggestions and comments.

src/libraries/UtilsLib.sol
MathisGD: the forge linter is buggy, it doesn't always raise an error... I'll open an issue

I don't think that we need a safe cast

certora/helpers/Utils.sol
QGarchery: I had some errors in my editor but it went away I think :/

test/LiquidationTest.sol
QGarchery: I kept it since it's just a test, and so we won't mind changing this if it causes issues (very low chance though)

src/Midnight.sol
QGarchery: I had an error in my editor's linter... fixed it now 👍

src/libraries/UtilsLib.sol
QGarchery: added a `forge-lint` comment, so that it doesn't break later

src/libraries/UtilsLib.sol
QGarchery: added a `forge-lint` comment so that it doesn't break later

src/libraries/UtilsLib.sol
```diff
@@ -68,17 +68,22 @@ library UtilsLib {
...
         assembly {
             res := sub(255, clz(bitmap))
         }
     }
 
-    function setBit(uint256 bitmap, uint256 bit) internal pure returns (uint256) {
-        return bitmap | (1 << bit);
+    /// @dev Assumes `bit` is less than 128.
+    function setBit(uint128 bitmap, uint256 bit) internal pure returns (uint128) {
+        // forge-lint: disable-next-item(unsafe-typecast) as bit < 128
```
MathisGD, line 80: we could have a rule that hooks on setBit to check that this is true

jhoenicke: It's part of the summary for setBit/clearBit: there is an `assert bit < 128`.

jhoenicke (approved):

Issue #543, opened 2026-03-19
state reconstruction from take simple enough?


adhusson:
```suggestion
            sellerPos.pendingFee,
            buyerDebtIncrease,
            sellerDebtIncrease
```
buyerPos.debt and sellerPos.debt might be difficult to re-compute without this. It is updated by `buyerPos.debt -= UtilsLib.toUint128(buyerDebtDecrease)` where `uint256 buyerDebtDecrease = UtilsLib.min(oldBuyerDebt, obligationUnits)` which require the debt from the previous interaction (difficult to get)

but at the same time, `Take` already logs a lot of information

_Originally posted by @peyha in https://github.com/morpho-org/midnight/pull/470#discussion_r2960260863_

MathisGD:
@julien-devatom told me that this is pretty annoying for indexing. let's fix it

PR #561, opened 2026-03-24
Update events

peyha:
Should solve #543 

- credit cannot be retrieved through events by adding updates because the formula of `updatePosition` is too complex so new credit values should be logged when its state value is updated (in take, withdraw and updatePosition)
- debt can easily be retrieved by adding debt increase/decrease in take, repay and liquidate. Note that this breaks consistency between credit and debt (because the former logs its new value while the latter logs its diff with the previous value). We could make it consistent by logging `newDebt` everywhere but not sure about the benefit. Same thing for `totalUnits`

### DIFF
--- a/src/libraries/EventsLib.sol
+++ b/src/libraries/EventsLib.sol
@@ -31,7 +31,10 @@ library EventsLib {
         address sellerReceiver,
         bytes32 group,
         uint256 consumed,
-        uint256 totalUnits,
+        uint256 buyerCredit,
+        uint256 sellerCredit,
+        uint256 buyerCreditIncrease,
+        uint256 sellerCreditDecrease,
         uint256 buyerPendingFee,
         uint256 sellerPendingFee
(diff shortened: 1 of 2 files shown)


src/libraries/EventsLib.sol
```diff
@@ -31,7 +31,10 @@ library EventsLib {
         address sellerReceiver,
         bytes32 group,
         uint256 consumed,
-        uint256 totalUnits,
+        uint256 buyerCredit,
+        uint256 sellerCredit,
+        uint256 buyerCreditIncrease,
+        uint256 sellerCreditDecrease,
```
MathisGD, line 37: curious why adding buyerCredit and sellerCredit too?

the increases are sufficient to compute the new state easily

see https://github.com/morpho-org/midnight/pull/563

MathisGD:
replaced by #563

PR #566, opened 2026-03-24
refactor pending fee computation

QGarchery:
References:
- [Original idea](https://github.com/morpho-org/midnight/pull/470#discussion_r2911949508)
- [corresponding thread](https://morpholabs.slack.com/archives/C08A40KKN9W/p1774365085566989)

Not completely convinced, but I think we could go for only changing the `updatePositionView` function with the specs that we have

### DIFF
--- a/certora/specs/Midnight.spec
+++ b/certora/specs/Midnight.spec
@@ -61,6 +61,7 @@ function summaryMulDiv(uint256 x, uint256 y, uint256 d) returns uint256 {
     uint256 r;
     require x == 0 => r == 0;
     require d > 0 && y <= d => r <= x;
+    require d > 0 && x <= d => r <= y;
     return r;
 }
 
--- a/certora/specs/Midnight.spec
+++ b/certora/specs/Midnight.spec
@@ -117,7 +117,12 @@ strong invariant totalUnitsEqualsSumNegativeDebtPlusWithdrawable(bytes32 id)
     to_mathint(totalUnits(id)) == sumDebt[id] + to_mathint(withdrawable(id));
 
 strong invariant pendingContinuousFeeBoundedByCredit(bytes32 id, address user)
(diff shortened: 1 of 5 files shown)


certora/specs/Midnight.spec
MathisGD: we don't need it twice right?

certora/specs/Midnight.spec
```diff
@@ -117,7 +117,12 @@ strong invariant totalUnitsEqualsSumNegativeDebtPlusWithdrawable(bytes32 id)
     to_mathint(totalUnits(id)) == sumDebt[id] + to_mathint(withdrawable(id));
 
 strong invariant pendingContinuousFeeBoundedByCredit(bytes32 id, address user)
-    pendingFee(id, user) <= creditOf(id, user);
+    pendingFee(id, user) <= creditOf(id, user)
+    {
+        preserved {
+            requireInvariant userLossIndexLeqObligationLossIndex(id, user);
+        }
+    }
```
QGarchery, line 125: Weird that we need to do this, but this is because before we were doing `credit - postSlashCredit` in `updatePosition` so the prover could directly use the fact that `postSlashCredit <= credit`. Now it can't, and to prove that fact it needs to know that the user loss index is smaller than or equal to the obligation index, so that `credit.mulDivDown(type(uint128).max - obligationState[id].lossIndex, type(uint128).max - lossIndex)` is smaller than `credit`. And this is needed because otherwise the prover comes up with a couter-example where the unsafe cast `uint128(postSlashCredit)` fails

MathisGD:
~nb: not needed anymore if we merge #568~

568 was then modified

MathisGD:
we decided to close because
- the cost is slightly less nice
- the spec is only slightly improved

if there were other arguments @QGarchery feel free to reopen

PR #568, opened 2026-03-25
refactor: credit diff in updatePosition

peyha:


src/Midnight.sol
```diff
@@ -616,12 +618,12 @@ contract Midnight is IMidnight {
             ? uint128(postSlashPending.mulDivDown(accrualEnd - _lastAccrual, obligation.maturity - _lastAccrual))
             : 0;
         // forge-lint: disable-next-item(unsafe-typecast) as credit and pending are <= uint128 position fields
-        return (uint128(postSlashCredit) - fee, uint128(postSlashPending) - fee, fee);
+        return (credit - uint128(postSlashCredit) + fee, uint128(postSlashPending) - fee, fee);
```
```suggestion
        return (credit - uint128(postSlashCredit) + fee, pending - uint128(postSlashPending) - fee, fee);
```

MathisGD (changes requested):

src/Midnight.sol
```diff
@@ -630,17 +635,17 @@ contract Midnight is IMidnight {
...
         _position.lossIndex = obligationState[id].lossIndex;
-        _position.pendingFee = newPending;
+        _position.pendingFee -= pendingFeeLost;
         _position.lastAccrual = uint128(block.timestamp);
         // The passive fee recipient's credit is increased without slashing them first, meaning that they will get
         // slashed a bit too much later.
         position[id][PASSIVE_FEE_RECIPIENT].credit += accruedFee;
 
-        emit EventsLib.UpdatePosition(id, user, newCredit, newPending, accruedFee);
+        emit EventsLib.UpdatePosition(id, user, creditLost, pendingFeeLost, accruedFee);
```
adhusson: i don't get the point of this (and the corresponding update of updatePositionView)

peyha: The SQL query to compute the current credit state will be
```
SELECT SUM(delta_credit), usr
FROM 
     (SELECT user, -creditLost as delta_credit
      from midnight.UpdatePosition
      UNION ALL 
      SELECT user, -units as delta_credit
      from midnight.Withdraw
      UNION ALL
      SELECT buyer, buyerCreditIncrease as delta_credit
      FROM midnight.Take
      UNION ALL
      SELECT seller, -sellerCreditDecrease as delta_credit
      FROM midnight.Take)
GROUP BY usr
```
Without this it would be fairly complicated

adhusson (approved):

src/libraries/EventsLib.sol
```diff
@@ -16,7 +16,7 @@ library EventsLib {
     event SetObligationContinuousFee(bytes32 indexed id_, uint256 newContinuousFee);
     event SetDefaultContinuousFee(address indexed loanToken, uint256 newContinuousFee);
     event UpdatePosition(
-        bytes32 indexed id_, address indexed user, uint256 newCredit, uint256 newPendingFee, uint256 accruedFee
+        bytes32 indexed id_, address indexed user, uint256 creditLost, uint256 pendingFeeLost, uint256 accruedFee
```
MathisGD:
```suggestion
        bytes32 indexed id_, address indexed user, uint256 creditDecrease, uint256 pendingFeeDecrease, uint256 accruedFee
```

src/Midnight.sol
```diff
@@ -305,8 +306,8 @@ contract Midnight is IMidnight {
             offer.group,
             newConsumed,
             _obligationState.totalUnits,
```
MathisGD: totalUnits is not a diff, that's ok?

peyha: I think it's ok and we can even remove it. It's updated through `_obligationState.totalUnits =
            UtilsLib.toUint128(_obligationState.totalUnits + buyerCreditIncrease - sellerCreditDecrease)` in Take and is always updated with diff that are already logged in the corresponding event. I'm removing it

MathisGD: oh, nice

peyha:
Looks like all the state can be recomputed efficiently with the current events [cf this](https://github.com/morpho-org/midnight/blob/feat/sql/queries.sql)

MathisGD (approved):

src/Midnight.sol
```diff
@@ -618,7 +620,7 @@ contract Midnight is IMidnight {
             ? uint128(postSlashPending.mulDivDown(accrualEnd - _lastAccrual, obligation.maturity - _lastAccrual))
             : 0;
         // forge-lint: disable-next-item(unsafe-typecast) as credit and pending are <= uint128 position fields
-        return (uint128(postSlashCredit) - fee, uint128(postSlashPending) - fee, fee);
+        return (credit - uint128(postSlashCredit) + fee, _pendingFee - uint128(postSlashPending) + fee, fee);
```
adhusson: Still not a fan of this it means you need 2 calls to get the updated credit. What about returning the final values (for other consumers) & computing the delta in the log?

peyha: that would work yes

peyha: fixed

src/Midnight.sol
```diff
@@ -632,17 +634,20 @@ contract Midnight is IMidnight {
...
         _position.lossIndex = obligationState[id].lossIndex;
-        _position.pendingFee = newPending;
+        _position.pendingFee = newPendingFee;
         _position.lastAccrual = uint128(block.timestamp);
         // The passive fee recipient's credit is increased without slashing them first, meaning that they will get
         // slashed a bit too much later.
         position[id][PASSIVE_FEE_RECIPIENT].credit += accruedFee;
 
-        emit EventsLib.UpdatePosition(id, user, newCredit, newPending, accruedFee);
+        emit EventsLib.UpdatePosition(id, user, creditDecrease, pendingFeeDecrease, accruedFee);
```
MathisGD, line 650:
```suggestion
        emit EventsLib.UpdatePosition(id, user, _position.credit - newCredit, _position.pendingFee - newPendingFee, accruedFee);
```

peyha: that means moving the event emission before the state update, we can do it but it's not common practice in our events

MathisGD: ah yes oops, nvm

MathisGD (approved):

src/Midnight.sol
```diff
@@ -632,17 +634,20 @@ contract Midnight is IMidnight {
     /// @dev Expects the id to correspond to the obligation's id.
     function _updatePosition(Obligation memory obligation, bytes32 id, address user) internal {
         Position storage _position = position[id][user];
-        (uint128 newCredit, uint128 newPending, uint128 accruedFee) = updatePositionView(obligation, id, user);
+        (uint128 newCredit, uint128 newPendingFee, uint128 accruedFee) = updatePositionView(obligation, id, user);
+
+        uint128 creditDecrease = _position.credit - newCredit;
+        uint128 pendingFeeDecrease = _position.pendingFee - newPendingFee;
```
adhusson, line 640:
```suggestion

        emit EventsLib.UpdatePosition(id, user, _position.credit - newCredit, _position.pendingFee - newPendingFee, accruedFee);
```

src/Midnight.sol
adhusson, line 650:
```suggestion
```

adhusson (approved):

QGarchery (commented):
LGTM

PR #569, opened 2026-03-25
Use bitmap summaries in collateralbitmap

jhoenicke:
The BitmapSummaries now defines only the summaries, nothing else. The invariant is proved in CollateralBitmap and the file now uses the summaries instead of bitprecise reasoning.

This is on top of the stay healthy pull request, which defined the CollateralBitmap.


MathisGD (approved):

lilCertora (approved):
Nice , i like it and it is useful to have access as a library to the summaries

PR #570, opened 2026-03-25
[Certora] Loss index properties

lilCertora:
properties about the loss index:

- only changed when bad debt happens
- cannot cause reverts
- is synced after updatePosition

### DIFF
--- a/certora/confs/LossIndex.conf
+++ b/certora/confs/LossIndex.conf
@@ -13,7 +13,7 @@
   "optimistic_loop": true,
   "loop_iter": 2,
   "optimistic_hashing": true,
-  "hashing_length_bound": 1024,
+  "hashing_length_bound": 2048,
   "rule_sanity": "basic",
   "smt_timeout": 7200,
   "prover_args": [
--- a/certora/confs/LossIndex.conf
+++ b/certora/confs/LossIndex.conf
@@ -1,0 +1,27 @@
+{
+  "files": [
+    "certora/helpers/Utils.sol",
+    "src/Midnight.sol"
+  ],
+  "parametric_contracts": [
+    "Midnight"
+  ],
+  "verify": "Midnight:certora/specs/LossIndex.spec",
+  "solc": "solc-0.8.31",
+  "solc_via_ir": true,
+  "solc_evm_version": "osaka",
+  "optimistic_loop": true,
+  "loop_iter": 2,
+  "optimistic_hashing": true,
+  "hashing_length_bound": 1024,
+  "rule_sanity": "basic",
+  "smt_timeout": 7200,
+  "prover_args": [
+    "-splitParallel true",
+    "-dontStopAtFirstSplitTimeout true",
+    "-destructiveOptimizations twostage",
+    "-s",
+    "[z3:def{randomSeed=1},z3:def{randomSeed=2},z3:def{randomSeed=3},z3:def{randomSeed=4},z3:def{randomSeed=5}]"
+  ],
+  "msg": "Loss Index"
+}
--- a/certora/confs/LossIndex.conf
+++ b/certora/confs/LossIndex.conf
@@ -3,9 +3,6 @@
     "certora/helpers/Utils.sol",
     "src/Midnight.sol"
   ],
-  "parametric_contracts": [
-    "Midnight"
-  ],
   "verify": "Midnight:certora/specs/LossIndex.spec",
   "solc": "solc-0.8.34",
   "solc_via_ir": true,
@@ -14,11 +11,8 @@
   "loop_iter": 2,
   "optimistic_hashing": true,
   "hashing_length_bound": 2048,
-  "rule_sanity": "basic",
-  "smt_timeout": 7200,
   "prover_args": [
     "-splitParallel true",
-    "-dontStopAtFirstSplitTimeout true",
     "-destructiveOptimizations twostage",
     "-s",
     "[z3:def{randomSeed=1},z3:def{randomSeed=2},z3:def{randomSeed=3},z3:def{randomSeed=4},z3:def{randomSeed=5}]"
--- a/certora/confs/LossIndex.conf
+++ b/certora/confs/LossIndex.conf
@@ -7,7 +7,7 @@
     "Midnight"
   ],
   "verify": "Midnight:certora/specs/LossIndex.spec",
-  "solc": "solc-0.8.31",
+  "solc": "solc-0.8.34",
   "solc_via_ir": true,
   "solc_evm_version": "osaka",
   "optimistic_loop": true,
--- a/certora/specs/BalanceEffects.spec
+++ b/certora/specs/BalanceEffects.spec
@@ -153,20 +153,6 @@ rule takeSellerEffects(env e, uint256 units, address taker, address takerCallbac
     assert creditOf(id, seller) >= sellerUpdatedCreditBefore - units;
 }
 
-/// TAKE (DEBT DIRECTION) ///
-
-/// In take, the buyer's debt can only decrease.
-rule takeOnlySellerCanNewlyBecomeBorrower(env e, uint256 units, address taker, address takerCallback, bytes takerCallbackData, address receiver, Midnight.Offer offer, bytes ratifierData, bytes32 root, bytes32[] proof) {
-    bytes32 id = toId(e, offer.obligation);
-    address buyer = offer.buy ? offer.maker : taker;
-
-    uint256 buyerDebtBefore = debtOf(id, buyer);
-
-    take(e, units, taker, takerCallback, takerCallbackData, receiver, offer, ratifierData, root, proof);
-
-    assert debtOf(id, buyer) <= buyerDebtBefore;
-}
-
 /// REPAY ///
 
 /// Repay decreases onBehalf's debt by exactly units and only changes position[id][onBehalf].debt
--- a/certora/specs/BalanceEffects.spec
+++ b/certora/specs/BalanceEffects.spec
@@ -37,11 +37,13 @@ methods {
 
 /// UPDATE POSITION ///
 
-/// updatePosition sets user's credit to the post-update value,
-/// only changes credit of user at the obligation id, and accrues fee to continuousFeeCredit.
+/// updatePosition can only decrease user's credit (through slashing and fee accrual),
+/// sets it to the post-update value, only changes credit of user at the obligation id,
+/// and accrues fee to continuousFeeCredit.
 rule updatePositionEffects(env e, Midnight.Obligation obligation, address user, bytes32 anyId, address anyUser) {
     bytes32 id = toId(e, obligation);
 
+    uint256 creditBefore = creditOf(id, user);
     uint128 updatedUserCredit;
     uint128 userFee;
     updatedUserCredit, _, userFee = updatePositionView(e, obligation, id, user);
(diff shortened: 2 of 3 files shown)


certora/specs/LossIndex.spec
```diff
@@ -0,0 +1,81 @@
...
+    updatePosition(e, obligation, user);
+
+    assert creditOf(id, user) <= creditBefore;
+}
+
+/// After `updatePosition`, the user's lossIndex is synced to the obligation's lossIndex.
+rule updatePositionSyncsLossIndex(env e, Midnight.Obligation obligation, address user) {
+    bytes32 id = summaryToId(obligation);
+
+    updatePosition(e, obligation, user);
```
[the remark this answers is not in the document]
MathisGD: this is not how CVL works

certora/specs/LossIndex.spec
```diff
@@ -0,0 +1,91 @@
...
+    assert lossIndexChanged <=> badDebtOccurred;
+}
+
+/// LOSS INDEX: UPDATE POSITION EFFECTS ///
+
+/// `updatePosition` can only decrease the user's credit through slashing and fee accrual.
+rule updatePositionDecreasesCredit(env e, Midnight.Obligation obligation, address user) {
+    bytes32 id = summaryToId(obligation);
+
+    require obligationCreated(id), "obligation must be created";
```
QGarchery, line 77:
```suggestion
```
not needed because otherwise `updatePosition` will revert
https://prover.certora.com/output/66603/809f3ce388d44ddfad6c935fc6c6ab3b?anonymousKey=ee94ce73fbcc3b07016b710de63e270b16369cf0

certora/specs/LossIndex.spec
```diff
@@ -0,0 +1,91 @@
...
+    bool lossIndexChanged = currentContract.obligationState[id].lossIndex != lossIndexBefore;
+    bool badDebtOccurred = totalUnits(id) < totalUnitsBefore;
+
+    assert lossIndexChanged <=> badDebtOccurred;
+}
+
+/// LOSS INDEX: UPDATE POSITION EFFECTS ///
+
+/// `updatePosition` can only decrease the user's credit through slashing and fee accrual.
+rule updatePositionDecreasesCredit(env e, Midnight.Obligation obligation, address user) {
```
QGarchery: I think this would fit better in the rule BalanceEffects.updatePositionEffects

certora/specs/LossIndex.spec
```diff
@@ -0,0 +1,91 @@
...
+    bytes32 id = summaryToId(obligation);
+
+    require obligationCreated(id), "obligation must be created";
+    require userLossIndex(id, user) <= currentContract.obligationState[id].lossIndex, "user lossIndex bounded by obligation lossIndex, already proved in Midnight.spec";
+    require pendingFee(id, user) <= creditOf(id, user), "pending fee bounded by credit, already proved in Midnight.spec";
```
QGarchery, line 79: Could we use `requireInvariant` for those ?

certora/specs/LossIndex.spec
```diff
@@ -0,0 +1,91 @@
...
+    // Callbacks and external calls assumed non-reentrant.
+    function SafeTransferLib.safeTransferFrom(address, address, address, uint256) internal => NONDET;
+    function SafeTransferLib.safeTransfer(address, address, uint256) internal => NONDET;
+    function _.onLiquidate(bytes32, Midnight.Obligation, uint256, uint256, uint256, address, bytes) external => NONDET;
+    function _.onRatify(Midnight.Offer, bytes32, bytes) external => NONDET;
+    function _.onBuy(bytes32, Midnight.Obligation, address, uint256, uint256, bytes) external => NONDET;
+    function _.onSell(bytes32, Midnight.Obligation, address, uint256, uint256, bytes) external => NONDET;
+    function _.canIncreaseCredit(address) external => NONDET;
+    function _.canIncreaseDebt(address) external => NONDET;
+    function _.canLiquidate(address) external => NONDET;
```
QGarchery:
```suggestion
```
This seems to work too: https://prover.certora.com/output/66603/496b5903babf4f6196fa83fb6d1bcf0d?anonymousKey=f0c1b2874969fb8ee6f971f78715e52a28cc1492

Let's keep the mention that the external calls are assumed to be non-reentrant though

certora/confs/LossIndex.conf
```diff
@@ -0,0 +1,27 @@
...
+  "verify": "Midnight:certora/specs/LossIndex.spec",
+  "solc": "solc-0.8.34",
+  "solc_via_ir": true,
+  "solc_evm_version": "osaka",
+  "optimistic_loop": true,
+  "loop_iter": 2,
+  "optimistic_hashing": true,
+  "hashing_length_bound": 2048,
+  "rule_sanity": "basic",
+  "smt_timeout": 7200,
```
QGarchery:
```suggestion
```
this should work too I think

certora/confs/LossIndex.conf
```diff
@@ -0,0 +1,27 @@
+{
+  "files": [
+    "certora/helpers/Utils.sol",
+    "src/Midnight.sol"
+  ],
+  "parametric_contracts": [
+    "Midnight"
+  ],
```
QGarchery:
```suggestion
```
there are no invariant in this spec

certora/confs/LossIndex.conf
```diff
@@ -0,0 +1,27 @@
...
+  "solc_evm_version": "osaka",
+  "optimistic_loop": true,
+  "loop_iter": 2,
+  "optimistic_hashing": true,
+  "hashing_length_bound": 2048,
+  "rule_sanity": "basic",
+  "smt_timeout": 7200,
+  "prover_args": [
+    "-splitParallel true",
+    "-dontStopAtFirstSplitTimeout true",
```
QGarchery:
```suggestion
```
as per the doc this is only useful if there exists a counterexample

certora/specs/LossIndex.spec
QGarchery, line 1: > cannot cause reverts

this is not proven in this spec right ?

certora/specs/LossIndex.spec
lilCertora: oh yes, that's true! thanks

certora/confs/LossIndex.conf
lilCertora: yes

certora/specs/LossIndex.spec
lilCertora: yes that's fair

certora/specs/LossIndex.spec
lilCertora: if we move it to Balance effect we don't need it, but using `requireInvariant` would mean having to import Midnight.spec with all the summaries which is something we don't want. That would be a good idea for a future work to have all the invariants at the same place in order to be able to require them using `requireInvariant` when needed tough

certora/confs/LossIndex.conf
lilCertora: yes I added that for the parametric rule but all the utils.sol functions are pure so it's ok

certora/confs/LossIndex.conf
lilCertora: yes forgot to remove it

certora/specs/LossIndex.spec
lilCertora: working on it

certora/specs/LossIndex.spec
lilCertora: yes, good idea, we just need the safeTransfer to prevent havoc on the "does not cause revert" new rules

certora/specs/BalanceEffects.spec
```diff
@@ -109,6 +112,20 @@ rule takeEffects(env e, uint256 units, address taker, address takerCallback, byt
     assert anyId != id || (anyUser != offer.maker && anyUser != taker) => creditOf(anyId, anyUser) == otherCreditBefore;
 }
 
+/// TAKE (DEBT DIRECTION) ///
+
+/// In take, the buyer's debt can only decrease.
```
lilCertora: not related to this PR

QGarchery: I don't think this rule fits very well in this file: the effects of the `take` function (including changes to the `debtOf` variable) are already described in `takeEffects`

certora/specs/LossIndex.spec
```diff
@@ -0,0 +1,105 @@
...
+    bytes32 id = summaryToId(obligation);
+
+    require data.length == 0, "no callback to avoid unrelated external call reverts";
+    require obligationCreated(id), "obligation must be created";
+    require obligation.liquidatorGate == 0, "Assumption:no liquidator gate";
+    require obligation.collateralParams.length > 0, "obligation has at least one collateral (enforced by touchObligation)";
+    require !liquidationLocked(id, borrower), "liquidation not locked (transient storage is zero at transaction start)";
+    require currentContract.position[id][borrower].activatedCollaterals == 0, "Assumption: no active collaterals: skip loop and maximize badDebt";
+    require currentContract.position[id][borrower].debt > 0, "borrower must have debt to enter badDebt > 0 block";
+    require currentContract.position[id][borrower].debt <= currentContract.obligationState[id].totalUnits, "position debt bounded by totalUnits (system invariant)";
```
QGarchery:
```suggestion
    require currentContract.position[id][borrower].debt <= currentContract.obligationState[id].totalUnits, "position debt bounded by totalUnits (see totalUnitsEqualsSumNegativeDebtPlusWithdrawable)";
```

certora/specs/LossIndex.spec
```diff
@@ -0,0 +1,105 @@
...
+
+/// Under valid state, the loss index slash computation in `updatePosition` does not revert.
+rule updatePositionDoesNotRevert(env e, Midnight.Obligation obligation, address user) {
+    bytes32 id = summaryToId(obligation);
+
+    require obligationCreated(id), "obligation must be created";
+    require userLossIndex(id, user) <= currentContract.obligationState[id].lossIndex, "user lossIndex bounded by obligation lossIndex, already proved in Midnight.spec";
+    require pendingFee(id, user) <= creditOf(id, user), "pending fee bounded by credit, already proved in Midnight.spec";
+    require currentContract.position[id][user].lastAccrual <= e.block.timestamp, "lastAccrual <= block.timestamp by timestamp monotonicity";
+    require to_mathint(currentContract.obligationState[id].continuousFeeCredit) + pendingFee(id, user) <= max_uint128, "continuousFeeCredit cannot overflow from a single position's fee";
```
QGarchery:
```suggestion
```
we could do this if we target `updatePositionView` instead (to isolate the loss index update of the user), and it allows to remove the `require obligationCreated(id), "obligation must be created";` as well

QGarchery (approved):

certora/specs/BalanceEffects.spec
lilCertora: yes, it's an error , I removed it

certora/specs/LossIndex.spec
```diff
@@ -0,0 +1,104 @@
...
+    require obligationCreated(id), "obligation must be created";
+    require obligation.liquidatorGate == 0, "Assumption:no liquidator gate";
+    require obligation.collateralParams.length > 0, "obligation has at least one collateral (enforced by touchObligation)";
+    require !liquidationLocked(id, borrower), "liquidation not locked (transient storage is zero at transaction start)";
+    require currentContract.position[id][borrower].activatedCollaterals == 0, "Assumption: no active collaterals: skip loop and maximize badDebt";
+    require currentContract.position[id][borrower].debt > 0, "borrower must have debt to enter badDebt > 0 block";
+    require currentContract.position[id][borrower].debt <= currentContract.obligationState[id].totalUnits, "position debt bounded by totalUnits (see totalUnitsEqualsSumNegativeDebtPlusWithdrawable)";
+    require e.msg.value == 0, "Midnight is not payable";
+
+    liquidate@withrevert(e, obligation, 0, 0, 0, borrower, borrower, borrower, data);
```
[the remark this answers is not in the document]
lilCertora: true

MathisGD (approved):

PR #571, opened 2026-03-25
refactor: set/clearBit

MathisGD:
removes the need for unsafe cast, but arguably a bit less clear what happens.

see the tests to see that it behaves as expected

### DIFF
--- a/test/UtilsLibTest.sol
+++ b/test/UtilsLibTest.sol
@@ -107,6 +107,38 @@ contract UtilsLibTest is Test {
         UtilsLib.toUint128(x);
     }
 
+    function _mask(uint256 bit) internal pure returns (uint128 m) {
+        assembly {
+            m := shl(bit, 1)
+        }
+    }
+
+    function testFuzzSetBit(uint128 bitmap, uint256 bit) public pure {
+        bit = bound(bit, 0, 127);
+        uint128 result = UtilsLib.setBit(bitmap, bit);
+        uint128 m = _mask(bit);
+        assertEq(result & m, m);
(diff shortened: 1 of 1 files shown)


src/libraries/UtilsLib.sol
MathisGD, line 1: actually, this produces compiler warnings interestingly

Image

jhoenicke: Does `uint128(1) << uint128(bit)` work, or does it trigger an unsafe cast warning for bit?

MathisGD: normally it will trigger a warning too, but unsure (will try). btw I'm surprised by the solc warning here, while it says nothing on all the other unsafe casts (maybe they judged that this is particularly misleading)

MathisGD: it seems to fix! but then forge lint with raise a warning for the unsafe cast of uint128...

at this point I'm considering switching to assembly 😂

src/libraries/UtilsLib.sol
```diff
@@ -77,13 +77,11 @@ library UtilsLib {
 
     /// @dev Assumes `bit` is less than 128.
     function setBit(uint128 bitmap, uint256 bit) internal pure returns (uint128) {
-        // forge-lint: disable-next-item(unsafe-typecast) as bit < 128
-        return uint128(bitmap | (1 << bit));
+        return bitmap | (uint128(1) << bit);
```
adhusson: i find the unsafe cast + lint comment good enough. The comment could repeat that its an assumption but its minor. Assembly OK too of course.

PR #574, opened 2026-03-26
[Certora] Certora/slot saving research

lilCertora:


certora/specs/UpdateBeforeCredit.spec
MathisGD, line 1: with this, we could have a function that reads the credit to compute something else before the update and things wouldn't break, right?

lilCertora:
latest run : https://prover.certora.com/output/7508195/2aa3cd379ed347538ade59f7b003e245/?anonymousKey=5b6d16b0109b93fcd7aa1a64b6bda7288e02064f

certora/specs/UpdateBeforeCredit.spec
lilCertora: yes that's the tradeoff, i think if you want to go this way the `bool mustUpdate` approach would be better, what do you think @jhoenicke ?

MathisGD: at least we should separate take from the other function, if you go for this route

jhoenicke: I would prefer the check for zero credit in _updatePosition, which requires the extra bool argument for updateTimestamps.

If you want to allow the non-zero check before _updatePosition and just that, one way would be to have an internal `hasCredit(bytes32 id, address user)` that just returns position.credit > 0.  Use this function in take to guard updatePosition.  Then summarize this function in this spec and implement it in CVL.  Then the sload hook is not triggered for the `hasCredit` function.  That is okay, as updatePosition does not change the value of `hasCredit` meaningfully (it may remove dust credit).

MathisGD:
we decided to move on with #532 without changing src, FYI

lilCertora:
> we decided to move on with #532 without changing src, FYI

i see #532 is closed now and #550 seems to be the one you moved to, we can close this one @MathisGD ?

MathisGD:
replaced by #608

PR #575, opened 2026-03-26
assets offer input

MathisGD:
fixes [this issue](https://morpholabs.slack.com/archives/C08A40KKN9W/p1773819678844489). 

credits to the Tenor team for the catch

### DIFF
--- a/certora/specs/Consume.spec
+++ b/certora/specs/Consume.spec
@@ -81,7 +81,7 @@ rule takeConsumedDelta(env e, uint256 units, address taker, address takerCallbac
 
 /// If consumed[offer.maker][offer.group] is already at or above maxUnits before a `take` in units mode,
 /// it remains unchanged.
-rule takeConsumedAtMaxUnchanged(env e, uint256 units, address taker, address takerCallback, bytes takerCallbackData, address receiver, Midnight.Offer offer, Midnight.Signature signature, bytes32 root, bytes32[] proof) {
+rule takeConsumedAtMaxUnchangedUnits(env e, uint256 units, address taker, address takerCallback, bytes takerCallbackData, address receiver, Midnight.Offer offer, Midnight.Signature signature, bytes32 root, bytes32[] proof) {
     require offer.maxSellerAssets == 0 && offer.maxBuyerAssets == 0;
 
     uint256 consumedBefore = consumed(offer.maker, offer.group);
@@ -89,6 +89,30 @@ rule takeConsumedAtMaxUnchanged(env e, uint256 units, address taker, address tak
     take(e, units, taker, takerCallback, takerCallbackData, receiver, offer, signature, root, proof);
 
     assert consumedBefore >= offer.maxUnits => consumed(offer.maker, offer.group) == consumedBefore;
+}
+
+/// If consumed is already at or above maxSellerAssets before a `take` in seller assets mode,
+/// it remains unchanged.
+rule takeConsumedAtMaxUnchangedSellerAssets(env e, uint256 units, address taker, address takerCallback, bytes takerCallbackData, address receiver, Midnight.Offer offer, Midnight.Signature signature, bytes32 root, bytes32[] proof) {
+    require offer.maxSellerAssets > 0;
+
+    uint256 consumedBefore = consumed(offer.maker, offer.group);
+
+    take(e, units, taker, takerCallback, takerCallbackData, receiver, offer, signature, root, proof);
+
+    assert consumedBefore >= offer.maxSellerAssets => consumed(offer.maker, offer.group) == consumedBefore;
+}
+
+/// If consumed is already at or above maxBuyerAssets before a `take` in buyer assets mode,
+/// it remains unchanged.
+rule takeConsumedAtMaxUnchangedBuyerAssets(env e, uint256 units, address taker, address takerCallback, bytes takerCallbackData, address receiver, Midnight.Offer offer, Midnight.Signature signature, bytes32 root, bytes32[] proof) {
+    require offer.maxSellerAssets == 0 && offer.maxBuyerAssets > 0;
+
+    uint256 consumedBefore = consumed(offer.maker, offer.group);
+
+    take(e, units, taker, takerCallback, takerCallbackData, receiver, offer, signature, root, proof);
+
+    assert consumedBefore >= offer.maxBuyerAssets => consumed(offer.maker, offer.group) == consumedBefore;
 }
 
 /// A fully-consumed offer in units mode only allows no-op takes.
--- a/certora/specs/Consume.spec
+++ b/certora/specs/Consume.spec
@@ -116,7 +116,7 @@ rule takeConsumedAtMaxUnchangedBuyerAssets(env e, uint256 units, address taker, 
 }
 
 /// A fully-consumed offer in units mode only allows no-op takes.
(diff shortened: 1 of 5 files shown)


src/Midnight.sol
```diff
@@ -256,8 +256,17 @@ contract Midnight is IMidnight {
...
         uint256 sellerAssets = offer.buy ? units.mulDivDown(sellerPrice, WAD) : units.mulDivUp(sellerPrice, WAD);
 
-        uint256 newConsumed = consumed[offer.maker][offer.group] += units;
-        require(newConsumed <= offer.maxUnits, "consumed");
+        uint256 newConsumed;
+        if (offer.maxSellerAssets > 0) {
+            newConsumed = consumed[offer.maker][offer.group] += sellerAssets;
+            require(newConsumed <= offer.maxSellerAssets, "consumed");
+        } else if (offer.maxBuyerAssets > 0) {
+            newConsumed = consumed[offer.maker][offer.group] += buyerAssets;
```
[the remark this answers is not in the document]
MathisGD: we should document that the protocol expects only one of the 3 max to be nonzero (and effective)

src/Midnight.sol
```diff
@@ -256,8 +256,17 @@ contract Midnight is IMidnight {
...
+            require(newConsumed <= offer.maxBuyerAssets, "consumed");
+        } else {
+            newConsumed = consumed[offer.maker][offer.group] += units;
```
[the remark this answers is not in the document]
MathisGD: we should document here that all offers within a same group should have the same effective max

/// @dev Groups are useful to have a global offered amount shared across multiple offers ("OCO").
    /// @dev To work as expected, all offers in a same group should have the same units and loan token.

src/Midnight.sol
```diff
@@ -256,8 +256,17 @@ contract Midnight is IMidnight {
         uint256 buyerAssets = offer.buy ? units.mulDivDown(buyerPrice, WAD) : units.mulDivUp(buyerPrice, WAD);
         uint256 sellerAssets = offer.buy ? units.mulDivDown(sellerPrice, WAD) : units.mulDivUp(sellerPrice, WAD);
 
-        uint256 newConsumed = consumed[offer.maker][offer.group] += units;
-        require(newConsumed <= offer.maxUnits, "consumed");
+        uint256 newConsumed;
+        if (offer.maxSellerAssets > 0) {
```
adhusson, line 263: because [making offers in different groups requires multiple signatures](https://morpholabs.slack.com/archives/C02N7CZ088N/p1764668580649929?thread_ts=1764667868.337599&cid=C02N7CZ088N) we decided to keep group at the offer level. But here it means offer can use different units for their consumed, it would be easier if the group and the max (max seller / max buyer / max units) was at the root level.

src/Midnight.sol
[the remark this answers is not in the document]
adhusson: I assume the expectation is that the only correct values for `(maxSellerAssets, maxBuyerAssets)` are `(1,0)`, `(0,1)` and `(0,0)`. If yes we could `require(!maxSellerAssets || maxBuyerAssets, "incoherent")`. Or use an enum `LimitingFactor (Units,SellerAssets,BuyerAssets)`

src/Midnight.sol
```diff
@@ -82,7 +82,8 @@ contract Midnight is IMidnight {
     mapping(bytes32 id => ObligationState) public obligationState;
 
     /// @dev Groups are useful to have a global offered amount shared across multiple offers ("OCO").
-    /// @dev To work as expected, all offers in a same group should have the same units and loan token.
+    /// @dev To work as expected, all offers in a same group should have the maxs and loan token.
```
```suggestion
    /// @dev To work as expected, all offers in a same group should have the same maxs and loan token.
```

src/Midnight.sol
MathisGD: why making offers in different groups require multiple signature? at the end we didn't implement this

adhusson: badly worded sorry, should have said "because [with groups at root level] making offers in different groups requires multiple signature..."

test/TakeTest.sol
```diff
@@ -598,6 +598,106 @@ contract TakeTest is BaseTest {
...
+        units = bound(units, 1, maxAssets);
+        deal(address(loanToken), lender, units);
+        collateralize(obligation, borrower, units);
+
+        lenderOffer.maxUnits = 0;
+        lenderOffer.maxSellerAssets = type(uint256).max;
+
+        (, uint256 sellerAssets,) = take(units, borrower, lenderOffer);
+
+        assertTrue(sellerAssets <= lenderOffer.maxSellerAssets);
```
peyha: always true because `lenderOffer.maxSellerAssets=type(uint256).max`

src/Midnight.sol
```diff
@@ -256,8 +258,17 @@ contract Midnight is IMidnight {
         uint256 buyerAssets = offer.buy ? units.mulDivDown(buyerPrice, WAD) : units.mulDivUp(buyerPrice, WAD);
         uint256 sellerAssets = offer.buy ? units.mulDivDown(sellerPrice, WAD) : units.mulDivUp(sellerPrice, WAD);
 
-        uint256 newConsumed = consumed[offer.maker][offer.group] += units;
-        require(newConsumed <= offer.maxUnits, "consumed");
+        uint256 newConsumed;
+        if (offer.maxSellerAssets > 0) {
+            newConsumed = consumed[offer.maker][offer.group] += sellerAssets;
+            require(newConsumed <= offer.maxSellerAssets, "consumed");
```
```suggestion
            require(newConsumed <= offer.maxSellerAssets, "consumed seller assets");
```
clearer when revert

src/interfaces/IMidnight.sol
```diff
@@ -36,6 +36,8 @@ struct Offer {
     bytes callbackData;
     address receiverIfMakerIsSeller;
     bool exitOnly;
+    uint256 maxSellerAssets;
+    uint256 maxBuyerAssets;
```
adhusson:
```suggestion
    uint256 start;
    uint256 expiry;
    uint256 tick;
    bytes32 group;
    bytes32 session;
    address callback;
    bytes callbackData;
    address receiverIfMakerIsSeller;
    bool exitOnly;
    uint256 maxUnits;
    uint256 maxSellerAssets;
    uint256 maxBuyerAssets;
```

certora/specs/Consume.spec
```diff
@@ -59,34 +59,66 @@ rule consumeNonDecreasing(env e, method f, calldataarg args, address user, bytes
...
     take(e, units, taker, takerCallback, takerCallbackData, receiver, offer, signature, root, proof);
 
-    assert consumed(offer.maker, offer.group) <= offer.maxUnits;
+    assert offer.maxSellerAssets > 0 => consumed(offer.maker, offer.group) <= offer.maxSellerAssets;
+    assert offer.maxBuyerAssets > 0 => consumed(offer.maker, offer.group) <= offer.maxBuyerAssets;
+    assert offer.maxSellerAssets == 0 && offer.maxBuyerAssets == 0 => consumed(offer.maker, offer.group) <= offer.maxUnits;
 }
 
-/// After a successful `take`, the change in consumed equals the units taken.
+/// After a successful `take` in units mode, the change in consumed equals the units taken.
```
adhusson:
```suggestion
/// After a successful `take`, the change in consumed equals the units taken.
```

adhusson (approved):

adhusson (approved):

peyha (approved):

PR #588, opened 2026-03-26
[ratification] separated ratification

adhusson:


src/Midnight.sol
```diff
@@ -110,9 +115,12 @@ contract Midnight is IMidnight {
     /// @dev Address that can set trading fees.
     address public feeSetter;
 
+    EcrecoverRatifier public immutable ECRECOVER_RATIFIER;
```
adhusson: small special treatment for ecreover for ease of use

src/Midnight.sol
```diff
@@ -554,6 +572,31 @@ contract Midnight is IMidnight {
         return (seizedAssets, repaidUnits);
     }
 
+    function setRatified(address onBehalf, bytes32 root, bool newRatified) external {
+        require(onBehalf == msg.sender || isAuthorized[onBehalf][msg.sender], "unauthorized");
+        ratified[onBehalf][root] = newRatified;
+        emit EventsLib.SetRatified(onBehalf, root, newRatified);
+    }
+
+    function setAuthorizedWithSig(Authorization memory authorization, Signature calldata signature) external {
```
MathisGD: I think now that we should not add this ([rationale](https://morpholabs.slack.com/archives/C08A40KKN9W/p1774966675871769?thread_ts=1774915247.151319&cid=C08A40KKN9W))

src/Midnight.sol
```diff
@@ -554,6 +572,31 @@ contract Midnight is IMidnight {
         return (seizedAssets, repaidUnits);
     }
 
+    function setRatified(address onBehalf, bytes32 root, bool newRatified) external {
```
MathisGD: if we go for modularization of ecrecover, does it makes sense to not even have this?

src/Midnight.sol
```diff
@@ -215,9 +220,22 @@ contract Midnight is IMidnight {
...
         require(offer.maker != taker, "buyer and seller cannot be the same");
-        require(signer(root, sig) == offer.maker, "invalid signature");
         require(UtilsLib.isLeaf(root, keccak256(abi.encode(offer)), proof), "invalid proof");
         require(offer.session == session[offer.maker], "invalid session");
+        if (offer.ratifier == address(0)) {
+            require(ratified[offer.maker][root], "unauthorized");
+        } else if (offer.ratifier == address(1)) {
+            Signature memory sig = abi.decode(ratifierData, (Signature));
+            address _signer = signer(root, sig);
+            require(_signer == offer.maker || isAuthorized[offer.maker][_signer], "invalid signature");
```
MathisGD: wondering if this shouldn't be a ratifier too

src/Midnight.sol
```diff
@@ -215,9 +220,22 @@ contract Midnight is IMidnight {
...
+        } else {
+            require(
+                isAuthorized[offer.maker][offer.ratifier]
```
MathisGD: why did you choose to re-use `isAuthorized`?

src/libraries/ConstantsLib.sol
```diff
@@ -9,6 +9,9 @@ uint32 constant MAX_CONTINUOUS_FEE = uint32(uint256(0.01e18) / uint256(365 days)
 uint256 constant TIME_TO_MAX_LIF = 15 minutes;
 bytes32 constant EIP712_DOMAIN_TYPEHASH = keccak256("EIP712Domain(uint256 chainId,address verifyingContract)");
 bytes32 constant ROOT_TYPEHASH = keccak256("Root(bytes32 root)");
+bytes32 constant AUTHORIZATION_TYPEHASH =
+    keccak256("Authorization(address authorizer,address authorizee,bool isAuthorized,uint256 nonce,uint256 deadline)");
+bytes32 constant CALLBACK_SUCCESS = keccak256("CALLBACK_SUCCESS");
```
MathisGD: what's the point of the callback success magic value?

src/Midnight.sol
adhusson: symmetry with `isAuthorized[maker][signer]`: ratifier has the same role of saying if your offer can be used by midnight or not

src/libraries/ConstantsLib.sol
MathisGD: some smart wallets don't revert on fallback, so if you put them as authorized it would be very problematic

src/Midnight.sol
adhusson: [no more special treatment](https://morpholabs.slack.com/archives/C02N7CZ088N/p1774955545143689)

src/libraries/ConstantsLib.sol
adhusson: [more details](https://github.com/morpho-org/midnight/pull/549#discussion_r2963482704)

src/Midnight.sol
```diff
@@ -225,9 +217,14 @@ contract Midnight is IMidnight {
...
         require(block.timestamp <= offer.expiry, "offer expired");
         require(offer.maker != taker, "buyer and seller cannot be the same");
-        require(signer(root, sig) == offer.maker, "invalid signature");
         require(UtilsLib.isLeaf(root, keccak256(abi.encode(offer)), proof), "invalid proof");
         require(offer.session == session[offer.maker], "invalid session");
+        require(
+            isAuthorized[offer.maker][offer.ratifier]
+                && IRatifier(offer.ratifier).onRatify(offer, root, ratifierData) == CALLBACK_SUCCESS,
+            "unauthorized"
+        );
```
MathisGD, line 290:
```suggestion
        require(isAuthorized[offer.maker][offer.ratifier], "ratifier not authorized");
        require(IRatifier(offer.ratifier).onRatify(offer, root, ratifierData) == CALLBACK_SUCCESS, "ratification failed");
```

src/Midnight.sol
```diff
@@ -225,9 +217,14 @@ contract Midnight is IMidnight {
         require(block.timestamp >= offer.start, "offer not started");
         require(block.timestamp <= offer.expiry, "offer expired");
         require(offer.maker != taker, "buyer and seller cannot be the same");
-        require(signer(root, sig) == offer.maker, "invalid signature");
         require(UtilsLib.isLeaf(root, keccak256(abi.encode(offer)), proof), "invalid proof");
```
MathisGD, line 285: arguably, this could be in the ratifier too

src/Midnight.sol
```diff
@@ -216,7 +208,7 @@ contract Midnight is IMidnight {
         bytes memory takerCallbackData,
         address receiverIfTakerIsSeller,
         Offer memory offer,
-        Signature memory sig,
+        bytes memory ratifierData,
```
MathisGD, line 276: passing arbitrary data opens the door to a lot of things, such as rate offers (closest tick), dutch orders, etc. we should make sure that it doesn't have non desirable second order effects

src/Midnight.sol
```diff
@@ -208,6 +199,7 @@ contract Midnight is IMidnight {
     /// position at the end.
     /// @dev The taker might not get the price they expected if the trading fee was just changed.
     /// @dev All sellerAssets are reachable with the units input, and all buyerAssets are reachable only if
+    /// @dev To work as expected, all offers in a same group should have the same ratifier.
```
MathisGD: why?

src/Midnight.sol
adhusson: > require(UtilsLib.isLeaf(root, keccak256(abi.encode(offer)), proof),

if this is in the ratifier then the proof should be part of the arbitrary data & we say people will be able to sign something other than proof merkle roots in the future.

src/Midnight.sol
adhusson: - it's all up to whitelisting for common usage but yes
- offer callbacks have arbitrary data too and the trade fails if it fails. An offer could include its own parameters in its callbackData, the only thing it can't include that `onRatify` receives is the root.

src/Midnight.sol
adhusson: It's mostly EIP712-specific so the comment is too broad. EIP712 says the `verifyingContract` should be the address of the contract doing the signature verification. So when you sign the root you specify which contract will verify that.

src/Midnight.sol
MathisGD: > whitelisting

but the whitlisting is per router and the morpho asso won't control all the routers

MathisGD: > offer callbacks have arbitrary data too and the trade fails if it fails. An offer could include its own parameters in its callbackData

right

src/Midnight.sol
adhusson: removed

test/TakeTest.sol
```diff
@@ -1017,3 +1280,31 @@ contract LendCallback is ICallbacks {
 
     function onLiquidate(bytes32, Obligation memory, uint256, uint256, uint256, address, bytes memory) external {}
 }
+
+contract RatifyCallback is IRatifier {
```
MathisGD, line 1572: let's use the real one, no?

test/OtherFunctionsTest.sol
```diff
@@ -161,7 +161,7 @@ contract OtherFunctionsTest is BaseTest {
         assertEq(ERC20(collateralToken).balanceOf(receiver), withdraw, "balance of receiver");
     }
 
-    function testSetConsumed(address user, bytes32 group, uint256 amount) public {
+    function testConsume(address user, bytes32 group, uint256 amount) public {
```
MathisGD:
```suggestion
    function testSetConsumed(address user, bytes32 group, uint256 amount) public {
```

test/SignatureTest.sol
```diff
@@ -1,28 +0,0 @@
-// SPDX-License-Identifier: GPL-2.0-or-later
-// Copyright (c) 2025 Morpho Association
-pragma solidity ^0.8.0;
-
-import {Signature} from "../src/interfaces/IMidnight.sol";
-import {Midnight} from "../src/Midnight.sol";
-import {ROOT_TYPEHASH} from "../src/libraries/ConstantsLib.sol";
-import {Test} from "../lib/forge-std/src/Test.sol";
-
-contract SignatureTest is Test, Midnight {
```
MathisGD, line 10: let's re-add these

src/interfaces/ICallbacks.sol
MathisGD, line 1: would separate IRatifier in an other contract

src/ratifiers/ApprovalRatifier.sol
```diff
@@ -0,0 +1,23 @@
...
+pragma solidity 0.8.31;
+
+import {IRatifier} from "../interfaces/ICallbacks.sol";
+import {Offer} from "../interfaces/IMidnight.sol";
+import {CALLBACK_SUCCESS} from "../libraries/ConstantsLib.sol";
+
+contract ApprovalRatifier is IRatifier {
+    event SetApproval(address indexed maker, bytes32 indexed root, bool newApproval);
+
+    mapping(address maker => mapping(bytes32 root => bool)) public approved;
```
MathisGD, line 12:
```suggestion
contract ManualRatifier is IRatifier {
    event SetApproval(address indexed maker, bytes32 indexed root, bool newApproval);

    mapping(address maker => mapping(bytes32 root => bool)) public isRatified;
```

src/ratifiers/EcrecoverRatifier.sol
```diff
@@ -0,0 +1,54 @@
...
+struct Signature {
+    uint8 v;
+    bytes32 r;
+    bytes32 s;
+}
+
+bytes32 constant EIP712_DOMAIN_TYPEHASH = keccak256("EIP712Domain(uint256 chainId,address verifyingContract)");
+bytes32 constant ROOT_TYPEHASH = keccak256("Root(bytes32 root)");
+bytes32 constant AUTHORIZATION_TYPEHASH =
+    keccak256("Authorization(address authorizer,address authorized,bool isAuthorized,uint256 nonce)");
```
MathisGD: let's put that in a IEcrecover

src/ratifiers/EcrecoverRatifier.sol
```diff
@@ -0,0 +1,54 @@
...
+    function setIsAuthorizedWithSig(address authorizer, address authorized, bool isAuthorized, Signature calldata sig)
+        external
+    {
+        require(msg.sender == authorizer || IMidnight(MIDNIGHT).isAuthorized(authorizer, msg.sender), "unauthorized");
+        bytes32 structHash =
+            keccak256(abi.encode(AUTHORIZATION_TYPEHASH, authorizer, authorized, isAuthorized, nonce[authorizer]++));
+        address _signer = signer(structHash, sig);
+        require(_signer == authorizer, "invalid signature");
+        IMidnight(MIDNIGHT).setIsAuthorized(authorizer, authorized, isAuthorized);
+    }
```
MathisGD: imo, should be a separated contract

src/ratifiers/EcrecoverRatifier.sol
```diff
@@ -0,0 +1,54 @@
...
+bytes32 constant AUTHORIZATION_TYPEHASH =
+    keccak256("Authorization(address authorizer,address authorized,bool isAuthorized,uint256 nonce)");
+
+contract EcrecoverRatifier is IRatifier {
+    address public immutable MIDNIGHT;
+    mapping(address => uint256) public nonce;
+
+    constructor(address _midnight) {
+        MIDNIGHT = _midnight;
+    }
```
MathisGD, line 15:
```suggestion
```

src/Midnight.sol
MathisGD: why do we pass the full obligation again?

src/ratifiers/ApprovalRatifier.sol
adhusson: `Manual` doesn't say much, I don't think it gives an idea of what it does. 

`Approval` is a known "alternative" to signature methods: you know approval is done by making a call and not for making a signature.

Otherwise wdyt `StorageRatifier`

src/Midnight.sol
adhusson: Only loan token is needed but apart from that it just depends on what we want to give the callback to decide what to do. 

I think it's cheap, and even if we don't like it and remove `id` + `offer.obligation`, people who want that info can include the obligation id in the arbitrary data anyway...

test/TakeTest.sol
adhusson: Actually we use the mock to record stuff and choose the return value. But most tests use the real one.

src/Midnight.sol
```diff
@@ -225,9 +217,13 @@ contract Midnight is IMidnight {
         require(block.timestamp >= offer.start, "offer not started");
         require(block.timestamp <= offer.expiry, "offer expired");
         require(offer.maker != taker, "buyer and seller cannot be the same");
-        require(signer(root, sig) == offer.maker, "invalid signature");
         require(UtilsLib.isLeaf(root, keccak256(abi.encode(offer)), proof), "invalid proof");
         require(offer.session == session[offer.maker], "invalid session");
+        require(isAuthorized[offer.maker][offer.ratifier], "ratifier not authorized");
+        require(
+            IRatifier(offer.ratifier).onRatify(offer, root, ratifierData) == CALLBACK_SUCCESS, "ratification failed"
+        );
```
MathisGD, line 290: should this be view?

adhusson: imo no, others are not view so if someone wants to do something we don't want (so dont whitelist) in `onRatify`, they can do it in buy/sell anyway except if it has to do with the root.

conversely if they want to do something we are ok with in `onRatify` why give them pain

src/authorizers/SetIsAuthorizedWithSig.sol
```diff
@@ -0,0 +1,28 @@
...
+    mapping(address => uint256) public nonce;
+
+    constructor(address _midnight) {
+        MIDNIGHT = _midnight;
+    }
+
+    function setIsAuthorizedWithSig(address authorizer, address authorized, bool isAuthorized, Signature calldata sig)
+        external
+    {
+        require(msg.sender == authorizer || IMidnight(MIDNIGHT).isAuthorized(authorizer, msg.sender), "unauthorized");
```
peyha: I don't get why we have this line, isn't the goal to allow anyone with a correct signature to authorize an account ?

peyha: morpho blue's [setIsAuthorizedWithSig ](https://github.com/morpho-org/morpho-blue/blob/57d444d9e243be21a80e8d4bf8794ebce4a089d9/src/Morpho.sol#L445) also has a nonce and a deadline in its `Authorization` object, the deadline looks like a good idea but not sure about the nonce

adhusson: indeed i messed that up

peyha: [related PR](https://github.com/morpho-org/morpho-blue/pull/289) on Blue for the nonce, and [related PR](https://github.com/morpho-org/morpho-blue/pull/114) for setAuthorized (less debate than I expected around the design)

src/interfaces/IMidnight.sol
```diff
@@ -65,4 +60,15 @@ struct Position {
     uint128[128] collateral;
 }
 
-interface IMidnight {}
+struct Authorization {
```
peyha: shouldn't this be part of IEcrecover ? it's not used by midnight

src/ratifiers/ApprovalRatifier.sol
peyha: in the end, feel like `ApprovalRatifier` works well here

src/authorizers/SetIsAuthorizedWithSig.sol
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    function setIsAuthorizedWithSig(Authorization memory authorization, Signature calldata signature) external {
+        require(block.timestamp <= authorization.deadline, "expired");
+        require(authorization.nonce == nonce[authorization.authorizer]++, "invalid nonce");
+
+        bytes32 hashStruct = keccak256(abi.encode(AUTHORIZATION_TYPEHASH, authorization));
+        bytes32 domainSeparator = keccak256(abi.encode(EIP712_DOMAIN_TYPEHASH, block.chainid, address(this)));
+        bytes32 digest = keccak256(bytes.concat("\x19\x01", domainSeparator, hashStruct));
+        address signatory = ecrecover(digest, signature.v, signature.r, signature.s);
```
peyha:
```suggestion
        address signer = ecrecover(digest, signature.v, signature.r, signature.s);
```
feels slightly better

src/Midnight.sol
adhusson: @MathisGD ok for you?

src/authorizers/EcrecoverAuthorizer.sol
```diff
@@ -0,0 +1,46 @@
...
+        emit AuthorizationWithSig(
+            msg.sender,
+            authorization.authorizer,
+            authorization.authorized,
+            authorization.isAuthorized,
+            authorization.nonce
+        );
+
+        IMidnight(MIDNIGHT)
+            .setIsAuthorized(authorization.authorizer, authorization.authorized, authorization.isAuthorized);
```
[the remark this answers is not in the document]
adhusson: by design

src/ratifiers/EcrecoverRatifier.sol
```diff
@@ -0,0 +1,27 @@
...
+
+    function onRatify(Offer memory offer, bytes32 root, bytes memory data) external view returns (bytes32) {
+        Signature memory sig = abi.decode(data, (Signature));
+        bytes32 structHash = keccak256(abi.encode(ROOT_TYPEHASH, root));
+        bytes32 domainSeparator = keccak256(abi.encode(EIP712_DOMAIN_TYPEHASH, block.chainid, address(this)));
+        bytes32 digest = keccak256(bytes.concat("\x19\x01", domainSeparator, structHash));
+        address _signer = ecrecover(digest, sig.v, sig.r, sig.s);
+        require(_signer != address(0), "invalid signature");
+        require(_signer == offer.maker || IMidnight(MIDNIGHT).isAuthorized(offer.maker, _signer), "invalid signature");
```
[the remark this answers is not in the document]
adhusson: another midnight could call any other contract's `onRatify` not just this one.

src/authorizers/SetIsAuthorizedWithSig.sol
```diff
@@ -0,0 +1,46 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Copyright (c) 2025 Morpho Association
+pragma solidity 0.8.31;
+
+import {IMidnight} from "../interfaces/IMidnight.sol";
+import {Authorization, Signature, AUTHORIZATION_TYPEHASH, EIP712_DOMAIN_TYPEHASH} from "../interfaces/IEcrecover.sol";
+
+contract SetIsAuthorizedWithSig {
+    event AuthorizationWithSig(
```
peyha:
```suggestion
    event SetAuthorizationWithSig(
```
closer to midnight's event

adhusson: will try to rename the contract

test/BaseTest.sol
```diff
@@ -66,6 +72,11 @@ abstract contract BaseTest is Test {
         (otherLender, _privateKey) = makeAddrAndKey("otherLender");
         privateKey[otherLender] = _privateKey;
 
+        authorize(borrower, address(ecrecoverRatifier));
+        authorize(lender, address(ecrecoverRatifier));
+        authorize(otherBorrower, address(ecrecoverRatifier));
+        authorize(otherLender, address(ecrecoverRatifier));
```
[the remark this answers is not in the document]
adhusson: its ok

src/interfaces/IEcrecover.sol
MathisGD, line 1: should be renamed and moved in authorizers

src/interfaces/IRatifier.sol
```diff
@@ -0,0 +1,9 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Copyright (c) 2025 Morpho Association
+pragma solidity >=0.5.0;
+
+import {Offer} from "./IMidnight.sol";
+
+interface IRatifier {
+    function onRatify(Offer memory offer, bytes32 root, bytes memory data) external returns (bytes32);
```
MathisGD, line 8: isn't it weird to pass both offer and root?

certora/confs/Ratification.conf
MathisGD, line 1: to rename (the spec as well) the same name as the contract

certora/specs/EcrecoverAuthorizer.spec
```diff
@@ -0,0 +1,38 @@
...
+
+    setIsAuthorized(e1, authorization, signature);
+
+    assert e1.block.timestamp <= authorization.deadline;
+    assert authorization.nonce == nonceBefore;
+
+    setIsAuthorized(e2, otherAuthorization, otherSignature);
+
+    assert otherAuthorization.nonce != nonceBefore;
+}
```
MathisGD, line 38: possible to write a full liveness spec (!lastReverted <=> A && B && ...)

but not urgent

src/interfaces/IEcrecover.sol
adhusson: it's used by both ecrecover authorizer and ecrecover ratifier. We could inline otherwise.

MathisGD: or duplicate. not a big deal anyway

src/interfaces/IRatifier.sol
adhusson: root is for verifying signature, offer is for verifying offer information. 

Some ratifiers will look at the root only others at the offer too. If you don't fully trust the signer you want to check some things. 

For instance that the callbacks are the right addresses.

certora/confs/Ratification.conf
adhusson: the spec is about midnight ratification in general (not about EcrecoverRatifier, and there is no spec for EcrecoverRatifier it didn't seem useful)

certora/specs/EcrecoverAuthorizer.spec
adhusson: It would require adding new conditions about signature verification, but checking the verification in tests seemed ok to me.

In this spec we get `A => lastReverted` and `B => lastReverted`, etc.

But just with these conditions we won't have `lastReverted => ~A || ~B` because there are other possible reasons for the revert, like the signer is address zero.

src/interfaces/IMidnight.sol
```diff
@@ -72,4 +67,7 @@ struct Position {
     uint128[128] collateral;
 }
 
-interface IMidnight {}
+interface IMidnight {
+    function isAuthorized(address authorizer, address authorized) external view returns (bool);
+    function setIsAuthorized(address onBehalf, address authorized, bool newIsAuthorized) external;
+}
```
MathisGD, line 73:
```suggestion
interface IMidnight {}
```

test/BundlerTest.sol
```diff
@@ -73,8 +75,10 @@ contract BundlerTest is BaseTest {
     }
 
     function _authorizeBundler() internal {
-        authorize(borrower, address(takeBundler));
```
MathisGD, line 76: Why can’t we modify the helper?

test/EcrecoverRatifierTest.sol
```diff
@@ -0,0 +1,95 @@
...
+        vm.expectRevert("invalid signature");
+        ecrecoverRatifier.onRatify(offer, _root, data);
+    }
+
+    function testOnRatifyInvalidSignature() public {
+        Offer memory offer = makeOffer(lender);
+        bytes32 _root = keccak256(abi.encode(offer));
+        bytes memory data = abi.encode(Signature({v: 27, r: bytes32(uint256(1)), s: bytes32(uint256(2))}));
+
+        vm.expectRevert();
```
MathisGD:
```suggestion
        vm.expectRevert("invalid signature");
```

MathisGD (approved):

peyha (approved):
LGTM

PR #592, opened 2026-03-27
trading fee rounding: +1

MathisGD:


src/Midnight.sol
```diff
@@ -256,8 +257,9 @@ contract Midnight is IMidnight {
         uint256 _tradingFee = tradingFee(id, timeToMaturity);
         uint256 sellerPrice = offer.buy ? offerPrice - _tradingFee : offerPrice;
         uint256 buyerPrice = sellerPrice + _tradingFee;
-        uint256 buyerAssets = offer.buy ? units.mulDivDown(buyerPrice, WAD) : units.mulDivUp(buyerPrice, WAD);
-        uint256 sellerAssets = offer.buy ? units.mulDivDown(sellerPrice, WAD) : units.mulDivUp(sellerPrice, WAD);
+        uint256 buyerAssets = offer.buy ? units.mulDivDown(buyerPrice, WAD) : units.mulDivUp(buyerPrice, WAD) + 1;
+        uint256 sellerAssets =
+            offer.buy ? UtilsLib.zeroFloorSub(units.mulDivDown(sellerPrice, WAD), 1) : units.mulDivUp(sellerPrice, WAD);
```
MathisGD, line 262: the zeroFloorSub is suspicious here, can't you still bypass the fees thanks to this

src/Midnight.sol
MathisGD, line 333: if receiving 1 each time is annoying. but then we would need to verify that this never underflows (shouldn't be too hard)
```suggestion
        SafeTransferLib.safeTransferFrom(offer.obligation.loanToken, buyer, feeRecipient, buyerAssets - sellerAssets - 1);
```

qdrn: It can underflow due to `zeroFloorSub` in the computation of `sellerAssets`. If you run the test with this line, it fails.

MathisGD:
https://morpholabs.slack.com/archives/C02N7CZ088N/p1774881825257579

PR #600, opened 2026-03-29
1 - lossIndex

MathisGD:


certora/specs/BalanceEffects.spec
```diff
@@ -138,6 +142,8 @@ rule takeEffects(env e, uint256 units, address taker, address takerCallback, byt
     bytes32 id = toId(e, offer.obligation);
     address passiveFeeRecipient = Utils.passiveFeeRecipient();
 
+    // The obligation must be created for updatePositionView to reflect the post-touchObligation state.
+    require Midnight.obligationCreated(id);
```
MathisGD: we could prove that it gives the same thing (because obligationNotCreated => credit==0)

adhusson (approved):
lgtm modulo codex comments

src/Midnight.sol
```diff
@@ -629,9 +625,7 @@ contract Midnight is IMidnight {
         Position storage _position = position[id][user];
         uint128 credit = _position.credit;
         uint128 _lossIndex = _position.lossIndex;
-        uint256 postSlashCredit = _lossIndex < type(uint128).max
-            ? credit.mulDivDown(type(uint128).max - obligationState[id].lossIndex, type(uint128).max - _lossIndex)
-            : 0;
+        uint256 postSlashCredit = _lossIndex > 0 ? credit.mulDivDown(obligationState[id].lossIndex, _lossIndex) : 0;
```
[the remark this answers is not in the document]
MathisGD: outdated

src/Midnight.sol
```diff
@@ -629,9 +627,8 @@ contract Midnight is IMidnight {
         Position storage _position = position[id][user];
         uint128 credit = _position.credit;
         uint128 _lossIndex = _position.lossIndex;
-        uint256 postSlashCredit = _lossIndex < type(uint128).max
-            ? credit.mulDivDown(type(uint128).max - obligationState[id].lossIndex, type(uint128).max - _lossIndex)
-            : 0;
+        uint256 postSlashCredit =
+            credit.mulDivDown(obligationState[id].lossIndex, _lossIndex > 0 ? _lossIndex : type(uint128).max);
```
MathisGD, line 631: note, the if-else here is very important for the fee recipient, because they can receive credit before their first update (so with lossIndex = 0)

it's also useful to avoid DOS when loss=100% (same as the if-else of before)

MathisGD:
https://morpholabs.slack.com/archives/C0ACSA8TNMA/p1774897616704409

PR #608, opened 2026-03-30
[Certora] Check that non-zero credit is not accessed before update even in take

jhoenicke:
This fixes the updateBeforeCredit spec to include take again.

### DIFF
--- a/src/Midnight.sol
+++ b/src/Midnight.sol
@@ -272,8 +272,8 @@ contract Midnight is IMidnight {
         Position storage buyerPos = position[id][buyer];
         Position storage sellerPos = position[id][seller];
 
-        if (buyerPos.credit > 0 || units > buyerPos.debt) _updatePosition(offer.obligation, id, buyer);
-        if (sellerPos.credit > 0) _updatePosition(offer.obligation, id, seller);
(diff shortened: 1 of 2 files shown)


src/Midnight.sol
```diff
@@ -673,6 +673,10 @@ contract Midnight is IMidnight {
         emit EventsLib.UpdatePosition(id, user, creditDecrease, pendingFeeDecrease, accruedFee);
     }
 
+    function hasCredit(bytes32 id, address user) internal view returns (bool) {
```
jhoenicke, line 676: I also tried using `Position storage` argument here.  Unfortunately storage arguments are not supported by the Certora Prover.

MathisGD, line 676: how painful would it be to do a munging for this rule?

jhoenicke: We can't do it with wrapper contracts, but would need to setup munging to copy the contract, apply a diff, and then run verification on the copy.  We don't have any setup for munging yet, and it adds the risk that we verify a different contract.

MathisGD (approved):
putting a high bounty on finding a solution that doesn't change src ^^

but worth it anyway

adhusson (approved):

PR #617, opened 2026-03-31
pull from buyer callbacks

adhusson:
see this [thread](https://github.com/morpho-org/midnight/pull/617#discussion_r3031887485) for the return value rationale.


src/Midnight.sol
```diff
@@ -337,13 +330,18 @@ contract Midnight is IMidnight {
             sellerCreditDecrease
         );
 
-        if (buyerCallback != address(0)) {
-            ICallbacks(buyerCallback)
-                .onBuy(id, offer.obligation, buyer, buyerAssets, sellerAssets, units, buyerCallbackData);
-        }
+        uint256 balanceBefore = IERC20(offer.obligation.loanToken).balanceOf(address(this));
```
adhusson: Against draining contracts that don't revert  on `onBuy`, use a balance check instead of `transferFrom`.

src/Midnight.sol
```diff
@@ -337,13 +330,18 @@ contract Midnight is IMidnight {
...
+
+        ICallbacks(buyerCallback)
```
adhusson: Even takers must use a callback. Could do something else like "if buyer is taker, pull from taker".

src/Midnight.sol
adhusson: fixable with magic value

src/callbacks/TransferFromCallback.sol
adhusson, line 1: this is mostly for takers

src/callbacks/IdleCallback.sol
```diff
@@ -0,0 +1,44 @@
...
+
+import {Obligation} from "../interfaces/IMidnight.sol";
+import {ICallbacks} from "../interfaces/ICallbacks.sol";
+import {IERC20} from "../interfaces/IERC20.sol";
+import {CALLBACK_SUCCESS} from "../libraries/ConstantsLib.sol";
+import {SafeTransferLib} from "../libraries/SafeTransferLib.sol";
+
+contract IdleCallback is ICallbacks {
+    address public immutable MIDNIGHT;
+    mapping(address user => mapping(address token => uint256)) public balances;
```
MathisGD: is there a particular reason for one callback for everybody (vs a factory of callbacks)

MathisGD: for "withdraw from blue" callback I prefer one per person otherwise this callback would have a massive position on blue, which can be weird

src/callbacks/IdleCallback.sol
```diff
@@ -0,0 +1,44 @@
...
+        SafeTransferLib.safeTransfer(token, msg.sender, amount);
+    }
+
+    function onBuy(bytes32, Obligation memory obligation, address buyer, uint256 buyerAssets, uint256, bytes memory)
+        external
+        returns (bytes32)
+    {
+        require(msg.sender == MIDNIGHT);
+        balances[buyer][obligation.loanToken] -= buyerAssets;
+        IERC20(obligation.loanToken).approve(MIDNIGHT, buyerAssets);
```
MathisGD: why not infinite approval?

src/callbacks/IdleCallback.sol
MathisGD, line 1: what is idleCallback useful for? Escrow? Because with no callback the buyer is the funding source

adhusson: also a callback example, good to have at least one

src/callbacks/IdleCallback.sol
adhusson: related to the singleton question:
- on ethereum glamsterdam will likely increase contract creation cost. But in the future contracts with big storage could cost more to interact with.
- on tempo contract creation is 500k gas base cost, + a 5x increase in gas per byte of code size

src/Midnight.sol
```diff
@@ -348,13 +349,23 @@ contract Midnight is IMidnight {
             sellerCreditDecrease
         );
 
+        address fundingSource;
         if (buyerCallback != address(0)) {
-            ICallbacks(buyerCallback).onBuy(id, offer.obligation, buyer, buyerAssets, units, buyerCallbackData);
+            require(
+                ICallbacks(buyerCallback).onBuy(id, offer.obligation, buyer, buyerAssets, units, buyerCallbackData)
+                    == CALLBACK_SUCCESS,
```
peyha, line 355: why is `CALLBACK_SUCCESS` required here ? (and not for onSell)

MathisGD: some smart-wallets don't revert on `fallback()`

so you could rug those who approved morpho by putting them as your callback.

Buyer callbacks need to auth the buyer, and this serves as a confirmation that it has been auth, in some way.

src/callbacks/IdleCallback.sol
MathisGD: let's see later

src/Midnight.sol
peyha: I see, then shouldn't we also have the same require for `onSell` to normalize the behavior between callback contracts in `take`(`onRatify` also has it)

test/OtherFunctionsTest.sol
```diff
@@ -576,6 +577,16 @@ contract OtherFunctionsTest is BaseTest {
         midnight.touchObligation(_obligation);
         assertEq(midnight.obligationCreated(toId(_obligation)), true, "obligation created with cursor 0.5");
     }
+
+    function testMidnightRevertsOnCallbacks(address sender, bytes calldata data) public {
```
peyha: what's the goal of this test ? ensuring that using midnight as a callback does not work ?

test/OtherFunctionsTest.sol
```diff
@@ -576,6 +577,16 @@ contract OtherFunctionsTest is BaseTest {
...
+        bytes4[3] memory selectors =
+            [ICallbacks.onBuy.selector, ICallbacks.onSell.selector, ICallbacks.onLiquidate.selector];
```
peyha:
```suggestion
            [ICallbacks.onBuy.selector, ICallbacks.onSell.selector];
```
No need to test onLiquidate because it's called on `msg.sender`

peyha (commented):
All good for me once issues are resolved.
The complexity pushed to the callback is very limited given that the callback already had to handle tokens and approvals but the buyer does not have to approve midnight anymore (only the callback) so it's ok

src/Midnight.sol
MathisGD: onSell technically doesn't need to auth. you can see it as a receiver. but I don't see a big deal with having them symmetric

test/OtherFunctionsTest.sol
MathisGD: yes, that you cannot rug midnight by putting it as a buyerCallback

src/libraries/ConstantsLib.sol
```diff
@@ -14,6 +14,7 @@ uint256 constant MAX_COLLATERALS_PER_BORROWER = 10;
 uint256 constant LIQUIDATION_CURSOR_LOW = 0.25e18;
 uint256 constant LIQUIDATION_CURSOR_HIGH = 0.5e18;
 address constant CONTINUOUS_FEE_RECIPIENT = address(uint160(uint256(keccak256("continuous fee recipient"))));
+bytes32 constant CALLBACK_SUCCESS = keccak256("MIDNIGHT CALLBACK SUCCESS");
```
MathisGD, line 17: would it be safer to have a hash of the midnight address and chain?

src/Midnight.sol
adhusson: agree i can add it

test/OtherFunctionsTest.sol
MathisGD: cheap to check, no?

peyha: yes, but in this case we can add `onRepay`

src/libraries/ConstantsLib.sol
adhusson: I don't really see it, contracts decide to say yes/no depending on the current chain state. Only scenario I see is another contract called Midnight exists? Then including the address would be useful. But it's so unlikely.

Chat did not find examples either, all EIPs / contracts that do magic value use hash(constant).

wdyt?

MathisGD: indeed, fine. maybe though we should document somewhere what checks should a buyer callback do to be safe

test/OtherFunctionsTest.sol
MathisGD: agree

src/Midnight.sol
```diff
@@ -348,13 +349,23 @@ contract Midnight is IMidnight {
...
+                "invalid callback"
+            );
+            fundingSource = buyerCallback;
+        } else {
+            fundingSource = offer.buy ? buyer : msg.sender;
```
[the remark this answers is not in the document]
MathisGD: that's the feature

src/Midnight.sol
```diff
@@ -348,16 +349,30 @@ contract Midnight is IMidnight {
             sellerCreditDecrease
         );
 
+        address fundingSource;
```
```suggestion
        address fundingSource = buyerCallback != address(0) ? buyerCallback : (offer.buy ? buyer : msg.sender);
```

MathisGD: parallel with receiver?
```suggestion
        address sender;
```

MathisGD (approved):

src/Midnight.sol
peyha: too close to `msg.sender` imo

MathisGD: searching an other name

src/Midnight.sol
```diff
@@ -349,23 +349,18 @@ contract Midnight is IMidnight {
...
-            fundingSource = buyerCallback;
-        } else {
-            fundingSource = offer.buy ? buyer : msg.sender;
         }
 
-        SafeTransferLib.safeTransferFrom(
-            offer.obligation.loanToken, fundingSource, address(this), buyerAssets - sellerAssets
-        );
+        address sender = buyerCallback != address(0) ? buyerCallback : (offer.buy ? buyer : msg.sender);
+        SafeTransferLib.safeTransferFrom(offer.obligation.loanToken, sender, address(this), buyerAssets - sellerAssets);
```
adhusson: prefer fundingSource, or `payer` but agree with @peyha sender weird

MathisGD: payer > funder?

MathisGD: payer is better

MathisGD (approved):

src/Midnight.sol
```diff
@@ -349,15 +350,24 @@ contract Midnight is IMidnight {
...
-            ICallbacks(buyerCallback).onBuy(id, offer.obligation, buyer, buyerAssets, units, buyerCallbackData);
+            require(
+                ICallbacks(buyerCallback).onBuy(id, offer.obligation, buyer, buyerAssets, units, buyerCallbackData)
+                    == CALLBACK_SUCCESS,
+                "invalid callback"
+            );
         }
 
-        SafeTransferLib.safeTransferFrom(offer.obligation.loanToken, buyer, address(this), buyerAssets - sellerAssets);
+        address payer = buyerCallback != address(0) ? buyerCallback : (offer.buy ? buyer : msg.sender);
```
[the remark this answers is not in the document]
MathisGD: this is the feature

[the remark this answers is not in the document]
peyha: indeed we should add a test in `BundlerTest` with sell offers

peyha (approved):

PR #619, opened 2026-04-01
modular ticks

MathisGD:
Rationale: https://morpholabs.slack.com/archives/C02N7CZ088N/p1774913820449649


test/TickLibTest.sol
```diff
@@ -14,8 +14,8 @@ contract TickLibTest is BaseTest {
 
     function testTickToPriceMinMax() public pure {
         assertEq(TickLib.tickToPrice(0), 0, "tick 0");
-        assertEq(TickLib.tickToPrice(1), 5e12, "tick 1");
-        assertEq(TickLib.tickToPrice(MAX_TICK - 1), 0.999995e18, "tick max - 1");
+        assertGt(TickLib.tickToPrice(14), 0, "first non-zero tick");
+        assertGt(TickLib.tickToPrice(MAX_TICK - 1), 0.999e18, "tick max - 1 near par");
```
MathisGD: we can adapt the tick, but we care about the last one that's not one
```suggestion
        assertEq(TickLib.tickToPrice(MAX_TICK - 1), 0.999e18, "tick max - 1 near par");
```

test/TakeAmountsTest.sol
```diff
@@ -90,6 +90,7 @@ contract TakeAmountsTest is BaseTest {
         uint256 tradingFee = _setFees(fee0, fee1);
         targetBuyerAssets = bound(targetBuyerAssets, 1, 1e30);
         tick = bound(tick, 1, _maxTick(tradingFee));
+        vm.assume(TickLib.tickToPrice(tick) > 0);
```
MathisGD: we should remove these

test/TakeTest.sol
```diff
@@ -203,7 +203,7 @@ contract TakeTest is BaseTest {
 
     function testBuy3(uint256 units, uint256 tick, uint256 existingUnits) public {
         units = bound(units, 0, maxAssets);
-        tick = bound(tick, 0, 600);
+        tick = bound(tick, 0, MAX_TICK / 4);
```
MathisGD: we should go up to maxTick

src/Midnight.sol
```diff
@@ -270,6 +304,10 @@ contract Midnight is IMidnight {
                 offer.receiverIfMakerIsSeller
             );
 
+        require(
+            offer.tick % (BASE_SPACING >> _obligationState.tickLevel) == 0, "tick not accessible at obligation level"
```
MathisGD: we could store the spacing directly
```suggestion
            offer.tick % obligation.spacing, "tick not accessible at obligation level"
```

src/interfaces/IMidnight.sol
```diff
@@ -51,6 +51,7 @@ struct ObligationState {
     uint128 lossIndex;
     uint256 withdrawable;
     bool created;
+    uint8 spacing;
     uint16[7] fees;
     uint32 continuousFee;
```
MathisGD: fits in one slot

src/Midnight.sol
```diff
@@ -157,6 +163,34 @@ contract Midnight is IMidnight {
...
+    /// @dev Sets the default spacing for new obligations with the given loan token.
+    /// @dev Unlike per-obligation spacing (which can only decrease), the default can be freely changed because it only
+    /// affects future obligations at creation time. Existing obligations and their signed offers are never impacted.
+    function setDefaultSpacing(address loanToken, uint256 newSpacing) external {
+        require(msg.sender == tickSetter, "only tick setter");
+        require(newSpacing > 0 && BASE_SPACING % newSpacing == 0, "invalid spacing");
+        // forge-lint: disable-next-line(unsafe-typecast) as newSpacing <= BASE_SPACING < type(uint8).max
+        defaultSpacing[loanToken] = uint8(newSpacing);
+        emit EventsLib.SetDefaultSpacing(loanToken, newSpacing);
+    }
```
MathisGD: not sure that it's useful

src/libraries/TickLib.sol
```diff
@@ -2,17 +2,17 @@
...
-int256 constant LN_ONE_PLUS_DELTA = 0.024692612590371501e18; // ln(1 + 0.025)
-uint256 constant MAX_TICK = 1046;
+int256 constant LN_ONE_PLUS_DELTA = 0.004987541511038968e18; // ln(1 + 0.005)
+uint256 constant MAX_TICK = 5844;
+uint256 constant BASE_SPACING = 4;
 
 library TickLib {
     using TickLib for uint256;
 
+    /// @dev Returns `x` / `d` rounded to the nearest integer with ties rounded down, without checking for overflow.
```
peyha: should be moved below errors

src/Midnight.sol
```diff
@@ -125,13 +125,21 @@ contract Midnight is IMidnight {
...
+    /// @dev Address that can set tick accessibility levels.
+    address public tickSetter;
+
+    /// @dev Default spacing per loan token. Applied when obligations are created.
+    mapping(address loanToken => uint8) public defaultSpacing;
+
     /// CONSTRUCTOR ///
 
     constructor() {
+        owner = msg.sender;
```
peyha: why use a new role for this ? it's also a roleSetter in the end

MathisGD: it's a rebasing issue

src/Midnight.sol
```diff
@@ -175,6 +176,24 @@ contract Midnight is IMidnight {
...
+        tickSetter = newTickSetter;
+        emit EventsLib.SetTickSetter(newTickSetter);
+    }
+
+    /// @dev Refines the tick spacing of a specific obligation. Can only decrease (more ticks become accessible).
+    function setObligationSpacing(bytes32 id, uint256 newSpacing) external {
+        require(msg.sender == tickSetter, "only tick setter");
+        require(newSpacing > 0 && BASE_SPACING % newSpacing == 0, "invalid spacing");
+        require(obligationState[id].created, "obligation not created");
+        require(newSpacing < obligationState[id].spacing, "can only decrease spacing");
```
peyha: not sure we need this, assessing that we need to change the spacing is difficult so we might need room for errors and it only affects future offers (just like the trading fee) so it should be fine in terms of user guarantees

src/Midnight.sol
```diff
@@ -231,6 +232,24 @@ contract Midnight is IMidnight {
...
+    function setTickSetter(address newTickSetter) external {
+        require(msg.sender == roleSetter, OnlyRoleSetter());
+        tickSetter = newTickSetter;
+        emit EventsLib.SetTickSetter(newTickSetter);
+    }
+
+    /// @dev Refines the tick spacing of a specific obligation. Can only decrease (more ticks become accessible).
+    function setObligationSpacing(bytes32 id, uint256 newSpacing) external {
+        require(msg.sender == tickSetter, "only tick setter");
+        require(newSpacing > 0 && BASE_SPACING % newSpacing == 0, "invalid spacing");
```
adhusson: need custom errors

src/libraries/TickLib.sol
```diff
@@ -2,8 +2,9 @@
 // Copyright (c) 2025 Morpho Association
 pragma solidity ^0.8.0;
 
-int256 constant LN_ONE_PLUS_DELTA = 0.024692612590371501e18; // floor(ln(1 + 0.025) * 1e18)
-uint256 constant MAX_TICK = 1046;
+int256 constant LN_ONE_PLUS_DELTA = 0.004987541511038968e18; // ln(1 + 0.005)
```
adhusson: I get `0.004987541511039073e18` (floor(ln(1 + 0.025) * 1e18)) with

```
>>> math.log(1.005)
0.004987541511038968
>>> from mpmath import mp, log
>>> mp.dps = 40
>>> print(f"0.{int(log(mp.mpf('1.005')) * 10**18):018d}")
0.004987541511039073
```

test/TickGatingTest.sol
adhusson, line 1: Should test that changing the grain in one obligation does not impact another, or spec it?

src/Midnight.sol
adhusson: if the spacing reincreases an offer can become invalid

MathisGD: we want the guarantee that it won't cancel any offers

src/Midnight.sol
```diff
@@ -230,6 +231,24 @@ contract Midnight is IMidnight {
...
+        require(obligationState[id].created, "obligation not created");
+        require(newSpacing < obligationState[id].spacing, "can only decrease spacing");
+        // forge-lint: disable-next-line(unsafe-typecast) as newSpacing <= BASE_SPACING < type(uint8).max
+        obligationState[id].spacing = uint8(newSpacing);
+        emit EventsLib.SetObligationSpacing(id, newSpacing);
+    }
+
+    /// @dev Overrides the fee of a specific obligation.
```
```suggestion
```

src/interfaces/IMidnight.sol
```diff
@@ -51,6 +51,7 @@ struct ObligationState {
     uint16 tradingFee5;
     uint16 tradingFee6;
     uint32 continuousFee;
+    uint8 spacing;
     bool created;
```
MathisGD:
```suggestion
    uint8 spacing;
```

src/libraries/TickLib.sol
qdrn: Same on Wolfram Alpha.

src/Midnight.sol
```diff
@@ -337,6 +356,8 @@ contract Midnight is IMidnight {
 
         (address buyer, address seller) = offer.buy ? (offer.maker, taker) : (taker, offer.maker);
 
+        require(offer.tick % _obligationState.spacing == 0, "tick not accessible at obligation level");
```
qdrn: custom error?

src/Midnight.sol
```diff
@@ -230,6 +231,24 @@ contract Midnight is IMidnight {
         emit EventsLib.SetFeeClaimer(newFeeClaimer);
     }
 
+    function setTickSetter(address newTickSetter) external {
```
qdrn: nit: maybe there's a better name than tick setter. The role does not set the tick but the grid? spacing?

src/Midnight.sol
qdrn: Another possibility is to add a timelock but pretty sure we don't want to go that route.

test/TickGatingTest.sol
MathisGD: felt that it was unecessary, it's pretty clear that it's isolated to me

src/Midnight.sol
```diff
@@ -857,7 +878,11 @@ contract Midnight is IMidnight {
     }
 
     function obligationCreated(bytes32 id) external view returns (bool) {
-        return obligationState[id].created;
+        return obligationState[id].spacing > 0;
+    }
```
adhusson: wdyt make this public and call `obligationCreated(id)` everywhere `obligationState[id].spacing > 0` is used, for clarity

src/Midnight.sol
```diff
@@ -231,13 +232,30 @@ contract Midnight is IMidnight {
...
+        spacingSetter = newSpacingSetter;
+        emit EventsLib.SetSpacingSetter(newSpacingSetter);
+    }
+
+    /// @dev Refines the tick spacing of a specific obligation. Can only decrease (more ticks become accessible).
+    function setObligationSpacing(bytes32 id, uint256 newSpacing) external {
+        require(msg.sender == spacingSetter, OnlySpacingSetter());
+        require(newSpacing > 0 && BASE_SPACING % newSpacing == 0, InvalidSpacing());
+        require(obligationState[id].spacing > 0, ObligationNotCreated());
+        require(newSpacing < obligationState[id].spacing, SpacingNotDecreasing());
```
```suggestion
        require(obligationState[id].spacing > 0, ObligationNotCreated());
        require(newSpacing < obligationState[id].spacing, SpacingNotDecreasing());
        require(newSpacing == 1 || newSpacing == 2, InvalidSpacing());
```

adhusson: (also the BASE_SPACING constant is maybe overkill)

QGarchery:
```suggestion
        require(obligationState[id].spacing > 0, ObligationNotCreated());
        require(newSpacing > 0 && obligationState[id].spacing % newSpacing == 0, InvalidSpacing());
```
because then it's correct without depending on the BASE_SPACING (original code did depend on it, because with a base spacing of 6 you could go from 3 to 2)

adhusson: this allows noop, but its ok (should clarify comment at the top in that case to nonincreasing)

test/TickLibTest.sol
```diff
@@ -26,13 +26,16 @@ contract TickLibTest is BaseTest {
...
+        // accommodates this while still verifying the geometric structure holds.
+        for (uint256 i = 1400; i <= 4600; i++) {
             uint256 previousReturn = _return(TickLib.tickToPrice(i - 1));
             uint256 currentReturn = _return(TickLib.tickToPrice(i));
             assertApproxEqRel(
                 currentReturn.mulDivDown(1e18, previousReturn),
-                0.975e18,
-                0.005e18,
+                1.005e18,
+                0.03e18,
```
QGarchery: This test is broken again, it goes in the wrong direction and the relative tolerance is way too big

test/TickLibTest.sol
```diff
@@ -14,8 +14,8 @@ contract TickLibTest is BaseTest {
 
     function testTickToPriceMinMax() public pure {
         assertEq(TickLib.tickToPrice(0), 0, "tick 0");
-        assertEq(TickLib.tickToPrice(1), 5e12, "tick 1");
-        assertEq(TickLib.tickToPrice(MAX_TICK - 1), 0.999995e18, "tick max - 1");
+        assertEq(TickLib.tickToPrice(14), 1e12, "first non-zero tick");
+        assertEq(TickLib.tickToPrice(MAX_TICK - 1), 1e18, "tick max - 1 near par");
```
QGarchery: Should we redefine MAX_TICK so that if you sutract 1 it gives a price different from 1e18 ?

src/Midnight.sol
```diff
@@ -189,6 +189,7 @@ contract Midnight is IMidnight {
     address public roleSetter;
     address public feeSetter;
     address public feeClaimer;
+    address public spacingSetter;
```
peyha: not a huge fan of using spacing without mentioning ticks but no big deal
```suggestion
    address public tickSpacingSetter;
```

test/TakeTest.sol
```diff
@@ -1390,32 +1390,36 @@ contract TakeTest is BaseTest {
...
         borrowerOffer.maxUnits = units;
+        uint256 price = TickLib.tickToPrice(0);
+        uint256 expectedAssets = units.mulDivUp(price, WAD);
+        deal(address(loanToken), lender, expectedAssets);
         collateralize(obligation, borrower, units);
         (uint256 buyerAssets, uint256 sellerAssets) = take(units, lender, borrowerOffer);
-        assertEq(buyerAssets, 0, "buyerAssets");
-        assertEq(sellerAssets, 0, "sellerAssets");
+        assertEq(buyerAssets, expectedAssets, "buyerAssets");
+        assertEq(sellerAssets, expectedAssets, "sellerAssets");
```
QGarchery: this change is not needed

QGarchery: same below as `tickToPrice(0)` is `0`

src/Midnight.sol
peyha, line 155: worth a comment here (or in a dedicated section)

test/TickLibTest.sol
```diff
@@ -105,37 +108,42 @@ contract TickLibTest is BaseTest {
         uint256 maxRelErrorWad;
         uint256 totalAbsErrorWad;
         uint256 totalRelErrorWad;
+        uint256 count;
 
-        for (uint256 tick = 0; tick <= MAX_TICK; tick++) {
+        // Sample every 10th tick to stay within memory limits.
+        for (uint256 tick = 0; tick <= MAX_TICK; tick += 10) {
```
QGarchery:
```suggestion
        for (uint256 tick = 0; tick <= MAX_TICK; tick++) {
```
it doesn't seem to be an issue when I test locally. We can remove `count` as well then

test/ticks_exact_gen.py
```diff
@@ -1,11 +1,12 @@
...
 
-MAX_TICK = 1046
+MAX_TICK = 5844
 
 def exact_price(tick):
-    """Reference tick to price implementation."""
-    return int(1e18 / (1 + 1.025 ** (MAX_TICK / 2 - tick)))
+    """Reference tick to price implementation matching Solidity rounding."""
+    raw = 1e18 / (1 + 1.005 ** (MAX_TICK / 2 - tick))
+    return round(raw / 1e12) * int(1e12)
```
QGarchery: We don't want the rounding here, this is something we want to test as well: making sure that it doesn't create too much difference with the "exact" version

src/Midnight.sol
adhusson: do you mean changing it everywhere (functions `spacing(id)` and `setObligationSpacing(id)`, event `SetObligationSpacing`, error `InvalidSpacing`)?

test/TickLibTest.sol
```diff
@@ -105,37 +108,42 @@ contract TickLibTest is BaseTest {
...
             uint256 solPrice = TickLib.tickToPrice(tick);
             uint256 exactPrice = exactPrices[tick];
 
             uint256 absErrorWad = absDiff(solPrice, exactPrice);
             maxAbsErrorWad = max(maxAbsErrorWad, absErrorWad);
             totalAbsErrorWad += absErrorWad;
-            uint256 relErrorWad = absDiff(solPrice, exactPrice) * 1e18 / exactPrice;
+            uint256 relErrorWad = exactPrice > 0 ? absDiff(solPrice, exactPrice) * 1e18 / exactPrice : 0;
```
QGarchery: remove this change as well once the above change has been applied

src/Midnight.sol
QGarchery: no strong opinion, current code may be ok because the error `ObligationNotCreated` is helping already

src/Midnight.sol
```diff
@@ -335,6 +353,8 @@ contract Midnight is IMidnight {
 
         (address buyer, address seller) = offer.buy ? (offer.maker, taker) : (taker, offer.maker);
 
+        require(offer.tick % _obligationState.spacing == 0, TickNotAccessible());
```
```suggestion
        require(offer.tick % _obligationState.spacing == 0, TickNotAccessible());
(address buyer, address seller) = offer.buy ? (offer.maker, taker) : (taker, offer.maker);


```

test/TickLibTest.sol
```diff
@@ -105,37 +108,42 @@ contract TickLibTest is BaseTest {
...
             if (solPrice > 0.01e18) {
-                assertLe(relErrorWad, 0.001e18, string.concat("Tick ", vm.toString(tick), " error exceeds 0.1%"));
+                assertLe(relErrorWad, 0.0015e18, string.concat("Tick ", vm.toString(tick), " error exceeds 0.15%"));
             }
 
-            // Check exact price is bracketed by adjacent sol prices (only where prices vary per-tick)
-            if (tick > 0 && tick < MAX_TICK) {
+            // Check exact price is bracketed by adjacent sol prices (only where the rounding step
+            // is small relative to the 0.1% geometric step).
+            if (solPrice > 0.01e18 && solPrice < 0.99e18) {
```
QGarchery: This change shouldn't be needed, and it's not needed it seems when I test locally

test/TickLibTest.sol
adhusson: it looks like 5818 which is not divisible by 4, too bad no? I mean usability wise it sounds worse than the last few ticks being 1.

the lowest one divisible by 4 that reaches 1e18 (but its predecessor also does) is 5820

src/Midnight.sol
adhusson: wdym?

peyha: i meant specifying the tick spacing thing in these comments

src/Midnight.sol
MathisGD: imo we should remove the function completely

adhusson: doing the opposite, no `obligationCreated`

src/Midnight.sol
peyha: Ideally everywhere but that would make the code less readable so not sure

test/MidnightBundlesTest.sol
```diff
@@ -1198,7 +1200,7 @@ contract MidnightBundlesTest is BaseTest {
     // Average price.
 
     function testBuyUnitsTargetAveragePriceExceeded(uint256 tick) public {
-        tick = bound(tick, 1, MAX_TICK);
+        tick = bound(tick, 4, MAX_TICK / BASE_SPACING) * BASE_SPACING;
```
peyha:
```suggestion
        tick = bound(tick, 1, MAX_TICK / BASE_SPACING) * BASE_SPACING;
```

src/Midnight.sol
adhusson: see in https://github.com/morpho-org/midnight/pull/862/

test/TakeTest.sol
adhusson: https://github.com/morpho-org/midnight/pull/862/

test/TickLibTest.sol
adhusson: https://github.com/morpho-org/midnight/pull/862/ with 5820 (so not optimal for subtracting / adding 1 to get non-max or non-zero, but closer)

test/TickLibTest.sol
adhusson: https://github.com/morpho-org/midnight/pull/862/

test/TickGatingTest.sol
```diff
@@ -0,0 +1,190 @@
...
+
+    function testSetSpacingSetterOnlyOwner() public {
+        vm.prank(lender);
+        vm.expectRevert(IMidnight.OnlyRoleSetter.selector);
+        midnight.setSpacingSetter(lender);
+    }
+
+    // --- Coarser ticks remain valid after refinement ---
+
+    function testCoarserTicksStillValidAfterRefinement() public {
```
peyha: should we fuzz ?

src/Midnight.sol
adhusson: https://github.com/morpho-org/midnight/pull/862/

adhusson: (but we could still delete BASE_SPACING)

test/TradingFeeTest.sol
```diff
@@ -83,7 +83,7 @@ contract TradingFeeTest is BaseTest {
 
     function testBuyUnits(uint256 tradingFee, uint256 sellerTick, uint256 units) public {
         units = bound(units, 0, MAX_DEBT);
-        sellerTick = bound(sellerTick, 0, MAX_TICK);
+        sellerTick = bound(sellerTick, 0, MAX_TICK / BASE_SPACING) * BASE_SPACING;
```
MathisGD: to clean, we could deactivate the spacing in the test (set as 1)

src/interfaces/IMidnight.sol
```diff
@@ -113,8 +119,9 @@ interface IMidnight {
 
     /// STORAGE GETTERS ///
     function position(bytes32 id, address user) external view returns (uint128 credit, uint128 pendingFee, uint128 lastLossFactor, uint128 lastAccrual, uint128 debt, uint128 collateralBitmap);
-    function obligationState(bytes32 id) external view returns (uint128 totalUnits, uint128 lossFactor, uint128 withdrawable, uint128 continuousFeeCredit, uint16 tradingFeeCbp0, uint16 tradingFeeCbp1, uint16 tradingFeeCbp2, uint16 tradingFeeCbp3, uint16 tradingFeeCbp4, uint16 tradingFeeCbp5, uint16 tradingFeeCbp6, uint32 continuousFee, bool created);
+    function obligationState(bytes32 id) external view returns (uint128 totalUnits, uint128 lossFactor, uint128 withdrawable, uint128 continuousFeeCredit, uint16 tradingFeeCbp0, uint16 tradingFeeCbp1, uint16 tradingFeeCbp2, uint16 tradingFeeCbp3, uint16 tradingFeeCbp4, uint16 tradingFeeCbp5, uint16 tradingFeeCbp6, uint32 continuousFee, uint8 spacing);
```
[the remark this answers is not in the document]
adhusson: done in https://github.com/morpho-org/midnight/pull/862

test/TickLibTest.sol
```diff
@@ -26,12 +26,12 @@ contract TickLibTest is BaseTest {
...
 
     function testReturnJumps() public pure {
-        for (uint256 i = 250; i <= 800; i++) {
+        for (uint256 i = 1400; i <= 4600; i++) {
             uint256 previousReturn = _return(TickLib.tickToPrice(i - 1));
             uint256 currentReturn = _return(TickLib.tickToPrice(i));
             assertApproxEqRel(
                 currentReturn.mulDivDown(1e18, previousReturn),
-                0.975e18,
+                0.995024875621890547e18,
```
QGarchery:
```suggestion
                0.995e18,
```
??

test/TickLibTest.sol
QGarchery: Sounds like it's the best we can do 👍

src/libraries/TickLib.sol
QGarchery, line 1: didn't check that the change delta from 2.5% to 0.5%  and the change of rounding from 5e-6 to 1e-6 was what was the quantitative research by @MattLsb recommended

QGarchery (commented):
Approving, modulo the tests are fixed as suggested above, so that we have enough confidence that the new tick set is close to the exact version

QGarchery (approved):

src/libraries/TickLib.sol
MathisGD: or completely remove?
```suggestion
    function priceToTick(uint256 price, uint256 spacing) internal pure returns (uint256) {
```

src/libraries/TickLib.sol
```diff
@@ -43,7 +43,7 @@ library TickLib {
             // forge-lint: disable-next-item(unsafe-typecast)
             return uint256(1e36)
                     .divHalfDownUnchecked(1e18 + wExp(LN_ONE_PLUS_DELTA * (int256(MAX_TICK / 2) - int256(tick))))
-                    .divHalfDownUnchecked(5e12) * 5e12;
+                    .divHalfDownUnchecked(1e12) * 1e12;
```
peyha: would it be worth introducing a var for this ?

src/libraries/TickLib.sol
adhusson: The code does not add bytes to the midnight contract and it would save time to users if we provide an inverse, so I don't see why we should remove it. But yes we should add a spacing parameter.

test/MidnightBundlesTest.sol
```diff
@@ -1233,7 +1235,7 @@ contract MidnightBundlesTest is BaseTest {
     }
 
     function testSellUnitsTargetAveragePriceTooLow(uint256 tick) public {
-        tick = bound(tick, 1, MAX_TICK);
+        tick = bound(tick, 4, MAX_TICK / DEFAULT_TICK_SPACING) * DEFAULT_TICK_SPACING;
```
peyha:
```suggestion
        tick = bound(tick, 1, MAX_TICK / DEFAULT_TICK_SPACING) * DEFAULT_TICK_SPACING;
```

test/MidnightBundlesTest.sol
```diff
@@ -1265,7 +1267,7 @@ contract MidnightBundlesTest is BaseTest {
     }
 
     function testBuyBuyerAssetsTargetAveragePriceExceeded(uint256 tick) public {
-        tick = bound(tick, 1, MAX_TICK);
+        tick = bound(tick, 4, MAX_TICK / DEFAULT_TICK_SPACING) * DEFAULT_TICK_SPACING;
```
peyha:
```suggestion
        tick = bound(tick, 1, MAX_TICK / DEFAULT_TICK_SPACING) * DEFAULT_TICK_SPACING;
```

test/MidnightBundlesTest.sol
```diff
@@ -1300,7 +1302,7 @@ contract MidnightBundlesTest is BaseTest {
     }
 
     function testSellSellerAssetsTargetAveragePriceTooLow(uint256 tick) public {
-        tick = bound(tick, 1, MAX_TICK);
+        tick = bound(tick, 4, MAX_TICK / DEFAULT_TICK_SPACING) * DEFAULT_TICK_SPACING;
```
peyha:
```suggestion
        tick = bound(tick, 1, MAX_TICK / DEFAULT_TICK_SPACING) * DEFAULT_TICK_SPACING;
```

peyha (approved):
lgtm, same as quentin did not look into the hardcoded values as it can be fine tuned later

src/libraries/TickLib.sol
adhusson:
```solidity
  /// @dev Returns the lowest tick with a higher price that is also a multiple of spacing.
  /// @dev spacing should divide MAX_TICK.
  function priceToTick(uint256 price, uint256 spacing) internal pure returns (uint256) {
      require(price <= 1e18, PriceGreaterThanOne());
      uint256 low = 0;
      uint256 high = MAX_TICK;
      while (low != high) {
          unchecked {
              uint256 mid = (low + high) / 2;
              if (tickToPrice(mid) < price) low = mid + 1;
              else high = mid;
          }
      }
      return (low + spacing - 1) / spacing * spacing;
  }
```

src/libraries/TickLib.sol
```diff
@@ -2,8 +2,10 @@
 // Copyright (c) 2025 Morpho Association
 pragma solidity ^0.8.0;
 
-int256 constant LN_ONE_PLUS_DELTA = 0.024692612590371501e18; // floor(ln(1 + 0.025) * 1e18)
-uint256 constant MAX_TICK = 1046;
+int256 constant LN_ONE_PLUS_DELTA = 0.004987541511039073e18; // floor(ln(1.005) * 1e18)
+uint256 constant MAX_TICK = 5820;
+// Minimum representable price increment in WAD (1e-6 WAD). Tick prices are rounded to multiples of this value.
+uint256 constant PRICE_ROUNDING_STEP = 1e12;
```
adhusson, line 8: If we do `0.8e12` then with `MAX_TICK = 5908`, tick 0 is the only tick with price 0, and `MAX_TICK` is the only tick with price 1e18

src/libraries/TickLib.sol
```diff
@@ -43,12 +45,13 @@ library TickLib {
...
             return uint256(1e36)
                     .divHalfDownUnchecked(1e18 + wExp(LN_ONE_PLUS_DELTA * (int256(MAX_TICK / 2) - int256(tick))))
-                    .divHalfDownUnchecked(5e12) * 5e12;
+                    .divHalfDownUnchecked(PRICE_ROUNDING_STEP) * PRICE_ROUNDING_STEP;
         }
     }
 
-    /// @dev Returns the lowest tick with a higher price.
-    function priceToTick(uint256 price) internal pure returns (uint256) {
+    /// @dev Returns the lowest tick with a higher price that is also a multiple of spacing.
```
QGarchery:
```suggestion
    /// @dev Among the ticks than are multiples of spacing, returns the lowest one with a price higher or equal.
```
- `higher` -> `higher or equal`, for example with price 0 it returns 0
- fix the ambiguity (higher among that set, or take the higher and return only if it is in the set)

src/libraries/TickLib.sol
adhusson: also `0.5e12` with max_tick `6096` it seems

adhusson: PR https://github.com/morpho-org/midnight/pull/865

PR #620, opened 2026-04-01
passive trading fee recipient

### LINKED ISSUE #618: Fix transfer to address(0)
credits to the router team for the finding

adhusson:
- no revert on zero to transfer for tokens like USDC
- prevent dos on loan tokens that blacklisted the fee recipient / no need to whitelist fee recipient for permission

[slack discussion](https://morpholabs.slack.com/archives/C08A40KKN9W/p1775044089911059)


src/Midnight.sol
```diff
@@ -112,6 +112,8 @@ contract Midnight is IMidnight {
     /// feeSetter.
     mapping(address loanToken => uint32) public defaultContinuousFee;
 
+    mapping(address token => uint256) public withdrawableTradingFee;
```
MathisGD: don't like "withdrawable" as it's already used.
```suggestion
    mapping(address token => uint256) public accumulatedTradingFee;
```

```suggestion
    mapping(address token => uint256) public claimableTradingFee;
```

MathisGD (commented):
was afraid about complexity, but seems worth it actually

src/libraries/EventsLib.sol
```diff
@@ -80,4 +80,6 @@ library EventsLib {
     event SetIsAuthorized(
         address indexed caller, address indexed onBehalf, address indexed authorized, bool newIsAuthorized
     );
+
+    event WithdrawTradingFee(address indexed caller, address indexed token, uint256 amount, address indexed receiver);
```
```suggestion
    event ClaimTradingFee(address indexed caller, address indexed token, uint256 amount, address indexed receiver);
```
slight pref because `withdraw` makes me think that the fee is yielding

MathisGD: yes, same as [here](https://github.com/morpho-org/midnight/pull/620#discussion_r3021824974)

src/Midnight.sol
```diff
@@ -342,7 +353,8 @@ contract Midnight is IMidnight {
                 .onBuy(id, offer.obligation, buyer, buyerAssets, sellerAssets, units, buyerCallbackData);
         }
 
-        SafeTransferLib.safeTransferFrom(offer.obligation.loanToken, buyer, feeRecipient, buyerAssets - sellerAssets);
+        SafeTransferLib.safeTransferFrom(offer.obligation.loanToken, buyer, address(this), buyerAssets - sellerAssets);
+        withdrawableTradingFee[offer.obligation.loanToken] += buyerAssets - sellerAssets;
```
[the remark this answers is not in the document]
adhusson: not sure if we care, without fixing this noop trades break on some tokens but not others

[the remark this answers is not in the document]
MathisGD: I think that's ok. we can document though

src/Midnight.sol
```diff
@@ -112,6 +112,8 @@ contract Midnight is IMidnight {
     /// feeSetter.
     mapping(address loanToken => uint32) public defaultContinuousFee;
 
+    mapping(address token => uint256) public claimableTradingFee;
+
     address public feeRecipient;
```
MathisGD: >Unclaimed fees now belong to whoever is the current feeRecipient at claim time, not necessarily the recipient that was configured when the fees accrued. Since owner can rotate feeRecipient, this extends owner control from “future fees” to “already accrued but unclaimed fees” too.

we should document this

src/Midnight.sol
```diff
@@ -342,7 +353,8 @@ contract Midnight is IMidnight {
                 .onBuy(id, offer.obligation, buyer, buyerAssets, sellerAssets, units, buyerCallbackData);
         }
 
-        SafeTransferLib.safeTransferFrom(offer.obligation.loanToken, buyer, feeRecipient, buyerAssets - sellerAssets);
+        SafeTransferLib.safeTransferFrom(offer.obligation.loanToken, buyer, address(this), buyerAssets - sellerAssets);
+        claimableTradingFee[offer.obligation.loanToken] += buyerAssets - sellerAssets;
```
MathisGD, line 357: ~this change changes the impact of fees on transfer. Not completely sure, but maybe they were supported before (probably with some caveats).~
~it's ok though, we can document that they are not supported~

withdrawable is broken already

src/Midnight.sol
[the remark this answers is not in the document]
MathisGD: let's keep for #622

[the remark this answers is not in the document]
peyha: this breaks very few tokens (from what i saw, it only breaks LEND the previous aave token) so it should be fine

src/Midnight.sol
```diff
@@ -342,7 +355,8 @@ contract Midnight is IMidnight {
...
```
MathisGD, line 357: let's move that with the state updates

peyha (approved):

src/Midnight.sol
```diff
@@ -200,6 +204,15 @@ contract Midnight is IMidnight {
         emit EventsLib.SetDefaultContinuousFee(loanToken, newContinuousFee);
     }
 
+    function claimTradingFee(address token, uint256 amount, address receiver) external {
```
qdrn, line 206: If there is a receiver then feeRecipient is not really a recipient 😁. Really useful to have flexibility here?

MathisGD: "feeClaimer" then. 

yes it's useful, you would typically have a contract or some people designated to do something with the fees

adhusson: It mirrors `withdraw` which also has a `receiver` and so for the continuous fee the fee recipient can choose a receiver. For me it's very standard and it maintains uniformity so it's worth the added argument.

MathisGD: still we could call the role feeClaimer, no?

src/Midnight.sol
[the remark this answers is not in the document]
adhusson: ok with no noop on no-0-transfer tokens

src/Midnight.sol
MathisGD: true that for the continuous fee it was already the case

adhusson: changing name

src/Midnight.sol
```diff
@@ -200,6 +204,15 @@ contract Midnight is IMidnight {
...
+        require(msg.sender == feeClaimer, "only fee claimer");
+        claimableTradingFee[token] -= amount;
+
+        emit EventsLib.ClaimTradingFee(msg.sender, token, amount, receiver);
+
```
MathisGD: useless style
```suggestion
        emit EventsLib.ClaimTradingFee(msg.sender, token, amount, receiver);
```

src/Midnight.sol
```diff
@@ -112,7 +112,11 @@ contract Midnight is IMidnight {
     /// feeSetter.
     mapping(address loanToken => uint32) public defaultContinuousFee;
 
-    address public feeRecipient;
+    /// @dev If a fee claimer does not claim a fee after a take, a future fee claimer could claim that fee in his
+    /// place.
```
```suggestion
    /// @dev When the claimer is set, the old claimer looses the unclaimed trading and continuous fees.
```

qdrn (approved):
lgtm

certora/specs/WithdrawableMonotonicity.spec
```diff
@@ -45,3 +46,17 @@ filtered {
...
+    uint256 before = claimableTradingFee(token);
+    claimTradingFee(e, token, amount, receiver);
+    assert claimableTradingFee(token) == before - amount;
+}
+
+rule claimableTradingFeeUnchanged(method f, env e, calldataarg args, address token) filtered { f -> !f.isView && f.selector != sig:take(uint256, address, address, bytes, address, Midnight.Offer, Midnight.Signature, bytes32, bytes32[]).selector && f.selector != sig:claimTradingFee(address, uint256, address).selector } {
+    uint256 before = claimableTradingFee(token);
+    f(e, args);
+    assert claimableTradingFee(token) == before;
+}
```
jhoenicke, line 62: Should there be another rule for takeIncreasesClaimableTradingFee?

certora/specs/Solvency.spec
```diff
@@ -131,12 +131,24 @@ hook Sstore obligationState[KEY bytes32 id].withdrawable uint256 newWithdrawable
...
+}
+
 /// INVARIANTS AND RULES ///
 
-// For any token, the balance of the contract is always greater than or equal to the sum of all collateral and withdrawable amounts for that token minus the flash loaned amount.
+// For any token, the balance of the contract is always greater than or equal to the sum of all collateral, withdrawable, and withdrawable trading fee amounts for that token minus the flash loaned amount.
 // Note: this invariant is strong, so it also holds before each external call.
 strong invariant tokenBalanceCorrect(address token)
-    tokenBalances[token][currentContract] >= collateralSum(token) + withdrawableSum(token) - flashloans[token]
+    tokenBalances[token][currentContract] >= collateralSum(token) + withdrawableSum(token) + claimableTradingFeeMirror[token] - flashloans[token]
```
jhoenicke: I'm not sure why you need the mirror.  It doesn't hurt, but wouldn't it have been simpler to just use claimableTradingFee(token)?

jhoenicke (approved):
Looks good.  I would really like to have the rule takeIncreasesClaimableTradingFee.   The other comment is optional.

MathisGD (changes requested):
there is a deep problem with the Solvency spec. let's discuss live

certora/specs/WithdrawableMonotonicity.spec
adhusson: > takeIncreasesClaimableTradingFee

i had one before but folded it in [takeInputOutputConsistency](https://github.com/morpho-org/midnight/pull/620/changes/cf060c52576ce46d45682374695f7152fa49570c#diff-cf745402ebda4dba7dd96ba8ce7036369b0cc770b93a6f6843f0b4ae367a7904R85), wdyt?

MathisGD (approved):

MathisGD:
Let's move forward this way and fix later

PR #628, opened 2026-04-02
deffered health check

MathisGD:


certora/helpers/MidnightWrapper.sol
```diff
@@ -36,6 +36,7 @@ contract MidnightWrapper is Midnight {
     }
 
     function isHealthyNoBitmap(Obligation memory obligation, bytes32 id, address borrower) public view returns (bool) {
+        if (_getDeferredCheck(borrower)) return true;
```
adhusson: The deferred checked should be obligation specific otherwise a borrower can withdraw collateral on other obligations in the callback.

src/Midnight.sol
```diff
@@ -780,6 +801,7 @@ contract Midnight is IMidnight {
     /// @dev This function does not call any oracle if debt is 0.
     /// @dev Expects the id to correspond to the obligation's id.
     function isHealthy(Obligation memory obligation, bytes32 id, address borrower) public view returns (bool) {
+        if (_getDeferredCheck(borrower)) return true;
```
MathisGD: It needs to impact health data too

MathisGD:
Replaced by #645

PR #632, opened 2026-04-02
minor changes

MathisGD:


src/Midnight.sol
MathisGD, line 1: we have a name clash between 
- collateral, an item of obligation.collaterals
- collateral, the amount of collateral of borrowers

src/Midnight.sol
```diff
@@ -492,11 +491,10 @@ contract Midnight is IMidnight {
             Collateral memory _collateral = obligation.collaterals[i];
             uint256 price = IOracle(_collateral.oracle).price();
             if (i == collateralIndex) liquidatedCollatPrice = price;
-            uint256 _collateralOf = _position.collateral[i];
-            maxDebt += _collateralOf.mulDivDown(price, ORACLE_PRICE_SCALE).mulDivDown(_collateral.lltv, WAD);
-            badDebt = badDebt.zeroFloorSub(
-                _collateralOf.mulDivUp(price, ORACLE_PRICE_SCALE).mulDivUp(WAD, _collateral.maxLif)
-            );
+            uint256 _collateral = _position.collateral[i];
+            maxDebt += _collateral.mulDivDown(price, ORACLE_PRICE_SCALE).mulDivDown(_collateral.lltv, WAD);
```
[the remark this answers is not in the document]
MathisGD: ik

src/Midnight.sol
MathisGD: I ended up renaming obligation.collateralsParams

peyha: I find it slightly clearer to rename the amount of collateral to `collateralAmounts` or `collateralBalances`instead

MathisGD: I thought that it was better like that because we have credit and debt

MathisGD: also arguably a "collateral" is only the asset, then you add LLTV and oracle

peyha: Both are fine to me, it just feels a bit more natural for a `collateral` variable to refer to a set of params than to the collateral balance

peyha: also the struct is still called `Collateral` so it's slightly misleading

MathisGD: right, will change

src/libraries/EventsLib.sol
```diff
@@ -61,25 +59,21 @@ library EventsLib {
         address indexed onBehalf,
         address receiver
     );
-
     event Liquidate(
-        address indexed caller,
+        address caller,
         bytes32 indexed id_,
-        address collateral,
+        address indexed collateral,
```
qdrn, line 65: What is the rational here?

MathisGD: in other events it's indexed, and the caller is not

peyha: It seems likely that people will filter Liquidation events on some collateral (e.g. total amount of WETH liquidated) than on a specific liquidator. Though it will still be possible to do it on dune for instance

test/TakeTest.sol
```diff
@@ -967,21 +967,16 @@ contract TakeTest is BaseTest {
 
 contract BorrowCallback is ICallbacks {
     bytes public recordedData;
-    bytes32 public recordedObligationId;
+    bytes32 public recordedid;
```
peyha:
```suggestion
    bytes32 public recordedId;
```

peyha (approved):

src/Midnight.sol
```diff
@@ -40,13 +40,13 @@ import {EventsLib} from "./libraries/EventsLib.sol";
...
-/// - `collaterals.length > 0`: at least one collateral is required.
-/// - `collaterals.length <= MAX_COLLATERALS` (128): at most 128 collaterals per obligation.
+/// - `collateralParams.length > 0`: at least one collateral is required.
+/// - `collateralParams.length <= MAX_COLLATERALS` (128): at most 128 collateralParams per obligation.
 /// - Collateral tokens must be non-zero and strictly sorted by address (ascending, no duplicates).
 /// - Each collateral's `lltv` must be one of the allowed tiers (see `isLltvAllowed` in ConstantsLib).
 /// - Each collateral's `maxLif` must equal `maxLif(lltv, LIQUIDATION_CURSOR_LOW)` or
 ///   `maxLif(lltv, LIQUIDATION_CURSOR_HIGH)`.
-/// @dev Additionally, a borrower can have collateral in at most `MAX_COLLATERALS_PER_BORROWER` (10) collaterals
+/// @dev Additionally, a borrower can have collateral in at most `MAX_COLLATERALS_PER_BORROWER` (10) collateralParams
```
qdrn:
```suggestion
/// @dev Additionally, within a single obligation, a borrower can use at most
/// `MAX_COLLATERALS_PER_BORROWER` (10) collaterals simultaneously.
```

test/BaseTest.sol
```diff
@@ -241,11 +241,11 @@ abstract contract BaseTest is Test {
         return sig(_root, privateKey[offers[0].maker]);
     }
 
-    function sortCollaterals(Collateral[] memory arr) internal pure returns (Collateral[] memory) {
+    function sortCollaterals(CollateralParams[] memory arr) internal pure returns (CollateralParams[] memory) {
```
qdrn: Arguably you are sorting params now.

qdrn (approved):
Maybe slight preference to keep collateral and use collateral amount as PA suggested but LGTM.

src/Midnight.sol
```diff
@@ -413,7 +412,7 @@ contract Midnight is IMidnight {
     {
         require(onBehalf == msg.sender || isAuthorized[onBehalf][msg.sender], "unauthorized");
         bytes32 id = touchObligation(obligation);
-        address collateralToken = obligation.collaterals[collateralIndex].token;
+        address collateralToken = obligation.collateralParams[collateralIndex].token;
 
         Position storage _position = position[id][onBehalf];
         uint256 oldCollateralOf = _position.collateral[collateralIndex];
```
qdrn: Now it's not consistent with the function.
Same point elsewhere in the code.

PR #633, opened 2026-04-02
[Certora] ContinuousFee Protections

bhargavbh:


bhargavbh:
@codex review

certora/specs/ContinuousFeeProtection.spec
```diff
@@ -0,0 +1,132 @@
...
+rule pendingFeeDecreasesProportionallyInSellOffer(env e, uint256 units, address taker, address takerCallback, bytes takerCallbackData, address receiver, Midnight.Offer offer, Midnight.Signature signature, bytes32 root, bytes32[] proof) {
+    bytes32 id;
+    uint128 postUpdateCredit;
+    uint128 postUpdatePendingFee;
+
+    postUpdateCredit, postUpdatePendingFee, _ = updatePositionView(e, offer.obligation, id, offer.maker);
+
+    require postUpdateCredit > 0 || postUpdatePendingFee == 0, "noRemainingContinuousFeeWithoutCredit";
+
+    take(e, units, taker, takerCallback, takerCallbackData, receiver, offer, signature, root, proof);
```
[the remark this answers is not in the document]
bhargavbh: proving time is <2m for this spec, not an issue.

certora/specs/ContinuousFeeProtection.spec
```diff
@@ -0,0 +1,132 @@
...
+
+// In a buy-offer take, the buyer's pendingFee increases by at most floor(creditIncrease * continuousFee * timeToMaturity / WAD).
+rule continuousFeeNotOverchargedInBuyOffer(env e, uint256 units, address taker, address takerCallback, bytes takerCallbackData, address receiver, Midnight.Offer offer, Midnight.Signature signature, bytes32 root, bytes32[] proof) {
+    bytes32 id;
+    uint128 postUpdateCredit;
+    uint128 postUpdatePendingFee;
+
+    postUpdateCredit, postUpdatePendingFee, _ = updatePositionView(e, offer.obligation, id, offer.maker);
+
+    take(e, units, taker, takerCallback, takerCallbackData, receiver, offer, signature, root, proof);
```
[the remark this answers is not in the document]
bhargavbh: not an issue, prover time is <2m for this spec

certora/specs/ContinuousFeeProtection.spec
```diff
@@ -0,0 +1,132 @@
...
+
+    uint256 creditAfter = creditOf(id, offer.maker);
+    uint256 pendingFeeAfter = pendingFee(id, offer.maker);
+
+    mathint creditDelta = to_mathint(postUpdateCredit) - to_mathint(creditAfter);
+    mathint pendingFeeDelta = to_mathint(postUpdatePendingFee) - to_mathint(pendingFeeAfter);
+
+    require offer.buy == false, "scope to sell offers";
+
+    assert postUpdateCredit == 0 || pendingFeeDelta == (to_mathint(postUpdatePendingFee) * creditDelta + to_mathint(postUpdateCredit) - 1) / to_mathint(postUpdateCredit);
```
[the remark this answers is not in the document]
bhargavbh: this is covered in another rule (noRemainingContinuousFeeWithoutCredit) and used here.

certora/specs/ContinuousFeeProtection.spec
```diff
@@ -0,0 +1,132 @@
...
+
+    require id == lastId, "id should be derived from obligation";
+
+    uint256 contFee = continuousFee(id);
+    uint256 timeToMaturity = e.block.timestamp <= offer.obligation.maturity ? assert_uint256(offer.obligation.maturity - e.block.timestamp) : 0;
+
+    mathint creditDelta = to_mathint(creditOf(id, offer.maker)) - to_mathint(postUpdateCredit);
+    mathint pendingFeeDelta = to_mathint(pendingFee(id, offer.maker)) - to_mathint(postUpdatePendingFee);
+
+    require offer.buy, "scope to buy offers";
```
jhoenicke: Move this line above the call to take.  It's a precondition of the call, so I think the rule is easier to read if it comes earlier.

certora/specs/ContinuousFeeProtection.spec
```diff
@@ -0,0 +1,132 @@
...
+
+    require id == lastId, "id should be derived from obligation";
+
+    uint256 creditAfter = creditOf(id, offer.maker);
+    uint256 pendingFeeAfter = pendingFee(id, offer.maker);
+
+    mathint creditDelta = to_mathint(postUpdateCredit) - to_mathint(creditAfter);
+    mathint pendingFeeDelta = to_mathint(postUpdatePendingFee) - to_mathint(pendingFeeAfter);
+
+    require offer.buy == false, "scope to sell offers";
```
jhoenicke: move line above the call.

Do you have a convention whether it should be written as `buy == false` or `!buy` ?

certora/specs/ContinuousFeeProtection.spec
```diff
@@ -0,0 +1,132 @@
...
+    bytes32 id;
+    uint256 creditBefore = creditOf(id, user);
+    uint256 pendingFeeBefore = pendingFee(id, user);
+
+    take(e, units, taker, takerCallback, takerCallbackData, receiver, offer, signature, root, proof);
+
+    require id == lastId, "id should be derived from obligation";
+
+    assert creditOf(id, user) == creditBefore;
+    assert pendingFee(id, user) == pendingFeeBefore;
```
jhoenicke: I think it's better to say that updatePositionView returns the same before and after the call for other users.  This would also say that the loss index or the lastAccrual is not affecting the user.

jhoenicke (approved):
The "NotOvercharged" checks are only for the maker.  Do we have some rules for the take how the fee should be for them?

I marked a few small issues I noticed.

certora/specs/ContinuousFeeProtection.spec
bhargavbh: now checks updatePositionView returns the same values before and after take()

certora/specs/ContinuousFeeProtection.spec
bhargavbh: generalised the rule. the require for scoping is no longer needed.

certora/specs/ContinuousFeeProtection.spec
bhargavbh: refactored the rule `continuousFeeNotOverchargedBuyer` to check the property for Buyers/Sellers, and not makers/takers. This now gets rid of the scoping require `require offer.buy`. We set buyer/seller to maker/taker based on whether offer.buy is true. 
Similarly for the rule `pendingFeeDecreasesProportionallySeller`. @jhoenicke

certora/specs/ContinuousFeeProtection.spec
```diff
@@ -0,0 +1,140 @@
...
+
+    require id == lastId, "id should be derived from obligation";
+
+    uint256 contFee = continuousFee(id);
+    uint256 timeToMaturity = e.block.timestamp <= offer.obligation.maturity ? assert_uint256(offer.obligation.maturity - e.block.timestamp) : 0;
+
+    mathint creditDelta = to_mathint(creditOf(id, buyer)) - to_mathint(postUpdateCredit);
+    mathint pendingFeeDelta = to_mathint(pendingFee(id, buyer)) - to_mathint(postUpdatePendingFee);
+
+    assert pendingFeeDelta <= (creditDelta * to_mathint(contFee) * to_mathint(timeToMaturity)) / WAD();
```
lilCertora: == is true here I think

certora/specs/ContinuousFee.spec
```diff
@@ -0,0 +1,140 @@
...
+
+    take(e, units, taker, takerCallback, takerCallbackData, receiver, offer, ratifierData, root, proof);
+
+    require id == lastId, "id should be derived from obligation";
+
+    assert continuousFeeCredit(id) == continuousFeeCreditBefore + buyerAccruedFee + sellerAccruedFee;
+}
+
+// updatePositionView()
+rule takeDoesNotAffectThirdParties(env e, uint256 units, address taker, address takerCallback, bytes takerCallbackData, address receiver, Midnight.Offer offer, bytes ratifierData, bytes32 root, bytes32[] proof, address user) {
```
lilCertora, line 130: we could also check the debtOf here to add some coverage maybe ?

lilCertora:
LGTM, some minor comments.
This could be extended to other functions in the future

lilCertora (approved):

certora/confs/ContinuousFeeProtection.conf
```diff
@@ -0,0 +1,24 @@
...
+  "loop_iter": 2,
+  "rule_sanity": "basic",
+  "smt_timeout": 7200,
+  "prover_args": [
+    "-depth 5",
+    "-timeout 3600",
+    "-splitParallel",
+    "true",
+    "-dontStopAtFirstSplitTimeout",
+    "true"
```
QGarchery:
```suggestion
    "-splitParallel true",
    "-dontStopAtFirstSplitTimeout true"
```
(nitpick)

certora/confs/ContinuousFeeProtection.conf
```diff
@@ -0,0 +1,24 @@
...
+  "verify": "Midnight:certora/specs/ContinuousFeeProtection.spec",
+  "solc": "solc-0.8.34",
+  "solc_evm_version": "osaka",
+  "optimistic_hashing": true,
+  "hashing_length_bound": 2048,
+  "solc_via_ir": true,
+  "optimistic_loop": true,
+  "loop_iter": 2,
+  "rule_sanity": "basic",
+  "smt_timeout": 7200,
```
QGarchery:
```suggestion
```
those are the default settings I think

certora/specs/ContinuousFeeProtection.spec
```diff
@@ -0,0 +1,140 @@
...
+    function _.onBuy(bytes32, Midnight.Obligation, address, uint256, uint256, bytes) external => NONDET;
+    function _.onSell(bytes32, Midnight.Obligation, address, uint256, uint256, bytes) external => NONDET;
+    function isHealthy(Midnight.Obligation memory, bytes32, address) internal returns (bool) => NONDET;
+    function _.canIncreaseCredit(address) external => NONDET;
+    function _.canIncreaseDebt(address) external => NONDET;
+    function TickLib.tickToPrice(uint256 tick) internal returns (uint256) => NONDET;
+    function tradingFee(bytes32 id, uint256 timeToMaturity) internal returns (uint256) => NONDET;
+
+    // the properties are about pendingFee/ credit accounting, and not liquidation interactions.
+    function UtilsLib.tExchange(uint256, bytes32, address, bool) internal returns (bool) => NONDET;
```
QGarchery: If `UtilsLib.tGet` is also summarized as NONDET, then those combined summaries are an over-approximation of the behavior of the transient storage and it would be safe to add in any spec

certora/specs/ContinuousFeeProtection.spec
```diff
@@ -0,0 +1,140 @@
...
+methods {
+    function IdLib.toId(Midnight.Obligation memory obligation, uint256 chainId, address midnight) internal returns (bytes32) => CVL_toId(obligation, chainId, midnight);
+
+    function creditOf(bytes32 id, address user) external returns (uint256) envfree;
+    function pendingFee(bytes32 id, address user) external returns (uint128) envfree;
+    function continuousFee(bytes32 id) external returns (uint32) envfree;
+    function continuousFeeCredit(bytes32 id) external returns (uint256) envfree;
+    function userLossIndex(bytes32 id, address user) external returns (uint128) envfree;
+    function lossIndex(bytes32 id) external returns (uint128) envfree;
+    function lastAccrual(bytes32 id, address user) external returns (uint128) envfree;
```
QGarchery:
```suggestion
```
unused it seems

certora/specs/ContinuousFeeProtection.spec
```diff
@@ -0,0 +1,140 @@
...
+    function UtilsLib.isLeaf(bytes32, bytes32, bytes32[] memory) internal returns (bool) => NONDET;
+    function IdLib.storeInCode(Midnight.Obligation memory) internal returns (address) => NONDET;
+    function SafeTransferLib.safeTransferFrom(address, address, address, uint256) internal => NONDET;
+    function _.onBuy(bytes32, Midnight.Obligation, address, uint256, uint256, bytes) external => NONDET;
+    function _.onSell(bytes32, Midnight.Obligation, address, uint256, uint256, bytes) external => NONDET;
+    function isHealthy(Midnight.Obligation memory, bytes32, address) internal returns (bool) => NONDET;
+    function _.canIncreaseCredit(address) external => NONDET;
+    function _.canIncreaseDebt(address) external => NONDET;
+    function TickLib.tickToPrice(uint256 tick) internal returns (uint256) => NONDET;
+    function tradingFee(bytes32 id, uint256 timeToMaturity) internal returns (uint256) => NONDET;
```
QGarchery:
```suggestion
    function isHealthy(Midnight.Obligation memory, bytes32, address) internal returns (bool) => NONDET;
    function TickLib.tickToPrice(uint256 tick) internal returns (uint256) => NONDET;
    function tradingFee(bytes32 id, uint256 timeToMaturity) internal returns (uint256) => NONDET;
    function UtilsLib.isLeaf(bytes32, bytes32, bytes32[] memory) internal returns (bool) => NONDET;
    // The protocol doesn't use the toObligation function.
    function IdLib.storeInCode(Midnight.Obligation memory) internal returns (address) => NONDET;
```
that way it's still pretty fast, and we only summarize view functions as NONDET, which is an over-approximation. Except `storeInCode` but we add an explanation why it's ok

certora/specs/ContinuousFee.spec
```diff
@@ -0,0 +1,140 @@
...
+rule continuousFeeCreditIncreasesByAccruedFees(env e, uint256 units, address taker, address takerCallback, bytes takerCallbackData, address receiver, Midnight.Offer offer, bytes ratifierData, bytes32 root, bytes32[] proof) {
+    address buyer = offer.buy ? offer.maker : taker;
+    address seller = offer.buy ? taker : offer.maker;
+
+    bytes32 id;
+    uint128 buyerAccruedFee;
+    uint128 sellerAccruedFee;
+
+    _, _, buyerAccruedFee = updatePositionView(e, offer.obligation, id, buyer);
+    _, _, sellerAccruedFee = updatePositionView(e, offer.obligation, id, seller);
```
QGarchery, line 118:
```suggestion
    _, _, takerAccruedFees = updatePositionView(e, offer.obligation, id, taker);
    _, _, makerAccruedFees = updatePositionView(e, offer.obligation, id, offer.maker);
```
that way we avoid to define `buyer` and `seller`. Thinking about it, maybe it's not worth it, since other rules are based on buyer/seller rather than taker/maker

certora/specs/ContinuousFeeProtection.spec
QGarchery: https://prover.certora.com/output/66603/52576d6f504341eba16ddecd3bdd678f?anonymousKey=d18d5f01ed0197a47ba0e3ef60708639f74c6977

certora/specs/ContinuousFee.spec
```diff
@@ -0,0 +1,140 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+methods {
```
QGarchery, line 3: Maybe worth including a comment like in BalanceEffects:
> Assume no reentrancy: callbacks and token transfers do not re-enter Midnight.
> This is justified because the properties we verify are about the effect of each function's own body on credit and debt, not the effect of the full transaction including callbacks.

certora/specs/ContinuousFeeProtection.spec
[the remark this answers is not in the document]
QGarchery: The reasoning is:
- if `postUpdateCredit` is 0, then `creditAfter` is 0, because the credit is non-increasing in this rule, using noRemainingContinuousFeeWithoutCredit, we get that `pendingFeeAfter` is 0 as well
- since we also have `postUpdatePendingFee == 0` (precisely because of the instance of noRemainingContinuousFeeWithoutCredit used in this rule), this means that `pendingFeeDelta` is 0
- so in both cases (`postUpdateCredit` is 0 or not), we are able to deduce the `pendingFeeDelta`

I think it's non-trivial enough to warrant a comment

certora/specs/ContinuousFee.spec
QGarchery, line 1: ContinuousFeeProtection sounds weird to me. I understand where it comes from (rounding in favor of the user), but it feels awkward to call it a protection, and there are other rules in this file. Renaming it `ContinuousFeeChanges` would make sense or just `ContinuousFee`

certora/specs/ContinuousFeeProtection.spec
```diff
@@ -0,0 +1,140 @@
...
+    uint256 continuousFeeCreditBefore = continuousFeeCredit(id);
+
+    take(e, units, taker, takerCallback, takerCallbackData, receiver, offer, ratifierData, root, proof);
+
+    require id == lastId, "id should be derived from obligation";
+
+    assert continuousFeeCredit(id) == continuousFeeCreditBefore + buyerAccruedFee + sellerAccruedFee;
+}
+
+// updatePositionView()
```
QGarchery:
```suggestion
```
leftover ?

certora/specs/ContinuousFee.spec
QGarchery: We already check credit/debt in BalanceEffects:takeEffects

QGarchery: maybe we should only check `pendingFee(id, user)` here actually

certora/confs/ContinuousFeeProtection.conf
bhargavbh: rule_sanity is enabled by default, thats fine. 
The docs say default smt_timeout is 300 https://docs.certora.com/en/latest/docs/prover/cli/options.html#smt-timeout
But it also recommends not setting it very high: 
" _The second use is when the solvers can prove the property, they just need more time. Usually, if the rule isn’t solved in 600 seconds, it will not be solved in 2,000 either. It is better to concentrate your efforts on simplifying the rule, the source code, add more summaries, or use other time-saving options._ "
@lilCertora any inputs/sugestions here?

certora/specs/ContinuousFeeProtection.spec
bhargavbh: unable to reply on the "We already check credit/debt in BalanceEffects:takeEffects" comment. Hence replying here:

The [assert](https://github.com/morpho-org/midnight/blob/04a9e199f5dc944f7948781149075b0ec7f2c337/certora/specs/BalanceEffects.spec#L109) in BalanceEffects:takeEffects is different. There we compare the storage but here we compare postUpdateCredit returned by updatePositionView. I'll clarify this in the comment.

certora/specs/ContinuousFee.spec
bhargavbh: lets name it ContinuousFee.spec. The additional WIP properties in https://github.com/morpho-org/midnight/pull/684 can also be accommodated here. The summaries and setup seem similar looking at them now.

certora/specs/ContinuousFee.spec
bhargavbh: this comment is no longer needed since we anyways got rid of the callback summaries
https://github.com/morpho-org/midnight/pull/633#discussion_r3080995499

certora/specs/ContinuousFeeProtection.spec
bhargavbh: correct, the RHS is nothing but the floor(creditIncrease * continuousFee * timeToMaturity / WAD).

certora/specs/ContinuousFee.spec
bhargavbh: agree. buyer/seller based rules feel more intuitive and readable

certora/specs/ContinuousFee.spec
bhargavbh: discussing in the thread above ^

certora/confs/ContinuousFeeProtection.conf
lilCertora: smt_timeout is per smt split leaf so yes as a general rule if not solved in 600 it won't be solved in 2000 but it's a general rule, midnight is quite a complex contract by itself so it depends. You can use `-timeout` in the prover args which is a per rule budget, we don't have timeout issues with this spec right ?

bhargavbh: yes, timeout is not an issue for this spec; hence the default of 300 should suffice. thanks

certora/specs/ContinuousFeeProtection.spec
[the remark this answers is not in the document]
bhargavbh: the rule is now more explicit about the postUpdateCredit==0 case and also added a comment.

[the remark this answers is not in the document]
bhargavbh: made the postUpdateCredit==0 branch more explicit and added a comment.

certora/confs/ContinuousFee.conf
```diff
@@ -0,0 +1,20 @@
...
+  "solc_evm_version": "osaka",
+  "optimistic_hashing": true,
+  "hashing_length_bound": 2048,
+  "solc_via_ir": true,
+  "optimistic_loop": true,
+  "loop_iter": 2,
+  "prover_args": [
+    "-depth 5",
+    "-timeout 3600",
+    "-splitParallel true",
```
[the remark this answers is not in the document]
bhargavbh: certora checks rule sanity by default: https://docs.certora.com/en/latest/docs/prover/checking/sanity.html

certora/specs/ContinuousFee.spec
QGarchery: Even if we got rid of the callbacks, they are summarized as HAVOC_ECF (default behavior for AUTO). So in a sense we are still assuming no reentrancy here. If we summarized them with HAVOC_ALL instead I think the rule would not pass

certora/confs/ContinuousFee.conf
```diff
@@ -0,0 +1,20 @@
+{
+  "files": [
+    "src/Midnight.sol"
+  ],
+  "verify": "Midnight:certora/specs/ContinuousFeeProtection.spec",
```
QGarchery:
```suggestion
  "verify": "Midnight:certora/specs/ContinuousFee.spec",
```

certora/specs/ContinuousFee.spec
```diff
@@ -7,25 +7,17 @@ methods {
     function pendingFee(bytes32 id, address user) external returns (uint128) envfree;
     function continuousFee(bytes32 id) external returns (uint32) envfree;
     function continuousFeeCredit(bytes32 id) external returns (uint256) envfree;
-    function userLossIndex(bytes32 id, address user) external returns (uint128) envfree;
-    function lossIndex(bytes32 id) external returns (uint128) envfree;
-    function lastAccrual(bytes32 id, address user) external returns (uint128) envfree;
 
-    function _.price() external => NONDET;
-    function _.onRatify(Midnight.Offer, bytes32, bytes) external => NONDET;
+    // Summarize internals irrelevant to continuous fee tracking.
```
QGarchery, line 13: This comment is only needed for storeInCode because the other functions are view (so NONDET is an over-approximation when we don't look at reverts)

QGarchery (approved):

certora/specs/ContinuousFee.spec
```diff
@@ -0,0 +1,140 @@
...
+
+// IdLib summary: remember the last id returned by toId.
+
+persistent ghost bytes32 lastId;
+
+function CVL_toId(Midnight.Obligation obligation, uint256 chainId, address midnight) returns bytes32 {
+    // non-deterministic id
+    bytes32 id;
+    lastId = id;
+    return id;
```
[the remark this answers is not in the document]
QGarchery: yes we should be careful about that, but it works here

certora/specs/ContinuousFee.spec
```diff
@@ -0,0 +1,140 @@
...
+    uint128 postUpdateCredit;
+    uint128 postUpdatePendingFee;
+
+    postUpdateCredit, postUpdatePendingFee, _ = updatePositionView(e, offer.obligation, id, buyer);
+
+    take(e, units, taker, takerCallback, takerCallbackData, receiver, offer, signature, root, proof);
+
+    require id == lastId, "id should be derived from obligation";
+
+    uint256 contFee = continuousFee(id);
```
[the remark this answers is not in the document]
QGarchery: Agree, we need to assume no callback here

certora/specs/ContinuousFee.spec
bhargavbh: i was wondering how the spec passed without summarising callbacks. makes sense now.

certora/specs/ContinuousFee.spec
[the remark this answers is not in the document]
bhargavbh: added comment on default HAVOC_ECF summary

certora/specs/ContinuousFeeProtection.spec
QGarchery: it seems like it was reverted in the latest commits, and we have the inequality again. Curious why

bhargavbh: with the summaries of transfers and callbacks removed, the assert with equality times out. assert with inequality passes in 4min. i tried summarising SafeTransferLiv.transferFrom but still times out: https://prover.certora.com/output/1438622/db03b950f3b74ba1a0eb28d0ff67a76a/?anonymousKey=049383bc4bcd120af9bca8f217d621faae52c0e5

Yet to find the root cause.

certora/specs/ContinuousFee.spec
```diff
@@ -0,0 +1,143 @@
...
+    function tradingFee(bytes32 id, uint256 timeToMaturity) internal returns (uint256) => NONDET;
+
+    // summaries over-approximating the behavior of transient storage.
+    function UtilsLib.tExchange(uint256, bytes32, address, bool) internal returns (bool) => NONDET;
+    function UtilsLib.tGet(uint256, bytes32, address) internal returns (bool) => NONDET;
+
+    // This spec assumes no reentrancy: callbacks do not re-enter Midnight.
+    function _.onBuy(bytes32, Midnight.Obligation, address, uint256, uint256, bytes) external => NONDET;
+    function _.onSell(bytes32, Midnight.Obligation, address, uint256, uint256, bytes) external => NONDET;
+    function _.onRatify(Midnight.Offer, bytes32, bytes) external => NONDET;
```
QGarchery: Would be surprised if this is what caused it to resolve the timeout (probably the seeds is what fixed the timeout), but I could be wrong

certora/specs/ContinuousFee.spec
```diff
@@ -0,0 +1,143 @@
...
+    function UtilsLib.isLeaf(bytes32, bytes32, bytes32[] memory) internal returns (bool) => NONDET;
+    function isHealthy(Midnight.Obligation memory, bytes32, address) internal returns (bool) => NONDET;
+    function TickLib.tickToPrice(uint256 tick) internal returns (uint256) => NONDET;
+    function tradingFee(bytes32 id, uint256 timeToMaturity) internal returns (uint256) => NONDET;
+
+    // summaries over-approximating the behavior of transient storage.
+    function UtilsLib.tExchange(uint256, bytes32, address, bool) internal returns (bool) => NONDET;
+    function UtilsLib.tGet(uint256, bytes32, address) internal returns (bool) => NONDET;
+
+    // This spec assumes no reentrancy: callbacks do not re-enter Midnight.
```
QGarchery:
```suggestion
    // This spec assumes no reentrancy: callbacks and tokens do not re-enter Midnight.
```
because token calls are still summarized as HAVOC_ECF automatically, which means no reentrancy

QGarchery (approved):

certora/specs/ContinuousFee.spec
bhargavbh: tru, the issue was random seeds for Z3, not the lack of summaries

certora/specs/ContinuousFee.spec
```diff
@@ -0,0 +1,140 @@
...
+
+    require id == lastId, "id should be derived from obligation";
+
+    uint256 contFee = continuousFee(id);
+    uint256 timeToMaturity = e.block.timestamp <= offer.obligation.maturity ? assert_uint256(offer.obligation.maturity - e.block.timestamp) : 0;
+
+    mathint creditDelta = to_mathint(creditOf(id, buyer)) - to_mathint(postUpdateCredit);
+    mathint pendingFeeDelta = to_mathint(pendingFee(id, buyer)) - to_mathint(postUpdatePendingFee);
+
+    assert pendingFeeDelta == (creditDelta * to_mathint(contFee) * to_mathint(timeToMaturity)) / WAD();
```
MathisGD: we want to prove that users don't overpay (right?). so I would rather have something of the form pendingFeeAfter <= ... 
```suggestion
    assert pendingFee(id, buyer) <= postUpdatePendingFee + (creditDelta * to_mathint(contFee) * to_mathint(timeToMaturity)) / WAD();
```

also, the equality in itself isn't that interesting

certora/specs/ContinuousFee.spec
```diff
@@ -0,0 +1,140 @@
...
+    uint256 pendingFeeAfter = pendingFee(id, seller);
+
+    require creditAfter > 0 || pendingFeeAfter == 0, "See noRemainingContinuousFeeWithoutCredit in Midnight.spec";
+
+    mathint creditDelta = to_mathint(postUpdateCredit) - to_mathint(creditAfter);
+    mathint pendingFeeDelta = to_mathint(postUpdatePendingFee) - to_mathint(pendingFeeAfter);
+
+    // When postUpdateCredit == 0: noRemainingContinuousFeeWithoutCredit gives postUpdatePendingFee == 0; credit is non-increasing for a seller, therefore creditAfter == 0;
+    // noRemainingContinuousFeeWithoutCredit gives pendingFeeAfter == 0; hence pendingFeeDelta == 0.
+    assert postUpdateCredit == 0 ? pendingFeeDelta == 0 : pendingFeeDelta == (to_mathint(postUpdatePendingFee) * creditDelta + to_mathint(postUpdateCredit) - 1) / to_mathint(postUpdateCredit);
```
MathisGD: same

certora/specs/ContinuousFee.spec
```diff
@@ -0,0 +1,158 @@
...
+
+    postUpdateCredit, postUpdatePendingFee, _ = updatePositionView(e, obligation, id, onBehalf);
+
+    withdraw(e, obligation, units, onBehalf, receiver);
+
+    require id == lastId, "id should be derived from obligation";
+
+    mathint pendingFeeDelta = to_mathint(postUpdatePendingFee) - to_mathint(pendingFee(id, onBehalf));
+
+    assert postUpdateCredit > 0 => pendingFeeDelta == (to_mathint(postUpdatePendingFee) * to_mathint(units) + to_mathint(postUpdateCredit) - 1) / to_mathint(postUpdateCredit);
```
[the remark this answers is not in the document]
QGarchery: Let's add a comment to say that this case is covered in noRemainingContinuousFeeWithoutCredit

certora/specs/ContinuousFee.spec
```diff
@@ -0,0 +1,158 @@
...
+    uint128 postUpdateCredit;
+    uint128 postUpdatePendingFee;
+
+    postUpdateCredit, postUpdatePendingFee, _ = updatePositionView(e, obligation, id, onBehalf);
+
+    withdraw(e, obligation, units, onBehalf, receiver);
+
+    require id == lastId, "id should be derived from obligation";
+
+    mathint pendingFeeDelta = to_mathint(postUpdatePendingFee) - to_mathint(pendingFee(id, onBehalf));
```
QGarchery:
```suggestion
    mathint pendingFeeDelta = postUpdatePendingFee - pendingFee(id, onBehalf);
```
we can remove all the `to_mathint` occurrences

QGarchery (approved):

certora/specs/ContinuousFee.spec
```diff
@@ -0,0 +1,158 @@
...
+    require id == lastId, "id should be derived from obligation";
+
+    uint256 contFee = continuousFee(id);
+    uint256 timeToMaturity = e.block.timestamp <= offer.obligation.maturity ? assert_uint256(offer.obligation.maturity - e.block.timestamp) : 0;
+
+    mathint creditDelta = to_mathint(creditOf(id, buyer)) - to_mathint(postUpdateCredit);
+    mathint pendingFeeDelta = to_mathint(pendingFee(id, buyer)) - to_mathint(postUpdatePendingFee);
+
+    //assert pendingFeeDelta == (creditDelta * to_mathint(contFee) * to_mathint(timeToMaturity)) / WAD();
+    assert pendingFee(id, buyer) <= postUpdatePendingFee + (creditDelta * to_mathint(contFee) * to_mathint(timeToMaturity)) / WAD();
```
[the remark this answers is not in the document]
QGarchery: We can also remove the commented `assert` statement

certora/specs/ContinuousFee.spec
bhargavbh: rearranged the assert to the suggested form. However, i feel if we can prove equality, we should. The comment stating "pendingFee increases by floor(creditIncrease*ContFee*ttm/WAD)" makes it clear that buyer is not overpaying.

certora/specs/ContinuousFee.spec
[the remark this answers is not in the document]
bhargavbh: for withdraw, we actually do not need to invoke `noRemainingContinuousFeeWithoutCredit`. 
postUpdateCredit==0 ->  pendingFee is unchanged in withdraw. 
Changed the assert to be consistent with the rule `pendingFeeDecreasesProportionallyForSeller`

certora/specs/ContinuousFee.spec
```diff
@@ -0,0 +1,154 @@
...
+    require postUpdateCredit > 0 || postUpdatePendingFee == 0, "See noRemainingContinuousFeeWithoutCredit in Midnight.spec";
+
+    take(e, units, taker, takerCallback, takerCallbackData, receiver, offer, ratifierData, root, proof);
+
+    require id == lastId, "id should be derived from obligation";
+
+    uint256 creditAfter = creditOf(id, seller);
+    uint256 pendingFeeAfter = pendingFee(id, seller);
+
+    require creditAfter > 0 || pendingFeeAfter == 0, "See noRemainingContinuousFeeWithoutCredit in Midnight.spec";
```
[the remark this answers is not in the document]
bhargavbh: using assert will not work here as its an invariant. we could go the requireInvariant route but pointing to the invariant as a reason string is an acceptable convention.

MathisGD (approved):

PR #634, opened 2026-04-02
merge health checks

MathisGD:
we don't loose much gas doing this

this will make #628 easier to review


src/Midnight.sol
```diff
@@ -766,22 +752,40 @@ contract Midnight is IMidnight {
...
+    }
+
+    /// @dev This function should be called with the id corresponding to the obligation.
+    /// @dev This function does not call any oracle if debt is 0.
+    /// @dev Expects the id to correspond to the obligation's id.
+    function isHealthy(Obligation memory obligation, bytes32 id, address borrower) public view returns (bool) {
+        if (position[id][borrower].debt == 0) {
+            return true;
+        } else {
+            (uint256 maxDebt,,) = healthData(obligation, id, borrower, 0);
```
adhusson: why index 0?

MathisGD: let's change to type(uint256).max

src/Midnight.sol
```diff
@@ -766,22 +752,40 @@ contract Midnight is IMidnight {
         return position[id][user].lastAccrual;
     }
 
-    /// @dev This function should be called with the id corresponding to the obligation.
-    /// @dev This function does not call any oracle if debt is 0.
+    /// @dev Computes health-related data for a position.
+    /// @dev Returns the max debt, the price of the collateral at collateralIndex, and the bad debt.
```
MathisGD: move this dev comment to be the last one

src/Midnight.sol
```diff
@@ -766,22 +752,40 @@ contract Midnight is IMidnight {
         return position[id][user].lastAccrual;
     }
 
-    /// @dev This function should be called with the id corresponding to the obligation.
-    /// @dev This function does not call any oracle if debt is 0.
+    /// @dev Computes health-related data for a position.
```
MathisGD: let's add a dev comment stating that the collateral price is zero if the collateral is not activated for the borrower

peyha (approved):
LGTM
Liquidation is cleaner. It also saves 100 bytes in the bytecode
Note that we lose the micro optimization in `isHealthy` of only checking collaterals up to the debt (the `maxDebt < debt`) part

MathisGD:
we could retrieve the early return actually, because badDebt is zero in the case and collatPrice is not needed (liquidation will revert). But it will need to be verified, and  I'm not sure that it's worth it (will save a bit of gas in very rare situations...)

certora/specs/NoDivisionByZero.spec
```diff
@@ -85,6 +85,8 @@ function equalsGlobalObligation(Midnight.Obligation obligation) returns (bool) {
 }
 
 function summaryToId(Midnight.Obligation obligation, uint256 chainId, address morpho) returns (bytes32) {
+    // Sound: touchObligation enforces maxLif >= WAD for all collaterals (ExactMath.spec).
+    require forall uint256 i. i < obligation.collaterals.length => obligation.collaterals[i].maxLif >= WAD();
```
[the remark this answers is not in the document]
MathisGD: actually it's still needed

qdrn (approved):

MathisGD:
Not needed since we go for #645

PR #640, opened 2026-04-03
isLiquidatable

MathisGD:


src/Midnight.sol
adhusson, line 1: `canBeLiquidated` is easier

MathisGD: image

MathisGD:
Not needed since we go for #645

PR #644, opened 2026-04-04
nested deferred hc

adhusson:


adhusson:
closing since https://github.com/morpho-org/midnight/pull/628 is closed

PR #656, opened 2026-04-05
[obligation state packing] all fees in one array

adhusson:


src/Midnight.sol
```diff
@@ -54,7 +54,7 @@ import {EventsLib} from "./libraries/EventsLib.sol";
 /// @dev Trading fee breakpoint indices: 0=0d, 1=1d, 2=7d, 3=30d, 4=90d, 5=180d, 6=360d.
 /// @dev For TTM > 360d, the trading fee is the fee at the 360d breakpoint.
 /// @dev Post-maturity, the trading fee is the fee at the 0d breakpoint.
-/// @dev Trading fees are stored divided by FEE_STEP (1e12) to fit in 16 bits.
+/// @dev Trading fees are stored divided by FEE_STEP (1e12) to fit in 32 bits.
```
adhusson, line 57: `FEE_STEP` could be 1e7 but keeping change minimal for now to make the idea clearer

test/LiquidationTest.sol
```diff
@@ -368,7 +368,7 @@ contract LiquidationTest is BaseTest {
         assertEq(midnight.totalUnits(id), debtAfterBadDebt, "total units");
         assertEq(midnight.creditOf(id, lender), units, "lender units");
         midnight.updatePosition(obligation, lender);
-        assertApproxEqAbs(midnight.creditOf(id, lender), debtAfterBadDebt, 1, "lender units after slashing");
+        assertApproxEqAbs(midnight.creditOf(id, lender), debtAfterBadDebt, 256, "lender units after slashing");
```
adhusson: Worst case overslash is now 256 units. For amounts <= 1e36 worst case overslash is still 1.

adhusson: not anymore

src/Midnight.sol
```diff
@@ -767,19 +770,23 @@ contract Midnight is IMidnight {
...
+        return obligationState[id].fees[7] != 0;
     }
 
     function withdrawable(bytes32 id) external view returns (uint256) {
         return obligationState[id].withdrawable;
     }
 
-    function fees(bytes32 id) external view returns (uint16[7] memory) {
-        return obligationState[id].fees;
+    function fees(bytes32 id) external view returns (uint32[7] memory tradingFees) {
```
MathisGD, line 780: Let's rename at the same time

adhusson:
did https://github.com/morpho-org/midnight/pull/651 instead

note that this PR also contains https://github.com/morpho-org/midnight/pull/656/changes/BASE..8bad6cab16a7bfa5074dba3057096e7fe2496545#diff-bc7f9788c87d9ce947db1da48e2c26ea767af169dda1aa6223af2a2b419a07eb which is better, 256 assets loss when the number of assets is > 1e36 cannot be significant

PR #673, opened 2026-04-07
interface

MathisGD:


src/interfaces/IMidnight.sol
```diff
@@ -68,6 +68,81 @@ struct Position {
 }
 
 interface IMidnight {
+    // forgefmt: disable-start
```
adhusson, line 67: i find it ok with the natural fmt

MathisGD: both work for me

adhusson: i would remove then, i thought maybe there was a particular reason not to fmt

MathisGD: cloc maxxing lol

src/libraries/EventsLib.sol
```diff
@@ -6,6 +6,7 @@ import {Obligation} from "../interfaces/IMidnight.sol";
 
 /// @dev id_ is used to avoid naming conflicts in indexers.
 library EventsLib {
+    /// forgefmt: disable-start
```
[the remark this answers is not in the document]
MathisGD:
```suggestion
    // forgefmt: disable-start
```

adhusson (approved):

peyha (approved):

PR #675, opened 2026-04-07
Fix post maturity exit

adhusson:


src/Midnight.sol
```diff
@@ -541,7 +541,10 @@ contract Midnight is IMidnight {
             bitmap = bitmap.clearBit(i);
         }
 
-        require(block.timestamp > obligation.maturity || originalDebt > maxDebt, "position is not liquidatable");
+        require(
+            originalDebt > 0 && (block.timestamp > obligation.maturity || originalDebt > maxDebt),
```
adhusson, line 545: It's a bit weird since `isHealthy` also checks debt > 0 implicitly. But the alternative is
- keep `isLiquidatable` as it was
- add `require(debt == 0 || !isLiquidatable(...))` in `take`

and then we can't just say not liquidatable after take

MathisGD: slight preference like this (simpler model)

MathisGD (approved):

peyha (commented):
lgtm, we could add a test to catch this

MathisGD:
they were added!

peyha:
good catch, I missed it

PR #680, opened 2026-04-08
check size in ci

MathisGD:


foundry.toml
```diff
@@ -1,7 +1,7 @@
 [profile.default]
 via_ir = true
 optimizer = true
-optimizer_runs = 500
+optimizer_runs = 58
```
peyha, line 4: why keep it so low ? to have some margin ?

MathisGD: it's the max that keeps the size below the limit

adhusson (approved):

foundry.toml
peyha: 1000 runs on my side if i'm not mistaken, not a huge issue though

peyha (approved):

foundry.toml
MathisGD: it's not passing with 1000 on my side. are you sure?

QGarchery (approved):

foundry.toml
peyha: just tried and 1000 does not work, probably got on the wrong branch

PR #681, opened 2026-04-10
flashloan callback return value

MathisGD:
https://morpholabs.slack.com/archives/C08A40KKN9W/p1775811715285949
we get the model: callback to msg.sender or return value is checked


adhusson (approved):

MathisGD:
#688

PR #682, opened 2026-04-12
fix: take bundler pulling assets

spennyp:
## Problem

Sell-offer takes reverted. `Midnight.take` resolves `payer = msg.sender` (the bundler) for sell offers with no buyer callback, but the bundler held no loan tokens. Every `safeTransferFrom` failed inside the take loop, was swallowed by try/catch, and the call reverted with `"insufficient liquidity"`. The take-lend flow was fully broken.

## Fix

On sell offers, pull the loan token from `msg.sender` up front, approve Midnight for the exact pull amount, run the bundle, then revoke the allowance and sweep any residual back to `msg.sender` on exit.

- **Pull from `msg.sender`** (not `taker`) — matches Midnight's convention.
- **Exact approve amount** — bounded exposure, not `type(uint256).max`.
- **Always revoke on exit** — `midnight` is a caller-supplied parameter, so a standing approval would let an attacker pass their own contract as `midnight` and trap a max allowance. The revoke enforces `allowance == 0` before and after every call.
- **`bundleTakeSellerAssets` restricted to buy offers** (`require(takes[0].offer.buy)`) — sell offers need the buyer-side amount up-front which the seller-asset target doesn't provide.
- **`_approve` helper** — low-level call for USDT-style tokens (no return value). No OZ retry path needed because `_sweepAndRevoke` zeros the allowance on exit, so approvals always start from a zero baseline. Probs replace this. 
- **Reentrancy guard**: a malicious maker can set `offer.callback` and, during the callback Midnight fires after pulling tokens from the bundler, reenter any entry point with `target=0` to drain `bundleTakeUnits`'s unconsumed slack. Added for now. 

> Targeting `mvp/limits` as this is what we are deploying for alpha. 
> Deployed to 0x487e6263188eFD547F6eFD9491AD902b5173ee72 on Base

### DIFF
--- a/src/periphery/TakeBundler.sol
+++ b/src/periphery/TakeBundler.sol
@@ -4,6 +4,8 @@ pragma solidity 0.8.34;
 
 import {Midnight} from "../Midnight.sol";
 import {Offer} from "../interfaces/IMidnight.sol";
+import {IERC20} from "../interfaces/IERC20.sol";
+import {SafeTransferLib} from "../libraries/SafeTransferLib.sol";
 import {UtilsLib} from "../libraries/UtilsLib.sol";
 import {TakeAmountsLib} from "./TakeAmountsLib.sol";
 
@@ -18,11 +20,26 @@ contract TakeBundler {
         bytes32[] proof;
     }
 
+    /// @dev Midnight's `take` fires maker-controlled callbacks after moving the loan token out of the bundler.
+    /// Without this guard, a malicious callback could reenter any entry point with `target=0` to trigger its sweep,
+    /// draining `bundleTakeUnits`'s residual balance (`maxBuyerAssets - totalBuyerAssets`) and shipping it to the
+    /// attacker via `msg.sender`. The shared transient lock blocks cross-function reentry too.
+    uint256 private transient _reentrancyLock;
+
+    modifier nonReentrant() {
+        require(_reentrancyLock == 0, "reentrancy");
+        _reentrancyLock = 1;
+        _;
+        _reentrancyLock = 0;
+    }
+
     /// @dev Iterates through orders, filling up to targetUnits units total.
     /// @dev Assumes offers are all buy or all sell and share the same obligation id.
     /// @dev The taker must have authorized this bundler and the msg.sender (if different from the taker) on Midnight.
     /// @dev The bundler skips every reason why `take` can revert (including ones that are not asynchrony related).
     /// @dev If taking an offer reverts, the bundler will completely skip this offer.
+    /// @dev For sell offers the bundler pulls `maxBuyerAssets` from msg.sender, whom must have previously approved the
+    /// bundler to spend their loan assets.
     function bundleTakeUnits(
         Midnight midnight,
         uint256 targetUnits,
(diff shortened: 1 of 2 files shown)


src/periphery/TakeBundler.sol
```diff
@@ -18,11 +20,26 @@ contract TakeBundler {
         bytes32[] proof;
     }
 
+    /// @dev Midnight's `take` fires maker-controlled callbacks after moving the loan token out of the bundler.
+    /// Without this guard, a malicious callback could reenter any entry point with `target=0` to trigger its sweep,
+    /// draining `bundleTakeUnits`'s residual balance (`maxBuyerAssets - totalBuyerAssets`) and shipping it to the
+    /// attacker via `msg.sender`. The shared transient lock blocks cross-function reentry too.
+    uint256 private transient _reentrancyLock;
```
spennyp, line 27: Think there is reentrancy risk via callbacks in `bundleTakeUnits` with residual. Put this here for alpha, but not sure if it's actually desired.

src/periphery/TakeBundler.sol
```diff
@@ -160,4 +198,17 @@ contract TakeBundler {
...
+
+    /// @dev Revokes Midnight's allowance and returns any residual loan-token balance to `msg.sender`.
+    function _sweepAndRevoke(Midnight midnight, address loanToken) private {
+        _approve(loanToken, address(midnight), 0);
+        uint256 residual = IERC20(loanToken).balanceOf(address(this));
+        if (residual > 0) SafeTransferLib.safeTransfer(loanToken, msg.sender, residual);
+    }
+
+    /// @dev Simplified version of OZ safeApprove.
+    function _approve(address token, address spender, uint256 value) private {
```
spennyp, line 210: Didn't want to bring in OZ lib

src/periphery/TakeBundler.sol
```diff
@@ -33,9 +50,15 @@ contract TakeBundler {
         uint256 maxBuyerAssets,
         uint256 minSellerAssets,
         uint256 maxSellerAssets
-    ) external {
+    ) external nonReentrant {
         require(taker == msg.sender || midnight.isAuthorized(taker, msg.sender), "unauthorized");
 
+        address loanToken = takes[0].offer.obligation.loanToken;
```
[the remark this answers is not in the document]
spennyp: Same behavior existed previously in `bundleTakeBuyerAssets` and `bundleTakeSellerAssets`. Probs all could benefit from:
```
require(takes.length > 0, "empty takes");
```

But, leaving for the scope of this PR

src/periphery/TakeBundler.sol
```diff
@@ -126,8 +163,9 @@ contract TakeBundler {
         Take[] calldata takes,
         uint256 minUnits,
         uint256 maxUnits
-    ) external {
+    ) external nonReentrant {
         require(taker == msg.sender || midnight.isAuthorized(taker, msg.sender), "unauthorized");
+        require(takes[0].offer.buy, "sell offers unsupported");
```
spennyp, line 168: Disallowed `bundleTakeSellerAssets` to use sell offers because we don't trivially know the corresponding buyer assets. Could compute accounting for fees and per take rounding.

Think fine for now as don't see a use case to target seller assets while taking sell offers anyways.

MathisGD:
replaced by #716

PR #685, opened 2026-04-13
[Certora] Liveness Properties (AND liveness docs fix)

lilCertora:
Liveness Properties : 
```
/// LIVENESS
/// @dev If an activated collateral oracle reverts on `price`, `liquidate`, `isHealthy`, `withdrawCollateral`  when the
/// borrower has debt, and `take` whenever the seller still has debt all revert.
/// @dev If an activated collateral oracle returns 0 on `price`, `isHealthy`, `withdrawCollateral` when the borrower has
/// debt, `take` whenever the seller still has debt, and `liquidate` with repaid input all revert.
/// @dev If `enterGate.canIncreaseCredit` reverts or returns false, `take` reverts if the buyer's credit increases.
/// @dev If `enterGate.canIncreaseDebt` reverts or returns false, `take` reverts if the seller's debt increases.
/// @dev If `liquidatorGate` reverts or returns false on `canLiquidate`, `liquidate` reverts.
/// @dev If a token pulled by Midnight reverts on `transferFrom` despite balances and approvals being right, `take`,
/// `repay`, `supplyCollateral`, `liquidate`, and `flashLoan` repayment revert when they need to pull that token.
/// @dev If a token sent by Midnight reverts on `transfer` despite balances being right, `withdraw`,
/// `withdrawCollateral`, fee claims, the collateral leg of `liquidate`, and `flashLoan` revert when they need to send
/// that token.
/// @dev If a callback reverts, or if a buy/sell callback returns something other than `CALLBACK_SUCCESS`,
/// callback-enabled `take`, `repay`, `liquidate`, and `flashLoan` revert.
```


certora/specs/Liveness.spec
```diff
@@ -0,0 +1,428 @@
...
+
+/// If any activated collateral oracle reverts on price, isHealthy reverts.
+rule oracleRevertCausesIsHealthyRevert(env e, Midnight.Obligation obligation, bytes32 id, address borrower) {
+    require forceOracleRevert, "oracle reverts on price";
+    require activatedCollaterals(id, borrower) != 0, "borrower has activated collaterals";
+
+    isHealthy@withrevert(e, obligation, id, borrower);
+    bool reverted = lastReverted;
+
+    require id == lastId, "id derived from obligation";
```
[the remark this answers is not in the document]
lilCertora: both id and lastId are unconstrained symbolics

certora/specs/Liveness.spec
```diff
@@ -0,0 +1,428 @@
...
+    f@withrevert(e, args);
+    assert lastReverted;
+}
+
+/// CALLBACK REVERT PROPAGATION ///
+
+/// If the callback reverts, callback-enabled take (with non-zero taker callback) reverts.
+rule callbackRevertCausesTakeRevert(env e, uint256 units, address taker, address takerCallback, bytes takerCallbackData, address receiver, Midnight.Offer offer, bytes ratifierData, bytes32 root, bytes32[] proof) {
+    require forceCallbackRevert, "callback reverts";
+    require takerCallback != 0, "callback-enabled take";
```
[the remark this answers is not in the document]
lilCertora: good

certora/specs/Liveness.spec
```diff
@@ -0,0 +1,428 @@
...
+    require activatedCollaterals(id, borrower) != 0, "borrower has activated collaterals";
+
+    liquidate@withrevert(e, obligation, collateralIndex, seizedAssets, repaidUnits, borrower, data);
+
+    require id == lastId, "id derived from obligation";
+
+    assert lastReverted;
+}
+
+/// If any activated collateral oracle reverts on price, withdrawCollateral reverts when the borrower has debt.
```
QGarchery: I think the fact that all oracles revert (instead of just one of them) hides a real problem in the comments in Midnight.sol: there can be a reverting oracle but it does not make `isHealthy` revert because that oracle is not called as soon as `maxDebt >= debt`

QGarchery: <details>
<summary> Test to put in OtherFunctionsTest.sol. Check that removing the optimization maxDebt < debt in the while loop of isHealthy makes this test fail </summary>

```
    function testWithdrawCollateralSucceedsWithRevertingActivatedOracle() public {
        RevertingOracle revertingOracle = new RevertingOracle();
        Oracle normalOracle = new Oracle();

        ERC20 tokenLow = new ERC20("", "");
        ERC20 tokenHigh = new ERC20("", "");
        if (address(tokenLow) > address(tokenHigh)) {
            (tokenLow, tokenHigh) = (tokenHigh, tokenLow);
        }

        Obligation memory _obligation;
        _obligation.loanToken = address(loanToken);
        _obligation.maturity = block.timestamp + 100;
        _obligation.rcfThreshold = 0;
        CollateralParams[] memory params = new CollateralParams[](2);
        params[0] = CollateralParams({
            token: address(tokenLow),
            lltv: 0.77e18,
            maxLif: maxLif(0.77e18, 0.25e18),
            oracle: address(revertingOracle)
        });
        params[1] = CollateralParams({
            token: address(tokenHigh),
            lltv: 0.77e18,
            maxLif: maxLif(0.77e18, 0.25e18),
            oracle: address(normalOracle)
        });
        _obligation.collateralParams = params;

        uint256 units = 100e18;
        uint256 collateralNeeded = units.mulDivUp(WAD, 0.77e18);

        deal(address(tokenHigh), address(this), collateralNeeded);
        tokenHigh.approve(address(midnight), type(uint256).max);
        midnight.supplyCollateral(_obligation, 1, collateralNeeded, borrower);

        deal(address(tokenLow), address(this), 1e18);
        tokenLow.approve(address(midnight), type(uint256).max);
        midnight.supplyCollateral(_obligation, 0, 1e18, borrower);

        setupObligation(_obligation, units);

        revertingOracle.stopOracle();

        vm.prank(borrower);
        midnight.withdrawCollateral(_obligation, 1, 0, borrower, borrower);
    }
```
</details>

certora/specs/Reverts.spec
```diff
@@ -0,0 +1,428 @@
...
+    require forceOracleRevert, "oracle reverts on price";
+
+    bytes32 id;
+    require activatedCollaterals(id, borrower) != 0, "borrower has activated collaterals";
+
+    liquidate@withrevert(e, obligation, collateralIndex, seizedAssets, repaidUnits, borrower, data);
+
+    require id == lastId, "id derived from obligation";
+
+    assert lastReverted;
```
QGarchery: It's not clear to me what is the semantic of `require id == lastId` when the liquidate call may have reverted. What would be the value of `lastId` if `liquidate` reverts, is it set or not ?

certora/specs/Liveness.spec
QGarchery: This applies to the next 2 rules as well

certora/specs/Reverts.spec
```diff
@@ -0,0 +1,428 @@
...
+    assert debtOf(id, borrower) > 0 => reverted;
+}
+
+/// If all oracles revert and take succeeds, the seller must have no debt.
+rule oracleRevertPreventsTakeWhenSellerHasDebt(env e, uint256 units, address taker, address takerCallback, bytes takerCallbackData, address receiver, Midnight.Offer offer, bytes ratifierData, bytes32 root, bytes32[] proof) {
+    require forceOracleRevert, "oracle reverts on price";
+
+    bytes32 id;
+    address seller = offer.buy ? taker : offer.maker;
+    require !liquidationLocked(id, seller), "seller is not liquidation locked";
```
QGarchery, line 239:
```suggestion
```
why is this needed ?

certora/specs/Liveness.spec
```diff
@@ -0,0 +1,428 @@
...
+    take(e, units, taker, takerCallback, takerCallbackData, receiver, offer, ratifierData, root, proof);
+
+    require id == lastId, "id derived from obligation";
+
+    assert debtOf(id, seller) == 0;
+}
+
+/// ORACLE RETURNS ZERO ///
+
+/// If the oracle returns 0, liquidate reverts when using repaid units as input.
```
QGarchery: Here also we require all oracles to return 0, when only the one corresponding to the liquidated collateral should be enough

certora/specs/Liveness.spec
```diff
@@ -0,0 +1,428 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+// Liveness properties from Midnight.sol (lines 85-99).
+// These rules verify that external dependency failures (oracle, gate, callback, token transfer)
+// correctly propagate as reverts to the affected user-facing functions.
+
+methods {
+    function multicall(bytes[]) external => HAVOC_ALL DELETE;
+
+    // envfree view functions
```
QGarchery:
```suggestion
```

certora/specs/Liveness.spec
```diff
@@ -0,0 +1,428 @@
...
+}
+
+function CVL_canIncreaseCredit() returns bool {
+    if (forceCanIncreaseCreditFalse) {
+        bool shouldRevert;
+        if (shouldRevert) {
+            revert();
+        }
+        return false;
+    }
```
QGarchery:
```suggestion
    bool shouldRevert;
    if (shouldRevert) {
        revert();
    }
    if (forceCanIncreaseCreditFalse) {
        return false;
    }
```
that way we make no assumption about the revert behavior of `canIncreaseCredit`. This idea could be applied to other functions as well

certora/specs/Liveness.spec
```diff
@@ -0,0 +1,428 @@
...
+
+    // Oracle summary: ghost-controlled
+    function _.price() external => CVL_oraclePrice() expect(uint256);
+
+    // Gate summaries: ghost-controlled
+    function _.canIncreaseCredit(address) external => CVL_canIncreaseCredit() expect(bool);
+    function _.canIncreaseDebt(address) external => CVL_canIncreaseDebt() expect(bool);
+    function _.canLiquidate(address) external => CVL_canLiquidate() expect(bool);
+
+    // Callback summaries: ghost-controlled
```
QGarchery: This comment (and similar ones above) is not very informative, why is that ghost-controlled, and why it's ok/sound ?
In particular we assume that callbacks make no state change

certora/specs/Liveness.spec
QGarchery: Given that this spec file has many non-trivial summaries, I think it makes sense to assume as little as possible in those summaries. So for example all functions could optionally revert (on top of the forced reverts), and also havoc the state when possible

certora/specs/Liveness.spec
```diff
@@ -0,0 +1,428 @@
...
+/// GATE BLOCKING ///
+
+/// If enterGate.canIncreaseCredit returns false and take succeeds, the buyer's credit does not increase.
+rule enterGateBlocksCreditIncrease(env e, uint256 units, address taker, address takerCallback, bytes takerCallbackData, address receiver, Midnight.Offer offer, bytes ratifierData, bytes32 root, bytes32[] proof) {
+    require forceCanIncreaseCreditFalse, "canIncreaseCredit blocked";
+    require offer.obligation.enterGate != 0, "enter gate is set";
+
+    bytes32 id;
+    address buyer = offer.buy ? offer.maker : taker;
+    uint256 creditBefore = creditOf(id, buyer);
```
QGarchery: could we even check that no credit increases ?

```suggestion
    uint256 creditBefore = creditOf(id, user);
```

certora/specs/Liveness.spec
```diff
@@ -0,0 +1,428 @@
...
+
+    require id == lastId, "id derived from obligation";
+    uint256 debtAfter = debtOf(id, seller);
+
+    assert debtAfter <= debtBefore;
+}
+
+/// If the liquidator gate returns false on canLiquidate, liquidate reverts.
+rule liquidatorGateBlocksLiquidation(env e, Midnight.Obligation obligation, uint256 collateralIndex, uint256 seizedAssets, uint256 repaidUnits, address borrower, bytes data) {
+    require forceCanLiquidateFalse, "canLiquidate blocked";
```
QGarchery: Instead of the mechanism "force every external call of a given type to return a given value", we could use instead "assume that external calls of a given type are deterministic". So here it would look like

```suggestion
    require !summaryCanLiquidate(obligation.liquidatorGate), "canLiquidate blocked";
```
Note that:
- it's a bit more natural, we do it elsewhere in this repo (for example for mulDiv in this file)
- it allows easily to not constrain every call (only to `obligation.liquidatorGate` here)
- it doesn't work for revert though

certora/specs/Reverts.spec
```diff
@@ -0,0 +1,428 @@
...
+
+    liquidate@withrevert(e, obligation, collateralIndex, 0, repaidUnits, borrower, data);
+
+    require id == lastId, "id derived from obligation";
+
+    assert lastReverted;
+}
+
+/// If the oracle returns 0 and the borrower has debt, isHealthy returns false.
+rule oracleZeroCausesIsHealthyReturnFalse(env e, Midnight.Obligation obligation, bytes32 id, address borrower) {
```
QGarchery: Here the spec should be changed, it says (wrongly) that isHealthy reverts:
```
/// @dev If an activated collateral oracle returns 0 on `price`, `isHealthy`, `withdrawCollateral` when the borrower has
/// debt, `take` whenever the seller still has debt, and `liquidate` with repaid input all revert.
```

certora/specs/Reverts.spec
lilCertora: Yes the `lastId` value retain for evm revert because it's a persistent ghost, here `require id == lastId` ensures the precondition (`activatedCollaterals != 0`) applies to the obligation actually being liquidated so that the revert from liquidate call is the one we're looking for

certora/specs/Reverts.spec
lilCertora: Without it, the prover [finds a counterexample](https://prover.certora.com/output/7508195/2441605836614d11867df795fbccfab3/?anonymousKey=ad94a11b58d0729ffc1fcb36da439e21619b403b) where `liquidationLocked(id, seller)` is true before the call. In that case take succeeds even if the seller has debt, because the `isLiquidatable` check at the end short-circuits on `!liquidationLocked` and never calls `isHealthy` , so no oracle is invoked. This is unreachable in practice because we use transient , so we filter it out with this require. I've looked deeper into that and I don't think about a concern around that for now, do you ?

certora/specs/Liveness.spec
lilCertora: yes , this is better this way, i've applied it to all the cvl functions

certora/specs/Liveness.spec
lilCertora: yes

certora/specs/Reverts.spec
```diff
@@ -0,0 +1,444 @@
...
+
+    assert debtOf(id, seller) == 0;
+}
+
+/// ORACLE RETURNS ZERO ///
+
+/// If an activated collateral oracle returns 0 on price, liquidate with repaid input reverts.
+rule oracleZeroCausesLiquidateWithRepaidRevert(env e, Midnight.Obligation obligation, uint256 collateralIndex, uint256 repaidUnits, address borrower, bytes data) {
+    require singleZeroOracle == obligation.collateralParams[collateralIndex].oracle, "oracle returns zero";
+    require repaidUnits > 0, "using repaid units as input";
```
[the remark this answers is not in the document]
lilCertora: division by zero should still revert

certora/specs/Reverts.spec
```diff
@@ -0,0 +1,444 @@
...
+rule oracleRevertPreventsTakeWhenSellerHasDebt(env e, uint256 units, address taker, address takerCallback, bytes takerCallbackData, address receiver, Midnight.Offer offer, bytes ratifierData, bytes32 root, bytes32[] proof, uint256 collateralIndex) {
+    require singleRevertingOracle == offer.obligation.collateralParams[collateralIndex].oracle, "oracle is reverting";
+
+    bytes32 id;
+    address seller = offer.buy ? taker : offer.maker;
+    require !liquidationLocked(id, seller), "seller is not liquidation locked";
+
+    uint128 bitmap = activatedCollaterals(id, seller);
+    require summaryGetBit(bitmap, collateralIndex), "collateralIndex is activated";
+    require forall uint256 bit. bit > collateralIndex && bit < 128 => !summaryGetBit(bitmap, bit), "collateralIndex is the MSB";
```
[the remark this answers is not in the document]
lilCertora: not feasible path

certora/specs/Reverts.spec
```diff
@@ -0,0 +1,433 @@
...
+
+/// If enterGate.canIncreaseCredit returns false and take succeeds, no user's credit increases.
+rule enterGateBlocksCreditIncrease(env e, uint256 units, address taker, address takerCallback, bytes takerCallbackData, address receiver, Midnight.Offer offer, bytes ratifierData, bytes32 root, bytes32[] proof, address user) {
+    require !summaryCanIncreaseCredit(offer.obligation.enterGate), "canIncreaseCredit blocked";
+    require offer.obligation.enterGate != 0, "enter gate is set";
+
+    bytes32 id;
+    uint256 creditBefore = creditOf(id, user);
+
+    take(e, units, taker, takerCallback, takerCallbackData, receiver, offer, ratifierData, root, proof);
```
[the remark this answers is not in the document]
lilCertora: misunderstanding

certora/specs/Reverts.spec
```diff
@@ -0,0 +1,433 @@
...
+    assert debtOf(id, onBehalf) > 0 => reverted;
+}
+
+/// If an activated collateral oracle reverts on price, isHealthy reverts when the borrower has debt.
+/// Assumption : collateralIndex is the MSB otherwise isHealthy short-circuit could find a CEX where first collateral is enough to have maxDebt >= debt and return true..
+rule oracleRevertCausesIsHealthyRevert(env e, Midnight.Obligation obligation, bytes32 id, address borrower, uint256 collateralIndex) {
+    require singleRevertingOracle == obligation.collateralParams[collateralIndex].oracle, "oracle is reverting";
+
+    uint128 bitmap = activatedCollaterals(id, borrower);
+    require summaryGetBit(bitmap, collateralIndex), "collateralIndex is activated";
```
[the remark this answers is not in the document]
lilCertora: same as before

certora/specs/Reverts.spec
```diff
@@ -0,0 +1,433 @@
...
+    if (shouldRevert || forceTransferRevert) {
+        revert();
+    }
+}
+
+/// ORACLE REVERT PROPAGATION ///
+
+/// If any activated collateral oracle reverts on price, liquidate reverts.
+rule oracleRevertCausesLiquidateRevert(env e, Midnight.Obligation obligation, uint256 collateralIndex, uint256 seizedAssets, uint256 repaidUnits, address borrower, bytes data, uint256 revertingCollateralIndex) {
+    require singleRevertingOracle == obligation.collateralParams[revertingCollateralIndex].oracle, "oracle is reverting";
```
bhargavbh, line 185: IIUC this rule states that liquidate reverts if the oracle of collat being liquidated reverts. Actually liquidate reverts if any of the activated collateral oracle reverts (not just the collat being liquidated). Can we try capturing this?

certora/specs/Liveness.spec
```diff
@@ -0,0 +1,433 @@
...
+    // revert propagation is sound because EVM rollback is state-independent; gate checks precede callbacks.
+    // Oracle rules are sound because re-entrant callbacks cannot deactivate collaterals without calling
+    // withdrawCollateral -> isHealthy which would hit the same reverting/zero oracle.
+    function _.onBuy(bytes32, Midnight.Obligation, address, uint256, uint256, bytes) external => CVL_callbackBytes32() expect(bytes32);
+    function _.onSell(bytes32, Midnight.Obligation, address, uint256, uint256, bytes) external => CVL_callbackBytes32() expect(bytes32);
+    function _.onRepay(bytes32, Midnight.Obligation, uint256, address, bytes) external => CVL_callbackVoid() expect void;
+    function _.onLiquidate(bytes32, Midnight.Obligation, uint256, uint256, uint256, address, bytes) external => CVL_callbackVoid() expect void;
+    function _.onFlashLoan(address, uint256, bytes) external => CVL_callbackVoid() expect void;
+
+    function _.onRatify(Midnight.Offer, bytes32, bytes) external => NONDET;
```
bhargavbh: curious why the same modelling approach as onSell and onBuy is not used for onRatify which also returns a bytes32. to show the property: "if onRatify returns non-CALLBACK_SUCCESS, take reverts"

bhargavbh (commented):
the properties in this spec are about "something goes wrong => function reverts". In vault-v2 and blue, such properties were in Reverts.spec. We should also rename it to Reverts.spec to be consistent. 
Do we plan to have the other direction of revert conditions (i.e. function reverts => X/Y/Z was not satisfied) for the core functions?

certora/specs/Reverts.spec
lilCertora: The rule already captures that as `revertingCollateralIndex` and `collateralIndex` are independent parameters

certora/specs/Liveness.spec
lilCertora: true, doing it

certora/specs/Reverts.spec
QGarchery: Ok it makes sense, but I guess it depends on when the ghost is set versus when the revert happens. In our case it seems to work well because the `toId`/`touchObligation` is called first. But maybe worth refactoring and using the same method as in Midnight.spec for the `toId` summary (also removes the need to write this require every time)

certora/specs/Reverts.spec
QGarchery: In theory you could have nested takes and only do the liquidation check at the end. So a given `take` call would not revert even if though every oracle reverts, but that's okay because the liquidation check that will come at the end will revert

certora/specs/Reverts.spec
```diff
@@ -0,0 +1,443 @@
...
+    function _.onFlashLoan(address, uint256, bytes) external => CVL_callbackVoid() expect void;
+
+    // Token transfers: routed through CVL functions to force revert per rule. Modeled as no-op on success
+    // (no balance tracking), which is sound for revert-propagation rules.
+    function SafeTransferLib.safeTransferFrom(address, address, address, uint256) internal => CVL_safeTransferFrom();
+    function SafeTransferLib.safeTransfer(address, address, uint256) internal => CVL_safeTransfer();
+
+    // Bitmap operations (msb, clearBit, setBit) are provided by BitmapSummaries.spec.
+    function IdLib.toId(Midnight.Obligation memory obligation, uint256 chainId, address midnight) internal returns (bytes32) => CVL_toId(obligation, chainId, midnight);
+    function IdLib.storeInCode(Midnight.Obligation memory) internal returns (address) => NONDET;
```
QGarchery, line 49:
```suggestion
    // The function `toObligation` is not used by the protocol.
    function IdLib.storeInCode(Midnight.Obligation memory) internal returns (address) => NONDET;
```

certora/specs/Reverts.spec
```diff
@@ -0,0 +1,443 @@
...
+    // Gates: routed through CVL functions using calledContract to identify which gate is being called.
+    // Return values are deterministic per gate address via ghost functions, so
+    // rules can constrain a specific gate's return value without affecting other gates. Each call can also
+    // nondeterministically revert, modeling that external gates can fail for any reason.
+    function _.canIncreaseCredit(address) external => CVL_canIncreaseCredit(calledContract) expect(bool);
+    function _.canIncreaseDebt(address) external => CVL_canIncreaseDebt(calledContract) expect(bool);
+    function _.canLiquidate(address) external => CVL_canLiquidate(calledContract) expect(bool);
+
+    // Callbacks: ghost-controlled to force revert/bad-return per rule. Modeled as pure (no state changes):
+    // revert propagation is sound because EVM rollback is state-independent; gate checks precede callbacks.
```
QGarchery: Not clear to me what "EVM rollback is state-independent" means and why it's important here

certora/specs/Reverts.spec
```diff
@@ -0,0 +1,443 @@
...
+
+    flashLoan@withrevert(e, token, assets, callback, data);
+
+    assert lastReverted;
+}
+
+/// If a buy/sell callback reverts or returns something other than CALLBACK_SUCCESS, callback-enabled take reverts.
+rule callbackRevertOrBadReturnCausesTakeRevert(env e, uint256 units, address taker, address takerCallback, bytes takerCallbackData, address receiver, Midnight.Offer offer, bytes ratifierData, bytes32 root, bytes32[] proof) {
+    require forceCallbackRevert || forceCallbackBadReturn, "callback reverts or returns bad value";
+    require takerCallback != 0 || offer.callback != 0, "callback-enabled take";
```
QGarchery:
```suggestion
    require takerCallback != 0 || offer.callback != 0 || offer.ratifier != 0, "callback-enabled take";
```
and remove the last rule

certora/specs/Reverts.spec
```diff
@@ -0,0 +1,443 @@
...
+
+    // Oracle: routed through CVL function to allow ghost flags to force specific behaviors (revert, return zero) per rule.
+    // calledContract is used to target a single oracle address for per-oracle revert control (used by oracle revert/zero rules).
+    function _.price() external => CVL_oraclePrice(calledContract) expect(uint256);
+
+    // Gates: routed through CVL functions using calledContract to identify which gate is being called.
+    // Return values are deterministic per gate address via ghost functions, so
+    // rules can constrain a specific gate's return value without affecting other gates. Each call can also
+    // nondeterministically revert, modeling that external gates can fail for any reason.
+    function _.canIncreaseCredit(address) external => CVL_canIncreaseCredit(calledContract) expect(bool);
```
QGarchery: Nitpick but it would make more sense to me to have

```suggestion
    function _.canIncreaseCredit(address) external => summaryCanIncreaseCredit(calledContract) expect(bool);
```
and then have `summaryCanIncreaseCredit` use the value of a `ghostCanIncreaseCredit`. Right now the ghost variable is called "summary" but this is a bit confusing because it's not what is used as a summary

certora/specs/Reverts.spec
```diff
@@ -0,0 +1,443 @@
...
+    function _.canIncreaseDebt(address) external => CVL_canIncreaseDebt(calledContract) expect(bool);
+    function _.canLiquidate(address) external => CVL_canLiquidate(calledContract) expect(bool);
```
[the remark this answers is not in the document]
lilCertora: Dropping the user argument is an over-approximation

certora/specs/Reverts.spec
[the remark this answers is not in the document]
lilCertora: false and the fix wouldn't work

lilCertora: yes thanks

certora/specs/Reverts.spec
lilCertora: yes this wasn't very clear too, i modified , thanks

certora/specs/Reverts.spec
lilCertora: yes, I added some clarity

QGarchery (approved):

src/Midnight.sol
```diff
@@ -102,7 +105,7 @@ import {EventsLib} from "./libraries/EventsLib.sol";
 /// @dev If a token sent by Midnight reverts on `transfer` despite balances being right, `withdraw`,
 /// `withdrawCollateral`, fee claims, the collateral leg of `liquidate`, and `flashLoan` revert when they need to send
 /// that token.
-/// @dev If a callback reverts, or if a buy/sell callback returns something other than `CALLBACK_SUCCESS`,
+/// @dev If a callback reverts, or if a buy/sell/ratify callback returns something other than `CALLBACK_SUCCESS`,
```
MathisGD: will need to re-fmt
```suggestion
/// @dev If a callback reverts or returns something other than `CALLBACK_SUCCESS`,
```

src/Midnight.sol
```diff
@@ -90,10 +90,13 @@ import {EventsLib} from "./libraries/EventsLib.sol";
...
 ///
 /// LIVENESS
-/// @dev If an activated collateral oracle reverts on `price`, `liquidate`, `isHealthy`, `withdrawCollateral`  when the
-/// borrower has debt, and `take` whenever the seller still has debt all revert.
-/// @dev If an activated collateral oracle returns 0 on `price`, `isHealthy`, `withdrawCollateral` when the borrower has
-/// debt, `take` whenever the seller still has debt, and `liquidate` with repaid input all revert.
+/// @dev If an activated collateral oracle reverts on `price`, `liquidate` reverts unconditionally.
+/// @dev If `isHealthy` is not short-circuited (i.e. the reverting oracle is reached before `maxDebt >= debt`),
+/// `isHealthy`, `withdrawCollateral` when the borrower has debt, and `take` whenever the seller still has debt
+/// all revert.
```
MathisGD: wdyt? more usable
```suggestion
/// @dev If an activated collateral oracle reverts on `price`, `isHealthy`, `withdrawCollateral` when the borrower has debt, and `take` whenever the seller still has debt might revert.
```

src/Midnight.sol
```diff
@@ -90,10 +90,13 @@ import {EventsLib} from "./libraries/EventsLib.sol";
...
+/// @dev If the liquidated collateral oracle returns 0 on `price`, `liquidate` with repaid input reverts.
+/// @dev If all activated collateral oracles return 0 on `price`, `isHealthy` returns false when the borrower has debt,
+/// `withdrawCollateral` reverts when the borrower has debt, and `take` reverts when the seller has debt.
```
MathisGD: that's not unexpected right? they are just unhealthy 
```suggestion
```

src/Midnight.sol
lilCertora: yes that's okay the "might revert" still captures the short circuit

QGarchery (approved):

certora/specs/Reverts.spec
```diff
@@ -0,0 +1,422 @@
...
+
+/// If any activated collateral oracle reverts on price, liquidate reverts.
+rule oracleRevertCausesLiquidateRevert(env e, Midnight.Obligation obligation, uint256 collateralIndex, uint256 seizedAssets, uint256 repaidUnits, address borrower, bytes data, uint256 revertingCollateralIndex) {
+    require singleRevertingOracle == obligation.collateralParams[revertingCollateralIndex].oracle, "oracle is reverting";
+
+    bytes32 id = summaryToId(obligation);
+    uint128 bitmap = activatedCollaterals(id, borrower);
+    require summaryGetBit(bitmap, revertingCollateralIndex), "revertingCollateralIndex is activated";
+
+    liquidate@withrevert(e, obligation, collateralIndex, seizedAssets, repaidUnits, borrower, data);
```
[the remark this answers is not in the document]
lilCertora: updated

certora/specs/Reverts.spec
```diff
@@ -0,0 +1,422 @@
...
+
+/// TOKEN TRANSFER REVERT PROPAGATION ///
+
+/// If transferFrom reverts, take, repay, supplyCollateral, liquidate, and flashLoan all revert.
+rule transferFromRevertPropagation(method f, env e, calldataarg args)
+filtered {
+    f -> f.selector == sig:take(uint256, address, address, bytes, address, Midnight.Offer, bytes, bytes32, bytes32[]).selector
+        || f.selector == sig:repay(Midnight.Obligation, uint256, address, bytes).selector
+        || f.selector == sig:supplyCollateral(Midnight.Obligation, uint256, uint256, address).selector
+        || f.selector == sig:liquidate(Midnight.Obligation, uint256, uint256, uint256, address, bytes).selector
```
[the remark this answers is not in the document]
lilCertora: updated

certora/specs/Reverts.spec
```diff
@@ -0,0 +1,422 @@
...
+}
+
+/// CALLBACK REVERT PROPAGATION ///
+
+/// If the callback reverts, callback-enabled repay (with non-empty data) reverts.
+rule callbackRevertCausesRepayRevert(env e, Midnight.Obligation obligation, uint256 units, address onBehalf, bytes data) {
+    require forceCallbackRevert, "callback reverts";
+    require data.length > 0, "callback-enabled repay";
+
+    repay@withrevert(e, obligation, units, onBehalf, data);
```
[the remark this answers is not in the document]
lilCertora: updated

certora/specs/Reverts.spec
```diff
@@ -0,0 +1,416 @@
...
+    f@withrevert(e, args);
+    assert lastReverted;
+}
+
+/// CALLBACK REVERT PROPAGATION ///
+
+/// If the callback reverts, callback-enabled repay (non-zero callback) reverts.
+rule callbackRevertCausesRepayRevert(env e, Midnight.Obligation obligation, uint256 units, address onBehalf, address callback, bytes data) {
+    require forceCallbackRevert, "callback reverts";
+    require callback != 0, "callback-enabled repay";
```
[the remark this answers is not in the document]
lilCertora: true , applied

certora/specs/Reverts.spec
```diff
@@ -0,0 +1,416 @@
...
+    require repaidUnits > 0, "using repaid units as input";
+
+    liquidate@withrevert(e, obligation, collateralIndex, 0, repaidUnits, borrower, receiver, callback, data);
+
+    assert lastReverted;
+}
+
+/// If all oracles return 0 and the borrower has debt, isHealthy returns false.
+/// Note: this forces ALL oracles to return 0 because isHealthy has a short-circuit (while maxDebt < debt).
+rule oracleZeroCausesIsHealthyReturnFalse(env e, Midnight.Obligation obligation, address borrower) {
```
MathisGD, line 264: not a big deal, but not really "revert"

certora/specs/Reverts.spec
```diff
@@ -0,0 +1,416 @@
...
+    require activatedCollaterals(id, borrower) != 0, "borrower has activated collaterals";
+
+    bool healthy = isHealthy(e, obligation, id, borrower);
+
+    assert debtOf(id, borrower) > 0 => !healthy;
+}
+
+/// If all oracles return 0, withdrawCollateral reverts when the borrower has debt.
+/// Note: same short-circuit limitation as oracleZeroCausesIsHealthyReturnFalse
+rule oracleZeroCausesWithdrawCollateralRevert(env e, Midnight.Obligation obligation, uint256 collateralIndex, uint256 assets, address onBehalf, address receiver) {
```
MathisGD: same, it's more like unHealthy => withdraw impossible

certora/specs/Reverts.spec
```diff
@@ -0,0 +1,416 @@
...
+    require forceOracleReturnZero, "all oracles return zero";
+
+    bytes32 id = summaryToId(offer.obligation);
+    address seller = offer.buy ? taker : offer.maker;
+    require !liquidationLocked(id, seller), "seller is not liquidation locked";
+
+    take@withrevert(e, units, taker, takerCallback, takerCallbackData, receiver, offer, ratifierData, root, proof);
+    bool reverted = lastReverted;
+
+    assert debtOf(id, seller) > 0 => reverted;
```
MathisGD: otherwise it's weird, no?
```suggestion
    assert debtOf(id, seller) == 0;
```

MathisGD (approved):

certora/specs/Reverts.spec
lilCertora: yes that's right

certora/specs/Reverts.spec
lilCertora: it's already saying "return false" so it's ok no ?

certora/specs/Reverts.spec
lilCertora: yes i agree

certora/specs/Reverts.spec
MathisGD: yes, it's more that it doesn't belong to this spec

lilCertora: I agree

MathisGD (approved):

PR #688, opened 2026-04-14
arbitrary callback targets

MathisGD:
- motivation: https://morpholabs.slack.com/archives/C08A40KKN9W/p1775811715285949
- changed the payer everywhere to follow what was done in `take`
- added the buyerCallback/callback in the events


src/Midnight.sol
```diff
@@ -439,8 +440,11 @@ contract Midnight is IMidnight {
...
-        if (data.length > 0) {
-            ICallbacks(msg.sender).onRepay(id, obligation, units, onBehalf, data);
+        if (callback != address(0)) {
+            require(
+                ICallbacks(callback).onRepay(id, obligation, units, onBehalf, data) == CALLBACK_SUCCESS,
+                WrongRepayCallbackReturnValue()
+            );
         }
 
         SafeTransferLib.safeTransferFrom(obligation.loanToken, msg.sender, address(this), units);
```
adhusson: should the transfer be from the callback?

src/Midnight.sol
```diff
@@ -617,9 +622,13 @@ contract Midnight is IMidnight {
...
+        if (callback != address(0)) {
+            require(
+                ICallbacks(callback)
+                    .onLiquidate(id, obligation, collateralIndex, seizedAssets, repaidUnits, borrower, data)
+                == CALLBACK_SUCCESS,
+                WrongLiquidateCallbackReturnValue()
+            );
         }
 
         SafeTransferLib.safeTransferFrom(obligation.loanToken, msg.sender, address(this), repaidUnits);
```
adhusson: should be taken from the callback?

src/Midnight.sol
```diff
@@ -617,9 +622,13 @@ contract Midnight is IMidnight {
 
         SafeTransferLib.safeTransfer(obligation.collateralParams[collateralIndex].token, msg.sender, seizedAssets);
```
adhusson: should be sent to the callback?

test/FlashloanTest.sol
adhusson, line 1: should update flashloan/liquidation/repay tests for cases where callback != msg.sender

src/Midnight.sol
QGarchery: If we see a repay as a buy take at price 1 when you have a borrow position, then I think we want `payer = callback != address(0) ? buyerCallback : msg.sender`

src/Midnight.sol
```diff
@@ -660,9 +664,12 @@ contract Midnight is IMidnight {
 
     function flashLoan(address token, uint256 assets, address callback, bytes calldata data) external {
         emit EventsLib.FlashLoan(msg.sender, token, assets);
-        SafeTransferLib.safeTransfer(token, msg.sender, assets);
-        IFlashLoanCallback(callback).onFlashLoan(token, assets, data);
-        SafeTransferLib.safeTransferFrom(token, msg.sender, address(this), assets);
+        SafeTransferLib.safeTransfer(token, callback, assets);
```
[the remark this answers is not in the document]
MathisGD: same for other functions, thoughts @QGarchery @peyha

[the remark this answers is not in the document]
MathisGD: actually I think that we should do it

src/libraries/EventsLib.sol
```diff
@@ -17,15 +17,15 @@ library EventsLib {
     event SetDefaultContinuousFee(address indexed loanToken, uint256 newContinuousFee);
     event UpdatePosition(bytes32 indexed id_, address indexed user, uint256 creditDecrease, uint256 pendingFeeDecrease, uint256 accruedFee);
     event ObligationCreated(bytes32 indexed id_, Obligation obligation);
-    event Take(address caller, bytes32 indexed id_, address indexed maker, address indexed taker, bool offerIsBuy, uint256 buyerAssets, uint256 sellerAssets, uint256 units, address sellerReceiver, bytes32 group, uint256 consumed, uint256 buyerPendingFeeIncrease, uint256 sellerPendingFeeDecrease, uint256 buyerCreditIncrease, uint256 sellerCreditDecrease);
+    event Take(address caller, bytes32 indexed id_, address indexed maker, address indexed taker, bool offerIsBuy, uint256 buyerAssets, uint256 sellerAssets, uint256 units, address sellerReceiver, bytes32 group, uint256 consumed, uint256 buyerPendingFeeIncrease, uint256 sellerPendingFeeDecrease, uint256 buyerCreditIncrease, uint256 sellerCreditDecrease, address buyerCallback);
```
MathisGD: not sure. rationale for changing would be to make it clear why we don't log sellerCallback
```suggestion
    event Take(address caller, bytes32 indexed id_, address indexed maker, address indexed taker, bool offerIsBuy, uint256 buyerAssets, uint256 sellerAssets, uint256 units, address sellerReceiver, bytes32 group, uint256 consumed, uint256 buyerPendingFeeIncrease, uint256 sellerPendingFeeDecrease, uint256 buyerCreditIncrease, uint256 sellerCreditDecrease, address payer);
```

src/Midnight.sol
```diff
@@ -437,20 +439,26 @@ contract Midnight is IMidnight {
...
+        if (callback != address(0)) {
+            require(
+                IRepayCallback(callback).onRepay(id, obligation, units, onBehalf, data) == CALLBACK_SUCCESS,
+                WrongRepayCallbackReturnValue()
+            );
         }
 
-        SafeTransferLib.safeTransferFrom(obligation.loanToken, msg.sender, address(this), units);
+        address payer = callback != address(0) ? callback : msg.sender;
+        SafeTransferLib.safeTransferFrom(obligation.loanToken, payer, address(this), units);
```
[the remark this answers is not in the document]
MathisGD: that's why there is the callback return value

src/Midnight.sol
```diff
@@ -102,9 +102,8 @@ import {EventsLib} from "./libraries/EventsLib.sol";
 /// @dev If a token sent by Midnight reverts on `transfer` despite balances being right, `withdraw`,
 /// `withdrawCollateral`, fee claims, the collateral leg of `liquidate`, and `flashLoan` revert when they need to send
 /// that token.
-/// @dev If a callback reverts, or if a buy/sell callback returns something other than `CALLBACK_SUCCESS`,
-/// callback-enabled `take`, `repay`, `liquidate`, and `flashLoan` revert.
-///
+/// @dev If a callback reverts, or if a callback returns something other than `CALLBACK_SUCCESS`, callback-enabled
```
```suggestion
/// @dev If a callback reverts or returns something other than `CALLBACK_SUCCESS`, callback-enabled
```

src/libraries/EventsLib.sol
peyha: we log callback addresses in `repay`, `liquidate` and `flashloan`so why wouldn't we do it here ? It could be useful for some data queries (e.g., how much volume on leverage callback?) and is quite cheap (256 gas)

peyha: also payer does not add much info and you could get it through transfer logs

MathisGD: I added it to the event just to have the payer. the payer is important if you want to index the allowance of the contract

MathisGD: I'm getting convinced that the payer should be logged, to clarify

peyha: ah it's for the allowance, then payer is simpler to understand. otherwise you have to check the condition `address payer = buyerCallback != address(0) ? buyerCallback : (offer.buy ? buyer : msg.sender)` in the indexer

MathisGD: changing right now

peyha (approved):
lgtm pending event discussion

src/Midnight.sol
```diff
@@ -659,10 +675,13 @@ contract Midnight is IMidnight {
...
-        emit EventsLib.FlashLoan(msg.sender, token, assets);
-        SafeTransferLib.safeTransfer(token, msg.sender, assets);
-        IFlashLoanCallback(callback).onFlashLoan(token, assets, data);
-        SafeTransferLib.safeTransferFrom(token, msg.sender, address(this), assets);
+        emit EventsLib.FlashLoan(msg.sender, token, assets, callback);
+        SafeTransferLib.safeTransfer(token, callback, assets);
+        require(
+            IFlashLoanCallback(callback).onFlashLoan(token, assets, data) == CALLBACK_SUCCESS,
+            WrongFlashLoanCallbackReturnValue()
+        );
```
QGarchery, line 663: Curious if you think that there could be a vulnerability if we don't check the return value (but I know we want to do it for consistency)

src/Midnight.sol
```diff
@@ -369,30 +369,33 @@ contract Midnight is IMidnight {
...
-            units,
-            receiver,
-            offer.group,
-            newConsumed,
-            buyerPendingFeeIncrease,
-            sellerPendingFeeDecrease,
-            buyerCreditIncrease,
-            sellerCreditDecrease
-        );
+        {
```
QGarchery: it doesn't seem needed to add the brackets when I test locally

src/Midnight.sol
```diff
@@ -620,17 +608,21 @@ contract Midnight is IMidnight {
             repaidUnits,
             borrower,
             badDebt,
-            _obligationState.lossIndex
+            _obligationState.lossIndex,
+            payer
         );
+        SafeTransferLib.safeTransfer(obligation.collateralParams[collateralIndex].token, payer, seizedAssets);
```
```suggestion
        SafeTransferLib.safeTransfer(obligation.collateralParams[collateralIndex].token, receiver, seizedAssets);
```
if you think about liquidate as repay + withdraw. But maybe it's not great as we lose the property that if `callback == msg.sender` then the computation is the same as before this PR and the same as on Blue

src/Midnight.sol
MathisGD: maybe not onchain, it means at least that you can create arbitrary large IN and OUT transfer events from any accounts that authorized midnight..

src/Midnight.sol
MathisGD: I don't think that would be used much. Also not sure that the abstraction repay+withdraw is that useful. Though maybe we should rename payer?

MathisGD: I changed my mind, it's really cheap

QGarchery (approved):

adhusson:
LGTM, added tests for sender != callback https://github.com/morpho-org/midnight/pull/765

PR #690, opened 2026-04-14
update position returns more values

adhusson:


src/Midnight.sol
```diff
@@ -720,17 +720,22 @@ contract Midnight is IMidnight {
...
+        return _updatePosition(obligation, id, user);
     }
 
     /// @dev Expects the obligation to be touched.
     /// @dev Expects the id to correspond to the obligation's id.
-    function _updatePosition(Obligation memory obligation, bytes32 id, address user) internal {
+    /// @dev Returns the new credit, new pending fee, and accrued fee after having updated the position.
+    function _updatePosition(Obligation memory obligation, bytes32 id, address user)
+        internal
+        returns (uint128 newCredit, uint128 newPendingFee, uint128 accruedFee)
```
MathisGD: why named return values? if no particular reason I would revert

adhusson: readability, it's either

```
function f() returns (uint a, uint b) {
  (a,b) = g();
  // more stuff
}
```

or

```
function f() returns (uint,uint) {
  (uint a, uint b) = g();
  // more stuff
  return (a,b);
}
```

wdyt on the readability aspect? not a strong preference for me happy to revert

MathisGD: the second one is better. in the first one if you have c and forget to set it (notably with an early return)...

MathisGD (approved):

peyha (approved):

src/Midnight.sol
QGarchery, line 1: We could add a test that indeed the return values are the new credit, new pending fee, and accrued fee after

QGarchery (approved):

MathisGD (approved):

PR #677, opened 2026-04-07
Custom errors 

peyha:

PR #691, opened 2026-04-14
use IMidnight interface

adhusson:


MathisGD (approved):

MathisGD:
~~should we change in tests too?~~

peyha (commented):
Need to wait for #677 first because it introduces a takeBundler interface

peyha (approved):

QGarchery (approved):

PR #693, opened 2026-04-14
bundler: allow partial fill

### LINKED ISSUE #692: bundler: FOK

MathisGD:
some questions:
- a partial fill of a limit order on an CLOB is not the same thing as here, because here you don't get a limit order in the book for the missing amount. should it actually behave this way? (would require some change at the midnight level I guess) -> only the simple behavior for now
- in terms of naming, is it ok to use allowPartialFill still? note that the default behavior on exchanges for limit orders is allow partial fills, and then you have an option immediate_or_cancel which allows you to cancel the remaining part if you want. ([source](https://docs.deribit.com/api-reference/trading/private-buy#parameter-time-in-force)) -> went for something very low level to avoid any confusion
- the min/max prices check has been refactored to account for this change

slack thread: https://morpholabs.slack.com/archives/C08A40KKN9W/p1776186464435669

fixes #692


src/periphery/TakeBundler.sol
adhusson, line 1: IOC is a [valid name](https://en.wikipedia.org/wiki/Immediate_or_cancel) for an order type not just an option.

Some ideas:
- The name "take units" is already pretty clear, it's not a limit order it's a take.
- Or rename to `bundleTakeImmediate` so it's even clearer (then `allowPartialFill` bool).
- 2 or 3 functions `bundleTakeFOK` `bundleTakeIOC`, possibly `bundleTakeLimit`.
- If 3 order types, an enum starts to make sense `{FillOrKill,ImmediateOrCancel,Limit}`


Note we are already not exactly a limit order since we are checking the average buy price, not stopping as soon as an offer has a price above a threshold. And if we added a "post unfilled amount" feature, at what price would that offer be?

src/periphery/TakeBundler.sol
```diff
@@ -32,7 +32,8 @@ contract TakeBundler {
         uint256 minBuyerAssets,
         uint256 maxBuyerAssets,
         uint256 minSellerAssets,
-        uint256 maxSellerAssets
+        uint256 maxSellerAssets,
+        bool allowPartialFill
```
peyha: Indeed this parameter is a bit misleading, there are actually two different kind of "partial fill" concept
- partial fill in the sense that `targetUnits` is not reached after all the takes
- partial fill in the sense of taking stricty less than `UtilsLib.min(targetUnits - totalFilledUnits, takes[i].units)` in some take
To me implementing one or the other depend on the nature of the book. If offers are very small in size compared to `targetUnits` we would need the first concept. If the book is very high frequency and some offers might be taken within the same block/the block before then the second concept would be useful. We can also go for both for full flexibility

src/periphery/TakeBundler.sol
```diff
@@ -101,7 +103,7 @@ contract TakeBundler is ITakeBundler {
             } catch {}
         }
 
-        require(totalFilledBuyerAssets == targetBuyerAssets, InsufficientLiquidity());
+        require(allowPartialFill || totalFilledBuyerAssets == targetBuyerAssets, InsufficientLiquidity());
```
QGarchery: If `allowPartialFill` is true, then the slippage check may not work as intended. In this function, the min/maxUnits should be scaled to the amount that was effectively taken

QGarchery (changes requested):

src/periphery/TakeBundler.sol
```diff
@@ -52,26 +55,28 @@ contract TakeBundler is ITakeBundler {
             } catch {}
         }
 
-        require(totalFilledUnits == targetUnits, InsufficientLiquidity());
-        require(totalBuyerAssets >= minBuyerAssets, BuyerAssetsBelowMin());
-        require(totalBuyerAssets <= maxBuyerAssets, BuyerAssetsAboveMax());
-        require(totalSellerAssets >= minSellerAssets, SellerAssetsBelowMin());
-        require(totalSellerAssets <= maxSellerAssets, SellerAssetsAboveMax());
+        require(skipRevertOnTargetNotMet || totalFilledUnits == targetUnits, InsufficientLiquidity());
+        require(totalBuyerAssets.mulDivDown(WAD, totalFilledUnits) >= minBuyerPrice, PriceBelowMin());
```
peyha: if we want to be really accurate, prob not worth it
```suggestion
        require(totalBuyerAssets * WAD >= minBuyerPrice * totalFilledUnits, PriceBelowMin());
```

MathisGD: see [this](https://morpholabs.slack.com/archives/C08A40KKN9W/p1776415516821789?thread_ts=1776414999.647309&cid=C08A40KKN9W), it's equivalent. and the current formula is nicer

MathisGD: caveat: not exactly equivalent, if totalFilledUnits==0. so let's do the change

MathisGD:
https://morpholabs.slack.com/archives/C08A40KKN9W/p1776688332469839?thread_ts=1776186464.435669&cid=C08A40KKN9W

PR #696, opened 2026-04-15
test with an old version of foundry

MathisGD:


.github/workflows/forge-old.yml
```diff
@@ -0,0 +1,33 @@
...
+        uses: actions/checkout@v4
+        with:
+          submodules: recursive
+      - name: Install Foundry (old)
+        uses: foundry-rs/foundry-toolchain@v1
+        with:
+          version: nightly-a44159a5c23d2699d3a390e6d4889b89a0e5a5e0
+      - name: Verify forge executable hash
+        env:
+          EXPECTED_FORGE_SHA256: 9366cb6647524cac03f6e7257e773eb6389d613c86fd0c47a5f523f1616cd23a
```
adhusson: verified against https://github.com/foundry-rs/foundry/releases/download/nightly-a44159a5c23d2699d3a390e6d4889b89a0e5a5e0/foundry_nightly_linux_amd64.tar.gz

adhusson (approved):

.github/workflows/forge-old.yml
MathisGD: sorry I had to change...

.github/workflows/forge-old.yml
```diff
@@ -0,0 +1,49 @@
...
+        run: |
+          forge_path="$(which forge)"
+          echo "forge path: $forge_path"
+          actual="$(sha256sum "$forge_path" | awk '{print $1}')"
+          echo "actual:   $actual"
+          echo "expected: $EXPECTED_FORGE_SHA256"
+          [ "$actual" = "$EXPECTED_FORGE_SHA256" ] || { echo "forge hash mismatch"; exit 1; }
+      - name: Run Forge tests (${{ matrix.name }})
+        run: forge test
```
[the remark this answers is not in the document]
MathisGD: it's the opposite no?

.github/workflows/forge-old.yml
```diff
@@ -0,0 +1,40 @@
...
+        uses: actions/checkout@v4
+        with:
+          submodules: recursive
+      - name: Install Foundry (old)
+        uses: foundry-rs/foundry-toolchain@v1
+        with:
+          version: v1.5.1
+      - name: Verify forge executable hash
+        env:
+          EXPECTED_FORGE_SHA256: a39fc1913b53dde6a35b603756f65d799e13817ef694107ec0d3704c20435cce
```
adhusson, line 24: OK https://github.com/foundry-rs/foundry/releases/download/v1.5.1/foundry_v1.5.1_linux_amd64.tar.gz

adhusson (approved):

.github/workflows/forge-old.yml
```diff
@@ -0,0 +1,40 @@
+name: forge-old
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
```
QGarchery, line 7:
```suggestion
```
(optional) maybe ok now, but if it adds too much time we could do that

.github/workflows/forge-old.yml
```diff
@@ -0,0 +1,40 @@
...
+          echo "actual:   $actual"
+          echo "expected: $EXPECTED_FORGE_SHA256"
+          [ "$actual" = "$EXPECTED_FORGE_SHA256" ] || { echo "forge hash mismatch"; exit 1; }
+      - name: Install solc 0.8.34 into svm cache
+        run: |
+          mkdir -p "$HOME/.svm/0.8.34"
+          curl -fsSL -o "$HOME/.svm/0.8.34/solc-0.8.34" \
+            https://github.com/ethereum/solidity/releases/download/v0.8.34/solc-static-linux
+          chmod +x "$HOME/.svm/0.8.34/solc-0.8.34"
+          "$HOME/.svm/0.8.34/solc-0.8.34" --version
```
QGarchery, line 38:
```suggestion
```
since the old forge version is trusted, it can be trusted to install the right solc binary too. If it's not trusted it could use a completely different binary anyway, no matter what is in `~/.svm`

QGarchery (approved):

.github/workflows/forge-old.yml
QGarchery, line 1: Another thing we could do is check that the same bytecode is produced for the contracts, with the old forge version and the new

.github/workflows/forge-old.yml
MathisGD: tests are really really fast compared to certora

MathisGD: also I feel that we don't check as much the main CI

.github/workflows/forge-old.yml
MathisGD: the problem is that old forge version don't know about solidity 0.8.34... we are forced to install by hand or use a newer version (newer than stable, actually)

.github/workflows/forge-old.yml
MathisGD: good idea, looking in to this

MathisGD: done

.github/workflows/forge-old.yml
```diff
@@ -0,0 +1,82 @@
...
+          echo "forge path: $forge_path"
+          actual="$(sha256sum "$forge_path" | awk '{print $1}')"
+          echo "actual:   $actual"
+          echo "expected: $EXPECTED_FORGE_SHA256"
+          [ "$actual" = "$EXPECTED_FORGE_SHA256" ] || { echo "forge hash mismatch"; exit 1; }
+      - name: Install solc 0.8.34 into svm cache
+        run: |
+          mkdir -p "$HOME/.svm/0.8.34"
+          curl -fsSL -o "$HOME/.svm/0.8.34/solc-0.8.34" \
+            https://github.com/ethereum/solidity/releases/download/v0.8.34/solc-static-linux
```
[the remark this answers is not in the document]
MathisGD: right, but wontfix

.github/workflows/forge-old.yml
```diff
@@ -0,0 +1,82 @@
...
+          chmod +x "$HOME/.svm/0.8.34/solc-0.8.34"
+          "$HOME/.svm/0.8.34/solc-0.8.34" --version
+      - name: Install Foundry (old)
+        uses: foundry-rs/foundry-toolchain@v1
+        with:
+          version: v1.5.1
+      - name: Build with Foundry (old)
+        run: forge build --force --no-metadata --out out-old --cache-path cache-old
```
QGarchery: It would be nice to do `forge test` to replace the `test:` job above. It can be done by renaming `out-old` -> `out`, `out` -> `out-new`

QGarchery (approved):

PR #698, opened 2026-04-15
cvl context

MathisGD:
written by an other LLM ^^

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -28,7 +28,7 @@ The primer below describes how CVL actually behaves — read it before reasoning
 - `calldataarg args` is an opaque bundle of arguments. You cannot inspect or constrain its fields; it only exists to call a `method f` parametrically.
 - `f(e, args)` targets `currentContract` by default. Call other contracts via `other.f(e, args)` where `other` is declared with `using OtherContract as other;`.
 - Addresses drawn nondeterministically may coincide with known contracts unless explicitly constrained (`require addr != currentContract`, etc.). The Prover does *not* assume addresses are distinct.
-- `envfree` tells the Prover a method doesn't read `env`; such calls omit the `env` argument. The Prover statically checks this, but a declaration mismatch with the actual Solidity is still possible and fails silently.
+- `envfree` tells the Prover a method doesn't read `env`; such calls omit the `env` argument. The Prover statically checks this.
 
 #### Ghosts, hooks, havoc
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -8,11 +8,11 @@ The primer below describes how CVL actually behaves — read it before reasoning
 #### Rules, invariants, `satisfy`
 
 - A `rule` passes iff every `assert` holds on every execution path that satisfies all preceding `require`s. A counterexample is a concrete trace (env, args, method choice) that satisfies the `require`s and falsifies an `assert`.
-- A rule can be *vacuous*: if the `require`s exclude every path, it passes trivially. Vacuity is silent unless a sanity rule is run.
+- A rule can be *vacuous*: if the `require`s exclude every path, it passes trivially. Vacuity is silent unless a sanity rule is run. Basic sanity is run by default.
 - `satisfy p` is the dual of `assert`: the Prover must find at least one feasible path where `p` holds. A rule with only `satisfy` and no `assert` therefore does not check any universal property.
-- CVL has both *weak* and *strong* invariants. A weak invariant is proven by induction over methods: base case checks it after the constructor; inductive step assumes it, runs an arbitrary method `f`, then asserts it again. By default invariants are checked only for `public`/`external` non-`view`/non-`pure` methods. A strong invariant adds checks around unresolved external calls that may modify the current contract's state. There is no temporal reasoning beyond this.
-- A `preserved` block injects extra `require`s into the inductive step for one method. Those assumptions are *not* verified elsewhere — the Prover trusts them. Unsound `preserved` is the most common source of fake invariant proofs.
-- `requireInvariant J` assumes another invariant `J` at that point. It's sound only because `J` was itself proven by the same induction scheme.
+- CVL has both *weak* and *strong* invariants. A weak invariant is proven by induction over methods: base case checks it after the constructor; inductive step assumes it, runs an arbitrary method `f`, then asserts it again. By default invariants are checked only for `public`/`external` non-`view`/non-`pure` methods. A strong invariant assumes the invariant before any unresolved external calls, havoc the state and assert the invariant after the call. There is no temporal reasoning beyond this.
+- A `preserved` block is a code block, that notably allows to inject extra `require`s into the inductive step for one method. Those assumptions are *not* checked by default, but they could be checked in other rules/invariants. Unsound `preserved` is a common source of fake invariant proofs.
+- `requireInvariant J` assumes another invariant `J` at the start of the rule/invariant. It's sound only because `J` was itself proven by the same induction scheme.
 - A *parametric rule* (one with a `method f` parameter) is expanded into one sub-rule per method in scope. A `filtered { f -> ... }` clause drops methods from that expansion; filtered-out methods are simply not checked.
 
 #### Reverts and path pruning
@@ -28,7 +28,7 @@ The primer below describes how CVL actually behaves — read it before reasoning
 - `calldataarg args` is an opaque bundle of arguments. You cannot inspect or constrain its fields; it only exists to call a `method f` parametrically.
 - `f(e, args)` targets `currentContract` by default. Call other contracts via `other.f(e, args)` where `other` is declared with `using OtherContract as other;`.
 - Addresses drawn nondeterministically may coincide with known contracts unless explicitly constrained (`require addr != currentContract`, etc.). The Prover does *not* assume addresses are distinct.
-- `envfree` tells the Prover a method doesn't read `env`; such calls omit the `env` argument. The Prover statically checks this, but a declaration mismatch with the actual Solidity is still possible and silently unsound.
+- `envfree` tells the Prover a method doesn't read `env`; such calls omit the `env` argument. The Prover statically checks this, but a declaration mismatch with the actual Solidity is still possible and fails silently.
 
 #### Ghosts, hooks, havoc
 
@@ -43,9 +43,9 @@ The primer below describes how CVL actually behaves — read it before reasoning
 #### Summaries and dispatch
 
 - The `methods { ... }` block declares how external calls are resolved. Exact-signature entries beat wildcard entries (`function _.foo() ...`).
-- `AUTO` (the default for unresolved calls): view-like methods are summarized as arbitrary returns, state-changing calls havoc.
+- `AUTO` (the default for unresolved calls): view/pure methods are summarized as NONDET, other external calls are summarized as HAVOC_ECF (see below).
 - `DISPATCHER(true)` / `DISPATCHER(false)`: on an unresolved interface call, the Prover considers every known contract with that selector. `true` is *optimistic* (only known impls are considered, assumed to be the full set); `false` is pessimistic (assumes an unknown impl could also exist). DISPATCHER does not apply to library calls.
-- `HAVOC_ALL` erases all contract state plus the return value — the maximally conservative summary. `HAVOC_ECF` (externally-controlled footprint) only havocs state that non-reentrant external callees could touch; it's sounder-looking but relies on the ECF assumption, which fails if the callee is actually the same contract.
+- `HAVOC_ALL` erases all contract state plus the return value — the maximally conservative summary. `HAVOC_ECF` (externally-controlled footprint) only havocs state that non-reentrant external callees could touch.
 - `NONDET` returns an unconstrained value and does not touch storage. For stateful methods it is unsound unless combined with a havoc.
 - A CVL function summary `=> cvlFn(args)` substitutes the CVL function body for the call. The Prover never looks at the Solidity body, so the summary must match the contract's semantics or proofs are unsound.
 - `DELETE` (e.g. `HAVOC_ALL DELETE`) additionally removes the callee's code from the scene, preventing path explosion from modeling its body.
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -1,5 +1,74 @@
...
+
+This repo uses the Certora Prover (CVL) to formally verify Solidity contracts.
+The primer below describes how CVL actually behaves — read it before reasoning about `.spec` files, because CVL semantics differ in subtle ways from Solidity and from what an LLM might assume by default.
+
+#### Rules, invariants, `satisfy`
+
+- A `rule` passes iff every `assert` holds on every execution path that satisfies all preceding `require`s. A counterexample is a concrete trace (env, args, method choice) that satisfies the `require`s and falsifies an `assert`.
+- A rule can be *vacuous*: if the `require`s exclude every path, it passes trivially. Vacuity is silent unless a sanity rule is run.
+- `satisfy p` is the dual of `assert`: the Prover must find at least one feasible path where `p` holds. A rule with only `satisfy` and no `assert` therefore does not check any universal property.
+- An `invariant I` is proven by induction over methods: base case checks `I` after the constructor; inductive step assumes `I`, runs an arbitrary method `f`, then asserts `I` again. There is no temporal reasoning beyond this.
```
adhusson: missing strong/weak invariant distinction

adhusson: and mention that it only checks non-view methods

AGENTS.md
```diff
@@ -1,5 +1,74 @@
...
+- `f(e, args)` targets `currentContract` by default. Call other contracts via `other.f(e, args)` where `other` is declared with `using OtherContract as other;`.
+- Addresses drawn nondeterministically may coincide with known contracts unless explicitly constrained (`require addr != currentContract`, etc.). The Prover does *not* assume addresses are distinct.
+- `envfree` tells the Prover a method doesn't read `env`; such calls omit the `env` argument. The Prover statically checks this, but a declaration mismatch with the actual Solidity is still possible and silently unsound.
+
+#### Ghosts, hooks, havoc
+
+- A ghost is an SMT variable (possibly a function `uint → uint`, etc.), not contract state. It exists only in the spec and can be updated by hooks or CVL assignments.
+- On an *unresolved* external call, the Prover havocs all non-persistent ghosts (assumes they take any value consistent with their axioms). `persistent ghost` declarations survive havoc.
+- Hooks fire on EVM-level events: `Sload`, `Sstore`, `CALL`, `REVERT`, etc. They match by storage slot / selector / opcode. Signature or layout drift silently disables a hook — it does not error.
```
adhusson: Should note it's not applied to CVL functions and not recursive

AGENTS.md
```diff
@@ -1,5 +1,74 @@
...
+- Direct field access on storage (`currentContract.x.y`) works for primitive/static-layout fields but not for dynamic arrays, `bytes`, or `string`.
+
+#### Quantifiers
+
+- `forall type v. e` / `exists type v. e` are first-class in CVL. The solver reasons about them via standard SMT quantifier handling; nested quantifiers and quantifiers over large domains drive solver blow-up.
+- Quantified expressions cannot call contract functions — they can only reference ghosts, storage, and CVL-pure expressions.
+
+#### Multi-contract setup
+
+- `currentContract` is the main verified contract (set in the `.conf`). Other contracts are linked by `using Foo as foo;`. Parametric rules iterate over methods of `currentContract` only unless explicitly told otherwise.
```
adhusson: Isn't it all contracts in the scene unless `parametric_contracts` is used in the configuration?

AGENTS.md
QGarchery, line 1: not clear how useful this will be. I would be more inclined to link the docs or a copy of it. 

Another option is to add progressively stuff to it when we feel that it misses the point. There is a lot of friction to do a commit/PR just for that, so we could have a doc somewhere with all our remarks

AGENTS.md
```diff
@@ -1,5 +1,74 @@
...
 
+### CVL Context
+
+This repo uses the Certora Prover (CVL) to formally verify Solidity contracts.
+The primer below describes how CVL actually behaves — read it before reasoning about `.spec` files, because CVL semantics differ in subtle ways from Solidity and from what an LLM might assume by default.
+
+#### Rules, invariants, `satisfy`
+
+- A `rule` passes iff every `assert` holds on every execution path that satisfies all preceding `require`s. A counterexample is a concrete trace (env, args, method choice) that satisfies the `require`s and falsifies an `assert`.
+- A rule can be *vacuous*: if the `require`s exclude every path, it passes trivially. Vacuity is silent unless a sanity rule is run.
```
QGarchery:
```suggestion
- A rule can be *vacuous*: if the `require`s exclude every path, it passes trivially. Vacuity is silent unless a sanity rule is run. Basic sanity is run by default.
```

AGENTS.md
```diff
@@ -1,5 +1,74 @@
...
+- `satisfy p` is the dual of `assert`: the Prover must find at least one feasible path where `p` holds. A rule with only `satisfy` and no `assert` therefore does not check any universal property.
+- CVL has both *weak* and *strong* invariants. A weak invariant is proven by induction over methods: base case checks it after the constructor; inductive step assumes it, runs an arbitrary method `f`, then asserts it again. By default invariants are checked only for `public`/`external` non-`view`/non-`pure` methods. A strong invariant adds checks around unresolved external calls that may modify the current contract's state. There is no temporal reasoning beyond this.
```
QGarchery:
```suggestion
- CVL has both *weak* and *strong* invariants. A weak invariant is proven by induction over methods: base case checks it after the constructor; inductive step assumes it, runs an arbitrary method `f`, then asserts it again. By default invariants are checked only for `public`/`external` non-`view`/non-`pure` methods. A strong invariant assumes the invariant before any unresolved external calls, havoc the state and assert the invariant after the call. There is no temporal reasoning beyond this.
```

AGENTS.md
```diff
@@ -1,5 +1,74 @@
...
+- A `preserved` block injects extra `require`s into the inductive step for one method. Those assumptions are *not* verified elsewhere — the Prover trusts them. Unsound `preserved` is the most common source of fake invariant proofs.
```
QGarchery:
```suggestion
- A `preserved` block is a code block, that notably allows to inject extra `require`s into the inductive step for one method. Those assumptions are *not* checked by default, but they could be checked in other rules/invariants. Unsound `preserved` is a common source of fake invariant proofs.
```

AGENTS.md
```diff
@@ -1,5 +1,74 @@
...
+- `requireInvariant J` assumes another invariant `J` at that point. It's sound only because `J` was itself proven by the same induction scheme.
```
QGarchery:
```suggestion
- `requireInvariant J` assumes another invariant `J` at the start of the rule/invariant. It's sound only because `J` was itself proven by the same induction scheme.
```

AGENTS.md
```diff
@@ -1,5 +1,74 @@
...
+- Without `@withrevert`, `lastReverted` after a call is always `false` (since the revert paths were pruned). Testing `lastReverted` after a plain call is meaningless.
+- Non-persistent ghosts are rolled back on revert; persistent ghosts are not.
+
+#### Methods, env, calldataarg
+
+- `env e` captures block number, timestamp, `msg.sender`, `msg.value`, etc. The Prover quantifies over all `env`s consistent with active `require`s.
+- `calldataarg args` is an opaque bundle of arguments. You cannot inspect or constrain its fields; it only exists to call a `method f` parametrically.
+- `f(e, args)` targets `currentContract` by default. Call other contracts via `other.f(e, args)` where `other` is declared with `using OtherContract as other;`.
+- Addresses drawn nondeterministically may coincide with known contracts unless explicitly constrained (`require addr != currentContract`, etc.). The Prover does *not* assume addresses are distinct.
+- `envfree` tells the Prover a method doesn't read `env`; such calls omit the `env` argument. The Prover statically checks this, but a declaration mismatch with the actual Solidity is still possible and silently unsound.
```
QGarchery:
```suggestion
- `envfree` tells the Prover a method doesn't read `env`; such calls omit the `env` argument. The Prover statically checks this, but a declaration mismatch with the actual Solidity is still possible and fails silently.
```
it's not unsound because then you just cannot use this function without specifying an environment

AGENTS.md
```diff
@@ -1,5 +1,74 @@
...
+- Inside an `Sstore` hook, the bound names conventionally written as old/new values refer to the pre-write and post-write values at that slot.
+- A two-state ghost function can be referenced as `g@old` / `g@new` inside `havoc g assuming ...`, letting you specify how the ghost changes across a havoc (e.g. `havoc g assuming g@new(x) == g@old(x) + 1`).
+- `axiom P` constrains the ghost in every state the Prover considers — adding an unsatisfiable axiom makes every rule vacuously pass. `init_state axiom P` only constrains the ghost in the base case of invariant induction, which is almost always what you want for "starts at zero"-style facts.
+- Ghost axioms may refer only to the ghost itself and quantified variables — not to Solidity or CVL functions.
+
+#### Summaries and dispatch
+
+- The `methods { ... }` block declares how external calls are resolved. Exact-signature entries beat wildcard entries (`function _.foo() ...`).
+- `AUTO` (the default for unresolved calls): view-like methods are summarized as arbitrary returns, state-changing calls havoc.
```
QGarchery:
```suggestion
- `AUTO` (the default for unresolved calls): view/pure methods are summarized as NONDET, other external calls are summarized as HAVOC_ECF (see below).
```

AGENTS.md
```diff
@@ -1,5 +1,74 @@
...
+- `DISPATCHER(true)` / `DISPATCHER(false)`: on an unresolved interface call, the Prover considers every known contract with that selector. `true` is *optimistic* (only known impls are considered, assumed to be the full set); `false` is pessimistic (assumes an unknown impl could also exist). DISPATCHER does not apply to library calls.
+- `HAVOC_ALL` erases all contract state plus the return value — the maximally conservative summary. `HAVOC_ECF` (externally-controlled footprint) only havocs state that non-reentrant external callees could touch; it's sounder-looking but relies on the ECF assumption, which fails if the callee is actually the same contract.
```
QGarchery:
```suggestion
- `HAVOC_ALL` erases all contract state plus the return value — the maximally conservative summary. `HAVOC_ECF` (externally-controlled footprint) only havocs state that non-reentrant external callees could touch.
```

AGENTS.md
```diff
@@ -1,5 +1,74 @@
...
+- `NONDET` returns an unconstrained value and does not touch storage. For stateful methods it is unsound unless combined with a havoc.
+- A CVL function summary `=> cvlFn(args)` substitutes the CVL function body for the call. The Prover never looks at the Solidity body, so the summary must match the contract's semantics or proofs are unsound.
+- `DELETE` (e.g. `HAVOC_ALL DELETE`) additionally removes the callee's code from the scene, preventing path explosion from modeling its body.
+
+#### Types and casts
+
+- `mathint` is unbounded signed integer; arithmetic never overflows. `uint256` / `int256` wrap modulo 2^n. Mixing them requires an explicit cast.
+- `to_mathint(x)` widens a machine int to `mathint` (always safe). `assert_uint256(m)` converts back and fails the current rule if `m` doesn't fit. `require_uint256(m)` converts back and *assumes* it fits (which can introduce vacuity if it never does).
```
QGarchery:
```suggestion
- `to_mathint(x)` widens a machine int to `mathint` (always safe). `assert_uint256(m)` converts back and fails the current rule if `m` doesn't fit. `require_uint256(m)` converts back and *assumes* it fits (which can introduce vacuity if it never does). Arithmetic operations in CVL automatically convert the result to mathint, so in most cases `to_mathint` is not needed and redundant.
```

QGarchery (approved):

AGENTS.md
```diff
@@ -1,5 +1,74 @@
...
+- Inside an `Sstore` hook, the bound names conventionally written as old/new values refer to the pre-write and post-write values at that slot.
+- A two-state ghost function can be referenced as `g@old` / `g@new` inside `havoc g assuming ...`, letting you specify how the ghost changes across a havoc (e.g. `havoc g assuming g@new(x) == g@old(x) + 1`).
+- `axiom P` constrains the ghost in every state the Prover considers — adding an unsatisfiable axiom makes every rule vacuously pass. `init_state axiom P` only constrains the ghost in the base case of invariant induction, which is almost always what you want for "starts at zero"-style facts.
+- Ghost axioms may refer only to the ghost itself and quantified variables — not to Solidity or CVL functions.
+
+#### Summaries and dispatch
+
+- The `methods { ... }` block declares how external calls are resolved. Exact-signature entries beat wildcard entries (`function _.foo() ...`).
+- `AUTO` (the default for unresolved calls): view/pure methods are summarized as NONDET, other external calls are summarized as HAVOC_ECF (see below).
```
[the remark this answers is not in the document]
MathisGD: it's exactly what's written

AGENTS.md
```diff
@@ -1,5 +1,74 @@
...
+- Without `@withrevert`, `lastReverted` after a call is always `false` (since the revert paths were pruned). Testing `lastReverted` after a plain call is meaningless.
+- Non-persistent ghosts are rolled back on revert; persistent ghosts are not.
+
+#### Methods, env, calldataarg
+
+- `env e` captures block number, timestamp, `msg.sender`, `msg.value`, etc. The Prover quantifies over all `env`s consistent with active `require`s.
+- `calldataarg args` is an opaque bundle of arguments. You cannot inspect or constrain its fields; it only exists to call a `method f` parametrically.
+- `f(e, args)` targets `currentContract` by default. Call other contracts via `other.f(e, args)` where `other` is declared with `using OtherContract as other;`.
+- Addresses drawn nondeterministically may coincide with known contracts unless explicitly constrained (`require addr != currentContract`, etc.). The Prover does *not* assume addresses are distinct.
+- `envfree` tells the Prover a method doesn't read `env`; such calls omit the `env` argument. The Prover statically checks this, but a declaration mismatch with the actual Solidity is still possible and fails silently.
```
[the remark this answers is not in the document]
MathisGD:
```suggestion
- `envfree` tells the Prover a method doesn't read `env`; such calls omit the `env` argument. The Prover statically checks this.
```

AGENTS.md
MathisGD: agree, let's try this and iterate on our method later

AGENTS.md
MathisGD: actually it doesn't fail silently right

adhusson (approved):

AGENTS.md
QGarchery: right, for envfree they are checked to be exist (for catch all methods it's not checked and there is only a warning)

PR #703, opened 2026-04-15
Small tweaks

QGarchery:
Screenshot 2026-04-15 at 20 40 48

### DIFF
--- a/.github/workflows/forge.yml
+++ b/.github/workflows/forge.yml
@@ -22,7 +22,7 @@ jobs:
           version: nightly-1e342ef7d296a674ea248e95ed12c152f7e00ee6
 
       - name: Check formatting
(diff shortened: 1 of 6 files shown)


.github/workflows/forge.yml
```diff
@@ -22,7 +22,7 @@ jobs:
           version: nightly-1e342ef7d296a674ea248e95ed12c152f7e00ee6
 
       - name: Check formatting
-        run: forge fmt --check
+        run: forge fmt src test script certora --check
```
QGarchery, line 25:
```suggestion
        run: forge fmt src test certora --check
```
we could do that, but I kept it because `script` is included by default

MathisGD (approved):

peyha (approved):

PR #704, opened 2026-04-15
separate callbacks

adhusson:
Contracts that want to implement the interface will need to declare the unused functions. Or not actually use this interface but then why have the interface. 

For instance the vault v2 adapter will have buy and sell callbacks but not liquidate and repay so it does

```solidity
    function onLiquidate(bytes32, Obligation memory, uint256, uint256, uint256, address, bytes memory) external pure {
        revert();
    }

    function onRepay(bytes32, Obligation memory, uint256, address, bytes memory) external pure {
        revert();
    }
```

### DIFF
--- a/src/interfaces/ICallbacks.sol
+++ b/src/interfaces/ICallbacks.sol
@@ -4,7 +4,7 @@ pragma solidity >=0.5.0;
 
 import {Obligation} from "./IMidnight.sol";
 
-interface ICallbacks {
+interface ITakeCallback {
     function onBuy(
         bytes32 id,
         Obligation memory obligation,
@@ -21,6 +21,9 @@ interface ICallbacks {
         uint256 units,
         bytes memory data
     ) external returns (bytes32);
+}
+
+interface ILiquidateCallback {
     function onLiquidate(
         bytes32 id,
         Obligation memory obligation,
@@ -30,6 +33,9 @@ interface ICallbacks {
         address borrower,
         bytes memory data
     ) external;
+}
+
+interface IRepayCallback {
     function onRepay(
         bytes32 obligationId,
         Obligation memory obligation,
--- a/src/interfaces/ICallbacks.sol
+++ b/src/interfaces/ICallbacks.sol
@@ -4,46 +4,23 @@ pragma solidity >=0.5.0;
 
 import {Obligation} from "./IMidnight.sol";
 
-interface ITakeCallback {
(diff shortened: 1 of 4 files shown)


src/interfaces/ICallbacks.sol
```diff
@@ -4,7 +4,7 @@ pragma solidity >=0.5.0;
 
 import {Obligation} from "./IMidnight.sol";
 
-interface ICallbacks {
+interface ITakeCallback {
```
adhusson: Went for a single interface for both take callbacks but not sure, could also split even more.

MathisGD: I would go for separated

src/interfaces/ICallbacks.sol
```diff
@@ -4,7 +4,7 @@ pragma solidity >=0.5.0;
 
 import {Obligation} from "./IMidnight.sol";
```
MathisGD, line 6: forge-fmt: disable-start? ^^

MathisGD (approved):

src/interfaces/ICallbacks.sol
peyha: as callback contracts should reasonably be sided (e.g., leverage/supply collateral on `onSell`, fetch liquidity on `onBuy`) I would also go for separated otherwise these callbacks will also implement a revert on the other side which is not nice

peyha (approved):

src/interfaces/ICallbacks.sol
MathisGD: I'm on it

MathisGD (approved):

test/OtherFunctionsTest.sol
```diff
@@ -654,10 +654,10 @@ contract OtherFunctionsTest is BaseTest {
 
     function testMidnightRevertsOnCallbacks(address msgSender, bytes calldata data) public {
         bytes4[4] memory selectors = [
-            ICallbacks.onBuy.selector,
-            ICallbacks.onSell.selector,
-            ICallbacks.onLiquidate.selector,
-            ICallbacks.onRepay.selector
+            IBuyCallback.onBuy.selector,
```
peyha, line 663: not sure if intended but the flashloan callback is missing (but its not critical)

peyha (approved):

QGarchery (approved):

PR #713, opened 2026-04-20
bundler: supply/withdraw collateral batch

MathisGD:
- withdraw collat on buy, supply collat on sell
- we could approve max on the first interaction with a token, but the code is more complex
- we could factorize the safe approve in a lib (but not in SafeTransferLib, because it's not used in midnight)

https://morpholabs.slack.com/archives/C08A40KKN9W/p1776693468452509

### DIFF
--- a/src/periphery/TakeBundler.sol
+++ b/src/periphery/TakeBundler.sol
@@ -238,6 +238,7 @@ contract TakeBundler is ITakeBundler {
         require(totalUnits <= maxUnits, UnitsAboveMax());
     }
 
+    /// @dev USDT won't break because the allowance is reset to 0 after supplyCollateral.
     function _safeApprove(address token, address spender, uint256 value) internal {
         (bool success, bytes memory returndata) = token.call(abi.encodeCall(IERC20.approve, (spender, value)));
         if (!success) {
--- a/src/periphery/TakeBundler.sol
+++ b/src/periphery/TakeBundler.sol
@@ -2,9 +2,11 @@
 // Copyright (c) 2025 Morpho Association
 pragma solidity 0.8.34;
 
-import {IMidnight} from "../interfaces/IMidnight.sol";
-import {ITakeBundler, Take} from "./interfaces/ITakeBundler.sol";
+import {IMidnight, Obligation} from "../interfaces/IMidnight.sol";
+import {IERC20} from "../interfaces/IERC20.sol";
+import {ITakeBundler, Take, CollateralTransfer} from "./interfaces/ITakeBundler.sol";
 import {UtilsLib} from "../libraries/UtilsLib.sol";
+import {SafeTransferLib} from "../libraries/SafeTransferLib.sol";
 import {TakeAmountsLib} from "./TakeAmountsLib.sol";
 
 contract TakeBundler is ITakeBundler {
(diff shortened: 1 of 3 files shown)


src/periphery/TakeBundler.sol
```diff
@@ -60,10 +76,22 @@ contract TakeBundler is ITakeBundler {
...
-        uint256 maxSellerAssets
+        uint256 maxSellerAssets,
+        CollateralTransfer[] calldata collateralSupplies
     ) external {
         require(taker == msg.sender || IMidnight(midnight).isAuthorized(taker, msg.sender), Unauthorized());
 
+        Obligation memory obligation = takes[0].offer.obligation;
+        for (uint256 i; i < collateralSupplies.length; i++) {
+            address token = obligation.collateralParams[collateralSupplies[i].collateralIndex].token;
+            SafeTransferLib.safeTransferFrom(token, taker, address(this), collateralSupplies[i].assets);
```
peyha: can add a dev comment that the taker has to authorize the bundler for the collateral tokens

src/periphery/TakeBundler.sol
```diff
@@ -60,10 +76,22 @@ contract TakeBundler is ITakeBundler {
...
+            _safeApprove(token, midnight, collateralSupplies[i].assets);
```
peyha, line 86: couldn't we factorize this by approving once whenever a colalteral token is processed for the first time ?

MathisGD: yes, but weirdly it's more code/logic

peyha: indeed and the gas save is negligible because the approval is reset to zero during the transfer so not worth it

src/periphery/TakeBundler.sol
```diff
@@ -184,4 +241,14 @@ contract TakeBundler is ITakeBundler {
         require(totalUnits >= minUnits, UnitsBelowMin());
         require(totalUnits <= maxUnits, UnitsAboveMax());
     }
+
+    function _safeApprove(address token, address spender, uint256 value) internal {
```
peyha, line 242: could arguably add this to `SafeTransferLib `

MathisGD: didn't want to, because it's not used in Midnight

test/BundlerTest.sol
```diff
@@ -260,11 +289,238 @@ contract BundlerTest is BaseTest {
...
+        vm.prank(borrower);
+        takeBundler.sellSellerAssetsTarget(
+            address(midnight), targetSellerAssets, borrower, borrower, takes, 0, type(uint256).max, supplies
+        );
+
+        assertEq(midnight.collateral(id, borrower, 0), supplyAmount, "collateral supplied");
+        assertEq(loanToken.balanceOf(borrower), targetSellerAssets, "borrower received seller assets");
+    }
+
+    function testSellUnitsTargetWithMultipleCollateralSupplies() public {
```
peyha: why do you only test `sellUnitsTarget` with multiple collats ? I'd say we either do all or nothing

MathisGD: refactored the tests

peyha (approved):

src/periphery/TakeBundler.sol
```diff
@@ -180,4 +237,14 @@ contract TakeBundler is ITakeBundler {
...
```
QGarchery, line 242: In bundler3 we settled on doing `forceApprove` which retries (by approving 0 and approving again) when the first approve failed. The reasoning was that there were potentially a lot of tokens that would fail otherwise. But this is because in bundler3 you could leave partial approvals, whereas here it does not seem possible (because you always do the transfer for the full amount right after)

QGarchery: Other discrepancy is that it doesn't check that the token has some code, but it shouldn't be an issue as well: the bundle should only contain valid tokens

QGarchery (approved):

src/periphery/TakeBundler.sol
MathisGD: added a comment for the first one

Issue #338, opened 2026-02-04
wallet decodable signature


MathisGD:
thread of @qdrn : https://morpholabs.slack.com/archives/C02N7CZ088N/p1765210304326109

PR #714, opened 2026-04-20
Wallet decodable signatures

QGarchery:
Fixes:
- #338 

Notes:
- works for tress that contain a number of leaves that is a power of 2. To that end the dummy offers should be used to pad the tree. The default empty offer works just fine, because its expiry is the timestamp 0, so any take is guaranteed to revert
- this PR is not gas optimized, to show the concept first using a simpler implementation
- 

Todo:
- [x] draft implem
- [ ] fix errors (stack too deep by forge and by certora)
- [ ] test with a different number of leaves, to show the padding
- [ ] test that indeed it makes possible to have "clear signing"
- [ ] optimize


QGarchery:
doing it in #715

PR #715, opened 2026-04-20
Wallet decodable signatures

### LINKED ISSUE #338: wallet decodable signature

QGarchery:
Fixes:
- #338 

Notes:
- works for and requires trees that contain a number of leaves that is a power of 2. 
  - To that end the dummy offers should be used to pad the tree. The default empty offer works just fine, because its expiry is the timestamp 0, so any take is guaranteed to revert
  - this is a downside in itself, since in the general case it requires to pad the offer tree with dummy offers. This represents +33% size in average (assuming random distribution, and all offers are the same size), but in practice it should be compressed very well
- looking at the first commit only can help to get a feel of what this PR is doing, without some unnecessary details

Todo:
- [x] draft implem
- [x] fix compilation errors (stack too deep by forge and by certora, bytecode size limit reached)
- [x] test with a different number of leaves, padding with empty offers
- [x] test that indeed it makes possible to have "clear signing", see `frontend/README.md`
- [x]  move `frontent/` to the `test/` folder
- [x] optimize (see [this](https://github.com/morpho-org/midnight/pull/715#discussion_r3114753605) and [this](https://github.com/morpho-org/midnight/pull/715#discussion_r3114672177))


src/libraries/UtilsLib.sol
```diff
@@ -62,6 +78,66 @@ library UtilsLib {
...
+                obligation.enterGate,
+                obligation.liquidatorGate
+            )
+        );
+    }
+
+    function hashOffer(Offer memory offer) internal pure returns (bytes32) {
+        return keccak256(
+            abi.encode(
+                abi.encode(
```
QGarchery: To fix a problem where we can't specify optimization runs in CVL, so that it ends up failing compilation with "stack too deep"

src/libraries/UtilsLib.sol
```diff
@@ -62,6 +78,66 @@ library UtilsLib {
...
+            )
+        );
+    }
+
+    function hashObligation(Obligation memory obligation) internal pure returns (bytes32) {
+        bytes32[] memory collateralParamsHashes = new bytes32[](obligation.collateralParams.length);
+        for (uint256 i = 0; i < obligation.collateralParams.length; i++) {
+            collateralParamsHashes[i] = hashCollateralParams(obligation.collateralParams[i]);
+        }
+        bytes32 collateralParamsArrayHash = keccak256(abi.encodePacked(collateralParamsHashes));
```
QGarchery: We could be more efficient here and compute `abi.encodePacked(collateralParamsHashes)` directly

src/interfaces/IMidnight.sol
```diff
@@ -63,6 +63,13 @@ struct Position {
     uint128[128] collateral;
 }
 
+/// @dev Struct used to avoid stack too deep errors in the take function.
+struct User {
+    address addr;
+    address callback;
+    bytes callbackData;
+}
```
QGarchery: If needed later we can go further and easily save 2 more stack slots with `buyerAssets`/`sellerAssets`

foundry.toml
```diff
@@ -1,7 +1,7 @@
 [profile.default]
 via_ir = true
 optimizer = true
-optimizer_runs = 300
+optimizer_runs = 100
```
QGarchery, line 4: To fix an issue where we reached the bytecode size limit. Another option is to define the following function:
```solidity
    function checkAuthorized(address onBehalf) internal view {
        require(onBehalf == msg.sender || isAuthorized[onBehalf][msg.sender], Unauthorized());
    }
```

src/libraries/UtilsLib.sol
```diff
@@ -62,6 +78,66 @@ library UtilsLib {
...
+
+        return keccak256(
+            abi.encode(
+                keccak256(OBLIGATION_TYPE),
```
QGarchery:
```suggestion
                OBLIGATION_TYPEHASH,
```
we could do that for improved gas usage

src/libraries/UtilsLib.sol
QGarchery: See https://morpholabs.slack.com/archives/C0ACSA8TNMA/p1776754812616349

src/Midnight.sol
```diff
@@ -274,12 +283,15 @@ contract Midnight is IMidnight {
...
+        require(UtilsLib.isLeaf(root, UtilsLib.hashOffer(offer), proof), InvalidProof());
         require(offer.session == session[offer.maker], InvalidSession());
         require(isAuthorized[offer.maker][offer.ratifier], RatifierUnauthorized());
         require(IRatifier(offer.ratifier).onRatify(offer, root, ratifierData) == CALLBACK_SUCCESS, RatifierFail());
 
-        (address buyer, address seller) = offer.buy ? (offer.maker, taker) : (taker, offer.maker);
+        User memory takeUser = User({addr: taker, callback: takerCallback, callbackData: takerCallbackData});
+        User memory makeUser = User({addr: offer.maker, callback: offer.callback, callbackData: offer.callbackData});
+        User memory buyer = offer.buy ? makeUser : takeUser;
+        User memory seller = offer.buy ? takeUser : makeUser;
```
QGarchery: The following changes in this file are only here to fix a stack too deep error

frontend/sign-root.ts
```diff
@@ -0,0 +1,176 @@
...
+      { name: "maxLif", type: "uint256" },
+      { name: "oracle", type: "address" },
+    ],
+  };
+}
+
+function defaultOffer(maker: string, number: string) {
+  return {
+    obligation: {
+      loanToken: "0x" + number.repeat(64),
```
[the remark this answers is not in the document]
QGarchery: this is only to show clear signing

frontend/sign-root.ts
```diff
@@ -0,0 +1,176 @@
...
+    maxSellerAssets: "0",
+    maxBuyerAssets: "0",
+  };
+}
+
+// WARNING: The root should be built by sorting the nodes in ascending order of their hash.
+function buildRoot(maker: string) {
+  return [
+    [defaultOffer(maker, "1"), defaultOffer(maker, "2")],
+    [defaultOffer(maker, "3"), defaultOffer(maker, "4")],
```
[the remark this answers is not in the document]
QGarchery: Yes this is a known limitation, documented at the top of the function. Will not fix as it will make things more complicated for no added benefit

src/interfaces/IMidnight.sol
MathisGD, line 22: Wdyt? I think that we discussed this in the past already. With this PR, it makes more sense because it really simplifies `hashObligation`.
```suggestion
    bytes32 id;
```

src/Midnight.sol
```diff
@@ -343,11 +349,6 @@ contract Midnight is IMidnight {
             SellerGatedFromIncreasingDebt()
         );
 
-        address buyerCallback = offer.buy ? offer.callback : takerCallback;
-        address sellerCallback = offer.buy ? takerCallback : offer.callback;
-        address payer = buyerCallback != address(0) ? buyerCallback : (offer.buy ? buyer : msg.sender);
-        address receiver = offer.buy ? receiverIfTakerIsSeller : offer.receiverIfMakerIsSeller;
```
MathisGD, line 349: we should move them back at the same place

src/Midnight.sol
```diff
@@ -274,21 +275,26 @@ contract Midnight is IMidnight {
...
+        require(UtilsLib.isLeaf(root, UtilsLib.hashOffer(offer), proof), InvalidProof());
         require(offer.session == session[offer.maker], InvalidSession());
         require(isAuthorized[offer.maker][offer.ratifier], RatifierUnauthorized());
         require(IRatifier(offer.ratifier).onRatify(offer, root, ratifierData) == CALLBACK_SUCCESS, RatifierFail());
 
         (address buyer, address seller) = offer.buy ? (offer.maker, taker) : (taker, offer.maker);
 
-        uint256 offerPrice = TickLib.tickToPrice(offer.tick);
         uint256 timeToMaturity = UtilsLib.zeroFloorSub(offer.obligation.maturity, block.timestamp);
+        uint256 offerPrice = TickLib.tickToPrice(offer.tick);
```
```suggestion
        uint256 offerPrice = TickLib.tickToPrice(offer.tick);
        uint256 timeToMaturity = UtilsLib.zeroFloorSub(offer.obligation.maturity, block.timestamp);
```

src/Midnight.sol
```diff
@@ -117,6 +117,7 @@ import {EventsLib} from "./libraries/EventsLib.sol";
 /// @dev No-ops are allowed.
 /// @dev NatSpec comments are included only when they bring clarity.
 ///
+
```
```suggestion
```

src/Midnight.sol
QGarchery: Why do you find it better ?

src/Midnight.sol
QGarchery: I'm confused, I thought that this was precisely what was changed in #719

src/interfaces/IMidnight.sol
QGarchery: I don't think that it will help, because `hashObligation` will be /has to be different from `toId`. The first one is defined by EIP712 (so it cannot be changed), the second one should contain the chain id and Midnight contract and other constraints to make `toObligation` work

src/libraries/UtilsLib.sol
```diff
@@ -62,6 +78,66 @@ library UtilsLib {
         }
     }
 
+    function hashCollateralParams(CollateralParams memory collateralParams) internal pure returns (bytes32) {
+        return keccak256(
+            abi.encode(
+                keccak256(COLLATERAL_PARAMS_TYPE),
```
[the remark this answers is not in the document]
QGarchery: duplicate of this: https://github.com/morpho-org/midnight/pull/715#discussion_r3114753605

src/libraries/UtilsLib.sol
```diff
@@ -42,6 +45,19 @@ library UtilsLib {
         return (x * y + (d - 1)) / d;
     }
 
+    function rootTypeHash(uint256 height) internal pure returns (bytes32) {
+        bytes memory brackets = new bytes(3 * height);
+        for (uint256 i = 0; i < height; i++) {
+            brackets[3 * i] = "[";
+            brackets[3 * i + 1] = "2";
+            brackets[3 * i + 2] = "]";
```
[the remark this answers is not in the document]
QGarchery: True, but it's only one keccak, and I find it very much worth the simplicity that it provides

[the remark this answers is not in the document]
QGarchery: @MathisGD suggested hardcoding the height

src/libraries/UtilsLib.sol
QGarchery: We will keep it in the end, because it fixes stack too deep nicely without altering the Midnight contract

src/libraries/UtilsLib.sol
QGarchery: seems very much worth it IMO

test/frontend/FrontendSignatureTest.sol
```diff
@@ -0,0 +1,76 @@
...
+        assertTrue(UtilsLib.isLeaf(_root, h2, proof2));
+
+        bytes32[] memory proof3 = new bytes32[](2);
+        proof3[0] = h2;
+        proof3[1] = left;
+        assertTrue(UtilsLib.isLeaf(_root, h3, proof3));
+
+        bytes memory ratifierData = abi.encode(Signature({v: SIG_V, r: SIG_R, s: SIG_S}), HEIGHT);
+        // Trick to pass the maker to the ratifier, without having the offers depend on the maker.
+        offers[0].maker = ACCOUNT;
```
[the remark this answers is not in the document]
QGarchery: The offer is not used for hashing in `Ecrecover.onRatify` it's only used to get the maker. Here we don't care that someone can sign an offer that he wasn't supposed to be able to sign. What we care about is that it is possible to see the data in the wallet, and that this data can be verified onchain

src/interfaces/IMidnight.sol
QGarchery: Actually the idea is to replace Obligation with bytes32, not to add a bytes32. This would help because we wouldn't need to hash the Obligation, so no need to hash CollateralParams either. 
I don't think this is the main motivation though, but I added it to #722 to keep it in mind

src/libraries/UtilsLib.sol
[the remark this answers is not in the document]
QGarchery: Decided that it was not worth hardcoding the height, because right now the signing of a single offer is very clear. But if we had a fixed height (allowing 1024 offers for example), then it wouldn't be the case and the single offer would be lost in all of them, when displayed by the wallet

src/libraries/UtilsLib.sol
QGarchery: deciding against it in the end, to avoid introducing assembly in the code while the current code is very clear

test/frontend/README.md
```diff
@@ -0,0 +1,8 @@
+# Usage
+
+Install with `cd frontend && npm install`.
```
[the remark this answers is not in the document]
QGarchery: fix this @codex

src/libraries/UtilsLib.sol
peyha, line 1: as I think mathis suggested we could use a `HashLib` because new functions essentially only compute hashes

src/interfaces/IEcrecover.sol
```diff
@@ -18,5 +18,18 @@ struct Authorization {
 
 bytes32 constant AUTHORIZATION_TYPEHASH =
     keccak256("Authorization(address authorizer,address authorized,bool isAuthorized,uint256 nonce,uint256 deadline)");
+
 bytes32 constant EIP712_DOMAIN_TYPEHASH = keccak256("EIP712Domain(uint256 chainId,address verifyingContract)");
-bytes32 constant ROOT_TYPEHASH = keccak256("Root(bytes32 root)");
+
+bytes constant COLLATERAL_PARAMS_TYPE = "CollateralParams(address token,uint256 lltv,uint256 maxLif,address oracle)";
+bytes32 constant COLLATERAL_PARAMS_TYPEHASH = keccak256(COLLATERAL_PARAMS_TYPE);
```
peyha:
```suggestion
bytes32 constant COLLATERAL_PARAMS_TYPEHASH = 0xaf44a88eb50ebdbbebd980e5a23045c44f61ece5f80ab708a1bbe8718102e6af
```
could also hardcode this one

src/libraries/UtilsLib.sol
MathisGD: I think that's fine here

src/libraries/UtilsLib.sol
```diff
@@ -62,6 +85,66 @@ library UtilsLib {
...
+                abi.encode(
+                    OFFER_TYPEHASH,
+                    hashObligation(offer.obligation),
+                    offer.buy,
+                    offer.maker,
+                    offer.start,
+                    offer.expiry,
+                    offer.tick
+                ),
+                abi.encode(
```
peyha:
```suggestion
                    offer.tick,
```
I thought you made this because abi.encode had a parameter nb limit but it hasn't

MathisGD, line 133: what was the reason to concat to abi.encode
```suggestion
```

src/libraries/UtilsLib.sol
```diff
@@ -42,6 +52,19 @@ library UtilsLib {
...
+    function rootTypeHash(uint256 height) internal pure returns (bytes32) {
+        bytes memory brackets = new bytes(3 * height);
+        for (uint256 i = 0; i < height; i++) {
+            brackets[3 * i] = "[";
+            brackets[3 * i + 1] = "2";
+            brackets[3 * i + 2] = "]";
+        }
+        return
+            keccak256(
+                bytes.concat("Root(Offer", brackets, " root)", COLLATERAL_PARAMS_TYPE, OBLIGATION_TYPE, OFFER_TYPE)
```
MathisGD: technically it's not a root anymore
```suggestion
                bytes.concat("OfferTree(Offer", brackets, " root)", COLLATERAL_PARAMS_TYPE, OBLIGATION_TYPE, OFFER_TYPE)
```

src/interfaces/IEcrecover.sol
QGarchery: It's a constant so we don't gain anything by harcoding it except: compilation time (but that's so small to notice), can help avoid compilation errors like stack too deep which is why I did it for the other but here there is no issue of stack too deep

src/libraries/UtilsLib.sol
QGarchery: I had it as a single abi.encode at first but it was creating stack too deep on the certora side (which modified the code to add sentinels). See this : https://github.com/morpho-org/midnight/pull/715#discussion_r3114670888

QGarchery: See above : https://github.com/morpho-org/midnight/pull/715#discussion_r3138577948

src/libraries/UtilsLib.sol
MathisGD: done in #734

test/BaseTest.sol
```diff
@@ -220,7 +220,7 @@ abstract contract BaseTest is Test {
     }
 
     function ratifierData(Offer[1] memory offers, address _signer) internal view returns (bytes memory) {
-        return abi.encode(signature(root(offers), privateKey[_signer], offers[0].ratifier));
+        return abi.encode(signature(root(offers), privateKey[_signer], offers[0].ratifier, 0), uint256(0));
```
MathisGD, line 223: shouldn't it be height 1?

MathisGD: actually no, tree of height zero is ok

certora/specs/Midnight.spec
MathisGD, line 1: I would add some general comments about what's signed and what. Right now it's pretty deep (you have to check the way offers are hashed to see it). Also it's not that clear as the type is built on the fly (the [2][2] thing)

MathisGD, line 1: approving pending #734

MathisGD (approved):

certora/specs/Midnight.spec
QGarchery: added some comments

MathisGD (approved):

src/libraries/UtilsLib.sol
```diff
@@ -42,6 +52,15 @@ library UtilsLib {
         return (x * y + (d - 1)) / d;
     }
 
+    function offerTreeTypeHash(uint256 height) internal pure returns (bytes32) {
```
peyha, line 55: small gas opti could be to set a maximum height and hardcode the offertreetypehash

src/Midnight.sol
```diff
@@ -60,6 +60,11 @@ import {EventsLib} from "./libraries/EventsLib.sol";
 /// @dev The session can be shuffled by the user to cancel all current offers easily and efficiently.
 /// @dev Offers should have the current session to be valid.
 ///
+/// ROOT
+/// @dev The root should correspond to the root of the offer tree, which is a Merkle tree of offers.
+/// @dev If the offers are well-sorted when given to the wallet, the EIP-712 digest will match the root of the tree.
```
peyha: could you please specify what you mean by well-sorted

test/BaseTest.sol
```diff
@@ -230,32 +230,70 @@ abstract contract BaseTest is Test {
     // assumes the offer is the first one!
     function proof(Offer[2] memory offers) internal pure returns (bytes32[] memory) {
         bytes32[] memory _path = new bytes32[](1);
-        _path[0] = keccak256(abi.encode(offers[1]));
+        _path[0] = UtilsLib.hashOffer(offers[1]);
+        return _path;
+    }
+
+    // 4 leaves, assumes the offer is the first one
+    function proofFirstLeaf(Offer[4] memory offers) internal pure returns (bytes32[] memory) {
```
peyha, line 238: wouldn't we want tests with a deeper depth ?

src/libraries/UtilsLib.sol
MathisGD: we discussed with Quentin, and thought that wouldn't be worth it

src/Midnight.sol
```suggestion
/// @dev If the offers are well-sorted (such that for all nodes, hash(left) < hash(right)) when given to the wallet, the EIP-712 digest will match the root of the tree.
```

src/libraries/UtilsLib.sol
QGarchery: I see 2 options, but I think both are worse than the current one:
- set a fixed height, but then you have plenty of null offers even in the case where you want to sign just one offer
- set a maximum height, and hardcode every typehash for heights lower than the maximum. But then the gas opti is very much non-trivial, since you can't use solidity arrays (storage is too costly) and you can't use control flow (jumps are too costly). So you basically have to re-code arrays in code (immutable arrays), which is a pain...

test/BaseTest.sol
QGarchery: For this test I wanted to have something as simple as possible, and the following helps: the offers and their order is fixed and don't depend on the input of the user. I think you can't have that with arbitrary height.
We could use height 3 if you want, but I'm not convinced it will be very useful (we already have an array of array, which is the main thing)

src/libraries/UtilsLib.sol
peyha: I see, probably not worth it at this point then

src/libraries/UtilsLib.sol
```diff
@@ -62,6 +81,66 @@ library UtilsLib {
         }
     }
```
peyha, line 83:
```suggestion
/// dev Computes the EIP-712 hash struct of a CollateralParams object
```
not necessary but could help

test/BaseTest.sol
peyha: tried with height 3 and it worked, convincing myself that it would be fine for higher heights

peyha (approved):
lgtm, don't forge to remove frontend before merging

MathisGD (approved):

src/interfaces/IEcrecover.sol
adhusson: btw also not usable in yul if not a literal

adhusson:
LGTM, also see https://github.com/morpho-org/midnight/pull/766 on top (remove frontend, reincrease optimizer runs, check constant literals)

PR #717, opened 2026-04-21
bundler: referral fees

MathisGD:
- see [this thread](https://morpholabs.slack.com/archives/C02N7CZ088N/p1776765655376129) for the max fees decision
- nb: even with no fees, sell functions now receive the funds and then transfer it to the receiver.
- considered fee in absolute amount (you can rebuild anything, but I think that 95% will use relative amounts so it's simpler for them)
- the fees maths are (without roundings, not scaled by wad):
  - target_S = filled_S * (1 - r)
  - filled_B = target_B * (1 - r)
- with roundings:
  - target_S = filled_S - floor(filled_S * r / WAD)
  - target_B = filled_B + floor(filled_B * r / (WAD - r))

### DIFF
--- a/src/periphery/TakeBundler.sol
+++ b/src/periphery/TakeBundler.sol
@@ -28,11 +28,13 @@ contract TakeBundler is ITakeBundler {
     ) external {
         require(taker == msg.sender || IMidnight(midnight).isAuthorized(taker, msg.sender), Unauthorized());
         require(referralFeePct <= WAD, MaxReferralFeePctExceeded());
+        address loanToken = takes[0].offer.obligation.loanToken;
 
         uint256 totalFilledUnits;
         uint256 totalFilledBuyerAssets;
         for (uint256 i; i < takes.length && totalFilledUnits < targetUnits; i++) {
             require(!takes[i].offer.buy, InconsistentSide());
+            require(takes[i].offer.obligation.loanToken == loanToken, InconsistentLoanToken());
             try IMidnight(midnight)
                 .take(
                     UtilsLib.min(targetUnits - totalFilledUnits, takes[i].units),
@@ -57,9 +59,7 @@ contract TakeBundler is ITakeBundler {
         require(totalFilledBuyerAssets <= maxBuyerAssets, BuyerAssetsAboveMax());
 
         uint256 referralFeeAssets = totalFilledBuyerAssets.mulDivDown(referralFeePct, WAD);
-        if (referralFeeAssets > 0) {
-            SafeTransferLib.safeTransfer(takes[0].offer.obligation.loanToken, referralFeeRecipient, referralFeeAssets);
-        }
+        if (referralFeeAssets > 0) SafeTransferLib.safeTransfer(loanToken, referralFeeRecipient, referralFeeAssets);
     }
 
     /// @dev See buyUnitsTarget.
@@ -76,11 +76,13 @@ contract TakeBundler is ITakeBundler {
     ) external {
         require(taker == msg.sender || IMidnight(midnight).isAuthorized(taker, msg.sender), Unauthorized());
         require(referralFeePct <= WAD, MaxReferralFeePctExceeded());
+        address loanToken = takes[0].offer.obligation.loanToken;
 
         uint256 totalFilledSellerAssets;
         uint256 totalFilledUnits;
         for (uint256 i; i < takes.length && totalFilledUnits < targetUnits; i++) {
             require(takes[i].offer.buy, InconsistentSide());
+            require(takes[i].offer.obligation.loanToken == loanToken, InconsistentLoanToken());
             try IMidnight(midnight)
                 .take(
                     UtilsLib.min(targetUnits - totalFilledUnits, takes[i].units),
@@ -104,7 +106,6 @@ contract TakeBundler is ITakeBundler {
         require(totalFilledSellerAssets >= minSellerAssets, SellerAssetsBelowMin());
         require(totalFilledSellerAssets <= maxSellerAssets, SellerAssetsAboveMax());
 
-        address loanToken = takes[0].offer.obligation.loanToken;
         uint256 referralFeeAssets = totalFilledSellerAssets.mulDivDown(referralFeePct, WAD);
         if (referralFeeAssets > 0) SafeTransferLib.safeTransfer(loanToken, referralFeeRecipient, referralFeeAssets);
         SafeTransferLib.safeTransfer(loanToken, receiver, totalFilledSellerAssets - referralFeeAssets);
@@ -123,6 +124,7 @@ contract TakeBundler is ITakeBundler {
     ) external {
         require(taker == msg.sender || IMidnight(midnight).isAuthorized(taker, msg.sender), Unauthorized());
         require(referralFeePct <= WAD, MaxReferralFeePctExceeded());
+        address loanToken = takes[0].offer.obligation.loanToken;
         // touchObligation to have the correct trading fees.
         bytes32 id = IMidnight(midnight).touchObligation(takes[0].offer.obligation);
 
@@ -130,6 +132,7 @@ contract TakeBundler is ITakeBundler {
         uint256 totalFilledUnits;
         for (uint256 i; i < takes.length && totalFilledBuyerAssets < targetBuyerAssets; i++) {
             require(!takes[i].offer.buy, InconsistentSide());
+            require(takes[i].offer.obligation.loanToken == loanToken, InconsistentLoanToken());
             try IMidnight(midnight)
                 .take(
                     UtilsLib.min(
@@ -159,9 +162,7 @@ contract TakeBundler is ITakeBundler {
         require(totalFilledUnits <= maxUnits, UnitsAboveMax());
 
         uint256 referralFeeAssets = totalFilledBuyerAssets.mulDivDown(referralFeePct, WAD);
-        if (referralFeeAssets > 0) {
-            SafeTransferLib.safeTransfer(takes[0].offer.obligation.loanToken, referralFeeRecipient, referralFeeAssets);
-        }
+        if (referralFeeAssets > 0) SafeTransferLib.safeTransfer(loanToken, referralFeeRecipient, referralFeeAssets);
     }
 
     /// @dev See buyUnitsTarget.
@@ -178,6 +179,7 @@ contract TakeBundler is ITakeBundler {
     ) external {
         require(taker == msg.sender || IMidnight(midnight).isAuthorized(taker, msg.sender), Unauthorized());
         require(referralFeePct <= WAD, MaxReferralFeePctExceeded());
+        address loanToken = takes[0].offer.obligation.loanToken;
         // touchObligation to have the correct trading fees.
         bytes32 id = IMidnight(midnight).touchObligation(takes[0].offer.obligation);
 
@@ -185,6 +187,7 @@ contract TakeBundler is ITakeBundler {
         uint256 totalFilledUnits;
         for (uint256 i; i < takes.length && totalFilledSellerAssets < targetSellerAssets; i++) {
             require(takes[i].offer.buy, InconsistentSide());
+            require(takes[i].offer.obligation.loanToken == loanToken, InconsistentLoanToken());
             try IMidnight(midnight)
                 .take(
                     UtilsLib.min(
@@ -213,7 +216,6 @@ contract TakeBundler is ITakeBundler {
         require(totalFilledUnits >= minUnits, UnitsBelowMin());
         require(totalFilledUnits <= maxUnits, UnitsAboveMax());
 
-        address loanToken = takes[0].offer.obligation.loanToken;
         uint256 referralFeeAssets = totalFilledSellerAssets.mulDivDown(referralFeePct, WAD);
         if (referralFeeAssets > 0) SafeTransferLib.safeTransfer(loanToken, referralFeeRecipient, referralFeeAssets);
         SafeTransferLib.safeTransfer(loanToken, receiver, totalFilledSellerAssets - referralFeeAssets);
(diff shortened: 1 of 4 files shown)


src/periphery/TakeBundler.sol
```diff
@@ -43,29 +48,35 @@ contract TakeBundler is ITakeBundler {
...
         }
 
         require(totalFilledUnits == targetUnits, InsufficientLiquidity());
-        require(totalBuyerAssets >= minBuyerAssets, BuyerAssetsBelowMin());
-        require(totalBuyerAssets <= maxBuyerAssets, BuyerAssetsAboveMax());
+        require(totalFilledBuyerAssets >= minBuyerAssets, BuyerAssetsBelowMin());
+        require(totalFilledBuyerAssets <= maxBuyerAssets, BuyerAssetsAboveMax());
+
+        uint256 referralFeeAssets = totalFilledBuyerAssets.mulDivDown(referralFeePct, WAD);
+        SafeTransferLib.safeTransfer(takes[0].offer.obligation.loanToken, referralFeeRecipient, referralFeeAssets);
```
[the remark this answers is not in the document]
MathisGD: right, but that's fine

src/periphery/TakeBundler.sol
```diff
@@ -83,13 +94,18 @@ contract TakeBundler is ITakeBundler {
...
         require(totalFilledUnits == targetUnits, InsufficientLiquidity());
-        require(totalSellerAssets >= minSellerAssets, SellerAssetsBelowMin());
-        require(totalSellerAssets <= maxSellerAssets, SellerAssetsAboveMax());
+        require(totalFilledSellerAssets >= minSellerAssets, SellerAssetsBelowMin());
+        require(totalFilledSellerAssets <= maxSellerAssets, SellerAssetsAboveMax());
+
+        address loanToken = takes[0].offer.obligation.loanToken;
+        uint256 referralFeeAssets = totalFilledSellerAssets.mulDivDown(referralFeePct, WAD);
+        SafeTransferLib.safeTransfer(loanToken, referralFeeRecipient, referralFeeAssets);
+        SafeTransferLib.safeTransfer(loanToken, receiver, totalFilledSellerAssets - referralFeeAssets);
```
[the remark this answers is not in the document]
MathisGD: fixing

src/periphery/TakeBundler.sol
```diff
@@ -43,38 +50,46 @@ contract TakeBundler is ITakeBundler {
...
         }
 
         require(totalFilledUnits == targetUnits, InsufficientLiquidity());
-        require(totalBuyerAssets >= minBuyerAssets, BuyerAssetsBelowMin());
-        require(totalBuyerAssets <= maxBuyerAssets, BuyerAssetsAboveMax());
+        require(totalFilledBuyerAssets >= minBuyerAssets, BuyerAssetsBelowMin());
+        require(totalFilledBuyerAssets <= maxBuyerAssets, BuyerAssetsAboveMax());
+
+        uint256 referralFeeAssets = totalFilledBuyerAssets.mulDivDown(referralFeePct, WAD);
+        if (referralFeeAssets > 0) SafeTransferLib.safeTransfer(loanToken, referralFeeRecipient, referralFeeAssets);
```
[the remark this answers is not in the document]
MathisGD: added as open question

src/periphery/TakeBundler.sol
```diff
@@ -20,14 +22,19 @@ contract TakeBundler is ITakeBundler {
...
         require(taker == msg.sender || IMidnight(midnight).isAuthorized(taker, msg.sender), Unauthorized());
+        require(referralFeePct <= WAD, MaxReferralFeePctExceeded());
+        address loanToken = takes[0].offer.obligation.loanToken;
 
         uint256 totalFilledUnits;
-        uint256 totalBuyerAssets;
+        uint256 totalFilledBuyerAssets;
         for (uint256 i; i < takes.length && totalFilledUnits < targetUnits; i++) {
             require(!takes[i].offer.buy, InconsistentSide());
+            require(takes[i].offer.obligation.loanToken == loanToken, InconsistentLoanToken());
```
QGarchery:
```suggestion
```
we already assume that the obligation id is the same across all obligations

src/periphery/interfaces/ITakeBundler.sol
```diff
@@ -16,19 +16,21 @@ interface ITakeBundler {
     /// ERRORS ///
     error BuyerAssetsAboveMax();
     error BuyerAssetsBelowMin();
+    error InconsistentLoanToken();
     error InconsistentSide();
     error InsufficientLiquidity();
     error SellerAssetsAboveMax();
     error SellerAssetsBelowMin();
+    error MaxReferralFeePctExceeded();
```
peyha:
```suggestion
    error PctOverflow()
```
or 
```suggestion
    error PctExceeded()
```
the current naming implies that there is a max referral fee

src/periphery/TakeBundler.sol
```diff
@@ -43,38 +50,46 @@ contract TakeBundler is ITakeBundler {
...
             } catch {}
         }
 
         require(totalFilledUnits == targetUnits, InsufficientLiquidity());
-        require(totalBuyerAssets >= minBuyerAssets, BuyerAssetsBelowMin());
-        require(totalBuyerAssets <= maxBuyerAssets, BuyerAssetsAboveMax());
+        require(totalFilledBuyerAssets >= minBuyerAssets, BuyerAssetsBelowMin());
+        require(totalFilledBuyerAssets <= maxBuyerAssets, BuyerAssetsAboveMax());
+
+        uint256 referralFeeAssets = totalFilledBuyerAssets.mulDivDown(referralFeePct, WAD);
```
QGarchery: The fee is not taken consistently on the different functions: an input decrease of x% does not provide the same value as an output increase of x%. Most likely not an issue for small values of x, but if x gets greater (and assuming that the front end puts the same fee value for every bundler function) the difference can be seen:
- as users will prefer one bundler function over the other 
- the front-end will not be able to charge more than 50% on this function, because of the check `referralFeePct <= WAD`


```suggestion
        uint256 referralFeeAssets = totalFilledBuyerAssets.mulDivDown(referralFeePct, WAD - referralFeePct);
```

src/periphery/TakeBundler.sol
peyha: If we check this, shouldn't we rather check that the whole obligation id is the same ? It's a bit more costly in gas but gives better guarantees

src/periphery/TakeBundler.sol
peyha, line 74: Shouldn't this be a transfer from ? 
in both cases it's worth documenting imo (if it's a transfer then the user should have transferred enough assets to the bundler, if it's transfer from then it should have approved it)

src/periphery/TakeBundler.sol
QGarchery: same as https://github.com/morpho-org/midnight/pull/717#discussion_r3117927835

src/periphery/TakeBundler.sol
```diff
@@ -129,33 +153,41 @@ contract TakeBundler is ITakeBundler {
...
         }
 
         require(totalFilledBuyerAssets == targetBuyerAssets, InsufficientLiquidity());
-        require(totalUnits >= minUnits, UnitsBelowMin());
-        require(totalUnits <= maxUnits, UnitsAboveMax());
+        require(totalFilledUnits >= minUnits, UnitsBelowMin());
+        require(totalFilledUnits <= maxUnits, UnitsAboveMax());
+
+        uint256 referralFeeAssets = totalFilledBuyerAssets.mulDivDown(referralFeePct, WAD);
+        if (referralFeeAssets > 0) SafeTransferLib.safeTransfer(loanToken, referralFeeRecipient, referralFeeAssets);
```
QGarchery, line 205: In this function `buyBuyerAssetsTarget` the user is targeting some `buyerAssets` but ends up paying `buyerAssets + referralFeeAssets` which defeats the purpose of the function. Shouldn't the `targetBuyerAssets` be considered as filled by `referralFeeAssets` before looping on all the offers ?

src/periphery/TakeBundler.sol
MathisGD: as discussed, we actually want to enable buy/sell with assets inputs cross obligations. it could be a feature. though I'll constrain for units inputs, as it makes less sense.

src/periphery/TakeBundler.sol
MathisGD: fixed now

src/periphery/TakeBundler.sol
```diff
@@ -126,16 +150,21 @@ contract TakeBundler is ITakeBundler {
...
-        bytes32 id = IMidnight(midnight).touchObligation(takes[0].offer.obligation);
+        require(referralFeePct < WAD, PctExceeded());
+        address loanToken = takes[0].offer.obligation.loanToken;
 
         uint256 totalFilledBuyerAssets;
-        uint256 totalUnits;
+        uint256 totalFilledUnits;
         for (uint256 i; i < takes.length && totalFilledBuyerAssets < targetBuyerAssets; i++) {
             require(!takes[i].offer.buy, InconsistentSide());
+            require(takes[i].offer.obligation.loanToken == loanToken, InconsistentLoanToken());
```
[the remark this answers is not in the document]
MathisGD: will fix in a PR on top

test/BundlerTest.sol
```diff
@@ -266,11 +291,248 @@ contract BundlerTest is BaseTest {
...
+        deal(address(loanToken), lender, 0);
+
+        uint256 price = TickLib.tickToPrice(MAX_TICK);
+        uint256 expectedFilledBuyerAssets = units.mulDivUp(price, WAD);
+        uint256 expectedFee = expectedFilledBuyerAssets.mulDivDown(referralFeePct, WAD - referralFeePct);
+
+        collateralize(obligation, lender, units);
+
+        Take[] memory takes = new Take[](1);
+        takes[0] = Take({
```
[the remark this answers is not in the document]
MathisGD: that's fine

peyha:
There's a small sign mistake in your formula, it should be `target_B = filled_B + floor(filled_B * r / (WAD - r))`

src/periphery/TakeBundler.sol
```diff
@@ -85,51 +107,71 @@ contract TakeBundler is ITakeBundler {
...
+        address referralFeeRecipient
     ) external {
         require(taker == msg.sender || IMidnight(midnight).isAuthorized(taker, msg.sender), Unauthorized());
-        // touchObligation to have the correct trading fees.
-        bytes32 id = IMidnight(midnight).touchObligation(takes[0].offer.obligation);
+        require(referralFeePct < WAD, PctExceeded());
+        address loanToken = takes[0].offer.obligation.loanToken;
+
+        uint256 referralFeeAssets = targetBuyerAssets.mulDivDown(referralFeePct, WAD);
+        uint256 preFeeTarget = targetBuyerAssets - referralFeeAssets;
```
QGarchery: I see why it's called preFeeTarget, but postFeeTarget would make more sense to me. It's a variable in the code only, so we can think of it as if the fees were paid upfront and then the rest to target after that would be the post fee target. We could even do the fee transfer here

src/periphery/TakeBundler.sol
QGarchery, line 1: With the fee, even without asynchrony, depending on the function, it's not necessarily easy to evaluate if a bundle makes sense:
- in `sellUnitsTarget`, compared to without the fee, each unit you are selling is giving you (1 - pct) times less value than what the offer advertises. So you can assume the fee is making your trade overall (1 - pct) times worse
- in `sellSellerAssetsTarget`, this is not the case. This is because you increase the size of the target so in addition you may have to take worse offers. For example let's say that I target 100 assets, and in the bundle there is an offer of 200 units at price 0.5 and 1000 units at price 0.1. Without fees I sell 200 units, with 50% fee do **not** sell 400 units as expected, I sell 1200 units (it targets 200 assets, and so it takes both offers completely).

I don't think there is an easy way to fix that in this PR, but it makes me change my mind about the average price check: at least users can be protected against such scenario

QGarchery (approved):

src/periphery/TakeBundler.sol
peyha: agreed, it makes me feel that the naming could be improved because it's ambiguous
for instance `fillTarget` would work

src/periphery/TakeBundler.sol
```diff
@@ -8,27 +8,36 @@ import {ITakeBundler, Take, CollateralTransfer} from "./interfaces/ITakeBundler.
...
 
 contract TakeBundler is ITakeBundler {
     using UtilsLib for uint256;
 
-    /// @dev Assumes offers are all share the same obligation id.
     /// @dev The taker must have authorized this bundler and the msg.sender (if different from the taker) on Midnight.
     /// @dev The bundler skips every reason why `take` can revert (including ones that are not asynchrony related).
     /// @dev If taking an offer reverts, the bundler will completely skip this offer.
+    /// @dev Taker receives exactly `targetUnits` units. Total loan-token cost is `totalFilledBuyerAssets + fee`, where
+    /// `fee = totalFilledBuyerAssets * pct / (WAD - pct)` (so `pct = fee / (totalFilledBuyerAssets + fee)`).
```
peyha: why do you say `pct = fee / (totalFilledBuyerAssets + fee)` ? not sure what info it adds

src/periphery/TakeBundler.sol
MathisGD: yes I'll go for a better name

peyha (approved):
LGTM, formula looks correct including rounding, pending naming issue

src/periphery/TakeBundler.sol
MathisGD: when for `targetFilled`

src/periphery/TakeBundler.sol
MathisGD: >I don't think there is an easy way to fix that in this PR, but it makes me change my mind about the average price check: at least users can be protected against such scenario

still, you control your worst price as you are passing the Takes (and the referral fee)

MathisGD: added a comment. let's discuss re-adding average check in a PR on top

PR #719, opened 2026-04-21
cleaner stack too deep fix

MathisGD:


QGarchery (approved):
let's fix the sizes issue in the base PR

PR #720, opened 2026-04-21
Remove problematic jq

QGarchery:
Which would allow to run the Certora CI again:
Screenshot 2026-04-21 at 17 43 53

### DIFF
--- a/.github/workflows/certora.yml
+++ b/.github/workflows/certora.yml
@@ -34,9 +34,6 @@ jobs:
           distribution: temurin
           java-version: "21"
 
-      - name: Install Jq
-        uses: sergeysova/jq-action@a3f0d4ff59cc1dddf023fc0b325dd75b10deec58 # v2
-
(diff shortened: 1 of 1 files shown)


.github/workflows/certora.yml
MathisGD, line 41: this will break right?

QGarchery: this action was useless, which is proven because the line `echo "confs=$(ls certora/confs/*.conf | xargs -I{} basename {} .conf | jq -nRc '[inputs]')" >> "$GITHUB_OUTPUT"` uses `jq` and was earlier in the job

peyha: ubuntu latest contains jq natively cf [this](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md)

peyha (approved):

MathisGD (approved):

PR #721, opened 2026-04-22
bundler: remove useless average checks

MathisGD:


peyha (approved):
LGTM, though `error SellerAssetsBelowMin` is not tested but it's not critical

QGarchery (approved):

PR #722, opened 2026-04-22
Refactor: use obligation id everywhere (even in Offer) instead of Obligation struct

peyha:
Probably requires a lot more work on the router side

TODO:
- [ ] adapt Certora files


src/Midnight.sol
```diff
@@ -731,10 +727,8 @@ contract Midnight is IMidnight {
 
     /// @dev Slashes the position and accrues the continuous fee.
     /// @dev Returns the new credit, new pending fee, and accrued fee after having updated the position.
-    function updatePosition(Obligation memory obligation, address user) external returns (uint128, uint128, uint128) {
-        bytes32 id = toId(obligation);
-        require(obligationState[id].created, ObligationNotCreated());
```
QGarchery, line 798: We should still check that the obligation is created, the new `updatePosition(bytes32, address)` function doesn't do it anymore

src/Midnight.sol
```diff
@@ -537,12 +529,6 @@ contract Midnight is IMidnight {
             bitmap = bitmap.clearBit(i);
         }
 
-        require(
-            originalDebt > 0 && !liquidationLocked(id, borrower)
-                && (block.timestamp > obligation.maturity || originalDebt > maxDebt),
-            NotLiquidatable()
-        );
```
QGarchery, line 601: We can keep it at the same place

src/Midnight.sol
QGarchery, line 1: I suspect it will be a pain to do adapt the CVL...

src/Midnight.sol
peyha: should have mentioned it but I moved this block to fix a stack too deep issue (it was only way i found to not have to implement additional structs)

src/Midnight.sol
peyha: agreed, that's why i didn't touch anything. But at this stage the PR is just a PoC

src/interfaces/IMidnight.sol
QGarchery, line 1: Additional motivation in [this thread](https://github.com/morpho-org/midnight/pull/715#discussion_r3118108074)

MathisGD:
abandoned, because doable on top (bundler level)

PR #724, opened 2026-04-22
bundler: pre-fee target

MathisGD:


MathisGD:
decided to go for this over #725 because it's not really more complex (only the inverse, but this is known stuff), and it's much simpler for people

MathisGD:
will fix directly in #717

PR #729, opened 2026-04-23
bundler: worst marginal price check

MathisGD:
- add a worst marginal price check, to be used as a sanity protection against router issues

https://morpholabs.slack.com/archives/C08A40KKN9W/p1776933228250279?thread_ts=1776931757.491219&cid=C08A40KKN9W


MathisGD:
discussed with router and front-end team, it's not needed

PR #730, opened 2026-04-23
bundler: supply/withdraw collat different obligations

MathisGD:
- buyerAssetsTarget and sellerAssetsTarget can be used with different obligations, so we need to be able to deposit/withdraw in different obligations.
- did it across all functions


src/periphery/TakeBundler.sol
peyha:
```suggestion
    /// @dev Assumes offers are all share the same obligation loan token.
```

MathisGD: actually it should be merged after #717

MathisGD:
it's broken, one cannot know in advance the amount that they will take in each obligation, so the all collateral management system would need to be rethought (see #793)

PR #732, opened 2026-04-24
Multi token flashloan

peyha:
not sure about the certora modifications


src/interfaces/IMidnight.sol
```diff
@@ -147,7 +147,7 @@ interface IMidnight {
     function setConsumed(bytes32 group, uint256 amount, address onBehalf) external;
     function shuffleSession(address onBehalf) external;
     function setIsAuthorized(address onBehalf, address authorized, bool newIsAuthorized) external;
-    function flashLoan(address token, uint256 assets, address callback, bytes calldata data) external;
+    function flashLoan(address[] calldata tokens, uint256[] calldata assets, address callback, bytes calldata data) external;
```
QGarchery:
```suggestion
    function flashLoan(address[] memory tokens, uint256[] memory assets, address callback, bytes calldata data) external;
```
it should do the same thing, this is essentially a convention that comes from things like [this](https://morpholabs.slack.com/archives/C02N7CSP91C/p1672161566913349)

src/Midnight.sol
```diff
@@ -660,14 +660,21 @@ contract Midnight is IMidnight {
         emit EventsLib.SetIsAuthorized(msg.sender, onBehalf, authorized, newIsAuthorized);
     }
 
-    function flashLoan(address token, uint256 assets, address callback, bytes calldata data) external {
-        emit EventsLib.FlashLoan(msg.sender, token, assets, callback);
-        SafeTransferLib.safeTransfer(token, callback, assets);
+    function flashLoan(address[] calldata tokens, uint256[] calldata assets, address callback, bytes calldata data)
```
QGarchery, line 663: Another option is to have a struct `FlashLoan` that has a field `token` and a field `assets`, and do:
```suggestion
    function flashLoan(FlashLoan[] calldata flashLoans, address callback, bytes calldata data)
```
No strong opinion, but then we wouldn't need to check `tokens.length == assets.length` and also it would be slightly more aligned with `multicall` and the `TakeBundler` function that have a `Take[]` parameter

QGarchery (approved):

MathisGD (approved):

src/interfaces/IMidnight.sol
adhusson: related https://morpholabs.slack.com/archives/C08A40KKN9W/p1747224968688519

any interface location can be paired with any implementation location

src/Midnight.sol
adhusson, line 1: Noting that this change is +353 bytes

adhusson (approved):

src/interfaces/IMidnight.sol
QGarchery: It would be nice to keep this convention then: very easy to check that in interfaces we always have `memory`, as this doesn't require to look at the implementation, and no risk to convey a false message ("ah this is calldata, so I can't call it with memory")

peyha: note that we also use calldata for dynamic args in `liquidate`, `repay` and `multicall` so not sure about this.
I don't see any strong argument to choose one or the other except consistency

peyha: actually, i think calldata is better for `tokens` and `amounts` because these arrays are only read element by element so solidity has no need to copy the whole array into memory

src/Midnight.sol
peyha: Slightly leaning toward keeping the current version because using custom struct would add unnecessary overhead (converting, understanding the struct). Also both balancer and aave v4 use an array of token and an array of amount in their flashloan functions

src/interfaces/IMidnight.sol
QGarchery: Yes the point is consistency

> actually, i think calldata is better for tokens and amounts because these arrays are only read element by element so solidity has no need to copy the whole array into memory

It doesn't change what is called, both do the same thing in the interface (not the same story in the implementation). See the thread linked above

MathisGD: let's apply this change

src/Midnight.sol
MathisGD: prefer double array too

src/interfaces/IMidnight.sol
peyha: shouldn't we apply it in another PR not linked to this one as it touches to other functions ?

MathisGD (approved):

PR #733, opened 2026-04-24
[Certora] role safety

lilCertora:
short PR with the same idea as [ownerSafety](https://github.com/morpho-org/vault-v2/blob/main/certora/specs/OwnerSafety.spec) in vault-v2


certora/specs/AdminSafety.spec
```diff
@@ -0,0 +1,193 @@
...
+
+/// FEE SETTER: LIVENESS ///
+
+rule feeSetterCanSetObligationTradingFee(env e, bytes32 id, uint256 index, uint256 newTradingFee) {
+    require(e.msg.sender == feeSetter(), "caller is the fee setter");
+    require(e.msg.value == 0, "no ETH sent");
+    require(index <= 6, "valid fee index");
+    require(newTradingFee <= maxTradingFee(index), "fee within bound");
+    require(newTradingFee % FEE_STEP() == 0, "fee is a multiple of FEE_STEP");
+    require(obligationCreated(id), "obligation exists");
```
MathisGD: why not the same syntax as usual?

assert !lastReverted <=> ...

MathisGD: btw it's not equivalent, because here we could over estimate

certora/specs/AdminSafety.spec
```diff
@@ -0,0 +1,193 @@
...
+/// FEE CLAIMER: LIVENESS ///
+/// Fee claimer access control is covered in OnlyAuthorizedCanChange.spec.
+
+rule feeClaimerCanClaimTradingFee(env e, address token, uint256 amount, address receiver) {
+    require(e.msg.sender == feeClaimer(), "caller is the fee claimer");
+    require(e.msg.value == 0, "no ETH sent");
+    require(amount <= claimableTradingFee(token), "enough claimable balance");
+
+    claimTradingFee@withrevert(e, token, amount, receiver);
+    assert !lastReverted;
```
MathisGD: maybe assert the token balance change?

certora/specs/Role.spec
```diff
@@ -0,0 +1,193 @@
...
+rule feeClaimerCanClaimContinuousFee(env e, Midnight.Obligation obligation, uint256 amount, address receiver) {
+    bytes32 id = toId(e, obligation);
+    require(e.msg.sender == feeClaimer(), "caller is the fee claimer");
+    require(e.msg.value == 0, "no ETH sent");
+    require(obligationCreated(id), "obligation exists");
+    require(amount <= withdrawable(id), "enough withdrawable");
+    require(amount <= totalUnits(id), "enough total units");
+    require(amount <= currentContract.obligationState[id].continuousFeeCredit, "enough continuous fee credit");
+
+    claimContinuousFee@withrevert(e, obligation, amount, receiver);
```
MathisGD, line 215: same

certora/specs/AdminSafety.spec
lilCertora: yes better this way , thanks!

certora/specs/AdminSafety.spec
lilCertora: done, thanks!

certora/specs/Role.spec
lilCertora: done, thanks!

certora/specs/Role.spec
QGarchery, line 1: I think that "Roles" would be a good name for this spec. It's not just about the admin, and it's not just about safety (there are liveness rules as well)

certora/specs/AdminSafety.spec
```diff
@@ -0,0 +1,209 @@
...
+definition defaultTradingFee(address loanToken, uint256 index) returns uint256 = assert_uint256(currentContract.defaultTradingFees[loanToken][index] * FEE_STEP());
+
+ghost mapping(address => mapping(address => mathint)) tokenBalance;
+
+function cvlSafeTransfer(address token, address receiver, uint256 amount) {
+    cvlSafeTransferFrom(token, currentContract, receiver, amount);
+}
+
+function cvlSafeTransferFrom(address token, address from, address to, uint256 amount) {
+    tokenBalance[token][from] = tokenBalance[token][from] - to_mathint(amount);
```
QGarchery:
```suggestion
    tokenBalance[token][from] = tokenBalance[token][from] - amount;
```
we can remove all `to_mathint()` occurences

certora/specs/AdminSafety.spec
```diff
@@ -0,0 +1,209 @@
...
+
+    assert feeClaimer() != before => e.msg.sender == roleSetterVal && f.selector == sig:setFeeClaimer(address).selector;
+}
+
+/// FEE SETTER: LIVENESS ///
+
+rule feeSetterCanSetObligationTradingFee(env e, bytes32 id, uint256 index, uint256 newTradingFee) {
+    address feeSetterBefore = feeSetter();
+    bool validIndex = index <= 6;
+    bool validFee = validIndex ? newTradingFee <= maxTradingFee(index) && newTradingFee % FEE_STEP() == 0 : false;
```
QGarchery:
```suggestion
    bool validFee = validIndex && newTradingFee <= maxTradingFee(index) && newTradingFee % FEE_STEP() == 0;
```
same below

certora/specs/AdminSafety.spec
```diff
@@ -0,0 +1,209 @@
...
+
+    setFeeClaimer@withrevert(e, newFeeClaimer);
+    assert !lastReverted <=> e.msg.sender == roleSetterBefore && e.msg.value == 0;
+    assert !lastReverted => feeClaimer() == newFeeClaimer;
+}
+
+/// ROLE SETTER: ACCESS CONTROL ///
+
+rule onlyRoleSetterCanChangeRoleSetter(env e, method f, calldataarg args) filtered { f -> !f.isView } {
+    address before = roleSetter();
```
QGarchery:
```suggestion
    address roleSetterBefore = roleSetter();
```
for consistency, there are occurrences below, with variables named `before` or `*Val`

certora/specs/AdminSafety.spec
```diff
@@ -0,0 +1,209 @@
...
+    uint32 before = currentContract.defaultContinuousFee[loanToken];
+    address feeSetterVal = feeSetter();
+
+    f(e, args);
+
+    assert currentContract.defaultContinuousFee[loanToken] != before => e.msg.sender == feeSetterVal && f.selector == sig:setDefaultContinuousFee(address, uint256).selector;
+}
+
+/// FEE CLAIMER: LIVENESS ///
+/// Fee claimer access control is covered in OnlyAuthorizedCanChange.spec.
```
QGarchery: Let's move those rules here. They don't belong in `OnlyAuthorizedCanChange` since they are not about the `isAuthorized` mapping

certora/specs/AdminSafety.spec
```diff
@@ -0,0 +1,209 @@
...
+    address feeSetterBefore = feeSetter();
+
+    setDefaultContinuousFee@withrevert(e, loanToken, newContinuousFee);
+    bool reverted = lastReverted;
+    assert !reverted <=> e.msg.sender == feeSetterBefore && e.msg.value == 0 && newContinuousFee <= MAX_CONTINUOUS_FEE();
+    assert !reverted => to_mathint(currentContract.defaultContinuousFee[loanToken]) == to_mathint(newContinuousFee);
+}
+
+/// FEE SETTER: ACCESS CONTROL ///
+/// Trading fee access control is covered in FeeBoundaries.spec.
```
QGarchery: Let's move those here, since those rules are not really about boundaries.
Also, can you rename according to the new convention please:

```suggestion
/// Trading fee access control is covered in TradingFeeBoundaries.spec.
```

certora/confs/AdminSafety.conf
```diff
@@ -0,0 +1,20 @@
+{
+  "files": [
+    "certora/helpers/Utils.sol",
```
QGarchery:
```suggestion
```

certora/specs/Role.spec
```diff
@@ -0,0 +1,209 @@
...
+
+    function _.price() external => NONDET;
+    function isHealthy(Midnight.Obligation memory, bytes32, address) internal returns (bool) => NONDET;
+    function _.onRatify(Midnight.Offer, bytes32, bytes) external => NONDET;
+    function IdLib.storeInCode(Midnight.Obligation memory) internal returns (address) => NONDET;
+    function TickLib.tickToPrice(uint256) internal returns (uint256) => NONDET;
+    function TickLib.wExp(int256) internal returns (uint256) => NONDET;
+    function UtilsLib.isLeaf(bytes32, bytes32, bytes32[] memory) internal returns (bool) => NONDET;
+    function UtilsLib.mulDivDown(uint256, uint256, uint256) internal returns (uint256) => NONDET;
+    function UtilsLib.mulDivUp(uint256, uint256, uint256) internal returns (uint256) => NONDET;
```
QGarchery: Do we need all those summaries ? For each of them we should explain why they are ok to make: mostly not used by the corresponding functions, except in rules such as `onlyRoleSetterCanChangeRoleSetter` that don't look at reverts

certora/specs/AdminSafety.spec
```diff
@@ -0,0 +1,209 @@
...
+    function UtilsLib.mulDivUp(uint256, uint256, uint256) internal returns (uint256) => NONDET;
+
+    // Assumption: token transfers do not revert and do not re-enter Midnight.
+    function SafeTransferLib.safeTransfer(address token, address receiver, uint256 amount) internal => cvlSafeTransfer(token, receiver, amount);
+    function SafeTransferLib.safeTransferFrom(address token, address from, address to, uint256 amount) internal => cvlSafeTransferFrom(token, from, to, amount);
+}
+
+/// HELPERS ///
+
+definition FEE_STEP() returns uint256 = 1000000000000;
```
QGarchery:
```suggestion
definition FEE_STEP() returns uint256 = 10 ^ 12;
```
(nitpick)

certora/specs/AdminSafety.spec
```diff
@@ -0,0 +1,209 @@
...
+    uint256 claimableBefore = claimableTradingFee(token);
+    mathint midnightBalanceBefore = tokenBalance[token][currentContract];
+    mathint receiverBalanceBefore = tokenBalance[token][receiver];
+
+    claimTradingFee@withrevert(e, token, amount, receiver);
+    bool reverted = lastReverted;
+    assert !reverted <=> e.msg.sender == feeClaimerBefore && e.msg.value == 0 && amount <= claimableBefore;
+    assert !reverted => claimableTradingFee(token) == claimableBefore - amount;
+    assert !reverted => tokenBalance[token][currentContract] == midnightBalanceBefore - (receiver == currentContract ? 0 : to_mathint(amount));
+    assert !reverted => tokenBalance[token][receiver] == receiverBalanceBefore + (receiver == currentContract ? 0 : to_mathint(amount));
```
QGarchery: (optional) we could also check that all other balances are unchanged

certora/specs/AdminSafety.spec
```diff
@@ -0,0 +1,209 @@
...
+    bool reverted = lastReverted;
+    assert !reverted <=> e.msg.sender == feeSetterBefore && e.msg.value == 0 && newContinuousFee <= MAX_CONTINUOUS_FEE();
+    assert !reverted => to_mathint(currentContract.defaultContinuousFee[loanToken]) == to_mathint(newContinuousFee);
+}
+
+/// FEE SETTER: ACCESS CONTROL ///
+/// Trading fee access control is covered in FeeBoundaries.spec.
+
+rule onlyFeeSetterCanChangeObligationContinuousFee(env e, method f, calldataarg args, bytes32 id) filtered { f -> !f.isView } {
+    require obligationCreated(id);
```
QGarchery: Let's add a reason string, and document it at the top of the rule

certora/specs/Role.spec
lilCertora: yes good idea

certora/specs/AdminSafety.spec
lilCertora: right

certora/specs/AdminSafety.spec
lilCertora: done

certora/specs/AdminSafety.spec
lilCertora: yes right it was not formatted consistently

certora/specs/Role.spec
lilCertora: done

certora/specs/AdminSafety.spec
lilCertora: done

certora/specs/AdminSafety.spec
lilCertora: done

certora/specs/Role.spec
lilCertora: yes that's right, I removed the non really useful ones and added explanations

certora/specs/AdminSafety.spec
QGarchery: We need to rename the corresponding file as well (`FeeBoundaries.spec` and `FeeBoundaries.conf`)

certora/specs/Role.spec
```diff
@@ -0,0 +1,228 @@
...
+    function maxTradingFee(uint256 index) external returns (uint256) envfree;
+
+    // Tick pricing is only used in unrelated offer-taking arithmetic.
+    function TickLib.tickToPrice(uint256) internal returns (uint256) => NONDET;
+
+    // Offer hashing only gates unrelated `take` executions; it does not affect role or fee-admin state.
+    function UtilsLib.hashOffer(Midnight.Offer memory) internal returns (bytes32) => NONDET;
+
+    // Merkle proof validity only gates unrelated `take` executions.
+    function UtilsLib.isLeaf(bytes32, bytes32, bytes32[] memory) internal returns (bool) => NONDET;
```
QGarchery:
```suggestion
    // Those functions are over-approximated, except for the reverting behavior. This is still sound as they are only used inside `take` but we don't look at the reverting behavior of `take` in this file.
    function TickLib.tickToPrice(uint256) internal returns (uint256) => NONDET;
    function UtilsLib.hashOffer(Midnight.Offer memory) internal returns (bytes32) => NONDET;
    function UtilsLib.isLeaf(bytes32, bytes32, bytes32[] memory) internal returns (bool) => NONDET;
```
to simplify a bit the explanation

certora/specs/Role.spec
```diff
@@ -0,0 +1,228 @@
...
+    bool reverted = lastReverted;
+    assert !reverted <=> e.msg.sender == feeSetterBefore && e.msg.value == 0 && newContinuousFee <= MAX_CONTINUOUS_FEE();
+    assert !reverted => currentContract.defaultContinuousFee[loanToken] == newContinuousFee;
+}
+
+/// FEE SETTER: ACCESS CONTROL ///
+/// Trading fee access control is covered in TradingFeeBoundaries.spec.
+
+rule onlyFeeSetterCanChangeObligationContinuousFee(env e, method f, calldataarg args, bytes32 id) filtered { f -> !f.isView } {
+    // Per-obligation continuous fees can only be changed for existing obligations.
```
QGarchery:
```suggestion

/// Once an obligation is created, only the fee setter can modify its continuous fees.
rule onlyFeeSetterCanChangeObligationContinuousFeePostCreation(env e, method f, calldataarg args, bytes32 id) filtered { f -> !f.isView } {
```
same comment as `onlyFeeSetterCanChangeObligationTradingFeesPostCreation`, and on top of the rule

QGarchery (approved):

certora/specs/Role.spec
```diff
@@ -0,0 +1,224 @@
...
+}
+
+/// ROLE SETTER: ACCESS CONTROL ///
+
+rule onlyRoleSetterCanChangeRoleSetter(env e, method f, calldataarg args) filtered { f -> !f.isView } {
+    address roleSetterBefore = roleSetter();
+
+    f(e, args);
+
+    assert roleSetter() != roleSetterBefore => e.msg.sender == roleSetterBefore && f.selector == sig:setRoleSetter(address).selector;
```
MathisGD, line 82: technically enough (with the rule above), but fine
```suggestion
    assert roleSetter() != roleSetterBefore => f.selector == sig:setRoleSetter(address).selector;
```

MathisGD (approved):

PR #734, opened 2026-04-24
wallet sig iteration

MathisGD:


src/ratifiers/interfaces/IEcrecoverRatifier.sol
```diff
@@ -4,6 +4,14 @@ pragma solidity ^0.8.0;
 
 import {IRatifier} from "../../interfaces/IRatifier.sol";
 
+struct Signature {
+    uint8 v;
+    bytes32 r;
+    bytes32 s;
```
peyha, line 10: not a huge fan of not centralizing this in an interface/lib
for instance we'll also need it in the adapter's `onRatify`

MathisGD: it's only 3 LOC, and usually we don't try to factorize interfaces

src/libraries/UtilsLib.sol
```diff
@@ -52,17 +52,13 @@ library UtilsLib {
...
 
-    function rootTypeHash(uint256 height) internal pure returns (bytes32) {
-        bytes memory brackets = new bytes(3 * height);
+    function offerTreeTypeHash(uint256 height) internal pure returns (bytes32) {
+        bytes memory typeStr = "OfferTree(Offer";
         for (uint256 i = 0; i < height; i++) {
-            brackets[3 * i] = "[";
-            brackets[3 * i + 1] = "2";
-            brackets[3 * i + 2] = "]";
+            typeStr = bytes.concat(typeStr, "[2]");
```
QGarchery, line 58: Slightly less efficient because it will allocate the `typeStr` variable at each iteration, but it is not expected to matter much

QGarchery (approved):

PR #738, opened 2026-04-28
document hardfork scenario (Spearbit #23)

MathisGD:
https://cantina.xyz/code/45b68aa9-34ba-4eef-ba96-8fa07c0b240a/findings/23


src/Midnight.sol
QGarchery, line 1: A similar comment could be added to Ecrecover contracts: signatures would not longer be applicable in case of a hard fork. Less critical though

Btw, it seems like we will always need a comment every time we use `block.chainid` for signatures. If the message doesn't change when `block.chainid` changes, then there is a replay "attack" in case of a hard fork. If the message recomputes the the `block.chainid` every time, then it stops working in some way in case of a hard fork

QGarchery (approved):

peyha (approved):

adhusson (commented):
LGTM

PR #739, opened 2026-04-28
prevent no-op liq inactive collateral (Spearbit-5)

MathisGD:
https://cantina.xyz/code/45b68aa9-34ba-4eef-ba96-8fa07c0b240a/findings/5

### DIFF
--- a/src/Midnight.sol
+++ b/src/Midnight.sol
@@ -519,6 +519,7 @@ contract Midnight is IMidnight {
         ObligationState storage _obligationState = obligationState[id];
         Position storage _position = position[id][borrower];
         require(UtilsLib.atMostOneNonZero(repaidUnits, seizedAssets), InconsistentInput());
+        require(_position.collateral[collateralIndex] > 0, InactiveCollateral());
         require(
             obligation.liquidatorGate == address(0)
                 || ILiquidatorGate(obligation.liquidatorGate).canLiquidate(msg.sender),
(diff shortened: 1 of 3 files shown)


src/Midnight.sol
```diff
@@ -519,6 +519,7 @@ contract Midnight is IMidnight {
         ObligationState storage _obligationState = obligationState[id];
         Position storage _position = position[id][borrower];
         require(UtilsLib.atMostOneNonZero(repaidUnits, seizedAssets), InconsistentInput());
+        require(_position.collateral[collateralIndex] > 0, InactiveCollateral());
```
Saw-mon-and-Natalie, line 522: nit: if `collateralIndex > 127` one would get compiler error instead.

MathisGD: right. that's fine, as this is an "impossible collateral", not an inactive collateral

peyha (approved):
LGTM, the previous version could emit Liquidate events with impossible collateral tokens potentially leading to weird behavior in indexers

QGarchery (approved):
See #764 as an alternative

src/Midnight.sol
adhusson, line 506: Needs some update on the fact that even with 0, an existing collat must be used.

MathisGD: I don't think that it's unexpected, not mandatory for me. Can add if you feel strongly

MathisGD:
replaced by https://github.com/morpho-org/midnight/pull/785

PR #741, opened 2026-04-28
docs: authorization considerations

MathisGD:
document what to look at when authorizing an account


src/Midnight.sol
```diff
@@ -70,6 +70,11 @@ import {EventsLib} from "./libraries/EventsLib.sol";
 /// @dev All functions that change the position, session, consumed and authorization are accessible to the user and to
 /// any account that has been authorized.
 /// @dev In particular, authorized accounts can authorize other accounts on behalf of the user.
+/// @dev When authorizing an account, one should consider:
+/// - The functions that the account can call, and on which contract. At least Midnight's functions should be
+/// considered, but other contracts might re-use Midnight's authorization mapping too (e.g ratifiers).
+/// - The calls that the account can receive. At least Midnight's callbacks should be considered (with
+/// `CALLBACK_SUCCESS` return value).
```
QGarchery: It's only `onRatify` that should be considered for authorization. Other callback methods should be considered, but not when authorizing an account. For example one should consider if a contract can be called with `onRepay`, if you pass that contract as the `callback` argument when calling `repay`

```suggestion
/// - Whether the account can return `CALLBACK_SUCCESS` when its `onRatify` function is called.
```

MathisGD: right, mb

src/Midnight.sol
```diff
@@ -70,6 +70,10 @@ import {EventsLib} from "./libraries/EventsLib.sol";
 /// @dev All functions that change the position, session, consumed and authorization are accessible to the user and to
 /// any account that has been authorized.
 /// @dev In particular, authorized accounts can authorize other accounts on behalf of the user.
+/// @dev When authorizing an account, one should consider:
+/// - The functions that the account can call, and on which contract. At least Midnight's functions should be
+/// considered, but other contracts might re-use Midnight's authorization mapping too (e.g ratifiers).
```
MathisGD: add signatures for EOAs

MathisGD: do a higher level description (and add authorizers)

src/Midnight.sol
```diff
@@ -70,6 +70,10 @@ import {EventsLib} from "./libraries/EventsLib.sol";
 /// @dev All functions that change the position, session, consumed and authorization are accessible to the user and to
 /// any account that has been authorized.
 /// @dev In particular, authorized accounts can authorize other accounts on behalf of the user.
```
adhusson: should be moved into the section below (as a "special consideration")

src/Midnight.sol
```diff
@@ -70,6 +70,10 @@ import {EventsLib} from "./libraries/EventsLib.sol";
...
+/// @dev When authorizing an account, one should consider:
+/// - The functions that the account can call, and on which contract. At least Midnight's functions should be
```
```suggestion
/// - All functions in contracts using the authorization mapping. Midnight functions are the most direct example but other contracts might re-use it (e.g. ratifiers)
```
would go for something like

src/Midnight.sol
```diff
@@ -88,7 +88,12 @@ import {EventsLib} from "./libraries/EventsLib.sol";
 /// AUTHORIZATIONS
 /// @dev All functions that change the position, session, consumed and authorization are accessible to the user and to
 /// any account that has been authorized.
-/// @dev In particular, authorized accounts can authorize other accounts on behalf of the user.
+/// @dev When authorizing an account, one should consider:
+/// - The targets/functions that the account can call. At least Midnight's functions should be considered, but other
+/// contracts might re-use Midnight's authorization mapping too (e.g ratifiers and authorizers). In particular,
+/// authorized accounts can authorize other accounts on behalf of the user.
+/// - Whether the account can return `CALLBACK_SUCCESS` when its `onRatify` function is called.
```
adhusson: More precise wording for authorizations that are specifically about ratification: what matters is not whether it can do it but when it does it.

```suggestion
/// - Under which conditions the account can return `CALLBACK_SUCCESS` when its `onRatify` function is called.
```

src/Midnight.sol
```diff
@@ -88,7 +88,12 @@ import {EventsLib} from "./libraries/EventsLib.sol";
 /// AUTHORIZATIONS
 /// @dev All functions that change the position, session, consumed and authorization are accessible to the user and to
 /// any account that has been authorized.
-/// @dev In particular, authorized accounts can authorize other accounts on behalf of the user.
+/// @dev When authorizing an account, one should consider:
+/// - The targets/functions that the account can call. At least Midnight's functions should be considered, but other
+/// contracts might re-use Midnight's authorization mapping too (e.g ratifiers and authorizers). In particular,
+/// authorized accounts can authorize other accounts on behalf of the user.
+/// - Under which conditions the account can return `CALLBACK_SUCCESS` when its `onRatify` function is called.
+/// - For EOAs, the signatures that it can sign.
```
adhusson: re-reading this line it's actually not clear to me

```suggestion
/// - For EOAs when a ratifier such as the EcrecoverRatifier has been authorized, the signatures that it can sign.
```
nitpick (but i find it less clear so we can prob discard it)

peyha (approved):

src/Midnight.sol
```suggestion
/// @dev When authorizing a smart-contract, one should consider:
/// - The targets/functions that the account can call. At least Midnight's functions should be considered, but other
/// contracts might re-use Midnight's authorization mapping too (e.g ratifiers and authorizers). In particular,
/// authorized accounts can authorize other accounts on behalf of the user.
/// - Under which conditions the account can return `CALLBACK_SUCCESS` when its `onRatify` function is called.
```

src/Midnight.sol
```diff
@@ -88,7 +88,12 @@ import {EventsLib} from "./libraries/EventsLib.sol";
 /// AUTHORIZATIONS
 /// @dev All functions that change the position, session, consumed and authorization are accessible to the user and to
 /// any account that has been authorized.
```
```suggestion
/// @dev All functions that change the position, session, consumed and authorization are accessible to the user and to
/// any account that has been authorized. Thus, to scope authorizations one should authorize a smart-contract with scoped behavior.
```

QGarchery (approved):

peyha (approved):

PR #742, opened 2026-04-28
require obligation's TTM<100y (Spearbit-3)

MathisGD:
- https://cantina.xyz/code/45b68aa9-34ba-4eef-ba96-8fa07c0b240a/findings/3 \n- thighter constraint: pending<=credit AND take won't revert for this (the feeSetter can't make take revert for this)


test/TakeTest.sol
```diff
@@ -1626,36 +1626,14 @@ contract TakeTest is BaseTest {
         );
     }
 
-    function testBuyerPendingFeeExceedsCredit() public {
-        // Use a very long maturity so continuousFee * TTM > WAD.
-        midnight.setDefaultContinuousFee(address(loanToken), MAX_CONTINUOUS_FEE);
-
+    function testMaturityTooFar() public {
```
peyha: could move the test out of `TakeTest` because it's not a take anymore

src/Midnight.sol
```diff
@@ -673,6 +672,7 @@ contract Midnight is IMidnight {
     /// @dev Returns the obligation id and creates the obligation if it doesn't exist yet.
     function touchObligation(Obligation memory obligation) public returns (bytes32) {
         bytes32 id = toId(obligation);
+        require(obligation.maturity <= block.timestamp + 100 * 365 days, MaturityTooFar());
```
QGarchery: Putting it here, instead of inside the `if (!obligationState[id].created)` block makes the verification go through, since the spec can now know, at each interaction, that the obligation maturity isn't too far. At the same time it costs gas at each interaction, instead of just at obligation creation. Now that I think about it, it's a nice way to do it, though a bit unexpected (for example if we do #722, then the main functions might not call `touchObligation`, and so they incidentally won't do this check)

Saw-mon-and-Natalie: also it does not cover the 2nd point from the finding:

> 2. Disallow creation of dead markets or document that creation of markets with past maturity is allowed.
> 
> Overall, one is hoping for checks of the form:
> 
> $$
> {\color{red}t_{creation} \leq} \space t_{m} \leq t_{creation} + \Delta t_{fixed}
> $$

MathisGD: @QGarchery I'll try to put it in the if and fix the certora.

@Saw-mon-and-Natalie I answered on the issue on the cantina platform.

test/OtherFunctionsTest.sol
```diff
@@ -364,6 +364,16 @@ contract OtherFunctionsTest is BaseTest {
         _obligation.rcfThreshold = 0;
     }
 
+    function testMaturityTooFar() public {
+        Obligation memory longObligation;
+        longObligation.loanToken = address(loanToken);
+        longObligation.maturity = block.timestamp + 100 * 365 days + 1;
```
peyha: could fuzz, no big deal

peyha (approved):
lgtm pending require that should be moved in the if

test/BaseTest.sol
```diff
@@ -348,6 +348,7 @@ abstract contract BaseTest is Test {
         }
         collateralParams = sortCollateralParams(collateralParams);
         obligation.collateralParams = collateralParams;
+        obligation.maturity = obligation.maturity % (block.timestamp + 100 * 365 days);
```
adhusson:
```suggestion
        obligation.maturity = bound(obligation.maturity, 0, block.timestamp + 100 * 365 days);
```

adhusson (commented):
lgtm pending upcoming changes

src/Midnight.sol
MathisGD: solved your issue with a little require in the certora spec, that we can solve later

adhusson (approved):

src/Midnight.sol
```diff
@@ -674,6 +673,7 @@ contract Midnight is IMidnight {
     function touchObligation(Obligation memory obligation) public returns (bytes32) {
         bytes32 id = toId(obligation);
         if (!obligationState[id].created) {
+            require(obligation.maturity <= block.timestamp + 100 * 365 days, MaturityTooFar());
```
QGarchery, line 732: Let's add a comment about this

QGarchery (approved):

certora/specs/Midnight.spec
```diff
@@ -123,8 +125,30 @@ rule userLossIndexMonotonicallyIncreases(bytes32 id, address user, method f, env
...
+    pendingFee(id, user) <= creditOf(id, user)
+    {
+        preserved with (env e) {
+            requireInvariant continuousFeeBounded(id);
+            requireInvariant defaultContinuousFeeBoundedAll();
+        }
+        preserved take(uint256 unitsInput, address taker, address takerCallbackAddress, bytes takerCallbackData, address receiverIfTakerIsSeller, Midnight.Offer offer, bytes ratifierData, bytes32 root, bytes32[] proof) with (env e) {
+            requireInvariant continuousFeeBounded(id);
+            requireInvariant defaultContinuousFeeBoundedAll();
+            require to_mathint(offer.obligation.maturity) <= to_mathint(e.block.timestamp) + MAX_TTM(); // TODO verify this cleanly
```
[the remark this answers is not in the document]
MathisGD: in #109

certora/specs/ContinuousFee.spec
```diff
@@ -49,6 +49,8 @@ rule continuousFeeNotOverchargedForBuyer(env e, uint256 units, address taker, ad
 
     postUpdateCredit, postUpdatePendingFee, _ = updatePositionView(e, offer.obligation, id, buyer);
 
+    require pendingFee(id, buyer) <= creditOf(id, buyer), "See pendingContinuousFeeBoundedByCredit in Midnight.spec";
```
MathisGD, line 52: to avoid take skipping updatePosition

adhusson (approved):

QGarchery (approved):
Approving, assuming the CI goes through

PR #743, opened 2026-04-28
handle max loss index (spearbit-9)

QGarchery:
https://cantina.xyz/code/45b68aa9-34ba-4eef-ba96-8fa07c0b240a/findings/9 

See [corresponding thread](https://morpholabs.slack.com/archives/C08A40KKN9W/p1777392118976379). And latest decision [here](https://morpholabs.slack.com/archives/C08A40KKN9W/p1777986402198879?thread_ts=1777570794.540109&cid=C08A40KKN9W)


src/Midnight.sol
```diff
@@ -77,8 +77,8 @@ import {EventsLib} from "./libraries/EventsLib.sol";
 /// for chains where the gas is cheaper than 1 asset of the loan token.
 /// @dev lossIndex is rounded up so lenders collectively lose a bit more on each bad debt realization.
 /// @dev slash rounds the credit down, so lenders lose a bit at each interaction.
-/// @dev If an obligation loses more than 99%+ of its value to bad debt over its lifetime, it won't function properly
-/// afterwards (bad debt can no longer be realized).
+/// @dev If an obligation loses almost all of its value to bad debt over its lifetime, such that the loss index is
```
peyha: not a huge fan of the phrasing, because it could be maxed out after multiple bad debt realization (e.g. 8xy times losing 90% of the value)

QGarchery: I don't see it as an issue because losing 800 times 90% implies that it has lost 99%+ of its value over its lifetime.

We have to have something that is comprehensible without being very deep in the code, no need to know exactly how the loss index works. So just removing the 99%+ thing is not great: readers could be alarmed that the obligation doesn't function properly in usual cases, when in reality it's a (very very very) extreme case

peyha: Isn't the concept of losing value over the obligation lifetime slightly misleading ? Suppose I have an obligation and that 800 times, a unique 1$ position creates 90% bad debt but then a large borrower (let's say 1M$) appear, what proportion of the obligation value was lost ?

src/Midnight.sol
```diff
@@ -77,8 +77,8 @@ import {EventsLib} from "./libraries/EventsLib.sol";
...
+/// maxed out, then the obligation won't function properly afterwards. Notably, credits would be worthless.
```
```suggestion
/// maxed out, then the obligation won't function properly afterwards. Notably, credit would be worthless.
```

MathisGD (approved):

src/Midnight.sol
MathisGD: what about this?
```suggestion
require(_lossIndex < type(uint128).max, ObligationIsDead());
uint256 postSlashCredit = 
    credit.mulDivDown(type(uint128).max - obligationState[id].lossIndex, type(uint128).max - _lossIndex);
```

MathisGD:
btw, see https://github.com/sherlock-audit/2026-04-morpho-midnight-apr-6th-2026/issues/23

src/Midnight.sol
MathisGD: can enable 

invariant index=max => totalUnits=0

and something around the line of "updatePosition is idempotent"

QGarchery: weird to error with `ObligationIsDead()` when the user loss index is checked and not the obligation loss index

QGarchery: other option is to simply remove the require compared to the code @MathisGD suggested (same behavior)

QGarchery: > index=max => totalUnits=0

This invariant doesn't hold because we round up the index. So we can have a bad debt of less than the total units and still reach the maximum loss index

QGarchery: See [corresponding thread](https://morpholabs.slack.com/archives/C08A40KKN9W/p1777392118976379)

src/Midnight.sol
QGarchery: Agree that it's slightly misleading, but we also give the details (loss index is maxed out) so readers have the full reasoning as well. And I think it's best to not remove it for the reason above, so not sure what to do

peyha: in the end, i think it's fine because it's an edge case and it's not too difficult to understand the subtlety

src/Midnight.sol
```diff
@@ -77,8 +77,8 @@ import {EventsLib} from "./libraries/EventsLib.sol";
 /// for chains where the gas is cheaper than 1 asset of the loan token.
 /// @dev lossIndex is rounded up so lenders collectively lose a bit more on each bad debt realization.
 /// @dev slash rounds the credit down, so lenders lose a bit at each interaction.
-/// @dev If an obligation loses more than 99%+ of its value to bad debt over its lifetime, it won't function properly
-/// afterwards (bad debt can no longer be realized).
+/// @dev If an obligation loses almost all of its value to bad debt over its lifetime, such that the loss index is
+/// maxed out, then the obligation won't function properly afterwards. Notably, credit positions would be stuck.
```
```suggestion
/// maxed out, lender interactons with the obligation will revert except collateral supply/withdraw. Notably, credit positions would be stuck.
```

peyha: not so sure about how to phrase it correctly, probably a bad idea

adhusson (approved):

peyha (approved):

certora/specs/NoDivisionByZero.spec
```diff
@@ -59,14 +59,9 @@ persistent ghost bytes32 globalId;
...
-// lossIndex < max: the protocol stop behaving correctly if this happens (documented).
-hook Sload uint128 value obligationState[KEY bytes32 id].lossIndex {
-    require value < max_uint128;
-}
-
-// Follows from userLossIndexLeqObligationLossIndex in Midnight.spec and the hook above.
+// Follows from userLossIndexLeqObligationLossIndex in Midnight.spec.
 hook Sload uint128 value position[KEY bytes32 id][KEY address user].lossIndex {
-    require value < max_uint128;
+    require value <= currentContract.obligationState[id].lossIndex;
```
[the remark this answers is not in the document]
QGarchery: This is proven in Midnight.spec, as stated above the hook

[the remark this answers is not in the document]
QGarchery: duplicate of https://github.com/morpho-org/midnight/pull/743#discussion_r3184167284

adhusson (approved):

src/Midnight.sol
```diff
@@ -101,8 +101,8 @@ import {EventsLib} from "./libraries/EventsLib.sol";
 /// for chains where the gas is cheaper than 1 asset of the loan token.
 /// @dev lossIndex is rounded up so lenders collectively lose a bit more on each bad debt realization.
 /// @dev slash rounds the credit down, so lenders lose a bit at each interaction.
-/// @dev If an obligation loses more than 99%+ of its value to bad debt over its lifetime, it won't function properly
-/// afterwards (bad debt can no longer be realized).
+/// @dev If an obligation loses almost all of its value to bad debt over its lifetime, such that the loss index is
+/// maxed out, then the obligation won't function properly afterwards. Notably, the take function would revert.
```
MathisGD: is it too ambitious to say exactly what doesn't work well (basically the relative value of the rounding)

certora/specs/Midnight.spec
```diff
@@ -118,6 +118,15 @@ rule userLossIndexMonotonicallyIncreases(bytes32 id, address user, method f, env
     assert lossIndexAfter >= lossIndexBefore;
 }
 
+rule creditCannotIncreaseWhenLossIndexIsMaxed(bytes32 id, address user, method f, env e, calldataarg args) {
```
MathisGD: debt neither? not mandatory

MathisGD (approved):

src/Midnight.sol
QGarchery: I tried something, let me know if you had something else in mind

src/Midnight.sol
```diff
@@ -103,9 +103,9 @@ import {EventsLib} from "./libraries/EventsLib.sol";
 /// for chains where the gas is cheaper than 1 asset of the loan token.
 /// @dev lossIndex is rounded up so lenders collectively lose a bit more on each bad debt realization.
 /// @dev slash rounds the credit down, so lenders lose a bit at each interaction.
-/// @dev If an obligation loses more than 99%+ of its value to bad debt over its lifetime, it won't function properly
-/// afterwards (bad debt can no longer be realized).
-///
+/// @dev If an obligation loses almost all of its value to bad debt over its lifetime, then the accounting of the loss
+/// may become so imprecise that extreme/total losses may be wrongly reported. In those cases the obligation doesn't
+/// function properly, and notably the take function reverts when the loss index is maxed out.
```
```suggestion
/// @dev If an obligation loses almost all of its value to bad debt over its lifetime, then the accounting of the loss
/// may become so imprecise (against users) and the take function reverts when the loss index is maxed out.
```

adhusson: As is not quite correct english. Also, imprecision can be super high at uint128.max-1 for instance.

MathisGD (approved):

src/Midnight.sol
QGarchery: I kept "In those cases the obligation doesn't function properly" because the issue also highlights that the repaid amount will not be able to be used by anyone. So I think it's better to be conservative (you shouldn't try to use this obligation anyway)

MathisGD (approved):

PR #747, opened 2026-04-29
docs: rcf

MathisGD:
clarify rcf threshold

### DIFF
--- a/src/Midnight.sol
+++ b/src/Midnight.sol
@@ -57,8 +57,8 @@ import {EventsLib} from "./libraries/EventsLib.sol";
 ///   newDebt >= newMaxDebt <=> debtOf - repaidUnits >= maxDebt - repaidUnits*LIF*LLTV
 ///                         <=> repaidUnits <= (debtOf-maxDebt) / (1 - LIF*LLTV).
 /// The maxRepaid computation is rounded up to avoid consecutive max liquidations, so the position could be slightly healthy after a liquidation.
-/// @dev The RCF is deactivated for small collateral amount, essentially to avoid liquidations that are too small
-/// (compared to the gas cost). More precisely, it is deactivated if the liquidation could leave a collateral with a
+/// @dev The RCF is deactivated for small collateral amount, essentially to mitigate issues with liquidations that are too small
+/// compared to the gas cost. More precisely, it is deactivated if the liquidation could leave a collateral with a
 /// value that would not be enough to repay rcfThreshold units. Which means:
 ///       minNewCollateral * liquidatedCollatPrice / LIF < rcfThreshold
 ///   <=> (collateral - maxRepaid * LIF / liquidatedCollatPrice)
--- a/src/Midnight.sol
+++ b/src/Midnight.sol
@@ -56,7 +56,7 @@ import {EventsLib} from "./libraries/EventsLib.sol";
 /// liquidation from putting the borrower back into health. Which means:
 ///   newDebt >= newMaxDebt <=> debtOf - repaidUnits >= maxDebt - repaidUnits*LIF*LLTV
 ///                         <=> repaidUnits <= (debtOf-maxDebt) / (1 - LIF*LLTV).
-/// (rounded up to avoid consecutive max liquidations, so the position could be slightly healthy after a liquidation).
+/// The maxRepaid computation is rounded up to avoid consecutive max liquidations, so the position could be slightly healthy after a liquidation.
 /// @dev The RCF is deactivated for small collateral amount, essentially to avoid liquidations that are too small
 /// (compared to the gas cost). More precisely, it is deactivated if the liquidation could leave a collateral with a
 /// value that would not be enough to repay rcfThreshold units. Which means:
--- a/src/Midnight.sol
+++ b/src/Midnight.sol
@@ -53,17 +53,18 @@ import {EventsLib} from "./libraries/EventsLib.sol";
 /// @dev Before maturity, the liquidation cannot put the borrower back into health (recovery close factor), unless
 /// the liquidation could leave a collateral with a value that would not be enough to repay rcfThreshold units.
 /// @dev The "recovery close factor" (RCF) limits the amount that can be liquidated. In particular, it prevents the
-/// liquidation from putting the borrower back into health. Which means:
+/// liquidation from putting the borrower back into health. Which means (omitting scaling and roundings):
 ///   newDebt >= newMaxDebt <=> debtOf - repaidUnits >= maxDebt - repaidUnits*LIF*LLTV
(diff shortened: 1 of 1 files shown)


src/Midnight.sol
```diff
@@ -47,6 +47,24 @@ import {EventsLib} from "./libraries/EventsLib.sol";
...
+///   newDebt >= newMaxDebt <=> debtOf - repaidUnits >= maxDebt - repaidUnits*LIF*LLTV
+///                         <=> repaidUnits <= (debtOf-maxDebt) / (1 - LIF*LLTV).
+/// (rounded up to avoid consecutive max liquidations, so the position could be slightly healthy after a liquidation).
+/// @dev The RCF is deactivated for small collateral amount, essentially to avoid liquidations that are too small
+/// (compared to the gas cost). More precisely, it is deactivated if the liquidation could leave a collateral with a
+/// value that would not be enough to repay rcfThreshold units. Which means:
+///   <=> minNewCollateral * liquidatedCollatPrice / ORACLE_PRICE_SCALE * WAD / lif < rcfThreshold
+///   <=> (collateral - maxRepaidUnits * lif * ORACLE_PRICE_SCALE / (WAD * liquidatedCollatPrice))
+///         * liquidatedCollatPrice / ORACLE_PRICE_SCALE * WAD / lif < rcfThreshold
+///   <=> collateral * liquidatedCollatPrice / ORACLE_PRICE_SCALE * WAD / lif - maxRepaidUnits < rcfThreshold
```
```suggestion
///       minNewCollateral * liquidatedCollatPrice / LIF < rcfThreshold
///   <=> (collateral - maxRepaid * LIF / liquidatedCollatPrice)
///         * liquidatedCollatPrice / LIF < rcfThreshold
///   <=> collateral * liquidatedCollatPrice / LIF - maxRepaid < rcfThreshold
```
to align with the notations above (LIF is capitalized, we don't use fixed-point arithmetic)

src/Midnight.sol
```diff
@@ -47,6 +47,24 @@ import {EventsLib} from "./libraries/EventsLib.sol";
...
+/// LIQUIDATIONS
+/// @dev Accounts with nonzero debt are liquidatable if they are unhealthy or if the maturity has passed.
+/// @dev If an account is healthy, the LIF grows linearly from 1 at maturity to maxLif at maturity + TIME_TO_MAX_LIF.
+/// @dev Before maturity, the liquidation cannot put the borrower back into health (recovery close factor), unless
+/// the liquidation could leave a collateral with a value that would not be enough to repay rcfThreshold units.
+/// @dev The "recovery close factor" (RCF) limits the amount that can be liquidated. In particular, it prevents the
+/// liquidation from putting the borrower back into health. Which means:
+///   newDebt >= newMaxDebt <=> debtOf - repaidUnits >= maxDebt - repaidUnits*LIF*LLTV
+///                         <=> repaidUnits <= (debtOf-maxDebt) / (1 - LIF*LLTV).
+/// (rounded up to avoid consecutive max liquidations, so the position could be slightly healthy after a liquidation).
```
```suggestion
/// The computation is rounded up to avoid consecutive max liquidations, so the position could be slightly healthy after a liquidation.
```
(style)

src/Midnight.sol
```diff
@@ -47,6 +47,24 @@ import {EventsLib} from "./libraries/EventsLib.sol";
...
+/// @dev The RCF is deactivated for small collateral amount, essentially to avoid liquidations that are too small
```
QGarchery: We should mention that it is deactivated after maturity as well

src/Midnight.sol
```diff
@@ -47,6 +47,24 @@ import {EventsLib} from "./libraries/EventsLib.sol";
...
+/// (compared to the gas cost). More precisely, it is deactivated if the liquidation could leave a collateral with a
```
```suggestion
/// @dev The RCF is deactivated for small collateral amount, essentially to mitigate issues with liquidations that are too small
/// compared to the gas cost. More precisely, it is deactivated if the liquidation could leave a collateral with a
```
because it doesn't prevent small liquidations (someone can still open a very small positions), it mitigates the problems that they create

src/Midnight.sol
peyha: not sure about removing all decimal scaling related term

src/Midnight.sol
```suggestion
/// The maxRepaid computation is rounded up to avoid consecutive max liquidations, so the position could be slightly healthy after a liquidation.
```

src/Midnight.sol
MathisGD: I think that makes sense as we also omit roundings

peyha (approved):

src/Midnight.sol
```diff
@@ -47,6 +47,25 @@ import {EventsLib} from "./libraries/EventsLib.sol";
...
 ///
+/// LIQUIDATIONS
+/// @dev Accounts with nonzero debt are liquidatable if they are unhealthy or if the maturity has passed.
+/// @dev If an account is healthy, the LIF grows linearly from 1 at maturity to maxLif at maturity + TIME_TO_MAX_LIF.
+/// @dev Before maturity, the liquidation cannot put the borrower back into health (recovery close factor), unless
+/// the liquidation could leave a collateral with a value that would not be enough to repay rcfThreshold units.
+/// @dev The "recovery close factor" (RCF) limits the amount that can be liquidated. In particular, it prevents the
+/// liquidation from putting the borrower back into health. Which means (omitting scaling and roundings):
+///   newDebt >= newMaxDebt <=> debtOf - repaidUnits >= maxDebt - repaidUnits*LIF*LLTV
+///                         <=> repaidUnits <= (debtOf-maxDebt) / (1 - LIF*LLTV).
```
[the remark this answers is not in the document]
QGarchery: I think it's fine. This is documentation, not code. And it actually leads to a correct understanding that the max repaid is infinite when LIF == LLTV == 1

QGarchery (approved):

src/Midnight.sol
```diff
@@ -47,6 +47,25 @@ import {EventsLib} from "./libraries/EventsLib.sol";
 /// the pending fee of existing lenders is not updated (=> their fee is fixed).
 /// @dev Absent bad debt, the face value of a lender's position is `credit - pendingFee`.
 ///
+/// LIQUIDATIONS
+/// @dev Accounts with nonzero debt are liquidatable if they are unhealthy or if the maturity has passed.
+/// @dev If an account is healthy, the LIF grows linearly from 1 at maturity to maxLif at maturity + TIME_TO_MAX_LIF.
+/// @dev Before maturity, the liquidation cannot put the borrower back into health (recovery close factor), unless
```
adhusson, line 53: it's more "before or at maturity"

adhusson (commented):
LGTM modulo small nit

PR #748, opened 2026-04-29
perf: constructor (Blackthorn-21)

MathisGD:
we don't really optimize for constructor gas efficiency. Though it's better to not use immutable variables in the constructor in general, as it's not clear if they have already been set of not

https://github.com/sherlock-audit/2026-04-morpho-midnight-apr-6th-2026/issues/21


peyha (approved):

QGarchery (approved):

adhusson (commented):
LGTM

PR #750, opened 2026-04-29
pass initiator to flashloan (Blackthorn-33)

MathisGD:
https://github.com/sherlock-audit/2026-04-morpho-midnight-apr-6th-2026/issues/33

### DIFF
--- a/src/Midnight.sol
+++ b/src/Midnight.sol
@@ -666,7 +666,7 @@ contract Midnight is IMidnight {
         emit EventsLib.FlashLoan(msg.sender, token, assets, callback);
         SafeTransferLib.safeTransfer(token, callback, assets);
         require(
-            IFlashLoanCallback(callback).onFlashLoan(token, assets, data) == CALLBACK_SUCCESS,
+            IFlashLoanCallback(callback).onFlashLoan(token, assets, msg.sender, data) == CALLBACK_SUCCESS,
             WrongFlashLoanCallbackReturnValue()
         );
         SafeTransferLib.safeTransferFrom(token, callback, address(this), assets);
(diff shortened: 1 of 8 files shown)


peyha (approved):
This looks useful

src/Midnight.sol
```diff
@@ -666,7 +666,7 @@ contract Midnight is IMidnight {
         emit EventsLib.FlashLoan(msg.sender, token, assets, callback);
         SafeTransferLib.safeTransfer(token, callback, assets);
         require(
-            IFlashLoanCallback(callback).onFlashLoan(token, assets, data) == CALLBACK_SUCCESS,
+            IFlashLoanCallback(callback).onFlashLoan(token, assets, msg.sender, data) == CALLBACK_SUCCESS,
```
MathisGD, line 669: nb: we already broke the "blue compatibility"

src/Midnight.sol
QGarchery, line 1: I see 2 main reasons why knowing the initiator is useful:
- to interact with the initiator. Let's say you create an arbitrage contract (your goal is that the price are up to date), but it is callable `onFlashLoan` only (because it needs some base funds), but want to remunerate the one who finds the opportunity
- for authorization. You want to gate who is calling the function that calls the callback. The same example above applies, where you want to only allow trusted actors for example

Maybe the second point doesn't matter much for other callbacks (intuitively someone that is authorized by onBehalf has the same trust privilege as onBehalf), but for the first point it may be useful. Main use case I have in ming is if we want to do a pre-liquidation on Midnight, and incentivize the account that triggered it

QGarchery: > Main use case I have in ming is if we want to do a pre-liquidation on Midnight, and incentivize the account that triggered it

so we would need to add the initiator to `onRepay` at least

adhusson: What about the remaining ones (`onLiquidate`, `onBuy`, `onSell`)? Not sure about either.

MathisGD:
https://github.com/sherlock-audit/2026-04-morpho-midnight-apr-6th-2026/issues/33#issuecomment-4381307163

PR #754, opened 2026-04-29
takeAmountLib reverts if offer is untakeable (Blackthorn-26)

MathisGD:
https://github.com/sherlock-audit/2026-04-morpho-midnight-apr-6th-2026/issues/26


src/periphery/TakeAmountsLib.sol
MathisGD, line 17: name could be improved
```suggestion
    function buyerAssetsInvert(address midnight, bytes32 id, Offer memory offer, uint256 targetBuyerAssets)
```

QGarchery, line 1: Unclear whether it's better to revert, or let the bundler skip this offer. 
It's very likely a mistake in the end to add this offer to the bundle. Only "legit" scenario I could think of is if the bundle is front-run by an increase of the trading fee such that the final price is < 0 ...

In any case, it means that this function respects more directly its spec "Returns the number of units to take to get the target buyer assets.", because they are no units in this case so reverting makes sense

QGarchery (approved):

src/periphery/TakeAmountsLib.sol
```diff
@@ -20,7 +20,8 @@ library TakeAmountsLib {
         uint256 offerPrice = TickLib.tickToPrice(offer.tick);
         uint256 tradingFee =
             IMidnight(midnight).tradingFee(id, UtilsLib.zeroFloorSub(offer.obligation.maturity, block.timestamp));
-        uint256 buyerPrice = offer.buy ? offerPrice : offerPrice + tradingFee;
+        uint256 sellerPrice = offer.buy ? offerPrice - tradingFee : offerPrice;
+        uint256 buyerPrice = sellerPrice + tradingFee;
```
adhusson, line 26:
```suggestion
        uint256 buyerPrice = (offer.buy ? offerPrice - tradingFee : offerPrice) + tradingFee;
```


```suggestion
        require(offerPrice >= tradingFee, PriceBelowFee());
        uint256 buyerPrice = offer.buy ? offerPrice : offerPrice + tradingFee;
```

src/periphery/TakeAmountsLib.sol
peyha: the current naming is good imo

src/periphery/TakeAmountsLib.sol
peyha: clearer way to revert

MathisGD: current code is closer to take's code.. not strong stance

QGarchery (approved):

src/periphery/TakeAmountsLib.sol
```diff
@@ -12,6 +12,7 @@ library TakeAmountsLib {
     /// @dev Forward: buyerAssets = offer.buy ? units.mulDivDown(buyerPrice, WAD) : units.mulDivUp(buyerPrice, WAD).
     /// @dev Assumes that id and offer.obligation match.
     /// @dev Reverts if buyerPrice > WAD, because not all buyerAssets are reachable then.
+    /// @dev Reverts if buyerPrice < tradingFee (midnight reverts too).
```
QGarchery:
```suggestion
    /// @dev Reverts if offerPrice < tradingFee in case of a buy offer (midnight reverts too).
```
because it doesn't revert when buyerPrice < tradingFee, if the taker is the buyer
Same below

QGarchery (changes requested):

peyha (approved):

src/periphery/TakeAmountsLib.sol
MathisGD: the comment was worth writing ^^

QGarchery (approved):

PR #761, opened 2026-04-29
document no-op callback (Blackthorn-39)

MathisGD:
https://github.com/sherlock-audit/2026-04-morpho-midnight-apr-6th-2026/issues/39

### DIFF
--- a/src/Midnight.sol
+++ b/src/Midnight.sol
@@ -139,7 +139,8 @@ import {EventsLib} from "./libraries/EventsLib.sol";
 /// MISC
 /// @dev The max amount of totalUnits, collateral, credit, and debt is type(uint128).max (~1e38).
 /// @dev Zero checks are not systematically performed.
-/// @dev No-ops are allowed.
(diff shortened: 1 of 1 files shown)


peyha (approved):

src/Midnight.sol
```diff
@@ -139,7 +139,8 @@ import {EventsLib} from "./libraries/EventsLib.sol";
 /// MISC
 /// @dev The max amount of totalUnits, collateral, credit, and debt is type(uint128).max (~1e38).
 /// @dev Zero checks are not systematically performed.
-/// @dev No-ops are allowed.
+/// @dev No-ops are allowed. In particular, Midnight can call the callback of "filled" (consumed=max) offers through a
+/// no-op take.
```
```suggestion
/// @dev No-ops are allowed. In particular, Midnight can call the callback of offers through a no-op take, even if
/// those offers are "filled" (consumed=max).
```
since the issue is also talking about take in general, not just for filled offers.
(should be well formatted already)

QGarchery (approved):

PR #762, opened 2026-04-29
document stale values (Blackthorn-41)

MathisGD:
https://github.com/sherlock-audit/2026-04-morpho-midnight-apr-6th-2026/issues/41


src/Midnight.sol
```diff
@@ -137,6 +137,9 @@ import {EventsLib} from "./libraries/EventsLib.sol";
 /// @dev When the claimer is set, the old claimer loses the unclaimed fees.
 ///
 /// MISC
+/// @dev position.creditOf is not up to date. One must use updatePositionView to get the up to date credit.
+/// @dev obligationState.tradingFee0-6 are zero before the obligation is created, and might get updated with the default
```
peyha: could potentially be moved to the `trading fee` part but no big deal

peyha (approved):

src/Midnight.sol
```diff
@@ -137,6 +137,9 @@ import {EventsLib} from "./libraries/EventsLib.sol";
 /// @dev When the claimer is set, the old claimer loses the unclaimed fees.
 ///
 /// MISC
+/// @dev position.creditOf is not up to date. One must use updatePositionView to get the up to date credit.
```
```suggestion
/// @dev creditOf is not up to date. One must use updatePositionView to get the up to date credit.
```
because it's position.credit or creditOf(), not position.creditOf

src/Midnight.sol
QGarchery: I would just remove this line. It feels weird to have it here, the reader won't get why it's there and we don't have the same thing for the continuous fee. Also I feel like it's already stated in the TRADING FEE section "A default trading fee (per loan token) is set on new obligations."

QGarchery (approved):

PR #764, opened 2026-04-30
Prevent no-op liq inactive collateral with bitmap (Spearbit-5)

QGarchery:
Alternative to #739 that allows to simplify the specs. Cost is that we need the `getBit` function now, so probably not worth it


certora/specs/BitmapSummaries.spec
```diff
@@ -4,6 +4,7 @@ methods {
     function UtilsLib.setBit(uint128 bitmap, uint256 bit) internal returns (uint128) => summarySetBit(bitmap, bit);
     function UtilsLib.clearBit(uint128 bitmap, uint256 bit) internal returns (uint128) => summaryClearBit(bitmap, bit);
     function UtilsLib.msb(uint128 bitmap) internal returns (uint256) => summaryMsb(bitmap);
+    function UtilsLib.getBit(uint128 bitmap, uint256 bit) internal returns (bool) => summaryGetBit(bitmap, bit);
```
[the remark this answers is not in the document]
QGarchery: Under specified means over-approximation, this is sound !

src/libraries/UtilsLib.sol
```diff
@@ -178,6 +178,10 @@ library UtilsLib {
         return uint128(bitmap & ~(1 << bit));
     }
 
+    function getBit(uint128 bitmap, uint256 bit) internal pure returns (bool) {
+        return bitmap & (1 << bit) != 0;
```
MathisGD, line 182: right?
```suggestion
        return bitmap & (1 << bit);
```

src/Midnight.sol
```diff
@@ -547,7 +547,9 @@ contract Midnight is IMidnight {
         bytes32 id = touchObligation(obligation);
         ObligationState storage _obligationState = obligationState[id];
         Position storage _position = position[id][borrower];
+        uint128 bitmap = _position.activatedCollaterals;
```
MathisGD, line 554:
```suggestion
        uint128 activatedCollaterals = _position.activatedCollaterals;
```

MathisGD (commented):
Both work for me, will let @peyha choose

src/libraries/UtilsLib.sol
adhusson: The typing breaks. 

We can use assembly. Either `and(bitmap, shl(bit, 1))`, but then it can return a `true` value that is not 1. It's OK but weird. Or `and(shr(bit, bitmap), 1)` which normalizes.

adhusson:
Small preference for this one. In principle you can't have collat=0 everywhere and >0 debt, but it still seems cleaner to directly require collateral activation.

Still, couldn't 739 spec be simplified as well by requiring invariant `collat > 0 => bit > 0` ? Not as much as here but still.

src/Midnight.sol
QGarchery: you want to change it everywhere ?

certora/specs/LiquidationBoundedByLIF.spec
```diff
@@ -69,43 +69,17 @@ function summaryMulDivUp(uint256 x, uint256 y, uint256 d) returns uint256 {
...
-
-/// Liquidation profit is bounded by maxLif (seizedAssets input)
-rule liquidationProfitBoundedSeizedAssets(env e, Midnight.Obligation obligation, uint256 collateralIndex, uint256 seizedAssets, address borrower, address receiver, address callback, bytes data) {
-    mathint maxLif = obligation.collateralParams[collateralIndex].maxLif;
-    require data.length == 0, "no callback for prover performance";
-    require maxLif >= WAD(), "maxLif must be at least 1x for profit boundedness (see touchObligation validation and ExactMath.spec)";
-
-    // Soundness: nonZeroCollateralsAreActivated is proven in CollateralBitmap.spec,
-    bytes32 id0 = summaryToId(obligation);
-    requireInvariant nonZeroCollateralsAreActivated(id0, borrower, collateralIndex);
```
QGarchery, line 104: @adhusson 
> Still, couldn't 739 spec be simplified as well by requiring invariant collat > 0 => bit > 0 ? Not as much as here but still.

It was done before here with `requireInvariant nonZeroCollateralsAreActivated`, but this makes this rule differ from the rule above. We could potentially be more concise in 739 spec still, by requiring it only if `seizedAssets > 0`, but it becomes a refactor rather than a real simplification. Not so different in the end though I agree

src/Midnight.sol
MathisGD: maybe. isn't it weird to have two names for the same thing?

src/Midnight.sol
```diff
@@ -551,7 +551,9 @@ contract Midnight is IMidnight {
...
         require(UtilsLib.atMostOneNonZero(repaidUnits, seizedAssets), InconsistentInput());
+        require(UtilsLib.getBit(bitmap, collateralIndex), InactiveCollateral());
```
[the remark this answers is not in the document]
QGarchery: Good thinking, but you can't have collat = 0 everywhere and some debt. This is because at the last interaction that worsens the health, the position was healthy. At that point it also had collat = 0 everywhere and some debt, but this is guaranteed to be unhealthy. Interesting property to try to verify though, adding it to the wish list.

Although it makes me think what's not great about adding this require is that it becomes a bit more annoying to realize the bad debt, since you have to choose a non-zero collateral

src/Midnight.sol
QGarchery: Agree, but I see where it comes from: you want a clear non-technical name for outside, but using "bitmap" is nice in the implementation since then it makes sense to use bitmap operations on it (like `clearBit` for example)

QGarchery: Not clear how to solve it, current version is not that bad since bitmap is only used very locally in the code.
Or we could decide to use collateralBitmap everywhere, both work for me

MathisGD: collateralBitmap everywhere is great

src/Midnight.sol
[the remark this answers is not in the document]
peyha: still looks ok to me, you can do it programmatically by using any bit set in `activatedCollateral`, might be annoying for EOAs though

MathisGD:
replaced by https://github.com/morpho-org/midnight/pull/785

Issue #759, opened 2026-04-29
rename offer.buy


MathisGD:
as it's only use in take and by the taker, it's sometimes misleading...

MathisGD:
at the end we won't do it

PR #766, opened 2026-05-01
Clean up wallet decodable signature tests

adhusson:
remove frontend (it seems this was an oversight in the previous PR, but maybe you decided to keep it at some point?), check hardcoded typehashes, increase back optimizer runs


test/UtilsLibTest.sol
```diff
@@ -158,4 +166,10 @@ contract UtilsLibTest is Test {
         assertApproxEqRel(TickLib.wExp(19 ether), 178482300.96318726092869632 ether, 0.001 ether, "exp(19)");
         assertApproxEqRel(TickLib.wExp(20 ether), 485165195.409790277969936384 ether, 0.001 ether, "exp(20)");
     }
+
+    function testTypehashes() public pure {
+        assertEq(COLLATERAL_PARAMS_TYPEHASH, keccak256(COLLATERAL_PARAMS_TYPE));
+        assertEq(OBLIGATION_TYPEHASH, keccak256(bytes.concat(OBLIGATION_TYPE, COLLATERAL_PARAMS_TYPE)));
+        assertEq(OFFER_TYPEHASH, keccak256(bytes.concat(OFFER_TYPE, COLLATERAL_PARAMS_TYPE, OBLIGATION_TYPE)));
```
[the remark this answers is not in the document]
adhusson: the repo defines the type so its the source of truth

test/frontend/README.md
MathisGD, line 1: we decided to keep the front-end at the end. it's pretty useful to check that things will work

adhusson: imo adding an actual frontend is not worth the added files. It's not even touched by ci. But keeping [FrontendSignatureTest.sol](https://github.com/morpho-org/midnight/pull/766/changes#diff-fa8a5a4249700ed54d36cd5c420369c7f509aed95a30b666f88c7c83e316a6ef) can make sense.

MathisGD: no strong stance

MathisGD (approved):

test/frontend/README.md
QGarchery: I think I like to keep it until the contract is deployed, then we can remove the whole folder. It's useful to test that we didn't break EIP 712 (manually though), for example after #759

adhusson: done, but 
- kept the sol file in normal tests (i assume we will keep it)
- we will need to manually check that frontend from time to time

MathisGD (approved):

PR #767, opened 2026-05-01
Multicollat documentation

bhargavbh:
multicollat documentation.

### DIFF
--- a/src/Midnight.sol
+++ b/src/Midnight.sol
@@ -118,14 +118,14 @@ import {EventsLib} from "./libraries/EventsLib.sol";
 /// @dev In particular, it does not prevent the user from exiting the obligation even when the entry gate is reverting.
 /// @dev The liquidator gate can prevent the user from liquidating borrowers in the obligation (and realizing bad debt).
 ///
-/// TOKEN REQUIREMENTS
+/// TOKEN SAFETY REQUIREMENTS
 /// @dev List of assumptions on tokens that guarantee that Midnight behaves as expected:
-/// - It should be ERC-20 compliant, except that it can omit return values on `transfer` and `transferFrom`.
+/// - It should be ERC-20 compliant, except that it can omit return values on `transfer` and `transferFrom`. In
+/// particular, it should not revert because a transfer is no-op.
 /// - Midnight's balance of the token should only decrease on `transfer` and `transferFrom`.
 /// - It should not re-enter Midnight on `transfer` nor `transferFrom`.
 /// - Midnight must send/receive exactly the requested amount on transfers.
-/// - It should not revert on `transfer` and `transferFrom` if balances and approvals are right.
-/// - It should not revert on no-op transfers.
+/// @dev See LIVENESS for liveness guarantees.
 ///
 /// LIVENESS
 /// @dev If an activated collateral oracle reverts on `price`, `liquidate` reverts unconditionally.
@@ -135,9 +135,10 @@ import {EventsLib} from "./libraries/EventsLib.sol";
 /// @dev If `enterGate.canIncreaseCredit` reverts or returns false, `take` reverts if the buyer's credit increases.
 /// @dev If `enterGate.canIncreaseDebt` reverts or returns false, `take` reverts if the seller's debt increases.
 /// @dev If `liquidatorGate` reverts or returns false on `canLiquidate`, `liquidate` reverts.
-/// @dev If a token pulled by Midnight reverts on `transferFrom` despite balances and approvals being right, `take`,
-/// `repay`, `supplyCollateral`, `liquidate`, and `flashLoan` repayment revert when they need to pull that token.
-/// @dev If a token sent by Midnight reverts on `transfer` despite balances being right, `withdraw`,
+/// @dev If a token pulled by Midnight reverts or returns false on `transferFrom` despite balances and approvals being
+/// right, `take`, `repay`, `supplyCollateral`, `liquidate`, and `flashLoan` repayment revert when they need to pull
+/// that token.
+/// @dev If a token sent by Midnight reverts or returns false on `transfer` despite balances being right, `withdraw`,
 /// `withdrawCollateral`, fee claims, the collateral leg of `liquidate`, and `flashLoan` revert when they need to send
 /// that token.
 /// @dev If a callback reverts or returns something other than `CALLBACK_SUCCESS`, `take`, `repay`, `liquidate`, and
--- a/src/Midnight.sol
+++ b/src/Midnight.sol
@@ -23,6 +23,7 @@ import {EventsLib} from "./libraries/EventsLib.sol";
 
 /// OBLIGATIONS
 /// @dev The following constraints are enforced on obligation creation (in `touchObligation`):
+/// - `maturity <= block.timestamp + 100 years`: the obligation must have a maturity that is not too far in the future.
 /// - `collateralParams.length > 0`: at least one collateral is required.
 /// - `collateralParams.length <= MAX_COLLATERALS` (128): at most 128 collateralParams per obligation.
 /// - Collateral tokens must be non-zero and strictly sorted by address (ascending, no duplicates).
@@ -397,7 +398,6 @@ contract Midnight is IMidnight {
         _obligationState.totalUnits =
             UtilsLib.toUint128(_obligationState.totalUnits + buyerCreditIncrease - sellerCreditDecrease);
 
-        require(buyerPos.pendingFee <= buyerPos.credit, BuyerPendingFeeExceedsCredit());
         if (offer.reduceOnly) {
             require(offer.buy ? buyerCreditIncrease == 0 : sellerDebtIncrease == 0, MakerCreditOrDebtIncreased());
         }
@@ -745,6 +745,7 @@ contract Midnight is IMidnight {
     function touchObligation(Obligation memory obligation) public returns (bytes32) {
         bytes32 id = toId(obligation);
         if (!obligationState[id].created) {
+            require(obligation.maturity <= block.timestamp + 100 * 365 days, MaturityTooFar());
             require(obligation.collateralParams.length > 0, NoCollateralParams());
             require(obligation.collateralParams.length <= MAX_COLLATERALS, TooManyCollateralParams());
             address previousCollateralToken;
--- a/src/Midnight.sol
+++ b/src/Midnight.sol
@@ -34,24 +34,16 @@ import {EventsLib} from "./libraries/EventsLib.sol";
 ///
 /// MULTICOLLATS
 /// @dev An obligation has a fixed set of permitted collaterals, each with its own lltv and maxLif.
-/// A borrower's debt is backed jointly by every collateral they have activated: debt capacity is the sum of each
-/// activated collateral's value, weighted by its lltv (`maxDebt = sum_i (collateral_i * price_i * lltv_i)`). A borrower
-/// can therefore back a single debt position with multiple collaterals at once.
-/// @dev Collateral composition remains changeable post-origination. A borrower can `supplyCollateral` and
-/// `withdrawCollateral` at any time, subject only to an instantaneous health check on withdrawal. In particular, the
-/// borrowers of multicollat obligations can completely change their collateral composition.
+/// A borrower's debt capacity is the sum of each activated collateral's value, weighted by its
+/// lltv (`maxDebt = sum_i (collateral_i * price_i * lltv_i)`).
+/// @dev A borrower can `supplyCollateral` and `withdrawCollateral` at any time, subject only to an instantaneous
+/// health check on withdrawal. In particular, the borrowers of multicollat obligations can completely
+///  change their collateral composition.
 /// @dev Liquidation iterates over all activated collaterals and reverts if any of their oracles reverts (see LIVENESS).
 /// A single reverting oracle blocks liquidation for every borrower with that collateral activated, and a borrower can
 /// activate such a collateral post-incident to block their own liquidation.
-/// @dev Collateral slot order is fixed by ascending token address at obligation creation. Sorted index `i` maps to
-/// `obligation.collateralParams[i]` and to bit `i` of `position.activatedCollaterals`. `isHealthy` and `liquidate`
-/// iterate MSB-first (highest-address collateral first).
-/// @dev `isHealthy` short-circuits as soon as `maxDebt >= debt`, but `liquidate` does not. Consequences:
-/// - `isHealthy` can return true without consulting a higher-bit collateral whose oracle reverts.
-/// - Post-maturity, `isLiquidatable` can return true while `liquidate` reverts on a broken oracle, so
-/// `isLiquidatable == true` does not imply `liquidate` succeeds.
-/// - A borrower can unknowingly reduce their effective health buffer by withdrawing a higher-address collateral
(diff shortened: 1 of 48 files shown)


src/Midnight.sol
```diff
@@ -32,6 +32,29 @@ import {EventsLib} from "./libraries/EventsLib.sol";
...
+/// @dev An obligation fixes its set of permitted collaterals at creation, each with its own lltv and maxLif.
+/// A borrower's debt is backed jointly by every collateral they have activated: debt capacity is the sum of each
+/// activated collateral's value, weighted by its lltv (`maxDebt = sum_i (collateral_i * price_i * lltv_i)`). A borrower
+/// can therefore back a single debt position with multiple collaterals at once.
+/// @dev Collateral composition remains changeable post-origination. A borrower can `supplyCollateral` and
+/// `withdrawCollateral` at any time, subject only to an instantaneous health check on withdrawal. Lenders therefore
+/// underwrite not just
+/// the collateral mix at origination but the borrower's option to recompose toward the riskiest permitted collateral
+/// at any point before maturity.
+/// @dev Liquidation iterates over all activated collaterals and reverts if any of their oracles reverts (see LIVENESS).
```
bhargavbh, line 40: this point overlaps with documentation in Liveness. Can get rid if we want to avoid redundancies.

src/Midnight.sol
```diff
@@ -32,6 +32,29 @@ import {EventsLib} from "./libraries/EventsLib.sol";
...
+/// A single reverting oracle blocks liquidation for every borrower with that collateral activated, and a borrower can
+/// activate such a collateral post-incident to block their own liquidation.
+/// @dev Collateral slot order is fixed by ascending token address at obligation creation. Sorted index `i` maps to
+/// `obligation.collateralParams[i]` and to bit `i` of `position.activatedCollaterals`. `isHealthy` and `liquidate`
+/// iterate MSB-first (highest-address collateral first).
+/// @dev `isHealthy` short-circuits as soon as `maxDebt >= debt`, but `liquidate` does not. Consequences:
+/// - `isHealthy` can return true without consulting a higher-bit collateral whose oracle reverts.
+/// - Post-maturity, `isLiquidatable` can return true while `liquidate` reverts on a broken oracle, so
```
bhargavbh: this point is not needed if we make isLiquidatable internal

src/Midnight.sol
```diff
@@ -32,6 +32,29 @@ import {EventsLib} from "./libraries/EventsLib.sol";
...
+/// MULTICOLLATS
+/// @dev An obligation fixes its set of permitted collaterals at creation, each with its own lltv and maxLif.
+/// A borrower's debt is backed jointly by every collateral they have activated: debt capacity is the sum of each
+/// activated collateral's value, weighted by its lltv (`maxDebt = sum_i (collateral_i * price_i * lltv_i)`). A borrower
+/// can therefore back a single debt position with multiple collaterals at once.
+/// @dev Collateral composition remains changeable post-origination. A borrower can `supplyCollateral` and
+/// `withdrawCollateral` at any time, subject only to an instantaneous health check on withdrawal. Lenders therefore
+/// underwrite not just
+/// the collateral mix at origination but the borrower's option to recompose toward the riskiest permitted collateral
+/// at any point before maturity.
```
MathisGD: I would make this point more general (a borrower can supply/withdraw collaterals at any time, as long as isHealthy passes), and "in particular, in multi-collateral obligations the borrower can change completely the collaterals composition".

"Lenders therefore underwrite not just the collateral mix at origination but the borrower's option to recompose toward the riskiest permitted collateral at any point before maturity" fees unnecessary, also "riskiest collateral" isn't well defined

src/Midnight.sol
```diff
@@ -32,6 +32,29 @@ import {EventsLib} from "./libraries/EventsLib.sol";
...
+/// @dev Liquidation iterates over all activated collaterals and reverts if any of their oracles reverts (see LIVENESS).
+/// A single reverting oracle blocks liquidation for every borrower with that collateral activated, and a borrower can
+/// activate such a collateral post-incident to block their own liquidation.
+/// @dev Collateral slot order is fixed by ascending token address at obligation creation. Sorted index `i` maps to
+/// `obligation.collateralParams[i]` and to bit `i` of `position.activatedCollaterals`. `isHealthy` and `liquidate`
+/// iterate MSB-first (highest-address collateral first).
```
MathisGD: why is that interesting to document?

src/Midnight.sol
```diff
@@ -32,6 +32,29 @@ import {EventsLib} from "./libraries/EventsLib.sol";
 /// @dev Additionally, within a single obligation, a borrower can use at most MAX_COLLATERALS_PER_BORROWER (10)
 /// collaterals simultaneously.
 ///
+/// MULTICOLLATS
+/// @dev An obligation fixes its set of permitted collaterals at creation, each with its own lltv and maxLif.
```
adhusson: I don't think it's necessary to talk about "creation", the data of an obligation always specifies the same thing.

```suggestion
/// @dev An obligation has a fixed set of permitted collaterals, each with its own lltv and maxLif.
```

src/Midnight.sol
```diff
@@ -32,6 +32,29 @@ import {EventsLib} from "./libraries/EventsLib.sol";
...
+/// underwrite not just
+/// the collateral mix at origination but the borrower's option to recompose toward the riskiest permitted collateral
+/// at any point before maturity.
+/// @dev Liquidation iterates over all activated collaterals and reverts if any of their oracles reverts (see LIVENESS).
+/// A single reverting oracle blocks liquidation for every borrower with that collateral activated, and a borrower can
+/// activate such a collateral post-incident to block their own liquidation.
+/// @dev Collateral slot order is fixed by ascending token address at obligation creation. Sorted index `i` maps to
+/// `obligation.collateralParams[i]` and to bit `i` of `position.activatedCollaterals`. `isHealthy` and `liquidate`
+/// iterate MSB-first (highest-address collateral first).
+/// @dev `isHealthy` short-circuits as soon as `maxDebt >= debt`, but `liquidate` does not. Consequences:
```
adhusson: Noting here we have a mix of using backquotes for variable names and not using it [elsewhere](https://github.com/morpho-org/midnight/blob/main/src/Midnight.sol#L51-L67), or [mixing both styles](https://github.com/morpho-org/midnight/blob/main/src/Midnight.sol#L140-L148). Should harmonize.

src/Midnight.sol
bhargavbh: got rid of "riskiest collateral" term. Skipped explicitly mentioning the implications of recomposition to risk assessors.

src/Midnight.sol
bhargavbh: precursor to the point on isHealthy short-circuiting and but liquidate does not. if is is inconsistent with rest of the documentation style, can get rid.

src/Midnight.sol
```diff
@@ -32,6 +32,27 @@ import {EventsLib} from "./libraries/EventsLib.sol";
...
+/// @dev Collateral composition remains changeable post-origination. A borrower can `supplyCollateral` and
+/// `withdrawCollateral` at any time, subject only to an instantaneous health check on withdrawal. In particular, the
+/// borrowers of multicollat obligations can completely change their collateral composition.
+/// @dev Liquidation iterates over all activated collaterals and reverts if any of their oracles reverts (see LIVENESS).
+/// A single reverting oracle blocks liquidation for every borrower with that collateral activated, and a borrower can
+/// activate such a collateral post-incident to block their own liquidation.
+/// @dev Collateral slot order is fixed by ascending token address at obligation creation. Sorted index `i` maps to
+/// `obligation.collateralParams[i]` and to bit `i` of `position.activatedCollaterals`. `isHealthy` and `liquidate`
+/// iterate MSB-first (highest-address collateral first).
+/// @dev `isHealthy` short-circuits as soon as `maxDebt >= debt`, but `liquidate` does not. Consequences:
```
MathisGD: shouldn't we just remove this short circuit?

src/Midnight.sol
```diff
@@ -32,6 +32,27 @@ import {EventsLib} from "./libraries/EventsLib.sol";
...
+/// - `isHealthy` can return true without consulting a higher-bit collateral whose oracle reverts.
+/// - Post-maturity, `isLiquidatable` can return true while `liquidate` reverts on a broken oracle, so
+/// `isLiquidatable == true` does not imply `liquidate` succeeds.
+/// - A borrower can unknowingly reduce their effective health buffer by withdrawing a higher-address collateral
+/// that masked a reverting/malfunctioning lower-address collateral in the instantaneous health check.
+///
```
MathisGD, line 45: should we add something about the fact that maxRepaid * LI isn't constant? so liquidators might prefer to liquidate some collaterals first

src/Midnight.sol
peyha: not very clear what short circuit mean without context so we should either develop or remove imo

src/Midnight.sol
```diff
@@ -32,6 +32,27 @@ import {EventsLib} from "./libraries/EventsLib.sol";
 /// @dev Additionally, within a single obligation, a borrower can use at most MAX_COLLATERALS_PER_BORROWER (10)
 /// collaterals simultaneously.
 ///
+/// MULTICOLLATS
+/// @dev An obligation has a fixed set of permitted collaterals, each with its own lltv and maxLif.
+/// A borrower's debt is backed jointly by every collateral they have activated: debt capacity is the sum of each
+/// activated collateral's value, weighted by its lltv (`maxDebt = sum_i (collateral_i * price_i * lltv_i)`). A borrower
+/// can therefore back a single debt position with multiple collaterals at once.
```
peyha: a bit too verbose to me (especially because it' slightly redundant with L32-33) but not sure if should be removed

src/Midnight.sol
peyha: i think it's fine especially because we wrote the rcf amount formulas

src/Midnight.sol
bhargavbh: if isLiquidatable is public then IMO we should keep the short-circuit point. Else, we can delete it.

src/Midnight.sol
bhargavbh: worth making it explicit: "maxRepaid*LI is not a constant across activated collaterals. Hence, liquidators may have a preference order in which collaterals are liquidated."

src/Midnight.sol
bhargavbh: shortened it. main point to convey is the maxDebt formula.

src/Midnight.sol
```diff
@@ -32,6 +32,19 @@ import {EventsLib} from "./libraries/EventsLib.sol";
...
+/// A borrower's debt capacity is the sum of each activated collateral's value, weighted by its
+/// lltv (`maxDebt = sum_i (collateral_i * price_i * lltv_i)`).
+/// @dev A borrower can `supplyCollateral` and `withdrawCollateral` at any time, subject only to an instantaneous
+/// health check on withdrawal. In particular, the borrowers of multicollat obligations can completely
+///  change their collateral composition.
+/// @dev Liquidation iterates over all activated collaterals and reverts if any of their oracles reverts (see LIVENESS).
+/// A single reverting oracle blocks liquidation for every borrower with that collateral activated, and a borrower can
+/// activate such a collateral post-incident to block their own liquidation.
+/// @dev Liquidator profits, i.e., maxRepaidUnits * (LIF-1) is not a constant across activated collaterals in
+/// RCF-constrained liquidations. Hence, liquidators may have a preference order over collaterals when liquidating.
```
MathisGD: also in non RCF-constrained liquidations, right? if the LIF is different. maybe we can explain both cases

bhargavbh: @MathisGD improved that point:
Liquidator profit (i.e., maxRepaidUnits * (LIF-1)) is not constant across activated collaterals. Hence, liquidators
may have a preference order over collaterals when liquidating. In the case of RCF-constrained liquidations, the
liquidator profits excluding gas costs is given by (LIF-1) * (debt - maxDebt) / (1 - LIF*LLTV), which is increasing in
LIF and decreasing in LLTV of the collateral being liquidated.

adhusson (approved):

src/Midnight.sol
```diff
@@ -33,6 +33,22 @@ import {EventsLib} from "./libraries/EventsLib.sol";
 /// @dev Additionally, within a single obligation, a borrower can use at most MAX_COLLATERALS_PER_BORROWER (10)
 /// collaterals simultaneously.
 ///
+/// MULTICOLLATS
+/// @dev An obligation has a fixed set of permitted collaterals, each with its own lltv and maxLif.
+/// A borrower's debt capacity is the sum of each activated collateral's value, weighted by its
+/// lltv (`maxDebt = sum_i (collateral_i * price_i * lltv_i)`).
```
MathisGD: would remove this as it's more a spec than documenting a particular point of attention
```suggestion
```

MathisGD (approved):

peyha (approved):

src/Midnight.sol
```diff
@@ -33,6 +33,22 @@ import {EventsLib} from "./libraries/EventsLib.sol";
...
+/// health check on withdrawal. In particular, the borrowers of multicollat obligations can completely
+/// change their collateral composition.
+/// @dev Liquidation iterates over all activated collaterals and reverts if any of their oracles reverts (see LIVENESS).
+/// A single reverting oracle blocks liquidation for every borrower with that collateral activated, and a borrower can
+/// activate such a collateral post-incident to block their own liquidation.
+/// @dev Liquidator profit (i.e., maxRepaidUnits * (LIF-1)) is not constant across activated collaterals. Hence,
+/// liquidators may have a preference order over collaterals when liquidating. In the case of RCF-constrained
+/// liquidations, the
+/// liquidator profits excluding gas costs is given by (LIF-1) * (debt - maxDebt) / (1 - LIF*LLTV), which is increasing
+/// in LIF and decreasing in LLTV of the collateral being liquidated.
```
```suggestion
/// @dev Liquidator profit (i.e., maxRepayable * (LIF-1)) is not constant across activated collaterals. Hence,
/// liquidators may have a preference order over collaterals when liquidating.
```
because it is not exactly maxRepaidUnits, it is also constrained by the collateral amount of the position. Same thing for the RCF part, so that's why I think it's simpler to omit it

QGarchery (approved):

src/Midnight.sol
```diff
@@ -33,6 +33,16 @@ import {EventsLib} from "./libraries/EventsLib.sol";
...
 /// collaterals simultaneously.
 ///
+/// MULTICOLLATS
+/// @dev A borrower can `supplyCollateral` and `withdrawCollateral` at any time, subject only to an instantaneous
+/// health check on withdrawal. In particular, the borrowers of multicollat obligations can completely
+/// change their collateral composition.
+/// @dev Liquidation iterates over all activated collaterals and reverts if any of their oracles reverts (see LIVENESS).
+/// A single reverting oracle blocks liquidation for every borrower with that collateral activated, and a borrower can
+/// activate such a collateral post-incident to block their own liquidation.
+/// @dev Liquidator profit (i.e., maxRepayable * (LIF-1)) is not constant across activated collaterals. Hence,
```
```suggestion
/// @dev The oracle-quoted liquidator incentive (i.e., maxRepayable * (LIF-1)) might not be constant across activated collaterals. Hence,
```

MathisGD (approved):

PR #768, opened 2026-05-01
misc smallupdates

adhusson:
- more precise rcf comment on at maturity liq
- remove wrong comment about stranded obligations after hard fork


MathisGD (approved):

peyha (approved):
good catch

QGarchery (approved):

PR #772, opened 2026-05-04
inline isLiquidatable

### LINKED ISSUE #760: inline isLiquidatable
because it hides multiple behaviors in take...

MathisGD:
rationale:
- it "hides" some stuff in take (notably why it's not possible to borrow after the maturity)
- not having to document the spec of isLiquidatable reverts vs liquidate reverts (~~~as we did it anyway)

keeps the liquidationLock check in take for [this](https://github.com/morpho-org/midnight/pull/772#discussion_r3189114689).

didn't put the require in the if above to have a clear mirror with the require in liquidate.

fixes #760

### DIFF
--- a/src/Midnight.sol
+++ b/src/Midnight.sol
@@ -436,12 +436,14 @@ contract Midnight is IMidnight {
                 WrongSellCallbackReturnValue()
             );
         }
-        if (!wasLocked) UtilsLib.tExchange(LIQUIDATION_LOCK_SLOT, id, seller, false);
-        require(
-            position[id][seller].debt == 0
-                || (block.timestamp <= offer.obligation.maturity && isHealthy(offer.obligation, id, seller)),
-            SellerIsLiquidatable()
-        );
+        if (!wasLocked) {
+            UtilsLib.tExchange(LIQUIDATION_LOCK_SLOT, id, seller, false);
+            require(
+                position[id][seller].debt == 0
+                    || (block.timestamp <= offer.obligation.maturity && isHealthy(offer.obligation, id, seller)),
+                SellerIsLiquidatable()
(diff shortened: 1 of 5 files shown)


adhusson:
why `isLiquidatable` non-public / inlined? last I remember the idea was that it made for a clean spec

src/Midnight.sol
```diff
@@ -435,7 +435,11 @@ contract Midnight is IMidnight {
             );
         }
         if (!wasLocked) UtilsLib.tExchange(LIQUIDATION_LOCK_SLOT, id, seller, false);
-        require(!isLiquidatable(offer.obligation, id, seller), SellerIsLiquidatable());
+        require(
+            position[id][seller].debt == 0
+                || (block.timestamp <= offer.obligation.maturity && isHealthy(offer.obligation, id, seller)),
+            SellerIsLiquidatable()
+        );
```
adhusson, line 452: should be inside !wasLocked

MathisGD: why? do we really need that the condition here is the same as the liquidatable condition?

adhusson: it keeps the nice use case highlighted by quentin here: https://morpholabs.slack.com/archives/C08A40KKN9W/p1776175303070179

MathisGD: let's discuss live

src/Midnight.sol
```suggestion
/// @dev Accounts are liquidatable only if the liquidation is not locked and they are either unhealthy or the maturity has passed.
/// @dev Liquidations can revert for other reasons, see LIVENESS.
```

QGarchery (approved):

adhusson (approved):

PR #776, opened 2026-05-04
[Certora] tweaks of MakerProtection

QGarchery:
[Corresponding thread](https://morpholabs.slack.com/archives/C0ACSA8TNMA/p1774951517402319)

Notes:
- removed `zeroUnitsTakeResultsInZeroAssets` as it is already proven in Midnight.spec:takeInputOutputConsistency
- removed `to_mathint`
- removed some summaries, explained others. Notably explains why we are assuming no reentrancy
- removed the ghost trading fee, using the real one instead, before the call to `take`
- changed `toId` to use a more concrete summary
- it might make sense to review this PR up to bbb31a814a0d1e9b3d11f752f99476e2f314d5e0, and then this commit separately, as it renames the files. Since `zeroUnitsTakeResultsInZeroAssets` was removed, this spec now talks about the spread that the trading fee creates: the spread is created by changing the taker price, the maker price doesn't change + what is the value of the spread


certora/specs/TradingFeeSpread.spec
```diff
@@ -0,0 +1,64 @@
...
+    assert offer.buy => buyerAssets * WAD() <= units * offerPrice;
+    assert !offer.buy => sellerAssets * WAD() >= units * offerPrice;
+}
+
+// The spread between what the buyer pays and what the seller receives is at least floor(units * fee / WAD) and at most ceil(units * fee / WAD).
+rule tradingFeeSpreadBounds(env e, uint256 units, address taker, address takerCallback, bytes takerCallbackData, address receiver, Midnight.Offer offer, bytes ratifierData, bytes32 root, bytes32[] proof) {
+    uint256 timeToMaturity = e.block.timestamp <= offer.obligation.maturity ? assert_uint256(offer.obligation.maturity - e.block.timestamp) : 0;
+
+    bytes32 id = summaryToId(offer.obligation);
+    uint256 fee = tradingFee(id, timeToMaturity);
```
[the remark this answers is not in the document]
QGarchery: This is not an issue, we can assume that the obligation has been created

lilCertora (approved):
LGTM

certora/specs/Reverts.spec
```diff
@@ -350,28 +350,43 @@ filtered {
...
     f@withrevert(e, args);
     assert lastReverted;
 }
 
-/// If transfer reverts, withdraw, withdrawCollateral, fee claims, liquidate, and flashLoan all revert.
+/// If transferFrom reverts, flashLoan reverts, assuming that the arrays are not empty.
+rule transferFromRevertPropagationFlashLoan(env e, address[] tokens, uint256[] assets, address callback, bytes data) {
+    require forceTransferFromRevert, "transferFrom reverts";
+    require tokens.length > 0, "assume tokens array is not empty";
+    flashLoan@withrevert(e, tokens, assets, callback, data);
```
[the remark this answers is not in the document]
QGarchery: It just considers more cases where the arrays can have different length, so this can only be sound

bhargavbh (approved):

certora/specs/TradingFeeSpread.spec
```diff
@@ -0,0 +1,62 @@
...
+    // Deterministic TickLib.tickToPrice summary to be able to reference the price in the rules.
+    function TickLib.tickToPrice(uint256 tick) internal returns (uint256) => summaryTickToPrice(tick);
+
+    // Sound summary since toObligation is not used by the protocol.
+    function IdLib.storeInCode(Midnight.Obligation memory, uint256) internal returns (address) => NONDET;
+
+    // Over-approximate view functions for prover performance.
+    function isHealthy(Midnight.Obligation memory, bytes32, address) internal returns (bool) => NONDET;
+
+    // Assume no reentrancy, because we need to know that the trading fee won't change in the onRatify callback. This allows to reference the trading fee in the rule tradingFeeSpreadBounds.
```
MathisGD, line 23: where is that assumed actually?

MathisGD (approved):

certora/specs/TradingFeeSpread.spec
QGarchery: It's assumed automatically, as `AUTO` will summarize external non-view calls as `HAVOC_ECF` which means that it can't modify the caller contract (Midnight in this case)

jhoenicke (approved):
The rules look good.  But they don't protect against taker paying too much or receiving too little.

Does the maker pay exactly floor/receive exactly ceil?   This together with the upper bound of the spread would ensure that taker pays no more than ceil(units*price) + ceil(units*fee).

MathisGD:
@jhoenicke the taker-buyer (resp. taker-seller) can pay as much as infinite price with units=0 and assets=1 (resp. price=0 with assets=0 and units=1). The bound depends on the quantity

certora/specs/TradingFeeSpread.spec
MathisGD: it's curious, it's not the soundest thing

QGarchery:
adding your suggestion in the wish list @jhoenicke

PR #785, opened 2026-05-05
Add clarification to liquidate function documentation (Spearbit-5)

MathisGD:


QGarchery (approved):

src/Midnight.sol
```diff
@@ -544,6 +544,8 @@ contract Midnight is IMidnight {
     /// @dev See LIQUIDATIONS section for more details.
     /// @dev At least one of `seizedAssets` or `repaidUnits` should be equal to zero.
     /// @dev Passing both 0 for `seizedAssets` and `repaidUnits` allows to realize bad debt with 0 token transferred.
+    /// @dev Liquidations with both 0 for `seizedAssets` and `repaidUnits` can be done with a collateral that is not
+    /// activated.
```
adhusson, line 548:
```suggestion
/// @dev Passing both 0 for `seizedAssets` and `repaidUnits` with any collateral index allows to realize bad debt with 0 token transferred.
```

MathisGD: I think that we want 2 comments, because there are really 2 separated points

peyha (approved):

src/Midnight.sol
adhusson: A bit overkill imo but no strong stance.

adhusson: (it's a special point for us because we thought about it, but the reader doesn't care he just needs to know how to realize bad debt)

MathisGD: the second event is targeted at the event listener, while the first one is targeted at the bad debt realizator

PR #789, opened 2026-05-05
document computation in TakeAmountsLib

MathisGD:


src/periphery/TakeAmountsLib.sol
QGarchery, line 1: Is it needed now that we have the comment "Reverts if offerPrice < tradingFee in case of a buy offer (midnight reverts too)." "

MathisGD: @adhusson wanted it, no strong take

QGarchery (approved):

src/periphery/TakeAmountsLib.sol
QGarchery: No strong take either, so I'm approving

adhusson: I find the shape weird enough (`a = x-y; b = a+y;` never use `a` again) to justify explaining what's going on

adhusson (approved):

PR #791, opened 2026-05-05
naming collateralBitmap

MathisGD:


src/interfaces/IMidnight.sol
```diff
@@ -103,7 +103,7 @@ interface IMidnight {
     error SellerGatedFromIncreasingDebt();
     error SellerIsLiquidatable();
     error TakerUnauthorized();
-    error TooManyActivatedCollaterals();
+    error TooManyCollateralBitmap();
```
QGarchery: Not very readable, I would revert this change

test/OtherFunctionsTest.sol
```diff
@@ -376,7 +376,7 @@ contract OtherFunctionsTest is BaseTest {
         midnight.withdrawCollateral(obligationWithRevertingOracle, 0, collateral, borrower, borrower);
     }
 
-    function testIsHealthyDoesNotShortCircuitActivatedCollaterals() public {
+    function testIsHealthyDoesNotShortCircuitCollateralBitmap() public {
```
QGarchery: To revert as well

certora/specs/CreatedObligations.spec
```diff
@@ -179,7 +179,7 @@ function noTradingFeesAreSet(bytes32 id) returns (bool) {
     return fees[0] == 0 && fees[1] == 0 && fees[2] == 0 && fees[3] == 0 && fees[4] == 0 && fees[5] == 0 && fees[6] == 0;
 }
 
-definition userHasNoActivatedCollaterals(bytes32 id, address user) returns bool = currentContract.position[id][user].activatedCollaterals == 0;
+definition userHasNoCollateralBitmap(bytes32 id, address user) returns bool = currentContract.position[id][user].collateralBitmap == 0;
```
QGarchery:
```suggestion
definition userHasEmptyCollateralBitmap(bytes32 id, address user) returns bool = currentContract.position[id][user].collateralBitmap == 0;
```

test/OtherFunctionsTest.sol
MathisGD: removed this test, it was path dependance

test/OtherFunctionsTest.sol
```diff
@@ -376,53 +376,7 @@ contract OtherFunctionsTest is BaseTest {
         midnight.withdrawCollateral(obligationWithRevertingOracle, 0, collateral, borrower, borrower);
     }
 
-    function testIsHealthyDoesNotShortCircuitActivatedCollaterals() public {
```
adhusson, line 379: why remove this?

adhusson (approved):

test/OtherFunctionsTest.sol
MathisGD: it's clear that it doesn't short circuit, the test was more path dependance

QGarchery (approved):

PR #793, opened 2026-05-06
bundler forces same obligation

MathisGD:
at the end we remove the multi obligation take because 1) it's a narrow use-case, you go a long way with async cross-obligation routing and 2) it create issues with collateral management (similar to partial fills), see #730

### DIFF
--- a/test/BundlerTest.sol
+++ b/test/BundlerTest.sol
@@ -227,6 +227,74 @@ contract BundlerTest is BaseTest {
         }
     }
 
+    function testBuyUnitsTargetInconsistentObligation() public {
+        Obligation memory otherObligation = obligation;
+        otherObligation.maturity = block.timestamp + 360 days;
+
+        offers[0].buy = false;
+        offers[0].maker = borrower;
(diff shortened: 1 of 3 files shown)


test/BundlerTest.sol
```diff
@@ -227,69 +227,45 @@ contract BundlerTest is BaseTest {
         }
     }
 
-    function testBuyBuyerAssetsTargetUsesEachOfferObligationId() public {
-        uint256 tick = TickLib.priceToTick(0.99e18);
-        uint256 offerPrice = TickLib.tickToPrice(tick);
-
+    function testBuyBuyerAssetsTargetInconsistentObligation() public {
```
QGarchery, line 306: We could have the same test for `buyUnitsTarget` and `sellUnitsTarget`

QGarchery (approved):

QGarchery (approved):

adhusson (approved):

peyha (commented):
lgtm

PR #799, opened 2026-05-06
save gas in offerTreeTypeHash

adhusson:
Gas comparison
- height 0: 1700 → 80.
- height 5: 3000 → 190.
- height 10: 4300 →80.

Also:
- Kept unused constants for ease of integration
- Wanted to pick the height=19 threshold (500k) because after that contract size is bigger.
- The "tree style if" saves additional gas on average vs. linear if, but it is also necessary, because due to inlining of pure functions a linear if with height 19 creates a stack too deep issue in callers.


src/libraries/UtilsLib.sol
```diff
@@ -53,12 +47,27 @@ library UtilsLib {
...
+        if (height == 11) return 0x93f78fd001e8bafca3563f60c2f0f17b7a4eae11cd7f2009af2488b318b32e30;
+        if (height == 12) return 0x7f4b555dd8cdba4396e42215d878847ba64c29ea68d852161264ccd43008e2df;
+        if (height == 13) return 0x3b3c022990f3110616cb95bee08380dee66e3a47a67d9c2c7ec1828102367957;
+        if (height == 14) return 0x2519b2c8c6476345d46d5f543ee11f421886319e2d68f2df4ec54344393d402f;
+        if (height == 15) return 0x342d2f85060d6c90578ac9783b38ba80e002837fc89448ae771ac246d4c68ce3;
+        if (height == 16) return 0xe3ff3face5c6307122b29d3f23b7106ab79793236005007af9e94bbc4d9836a7;
+        if (height == 17) return 0x9a955710808f486fc37d0607e120c223cc8b2029dab35c6ec7840ff9aea1ff52;
+        if (height == 18) return 0x419a52abde7cb2ff4d4eeb2f65b93dc2e05a12ea6f7a9c379b0b25558c34741f;
+        if (height == 19) return 0x9cf0a4c02fe1d55648b3c157e49872c82d1bc404e3c5359c02aba6cb0d934ddb;
+        revert InvalidOfferTreeHeight();
```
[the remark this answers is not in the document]
adhusson: i think it's ok to fail above 500k offer trees

src/libraries/UtilsLib.sol
QGarchery, line 1: The gas cost if we go back to the [initial implementation](https://github.com/morpho-org/midnight/blob/47b2bee499040d9d0bd485f3ad716b0526c38fab/src/libraries/UtilsLib.sol#L48-L56) should be close to 2k for all heights (untested), it may be worth it

adhusson: For me given

height 0: 1700 → 80.
height 5: 3000 → 190.
height 10: 4300 →80.

its worth doing

src/libraries/UtilsLib.sol
```diff
@@ -53,12 +47,29 @@ library UtilsLib {
     }
 
     function offerTreeTypeHash(uint256 height) internal pure returns (bytes32) {
-        bytes memory offerTreeType = "OfferTree(Offer";
-        for (uint256 i = 0; i < height; i++) {
-            offerTreeType = bytes.concat(offerTreeType, "[2]");
+        if (height < 10) {
```
MathisGD: it should be the opposite (skip to 10 if >10) right? btw we could also not have this imo

adhusson: > it should be the opposite (skip to 10 if >10) right?

not clear what you mean, for gas?

src/ratifiers/MerkleLib.sol
```diff
@@ -2,18 +2,35 @@
 // Copyright (c) 2025 Morpho Association
 pragma solidity ^0.8.0;
 
-import {COLLATERAL_PARAMS_TYPE, OBLIGATION_TYPE, OFFER_TYPE} from "../libraries/ConstantsLib.sol";
-
 /// @dev Helpers for verifying signed Merkle trees of offers.
 library MerkleLib {
+    error InvalidOfferTreeHeight();
```
MathisGD:
```suggestion
    error TreeTooHigh();
```

src/ratifiers/MerkleLib.sol
```diff
@@ -2,18 +2,35 @@
...
+
     /// @dev Returns the EIP-712 typehash of OfferTree(Offer[2]...[2] offerTree) with height levels.
```
MathisGD, line 9:
```suggestion
    /// @dev Returns the EIP-712 typehash of OfferTree(Offer[2]...[2] offerTree) with height levels.
    /// @dev Same as keccak256(bytes.concat("OfferTree(Offer[2]...[2] offerTree)", COLLATERAL_PARAMS_TYPE, OBLIGATION_TYPE, OFFER_TYPE))
```

src/ratifiers/MerkleLib.sol
```diff
@@ -2,18 +2,35 @@
...
+            if (height == 1) return 0x5d2cda016e26c76fc76abd14730b171e5728b985396fdd915f578684ebbe7c08;
+            if (height == 2) return 0xefac405c5f69bfadf16128c7517214821f43ed43418cb105786797e775c82208;
+            if (height == 3) return 0x31c93107b446078165ed61fdc00f9d9fcd4f3caaa1190a450d2ea6f5458eaccf;
+            if (height == 4) return 0xe0dee184a8590b1f885af247ca09e36c68586dbffa6867fa9b3b59558beffc15;
+            if (height == 5) return 0xd6e86cf62241738635bac0675446526dba2111139cd29cbe51540d1951f0f9d9;
+            if (height == 6) return 0x60b0a7b9b8026fb0843b29674731a80da02604c07e87380acba1e834127c036d;
+            if (height == 7) return 0xb57b5627fd51b740272bee7750d8a0f52856db5761767bbc90795d1166400f91;
+            if (height == 8) return 0xd9ea3b8d089568ca3a3e0858ec120441ef7ac3714712f451d18de174961ac92f;
+            return 0x7b1ceacb8b39729bf3887cec726d34a4b6944eb84eb8e9cdfe0d7070ad75ca30;
         }
```
MathisGD, line 38:
```suggestion
        } else {
```

src/ratifiers/MerkleLib.sol
```diff
@@ -2,18 +2,35 @@
...
 /// @dev Helpers for verifying signed Merkle trees of offers.
 library MerkleLib {
+    error InvalidOfferTreeHeight();
+
     /// @dev Returns the EIP-712 typehash of OfferTree(Offer[2]...[2] offerTree) with height levels.
     function offerTreeTypeHash(uint256 height) internal pure returns (bytes32) {
-        bytes memory offerTreeType = "OfferTree(Offer";
-        for (uint256 i = 0; i < height; i++) {
-            offerTreeType = bytes.concat(offerTreeType, "[2]");
+        if (height < 10) {
```
MathisGD: gas to try
```suggestion
        bytes32[10] typeHashArray = [...];
        return typeHashArray[height];
```

MathisGD (approved):

src/ratifiers/MerkleLib.sol
```diff
@@ -2,18 +2,35 @@
...
+        if (height == 10) return 0x56d6ab434a3896c1d1c814fe3e7c22fc7c1f9b9177a50f2f79e754bb2199cdb0;
+        if (height == 11) return 0x93f78fd001e8bafca3563f60c2f0f17b7a4eae11cd7f2009af2488b318b32e30;
+        if (height == 12) return 0x7f4b555dd8cdba4396e42215d878847ba64c29ea68d852161264ccd43008e2df;
+        if (height == 13) return 0x3b3c022990f3110616cb95bee08380dee66e3a47a67d9c2c7ec1828102367957;
+        if (height == 14) return 0x2519b2c8c6476345d46d5f543ee11f421886319e2d68f2df4ec54344393d402f;
+        if (height == 15) return 0x342d2f85060d6c90578ac9783b38ba80e002837fc89448ae771ac246d4c68ce3;
+        if (height == 16) return 0xe3ff3face5c6307122b29d3f23b7106ab79793236005007af9e94bbc4d9836a7;
+        if (height == 17) return 0x9a955710808f486fc37d0607e120c223cc8b2029dab35c6ec7840ff9aea1ff52;
+        if (height == 18) return 0x419a52abde7cb2ff4d4eeb2f65b93dc2e05a12ea6f7a9c379b0b25558c34741f;
+        if (height == 19) return 0x9cf0a4c02fe1d55648b3c157e49872c82d1bc404e3c5359c02aba6cb0d934ddb;
```
MathisGD: stopping at 16 would be fine

MathisGD: let's keep 20

MathisGD (approved):

src/ratifiers/MerkleLib.sol
```diff
@@ -2,18 +2,38 @@
...
+    error TreeTooHigh();
+
     /// @dev Returns the EIP-712 typehash of OfferTree(Offer[2]...[2] offerTree) with height levels.
+    /// @dev Same as keccak256(bytes.concat("OfferTree(Offer[2]...[2] offerTree)", COLLATERAL_PARAMS_TYPE,
+    /// OBLIGATION_TYPE, OFFER_TYPE))
     function offerTreeTypeHash(uint256 height) internal pure returns (bytes32) {
-        bytes memory offerTreeType = "OfferTree(Offer";
-        for (uint256 i = 0; i < height; i++) {
-            offerTreeType = bytes.concat(offerTreeType, "[2]");
+        if (height < 10) {
```
QGarchery: We have 9 `if` in the branch `height < 10` while we have 10 `if` in the else branch. Could we add one to the if branch, which will also make so the max height is 20 included (more round, ~1M offers)

src/ratifiers/MerkleLib.sol
```diff
@@ -2,18 +2,38 @@
...
 
-import {COLLATERAL_PARAMS_TYPE, OBLIGATION_TYPE, OFFER_TYPE} from "../libraries/ConstantsLib.sol";
-
 /// @dev Helpers for verifying signed Merkle trees of offers.
 library MerkleLib {
+    error TreeTooHigh();
+
     /// @dev Returns the EIP-712 typehash of OfferTree(Offer[2]...[2] offerTree) with height levels.
+    /// @dev Same as keccak256(bytes.concat("OfferTree(Offer[2]...[2] offerTree)", COLLATERAL_PARAMS_TYPE,
+    /// OBLIGATION_TYPE, OFFER_TYPE))
```
QGarchery: let's also mention that it reverts if the height is too big

src/ratifiers/MerkleLib.sol
QGarchery: contract size does not matter because this is called in the ratifiers

adhusson: note 19 was chosen as the limit after which contract size became higher, but it doesn't really matter here

QGarchery (approved):

MathisGD (approved):

PR #805, opened 2026-05-06
backquotes and clean formatting

### LINKED ISSUE #770: Harmonize documentation style re: backquotes
> Noting here we have a mix of using backquotes for variable names and not using it [elsewhere](https://github.com/morpho-org/midnight/blob/main/src/Midnight.sol#L51-L67), or [mixing both styles](https://github.com/morpho-org/midnight/blob/main/src/Midnight.sol#L140-L148). Should harmonize. 

 _Originally posted by @adhusson in [#767](https://github.com/morpho-org/midnight/pull/767/changes#r3176813376)_

MathisGD:
fixes #770

### DIFF
--- a/src/Midnight.sol
+++ b/src/Midnight.sol
@@ -22,14 +22,14 @@ import {IEnterGate, ILiquidatorGate} from "./interfaces/IGate.sol";
 import {EventsLib} from "./libraries/EventsLib.sol";
 
 /// OBLIGATIONS
-/// @dev The following constraints are enforced on obligation creation (in `touchObligation`):
-/// - `maturity <= block.timestamp + 100 years`: the obligation must have a maturity that is not too far in the future.
-/// - `collateralParams.length > 0`: at least one collateral is required.
-/// - `collateralParams.length <= MAX_COLLATERALS` (128): at most 128 collateralParams per obligation.
+/// @dev The following constraints are enforced on obligation creation (in touchObligation):
+/// - maturity <= block.timestamp + 100 years: the obligation must have a maturity that is not too far in the future.
+/// - collateralParams.length > 0: at least one collateral is required.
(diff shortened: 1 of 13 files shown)


QGarchery (approved):

peyha (approved):
not a huge fan of backquote in comment but like it to highlight formulas such as `filledBuyerAssets + filledBuyerAssets * pct / (WAD - pct)`, it stays readable enough though

src/Midnight.sol
```diff
@@ -306,8 +305,8 @@ contract Midnight is IMidnight {
     /// to perform atomic price checks.
     /// @dev Taking buy offers with price < trading fee will revert.
     /// @dev In particular, if the trading fee gets increased, it might implicitely cancel offers with very low price.
-    /// @dev All sellerAssets are reachable with the units input, and all buyerAssets are reachable only if
-    /// buyerPrice <= WAD.
+    /// @dev All sellerAssets are reachable with the units input, and all buyerAssets are reachable only if buyerPrice
+    /// <= WAD.
```
QGarchery, line 309: A bit unfortunate here

QGarchery: Removing wrapping will fix this :)

MathisGD: Let's open a PR to see

src/libraries/UtilsLib.sol
```diff
@@ -115,7 +115,7 @@ library UtilsLib {
     }
 
     /// @dev Computes the EIP-712 hash struct of an Offer.
-    /// @dev Split into two `abi.encode`s to avoid stack-too-deep under via-ir without optimizer.
+    /// @dev Split into two abi.encodes to avoid stack-too-deep under via-ir without optimizer.
```
adhusson, line 118:
```suggestion
    /// @dev Split into two abi.encode calls to avoid stack-too-deep under via-ir without optimizer.
```

PR #813, opened 2026-05-07
permit/permit2

### LINKED ISSUE #804: bundler: permit/permit2

MathisGD:
fixes #804 

hardcodes permit2 address as it's the same everywhere (even abstract, it seems). and if a chain doesn't have it here, then no contract would be at this address so there is no risk


src/periphery/interfaces/ITakeBundler.sol
```diff
@@ -17,19 +17,43 @@ struct CollateralTransfer {
...
+/// is fixed by the surrounding call (`maxBuyerAssets` / `targetBuyerAssets` / `collateralSupplies[i]`).
+/// - None: no permit is run; the bundler pulls via a standard ERC20 `transferFrom`. `data` is ignored.
+/// - ERC2612: `data = abi.encode(uint256 deadline, uint8 v, bytes32 r, bytes32 s)`. Bundler calls
+///   `token.permit(owner=msg.sender, spender=this, value=<pull amount>, deadline, v, r, s)`, then
+///   performs a standard ERC20 pull. Permit reverts are tolerated to handle a third party having
+///   already consumed the permit.
+/// - Permit2: `data = abi.encode(uint256 nonce, uint256 deadline, bytes signature)`. Bundler calls
+///   `Permit2.permitTransferFrom(permit{token, amount=<pull amount>, nonce, deadline},
+///   transferDetails{to=this, requestedAmount=<pull amount>}, owner=msg.sender, signature)`. The
+///   call IS the pull, so no separate transfer step is performed. Reverts propagate.
```
MathisGD:
```suggestion
```

src/periphery/MidnightBundles.sol
```diff
@@ -292,12 +303,38 @@ contract TakeBundler is ITakeBundler {
...
     function _forceApproveMax(address token, address spender) internal {
         if (IERC20(token).allowance(address(this), spender) >= type(uint96).max / 2) return;
         _safeApprove(token, spender, 0);
         _safeApprove(token, spender, type(uint256).max);
     }
+
+    /// @dev Pulls `amount` of `token` from `from` to this bundler, optionally using ERC2612 or Permit2.
+    /// @dev `permit.kind == PermitKind.None` skips the permit and falls back to a standard ERC20 pull.
+    /// @dev The signed value (ERC2612) and `permitted.amount` (Permit2) are both bound to `amount`, so the
+    /// signature must authorize exactly the bundler's pull.
```
adhusson:
```suggestion
    /// @dev Pulls `amount` of `token` from `from` to this bundler.
    /// @dev Can apply an ERC-2612 permit or a Permit2 transfer permit before the transfer.
    /// @dev The signed permit values must be for exactly the amount pulled by the bundler.
```

MathisGD: >Can apply an ERC-2612 permit or a Permit2 transfer permit before the transfer.

this is not exact

MathisGD:
```suggestion
    /// @dev Pulls `amount` of `token` from `from` to this bundler, optionally using ERC2612 or Permit2.
    /// @dev The signed value (ERC2612) and `permitted.amount` (Permit2) are both bound to `amount`, so the
    /// signature must authorize exactly the bundler's pull.
```

adhusson: > this is not exact

?

.github/workflows/forge.yml
```diff
@@ -33,7 +33,9 @@ jobs:
         uses: foundry-rs/foundry-toolchain@8789b3e21e6c11b2697f5eb56eddae542f746c10 # v1
 
       - name: Run forge lint
-        run: forge lint src test script certora --deny notes
+        run: |
+          forge lint --deny notes
+          forge lint certora --deny notes
```
MathisGD:
```suggestion
        run: forge lint src test script certora --deny notes
```

src/periphery/MidnightBundles.sol
```diff
@@ -95,17 +101,19 @@ contract TakeBundler is ITakeBundler {
...
         require(referralFeePct < WAD, PctExceeded());
+        require(collateralPermits.length == collateralSupplies.length, InvalidPermitArrayLength());
         address loanToken = takes[0].offer.obligation.loanToken;
         bytes32 id = IMidnight(midnight).toId(takes[0].offer.obligation);
 
         Obligation memory obligation = takes[0].offer.obligation;
         for (uint256 i; i < collateralSupplies.length; i++) {
             address token = obligation.collateralParams[collateralSupplies[i].collateralIndex].token;
-            SafeTransferLib.safeTransferFrom(token, msg.sender, address(this), collateralSupplies[i].assets);
+            _pullToken(token, msg.sender, collateralSupplies[i].assets, collateralPermits[i]);
```
adhusson: with permit2 we could require a single signature and transfer everything at once using `PermitBatchTransferFrom`

.github/workflows/forge.yml
```diff
@@ -33,7 +33,7 @@ jobs:
         uses: foundry-rs/foundry-toolchain@8789b3e21e6c11b2697f5eb56eddae542f746c10 # v1
 
       - name: Run forge lint
-        run: forge lint src test script certora --deny notes
+        run: forge lint src test script certora
```
adhusson: Not what we want. At least we want `--deny warnings`.

Then this will fail on `Permit2.sol` because the lint `ignore` field of foundry.toml will get ignored, and we have 2 solutions:

- either separate the auto forge lint (which respects `ignore`) from the certora lint:

```bash
forge lint --deny notes
forge lint certora --deny notes
```

- or add forge lint silencers inside Permit2.sol

src/periphery/MidnightBundles.sol
MathisGD: Good point. It it worth having two pullTokens though?

.github/workflows/forge.yml
MathisGD: i reverted everything

src/periphery/MidnightBundles.sol
```diff
@@ -342,4 +310,44 @@ contract MidnightBundles is IMidnightBundles {
...
+                    signature
+                );
+        } else {
+            SafeTransferLib.safeTransferFrom(token, from, address(this), amount);
+        }
+    }
+
+    /// @dev External self-call helper. Reverts unless invoked by the contract itself.
+    /// @dev Wraps `IMidnight.take` in try/catch and returns zeros on revert. Routed through `this.tryTake`
+    /// so the take call's wide argument list is not inlined into the trading functions.
```
MathisGD:
```suggestion
```

src/periphery/MidnightBundles.sol
```diff
@@ -342,4 +310,41 @@ contract MidnightBundles is IMidnightBundles {
...
+    {
+        require(msg.sender == address(this), Unauthorized());
+        try IMidnight(midnight).take(units, taker, address(0), "", receiver, take_.offer, take_.ratifierData) returns (
+            uint256 b, uint256 s, uint256 u
+        ) {
+            return (b, s, u);
+        } catch {
+            return (0, 0, 0);
+        }
+    }
```
MathisGD: let's try to remove this

src/periphery/MidnightBundles.sol
```diff
@@ -25,39 +30,40 @@ contract MidnightBundles is IMidnightBundles {
         uint256 targetUnits,
         uint256 maxBuyerAssets,
         address taker,
-        Take[] calldata takes,
-        CollateralTransfer[] calldata collateralWithdrawals,
+        Take[] memory takes,
+        CollateralTransfer[] memory collateralWithdrawals,
```
MathisGD: this helped with the stack too deep

src/periphery/MidnightBundles.sol
```diff
@@ -342,4 +342,27 @@ contract MidnightBundles is IMidnightBundles {
...
         _safeApprove(token, spender, type(uint256).max);
     }
+
+    /// @dev Pulls `amount` of `token` from `from` to this bundler, optionally using ERC2612 or Permit2.
+    function _pullToken(address token, address from, uint256 amount, TokenPermit memory permit) internal {
+        if (permit.kind == PermitKind.ERC2612) {
+            (uint256 deadline, uint8 v, bytes32 r, bytes32 s) =
+                abi.decode(permit.data, (uint256, uint8, bytes32, bytes32));
+            // Tolerate revert: a third party may have already consumed the permit.
+            try IERC20Permit(token).permit(from, address(this), amount, deadline, v, r, s) {} catch {}
```
peyha, line 358: i don't know how common it is to tolerate revert in this case but it could make errors less relevant. For instance, if the permit was failing because of an expired deadline it would still go through and only revert during the transfer through an underflow. Unfortunately I don't think there's a standard for permit errors when the permit has already been consumed but we could still check the current nonce of the user before using permit to catch this exact situation

src/periphery/MidnightBundles.sol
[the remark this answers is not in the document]
adhusson: callers expect most of their offers to be used, I would say the wasted gas for exceptional congestion cases is not worth the effort

src/periphery/MidnightBundles.sol
peyha: just checked, it seems pretty standard to do the try catch so resolving

src/periphery/MidnightBundles.sol
```diff
@@ -158,48 +166,42 @@ contract MidnightBundles is IMidnightBundles {
...
 
         uint256 filledBuyerAssets;
         uint256 filledUnits;
         for (uint256 i; i < takes.length && filledBuyerAssets < targetFilledBuyerAssets; i++) {
-            require(!takes[i].offer.buy, InconsistentSide());
-            require(IMidnight(midnight).toId(takes[i].offer.obligation) == id, InconsistentObligation());
+            Take memory take_ = takes[i];
+            require(!take_.offer.buy, InconsistentSide());
+            require(IMidnight(midnight).toId(take_.offer.obligation) == id, InconsistentObligation());
+            uint256 unitsToTake = UtilsLib.min(
```
QGarchery, line 189: (optional) we could extract `unitsToTake` in other functions as well

src/periphery/MidnightBundles.sol
```diff
@@ -25,39 +30,40 @@ contract MidnightBundles is IMidnightBundles {
...
         address taker,
-        Take[] calldata takes,
-        CollateralTransfer[] calldata collateralWithdrawals,
+        Take[] memory takes,
+        CollateralTransfer[] memory collateralWithdrawals,
         address collateralReceiver,
         uint256 referralFeePct,
-        address referralFeeRecipient
+        address referralFeeRecipient,
+        TokenPermit memory loanTokenPermit
```
QGarchery: (optional) we could move the `TokenPermit` before `Take`. It would be more intuitive as it happens before the `take` and is either related to the token of `Take` or to the tokens of `CollateralTransfer`

src/periphery/MidnightBundles.sol
```diff
@@ -93,55 +100,56 @@ contract MidnightBundles is IMidnightBundles {
...
                     address(0),
                     "",
                     address(this),
-                    takes[i].offer,
-                    takes[i].ratifierData
+                    take_.offer,
+                    take_.ratifierData
                 ) returns (
-                uint256, uint256 resSellerAssets, uint256 resUnits
+                uint256, uint256 s, uint256 u
```
peyha: why this change ?

test/MidnightBundlesTest.sol
peyha, line 1: should permit be tested ? currently only permit2 is

src/periphery/MidnightBundles.sol
```diff
@@ -93,42 +100,44 @@ contract MidnightBundles is IMidnightBundles {
...
+        require(collateralPermits.length == collateralSupplies.length, InvalidPermitArrayLength());
+        address loanToken = takes[0].offer.obligation.loanToken;
         bytes32 id = IMidnight(midnight).toId(takes[0].offer.obligation);
 
+        Obligation memory obligation = takes[0].offer.obligation;
         for (uint256 i; i < collateralSupplies.length; i++) {
-            address token = takes[0].offer.obligation.collateralParams[collateralSupplies[i].collateralIndex].token;
-            SafeTransferLib.safeTransferFrom(token, msg.sender, address(this), collateralSupplies[i].assets);
+            address token = obligation.collateralParams[collateralSupplies[i].collateralIndex].token;
+            _pullToken(token, msg.sender, collateralSupplies[i].assets, collateralPermits[i]);
```
peyha: maybe worth a comment that each collateral should have a permit (potentially none if the approval has already been done)

MathisGD: maybe we could split `CollateralTransfer` into two, and incorporate the permit struct?

src/periphery/MidnightBundles.sol
```diff
@@ -13,6 +15,9 @@ import {WAD} from "../libraries/ConstantsLib.sol";
 contract MidnightBundles is IMidnightBundles {
     using UtilsLib for uint256;
 
+    /// @dev Canonical Permit2 deployment address (same on every chain).
+    address public constant PERMIT2 = 0x000000000022D473030F116dDEE9F6B43aC78BA3;
```
QGarchery: I think this is not true for zkSync chains, notably we deployed the v1 stack on Abstract

src/periphery/MidnightBundles.sol
peyha: so then we would have `CollateralTransferWithPermit` and `CollateralTransfer` ? I think it would work yes

src/periphery/MidnightBundles.sol
peyha: true, it's at address `0x0000000000225E31D15943971F47aD3022F714Fa` there, probably worth passing in the constructor

peyha: claude told me it was the only exception (though there might be small chains on which it wasn't deployed at all)

MathisGD: I would rather have the user pass it each time, it makes deployments much simpler

QGarchery: I realize that this is quite standard to hardcode the address. The [Permit2Lib does it](https://github.com/Uniswap/permit2/blob/cc56ad0f3439c502c246fc5cfcc3db92bb8b7219/src/libraries/Permit2Lib.sol#L25-L27), and we use it in `bundler3`

peyha: there's a problem with the bundler3 general adapter 1 on abstract then

test/vendor/Permit2.sol
QGarchery, line 1: it would be easier to verify if this was a lib we import

test/erc20s/ERC20Permit.sol
```diff
@@ -0,0 +1,36 @@
...
+    function permit(address owner, address spender, uint256 value, uint256 deadline, uint8 v, bytes32 r, bytes32 s)
+        external
+    {
+        require(block.timestamp <= deadline, "permit expired");
+        bytes32 structHash = keccak256(abi.encode(PERMIT_TYPEHASH, owner, spender, value, nonces[owner]++, deadline));
+        bytes32 digest = keccak256(abi.encodePacked("\x19\x01", DOMAIN_SEPARATOR(), structHash));
+        require(ecrecover(digest, v, r, s) == owner, "invalid permit");
+        allowance[owner][spender] = value;
+    }
+}
```
QGarchery, line 14:
```suggestion
import {PermitExt} from "./PermitExt.sol";

contract ERC20Permit is ERC20, PermitExt {
    constructor(string memory _name, string memory _symbol) ERC20(_name, _symbol) PermitExt(_name) {}

    function _setAllowance(address owner, address spender, uint256 value) internal override {
        allowance[owner][spender] = value;
    }
}
```
(optional) and changing the `name` variable in `PermitExt` to something else like `_name`

test/erc20s/PermitExt.sol
```diff
@@ -0,0 +1,39 @@
...
+
+    function _setAllowance(address owner, address spender, uint256 value) internal virtual;
+
+    function permit(address owner, address spender, uint256 value, uint256 deadline, uint8 v, bytes32 r, bytes32 s)
+        external
+    {
+        require(block.timestamp <= deadline, "permit expired");
+        bytes32 structHash = keccak256(abi.encode(PERMIT_TYPEHASH, owner, spender, value, nonces[owner]++, deadline));
+        bytes32 digest = keccak256(abi.encodePacked("\x19\x01", DOMAIN_SEPARATOR(), structHash));
+        require(ecrecover(digest, v, r, s) == owner, "invalid permit");
```
QGarchery, line 36:
```suggestion
        require(ecrecover(digest, v, r, s) == owner, "invalid permit");
        require(owner != address(0), "bad signature");
```
EIP2612 specifies that it should revert in that case

test/vendor/Permit2.sol
MathisGD: will do once we are in a separated repo

test/erc20s/ERC20Permit.sol
QGarchery: See also change below (require that the signer is not address(0))

test/vendor/Permit2.sol
MathisGD: see https://github.com/morpho-org/bundles/issues/1

src/periphery/MidnightBundles.sol
```diff
@@ -351,4 +337,27 @@ contract MidnightBundles is IMidnightBundles {
...
+        if (permit.kind == PermitKind.ERC2612) {
+            (uint256 deadline, uint8 v, bytes32 r, bytes32 s) =
+                abi.decode(permit.data, (uint256, uint8, bytes32, bytes32));
+            // Tolerate revert: a third party may have already consumed the permit.
+            try IERC20Permit(token).permit(from, address(this), amount, deadline, v, r, s) {} catch {}
+            SafeTransferLib.safeTransferFrom(token, from, address(this), amount);
+        } else if (permit.kind == PermitKind.Permit2) {
+            (address permit2, uint256 nonce, uint256 deadline, bytes memory signature) =
+                abi.decode(permit.data, (address, uint256, uint256, bytes));
+            IPermit2(permit2)
```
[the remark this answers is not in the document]
MathisGD: that's fine

src/periphery/MidnightBundles.sol
QGarchery, line 1: Code LGTM. Main concern is that it's a shift from how we usually deal with approvals. Because it uses SignatureTransfer instead of AllowanceTransfer, it will necessarily require the user to sign the approval every time: the permit2 case (and the permit one that mirrors it in the current code) only allows to approve the exact amount that will be pulled. 
This can be a pain, especially for hardware wallets that cannot sign multiple messages at once

MathisGD: https://morpholabs.slack.com/archives/C08AAMKH524/p1778579957326139?thread_ts=1778143517.167099&cid=C08AAMKH524

MathisGD: see the thread, the interface already makes people sign just the required amount. so I think that's fine

src/periphery/MidnightBundles.sol
```diff
@@ -39,23 +49,17 @@ contract MidnightBundles is IMidnightBundles {
...
         _forceApproveMax(loanToken, midnight);
-        SafeTransferLib.safeTransferFrom(loanToken, msg.sender, address(this), maxBuyerAssets);
+        _pullToken(loanToken, msg.sender, maxBuyerAssets, loanTokenPermit);
 
         uint256 filledUnits;
         uint256 filledBuyerAssets;
         for (uint256 i; i < takes.length && filledUnits < targetUnits; i++) {
-            require(!takes[i].offer.buy, InconsistentSide());
-            require(IMidnight(midnight).toId(takes[i].offer.obligation) == id, InconsistentObligation());
+            Take memory take_ = takes[i];
```
adhusson: this is for the stack?

src/periphery/MidnightBundles.sol
```diff
@@ -359,4 +345,27 @@ contract MidnightBundles is IMidnightBundles {
...
+        if (permit.kind == PermitKind.ERC2612) {
+            (uint256 deadline, uint8 v, bytes32 r, bytes32 s) =
+                abi.decode(permit.data, (uint256, uint8, bytes32, bytes32));
+            // Tolerate revert: a third party may have already consumed the permit.
+            try IERC20Permit(token).permit(from, address(this), amount, deadline, v, r, s) {} catch {}
+            SafeTransferLib.safeTransferFrom(token, from, address(this), amount);
+        } else if (permit.kind == PermitKind.Permit2) {
+            (address permit2, uint256 nonce, uint256 deadline, bytes memory signature) =
+                abi.decode(permit.data, (address, uint256, uint256, bytes));
+            IPermit2(permit2)
```
QGarchery: This is very scary (the user chooses `permit2` here) as it allows both hidden arbitrary callbacks in many places, and also because it allows to bypass the transfer. Looking at the code, I think both are safe for the same reason: every token that is supposed to be pulled by `_pullToken` will be used in some way, and if there is too little left it will revert the whole bundle

QGarchery (approved):

src/periphery/MidnightBundles.sol
adhusson: even if the called contract does the required transfer it can also do additional arbitrary things. The call happens after the authorization check & approval, but before everyhting, except when supplying collateral where it happens between collateral supplies. I'm not seeing a problem but indeed it is a new degree of freedom.

MathisGD: let's actually hardcode the permit2 address.

src/periphery/MidnightBundles.sol
MathisGD: removed

QGarchery (approved):

peyha (approved):
LGTM, note that permit is not tested as extensively as permit2 (only on 1 bundle endpoint)

adhusson (approved):

src/periphery/MidnightBundles.sol
Saw-mon-and-Natalie: > [Permit2Lib does it](https://github.com/Uniswap/permit2/blob/cc56ad0f3439c502c246fc5cfcc3db92bb8b7219/src/libraries/Permit2Lib.sol#L25-L27)

@MathisGD you should also mention this to Alice from Uniswap.

PR #814, opened 2026-05-07
style lastLossFactor

### LINKED ISSUE #802: rename globalLossFactor
it's bit annoying that this has the exact same name. we notably call the Midnight getter "userLossFactor", not ideal...

_Originally posted by @MathisGD in https://github.com/morpho-org/midnight/pull/783#discussion_r3191710773_

MathisGD:
fixes #802

### DIFF
--- a/src/Midnight.sol
+++ b/src/Midnight.sol
@@ -613,17 +613,16 @@ contract Midnight is IMidnight {
         if (badDebt > 0) {
             // forge-lint: disable-next-item(unsafe-typecast) as badDebt <= _position.debt
             _position.debt -= uint128(badDebt);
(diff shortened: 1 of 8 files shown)


peyha (approved):

src/Midnight.sol
QGarchery, line 1: I realize that it may be a bit confusing in `liquidate` to have `oldLossFactor`, which is close in meaning to `lastLossFactor`, but it actually refers to the obligation loss factor and not the user loss factor. Maybe rename to use just `_lossFactor` ? (and same for total units)

QGarchery (approved):

PR #815, opened 2026-05-07
Permit2 test

adhusson:


test/BundlerTest.sol
```diff
@@ -23,10 +30,36 @@ contract BundlerTest is BaseTest {
...
+            )
+        );
+        bytes32 digest = keccak256(
+            abi.encodePacked("\x19\x01", VendorPermit2(takeBundler.PERMIT2()).DOMAIN_SEPARATOR(), permitHash)
+        );
+        (uint8 v, bytes32 r, bytes32 s) = vm.sign(privateKey[owner], digest);
+        return TokenPermit({kind: PermitKind.Permit2, data: abi.encode(nonce, deadline, abi.encodePacked(r, s, v))});
+    }
+
     function setUp() public override {
```
MathisGD, line 31: let's move at the top

test/BundlerTest.sol
```diff
@@ -256,6 +289,103 @@ contract BundlerTest is BaseTest {
         }
     }
 
+    function testBuyBuyerAssetsTargetPermit2() public {
```
MathisGD, line 292: let's approve zero at the beginning

adhusson: you mean move `loanToken.approve(address(takeBundler), 0);` up? not clear what the purpose is

MathisGD: right now the lender has tokens + has approved so the test could be not testing the right thing

adhusson: the approve zero happens, it's later in the test, so that is covered. 

I will move it to the top for clarity.

MathisGD (approved):

PR #816, opened 2026-05-07
[Certora] OfferTree Soundness

bhargavbh:
Objective is to show that a successful `take` can only settle an offer that was genuinely committed in the signed tree.
We reason about `OfferTree`, a model of the tree built only through the `newLeaf` and `newInternalNode` primitives. Leaves are keyed by `HashLib.hashOffer(offer)` and store a fixed-size pre-image of the offer, so `isWellFormed` re-hashes a leaf with a single bounded keccak instead of looping over the offer's dynamic members, which keeps the proofs bounded regardless of offer size.

- `OfferTreeWellFormed.spec` checks that the primitives only ever build well-formed trees: every node is empty, a leaf carrying a genuine `hashOffer`, or an internal node correctly hashing its two children.
- `OfferTreeMembership.spec` checks the main soundness result: for any well-formed tree, if a Merkle proof verifies an offer's hash against the root via `isLeaf`, then the offer is registered as a leaf. Equivalently, no valid proof can be forged for an offer that is not in the tree.

The verification setup and technique is inspired from the Merkle Tree Membership soundness spec in [Universal Rewards Distributor](https://github.com/morpho-org/universal-rewards-distributor/)

[corresponding thread](https://morpholabs.slack.com/archives/C0ACSA8TNMA/p1782129714266809)
[motivation](https://morpholabs.slack.com/archives/C0B70AVMM0E/p1788248396017159)


certora/specs/Ratification.spec
```diff
@@ -1,11 +1,27 @@
...
+
 methods {
     function multicall(bytes[]) external => HAVOC_ALL DELETE;
 
     function isAuthorized(address authorizer, address authorized) external returns (bool) envfree;
 
-    // Over-approximate view functions.
+    function Utils.callbackSuccess() external returns (bytes32) envfree;
+
+    function _.isRatified(Midnight.Offer, bytes) external => DISPATCHER(true);
```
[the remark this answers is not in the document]
bhargavbh: its exactly what we want, verifying against all existing ratifier implementations.

certora/README.md
QGarchery, line 1: Let's keep the following convention: one sentence per line

certora/README.md
```diff
@@ -43,6 +43,39 @@ How offers are consumed when taken.
...
+
+- [`OfferTreeWellFormed.spec`](specs/OfferTreeWellFormed.spec) checks that the primitives only ever build well-formed trees: every node is empty, a leaf carrying a genuine `hashOffer`, or an internal node correctly hashing its two children. In particular, there is no restriction for the left and right children of a parent node to be sorted.
+- [`OfferTreeMembership.spec`](specs/OfferTreeMembership.spec) checks the main soundness result: for any well-formed tree, if a Merkle proof verifies an offer's hash against the root, then the offer is registered as a leaf. Equivalently, no valid proof can be forged for an offer that is not in the tree.
+- [`Ratification.spec`](specs/Ratification.spec) connects this to the on-chain path: every successful `isRatified` (across all ratifier implementations) and every successful `take` actually invokes `HashLib.isLeaf` and it returns true.
```
QGarchery:
```suggestion
- [`Ratification.spec`](specs/Ratification.spec) connects this to the on-chain path: every successful `isRatified` (across all ratifier implementations) and every successful `take` actually verifies the Merkle tree membership of an offer, taking as input the offer's hash and checking it against the root.
```
Because `isLeaf` is an implementation detail that is not mentioned before

certora/README.md
```diff
@@ -43,6 +43,39 @@ How offers are consumed when taken.
...
+Combining the three: a successful `take` runs a Merkle membership check against the ratified root (`Ratification`), which for a well-formed root implies the offer is genuinely a leaf of that tree (`OfferTreeMembership`), and the root is well-formed because it is built only from the well-formedness-preserving primitives (`OfferTreeWellFormed`).
+
+### Checking a concrete root
+
+The membership result is stated for any well-formed root; the checker in [`checker`](checker) lets anyone confirm what a specific root commits to, by rebuilding the tree through the same verified primitives.
+
+1. Write a `proofs.json` listing the claimed `root` and the offers (`leaves`), padded to a power of two with empty offers.
+2. Generate the certificate, from the repository root:
+   ```
+   python certora/checker/create_certificate.py proofs.json
```
QGarchery: In the URD the `proofs.json` was obtained as an artifact of the creation of the rewards distribution program (see the `proofs.json` files under [this folder](https://github.com/morpho-org/morpho-optimizers-rewards/tree/main/distribution)). Since here it doesn't work the same it makes less sense to take a `proofs.json` file. I'm thinking that we may prefer the full offer tree, which is what is produced when doing clear signing (see `test/frontend/README.md`)

certora/helpers/OfferTree.sol
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    function hashNode(bytes32 left, bytes32 right) public pure returns (bytes32) {
+        return HashLib.hashNode(left, right);
+    }
+
+    function _hashLeaf(bytes32 id) public view returns (bytes32) {
+        return _hashLeaf(tree[id].leaf);
+    }
+
```
QGarchery:
```suggestion
```

certora/helpers/OfferTree.sol
```diff
@@ -0,0 +1,182 @@
...
+    /* PURE AND VIEW FUNCTIONS */
+
+    function _isEmpty(Node storage n) internal view returns (bool) {
+        return n.left == 0 && n.right == 0 && n.hashNode == 0;
+    }
+
+    function isEmpty(bytes32 id) public view returns (bool) {
+        return _isEmpty(tree[id]);
+    }
```
QGarchery:
```suggestion
    function isEmpty(Node storage n) internal view returns (bool) {
        return n.left == 0 && n.right == 0 && n.hashNode == 0;
    }
```
(and change `_isEmpty` occurrences to `isEmpty`)

certora/helpers/OfferTree.sol
```diff
@@ -0,0 +1,182 @@
...
+        l.callbackDataHash = keccak256(offer.callbackData);
+        l.receiverIfMakerIsSeller = offer.receiverIfMakerIsSeller;
+        l.ratifier = offer.ratifier;
+        l.reduceOnly = offer.reduceOnly;
+        l.maxUnits = offer.maxUnits;
+        l.maxAssets = offer.maxAssets;
+        n.hashNode = id;
+    }
+
+    function newInternalNode(bytes32 id, bytes32 left, bytes32 right) public {
```
jhoenicke, line 64: For a leafnode, the id equals the hash (`tree[i].hashNode == id`).  But for internal nodes you allow the id to be different from the hashNode.  Is there a reason for it?

certora/helpers/OfferTree.sol
```diff
@@ -0,0 +1,182 @@
...
+import {HashLib, OFFER_TYPEHASH} from "../../src/ratifiers/libraries/HashLib.sol";
+
+// A leaf's offer reaches HashLib.hashOffer only through fixed-size fields: every dynamic member of Offer
+// (market.collateralParams and callbackData) is already reduced to a single bytes32 before the final
+// keccak (hashMarket(market) and keccak256(callbackData)). So we store that fixed-size pre-image instead
+// of the raw Offer. This keeps every Node fixed-size, so isWellFormed re-hashes a leaf with one bounded
+// keccak (no dynamic-array reads, no loops) — the property that lets the wellFormed invariant scale, exactly
+// as it does for URD's fixed-size leaves. _hashLeaf mirrors hashOffer's outer keccak field-for-field, so a
+// leaf's id is the real hashOffer value; the wellFormed invariant's newLeaf case checks this mirror holds
+// (id == hashOffer(offer) and id == _hashLeaf(leaf)), so the correspondence is verified, not assumed.
```
QGarchery: It really checks: `hashNode == _hashLeaf(leaf)` and `id == _hashLeaf(leaf)`. But it doesn't verify that `hashNode == hashOffer(offer)` (which it can't do because it doesn't have the offer), so this part is really assumed

certora/helpers/OfferTree.sol
```diff
@@ -0,0 +1,182 @@
...
+        Node storage n = tree[id];
+        if (_isEmpty(n)) return true;
+        if (n.left == 0 && n.right == 0) {
+            bytes32 expected = _hashLeaf(n.leaf);
+            return n.hashNode == expected && id == expected;
+        }
+        if (n.left != 0 && n.right != 0) {
+            bytes32 leftHash = tree[n.left].hashNode;
+            bytes32 rightHash = tree[n.right].hashNode;
+            return leftHash != 0 && rightHash != 0 && n.hashNode == HashLib.hashNode(leftHash, rightHash);
```
QGarchery: Here we are verifying that the hashes are not zero, which is not exactly the definition of `isEmpty`. Should we simplify `isEmpty` to only `hashNode == 0` ?

certora/helpers/OfferTree.sol
```diff
@@ -0,0 +1,182 @@
...
+                l.maxUnits,
+                l.maxAssets
+            )
+        );
+    }
+
+    // The specification of a well-formed tree is the following:
+    //   - empty nodes are well-formed
+    //   - leaves have the correct identifier and hashing
+    //   - internal nodes have the correct hashing
```
QGarchery: Let's avoid using the word "correct", it hides the meaning
```suggestion
    //   - leaves have the hash of the leaf data as identifier, which is also their stored hash
    //   - internal nodes have the hash of their children has their stored hash
```

certora/helpers/OfferTree.sol
bhargavbh: yes, the id is left arbitrary for internal nodes intentionally because it does not affect the soundness proof. you can think of ids as just keys, and the property for well-formedness does not depend on it. The proof relies on the check hashNode == hashNode(leftHash, rightHash) [here]( 
https://github.com/morpho-org/morpho-v2/blob/a1b4b8ecabf4e126d40b2e295a800c659cc61b08/certora/helpers/OfferTree.sol#L164). I do not think it harms if we add it as a require but its not necessary.

certora/specs/Ratification.spec
QGarchery, line 1: this is too many changes to Ratification.spec IMO (the new assumptions we make are impacting other rules).
We could either:
1. try to reduce the number of assumptions (do we need all of them ?)
2. move it to a separate spec
3. remove those changes, the added rules are almost trivial, even though it is an important property

I would favor option 3, if we cannot do option 1 and remove a lot of those assumptions

certora/helpers/OfferTree.sol
QGarchery: actually it does verify it with the rule `hashLeafReproducesHashOffer`. Maybe it deserves a longer explanation though

certora/helpers/OfferTree.sol
QGarchery: Initially this constraint was there to prevent having the same address multiple times in the tree. 

We don't have this requirement, but it's weirdly useful here. Assume that the id of the leaves was not necessarily the same as the hash, and you remove the corresponding check in `isWellFormed` (the check `id == expected`). Now you could have a function that would change `marketHash` and adapt `hashNode` accordingly (as the `hashLeaf` of the new params with the changed `marketHash`). This would not break the invariant `isWellFormed`, but it would allow to forge another root.

Another more straightforward way to avoid the above scenario, is to check that the tree is immutable.

certora/specs/OfferTreeMembership.spec
```diff
@@ -0,0 +1,20 @@
...
+    function isLeaf(bytes32, bytes32, uint256, bytes32[]) external returns (bool) envfree;
+    function isLeafNode(bytes32) external returns (bool) envfree;
+    function wellFormedPath(bytes32, uint256, uint256) external returns (bool) envfree;
+}
+
+// Soundness: if a Merkle proof verifies along a well-formed path, hashOffer(offer) must be a leaf node.
+rule membershipSoundness(OfferTree.Offer offer, bytes32 root, uint256 leafIndex, bytes32[] proof) {
+    bytes32 node;
+    require getHash(node) == root, "root is the hash of node";
+    require wellFormedPath(node, leafIndex, proof.length), "the path from the root to the leaf is well-formed";
```
jhoenicke, line 15: This require is okay, because it only requires the invariant isWellformed(id) for some ids and the invariant guarantees this.  Should we add a comment that explains this?

jhoenicke (approved):
Looks good.  I also tried it with a broken hash function that allows collision of node and leaf and at least that found a counterexample.

certora/confs/OfferTreeMembership.conf
```diff
@@ -0,0 +1,21 @@
...
+  "solc": "solc-0.8.34",
+  "solc_via_ir": true,
+  "solc_optimize": "200",
+  "solc_evm_version": "osaka",
+  "optimistic_loop": true,
+  "loop_iter": 3,
+  "optimistic_hashing": true,
+  "hashing_length_bound": 1024,
+  "rule_sanity": "basic",
+  "smt_timeout": 7200,
```
QGarchery:
```suggestion
```

certora/confs/OfferTreeWellFormed.conf
```diff
@@ -0,0 +1,21 @@
...
+  "solc": "solc-0.8.34",
+  "solc_via_ir": true,
+  "solc_optimize": "200",
+  "solc_evm_version": "osaka",
+  "optimistic_loop": true,
+  "loop_iter": 3,
+  "optimistic_hashing": true,
+  "hashing_length_bound": 1024,
+  "rule_sanity": "basic",
+  "smt_timeout": 7200,
```
QGarchery:
```suggestion
```

certora/specs/OfferTreeWellFormed.spec
QGarchery, line 1: We could show that isEmpty means that it is .. empty. That is every field is the default one, not just the hash. This could be done by changing the definition of `isWellFormed` in the `isEmpty` case

certora/helpers/GenerateRoot.sol
```diff
@@ -0,0 +1,44 @@
...
+            for (uint256 i = 0; i < levelLength; i++) {
+                bytes32 left = level[2 * i];
+                bytes32 right = level[2 * i + 1];
+                bytes32 nodeHash = HashLib.hashNode(left, right);
+                Node storage n = tree[nodeHash];
+                if (isEmpty(n)) {
+                    newInternalNode(nodeHash, left, right);
+                } else {
+                    require(n.left == left && n.right == right && n.hash == nodeHash, "internal node id collision");
+                }
```
QGarchery: I don't think we need to protect against hashing collision (btw we already assume in this spec that we can't find one)

QGarchery:
```suggestion
                if (isEmpty(n)) {
                    newInternalNode(nodeHash, left, right);
                }
```
I don't think we need to protect against hashing collision (btw we already assume in this spec that we can't find one)

certora/helpers/GenerateRoot.sol
```diff
@@ -0,0 +1,44 @@
...
+    function generateRoot(Offer[] memory leaves) public returns (bytes32) {
+        require(leaves.length > 0 && (leaves.length & (leaves.length - 1)) == 0, "invalid leaves length");
+
+        bytes32[] memory level = new bytes32[](leaves.length);
+        for (uint256 i = 0; i < leaves.length; i++) {
+            bytes32 leafHash = HashLib.hashOffer(leaves[i]);
+            if (isEmpty(tree[leafHash])) {
+                newLeaf(leaves[i]);
+            } else {
+                require(isLeafNode(leafHash), "leaf id collision");
```
QGarchery:
```suggestion
```
same

certora/helpers/GenerateRoot.sol
```diff
@@ -0,0 +1,44 @@
...
+                if (isEmpty(n)) {
+                    newInternalNode(nodeHash, left, right);
+                } else {
+                    require(n.left == left && n.right == right && n.hash == nodeHash, "internal node id collision");
+                }
+                level[i] = nodeHash;
+            }
+        }
+
+        return level[0];
```
QGarchery, line 26:
```suggestion
        return getHash(level[0]);
```
this is actually critical ! We want to return a hash that corresponds to a node of a tree, because that is what is guaranteed to be a well-formed tree. 

This is a very nice trick to reduce the trust base, where the only line you need to look at in this contract for correctness is just this one:
- the front end uses `GenerateRoot.sol`, and checks that the return value is the root that should be signed
- so it knows that the root is the hash of a node of `OfferTree` (the helper contract), which is exactly what is needed for `membershipSoundness`: `require getHash(node) == root, "root is the hash of node"`

certora/helpers/OfferTree.sol
```diff
@@ -0,0 +1,159 @@
...
+contract OfferTree {
+    struct Node {
+        bytes32 left;
+        bytes32 right;
+        Leaf leaf;
+        // Offer hash for leaves and hash of the children for internal nodes.
+        bytes32 hash;
+    }
+
+    // Leaf ids are offer hashes. Internal node ids may be arbitrary.
```
QGarchery: We could use the hash as the id of internal nodes. Since this is what GenerateRoot uses we don't lose the use case

certora/helpers/GenerateRoot.sol
QGarchery, line 1: Let's have a test for this file. What would be interesting is a tree with duplicate offers that are siblings, since that will happen with empty offers to fill the tree up to a power of 2

certora/helpers/GenerateRoot.sol
```diff
@@ -0,0 +1,44 @@
...
+contract GenerateRoot is OfferTree {
+    // Build a perfect tree from a non-empty power-of-two list. Duplicate hashes share nodes.
+    function generateRoot(Offer[] memory leaves) public returns (bytes32) {
+        require(leaves.length > 0 && (leaves.length & (leaves.length - 1)) == 0, "invalid leaves length");
+
+        bytes32[] memory level = new bytes32[](leaves.length);
+        for (uint256 i = 0; i < leaves.length; i++) {
+            bytes32 leafHash = HashLib.hashOffer(leaves[i]);
+            if (isEmpty(tree[leafHash])) {
+                newLeaf(leaves[i]);
```
QGarchery: This does not compute a tree, it actually computes a DAG. I think it's the right thing to compute: we really want to have duplicate offers, which also leads to having duplicate internal nodes. We could model that data structure instead, by internalizing the fact the creation of a leaf/node is idempotent (instead of skipping the creation here). No strong opinion on this one, I let you decide what you think is clearer

certora/helpers/GenerateRoot.sol
QGarchery, line 1: As discussed: let's separate the OfferTree contract instead of inheriting from it

bhargavbh:
All good suggestions. Incorporated the following changes which simplifies the model quite a bit without impacting the claim:
- Hash IDs: Internal-node IDs equal their hashes.
- Idempotent creation: Constructors return hash IDs and reuse existing nodes.
- Simpler membership: Use root directly; remove the separate node variable and getHash.
- Generator does not inherit OfferTree: Compute roots in memory, independently of OfferTree.
- Cleaner configs: Remove explicit sanity and timeout settings.
- Regression tests: Cover root agreement, idempotency, missing children, duplicates, membership proofs, and storage isolation.
cc @QGarchery

certora/helpers/GenerateRoot.sol
bhargavbh: with the simplified model of hash as id, this is no longer needed. level[0] already is the root-hash

certora/specs/OfferTreeWellFormed.spec
bhargavbh: the membership rule does not depend on it. if you have a strong opinion we can change, do we want to establish empty entries means all foelds are zeroed?

certora/confs/OfferTreeWellFormed.conf
```diff
@@ -11,8 +11,7 @@
   "loop_iter": 3,
   "optimistic_hashing": true,
   "hashing_length_bound": 1024,
-  "rule_sanity": "basic",
-  "smt_timeout": 7200,
+  "disable_auto_cache_key_gen": true,
```
QGarchery, line 14: What is that flag doing ?

bhargavbh: the CI was [failing](https://prover.certora.com/output/21317/74819ff3819e45838dba06aa4cf81b3c?anonymousKey=3171dc81cd928539874e50646d2f3cf45a0b63a1) with `"Encountered an unexpected error in Prover, please report in certora.com. Error code 2463013129."`
and info:
`Contract preprocessing step fetched from cache.`

IIUC, this flag ensures the proofs are rerun and not fetched from cache.

test/OfferTreeTest.sol
```diff
@@ -0,0 +1,176 @@
...
+        Offer memory rightOffer;
+        rightOffer.tick = rightTick;
+        bytes32 left = HashLib.hashOffer(leftOffer);
+        bytes32 right = HashLib.hashOffer(rightOffer);
+        assertEq(tree.newLeaf(leftOffer), left);
+        assertEq(tree.newLeaf(rightOffer), right);
+        assertEq(tree.newLeaf(leftOffer), left);
+        assertFalse(tree.isEmpty(left));
+        assertTrue(tree.isLeafNode(left));
+        assertTrue(tree.isWellFormed(left));
```
QGarchery:
```suggestion
```
let's remove all the `isWellFormed` assertions, it clutters the tests and it's formally verified after all

certora/helpers/OfferTree.sol
```diff
@@ -0,0 +1,163 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Copyright (c) 2025 Morpho Association
+pragma solidity ^0.8.0;
+
+import {Offer} from "../../src/interfaces/IMidnight.sol";
+import {HashLib, OFFER_TYPEHASH} from "../../src/ratifiers/libraries/HashLib.sol";
+
+// Fixed-size pre-image of HashLib.hashOffer. Dynamic fields are stored as hashes so CVL can re-hash a leaf
+// without iterating over dynamic data.
```
QGarchery:
```suggestion
// Fixed-size pre-image of HashLib.hashOffer. Dynamic fields are stored as hashes so CVL can re-hash a leaf without iterating over dynamic data.
```

certora/helpers/OfferTree.sol
```diff
@@ -0,0 +1,163 @@
...
+                l.reduceOnly,
+                l.maxUnits,
+                l.maxAssets,
+                l.continuousFeeCap
+            )
+        );
+    }
+
+    // Empty nodes are fully zeroed. Populated nodes have their hash as identifier and hash their leaf data or their
+    // two non-empty children.
```
QGarchery:
```suggestion
    // Empty nodes are fully zeroed. Populated nodes have their hash as identifier and hash their leaf data or their two non-empty children.
```

certora/specs/OfferTreeMembership.spec
```diff
@@ -0,0 +1,61 @@
...
+
+// Check the generator itself, without assuming that its returned root or path is well-formed.
+// This covers 1, 2, or 4 input offers and proofs of at most 4 siblings. Solidity loops (including
+// collateral hashing) remain subject to loop_iter = 4 and the configured hash assumptions.
```
QGarchery:
```suggestion
// This covers 1, 2, or 4 input offers and proofs of at most 4 siblings. Solidity loops (including collateral hashing) remain subject to loop_iter = 4 and the configured hash assumptions.
```

certora/specs/OfferTreeMembership.spec
```diff
@@ -0,0 +1,61 @@
...
+rule generatedRootMembershipSoundness(OfferTree.Offer[] offers, OfferTree.Offer offer, uint256 leafIndex, bytes32[] proof) {
```
QGarchery: Not a huge fan of this rule, it adds some complexity and the a few things it does are not very relevant:
- some sort of sanity check with the 3 `satisfy` at the end, but we already have sanity checks enabled
- make sure that GenerateRoot returns a node, but this can be trusted IMO if we return a node at the end
- does some unit tests with `assert inFirst || inSecond || inSecondPair`

test/OfferTreeTest.sol
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    function testGenerateRootRejectsInvalidLengths() public {
+        vm.expectRevert("invalid leaves length");
+        generator.generateRoot(new Offer[](0));
+        vm.expectRevert("invalid leaves length");
+        generator.generateRoot(new Offer[](3));
+    }
+
+    function testGenerateRootOnlyPopulatesConfiguredTree() public {
```
QGarchery: This doesn't really make sense as a test, but this is actually more relevant for formal verification !
I think we should drop it, and optionally add it as a rule

QGarchery (approved):

certora/confs/OfferTreeWellFormed.conf
QGarchery: Ok makes sense. So once it is reported to Certora, it could be fixed and then we could remove this flag

certora/specs/OfferTreeMembership.spec
bhargavbh: removed

test/OfferTreeTest.sol
bhargavbh: decided to skip adding the rule; its very clear from GenerateRoot

certora/specs/OfferTreeMembership.spec
QGarchery: > make sure that GenerateRoot returns a node, but this can be trusted IMO if we return a node at the end

so we still need to return the hash of a node to have about the same benefit as this rule had

bhargavbh: correct, we return the hash of the node in generateRoot

certora/specs/OfferTreeMembership.spec
```diff
@@ -0,0 +1,61 @@
...
+    function isLeaf(bytes32, bytes32, uint256, bytes32[]) external returns (bool) envfree;
+    function isLeafNode(bytes32) external returns (bool) envfree;
+    function isWellFormed(bytes32) external returns (bool) envfree;
+    function wellFormedPath(bytes32, uint256, uint256) external returns (bool) envfree;
+}
+
+// Soundness: if a Merkle proof verifies along a well-formed path, hashOffer(offer) must be a leaf node.
+rule membershipSoundness(OfferTree.Offer offer, bytes32 root, uint256 leafIndex, bytes32[] proof) {
+    require !isEmpty(root), "root is a populated node";
+    require wellFormedPath(root, leafIndex, proof.length), "the path from the root to the leaf is well-formed";
```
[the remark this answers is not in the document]
bhargavbh: we intentionally chose to narrow down the scope of this PR to membership soundness. The soundness at ratifier level can be added later if necessary.

certora/confs/OfferTreeWellFormed.conf
bhargavbh: created an issue https://github.com/morpho-org/midnight/issues/1207
resolving this thread for now

PR #817, opened 2026-05-07
separate hashlib

MathisGD:


src/ratifiers/HashLib.sol
peyha, line 1: MerkleLib & HashLib should prob be in the same folder

peyha (approved):

PR #803, opened 2026-05-06
Clarify documentation on creditOf and related values

MathisGD:
Updated documentation to clarify that creditOf, pendingFee, and lossIndex are not up to date.

PR #406, opened 2026-02-25
bundler first iteration

MathisGD:
first version of the bundler. see the limitations at the top of the contract

PR #819, opened 2026-05-07
Document up to date values

QGarchery:
Follow-up to #803

### DIFF
--- a/src/Midnight.sol
+++ b/src/Midnight.sol
@@ -159,7 +159,6 @@ import {EventsLib} from "./libraries/EventsLib.sol";
 ///
 /// MISC
 /// @dev creditOf, pendingFee, and lossFactor are not up to date. Use updatePositionView to get the up-to-date values.
-/// @dev Trading fees and the continuous fee are 0 until the obligation is created, then set to the default values.
 /// @dev The max amount of totalUnits, collateral, credit, and debt is type(uint128).max (~1e38).
 /// @dev Zero checks are not systematically performed.
 /// @dev No-ops are allowed. In particular, Midnight can call the callback of offers through a no-op take, even if those
@@ -888,6 +887,7 @@ contract Midnight is IMidnight {
         return obligationState[id].withdrawable;
     }
 
+    /// @dev The trading fees are 0 until the obligation is created, then set to the default value.
     function tradingFees(bytes32 id) external view returns (uint16[7] memory) {
         return [
             obligationState[id].tradingFee0,
--- a/src/Midnight.sol
+++ b/src/Midnight.sol
@@ -158,7 +158,7 @@ import {EventsLib} from "./libraries/EventsLib.sol";
 /// @dev When the claimer is set, the old claimer loses the unclaimed fees.
 ///
 /// MISC
-/// @dev Credit, pending fee, and loss factor are not up to date. Use updatePositionView to get the up-to-date values.
+/// @dev creditOf, pendingFee, and lossFactor are not up to date. Use updatePositionView to get the up-to-date values.
 /// @dev Trading fees and the continuous fee are 0 until the obligation is created, then set to the default values.
 /// @dev The max amount of totalUnits, collateral, credit, and debt is type(uint128).max (~1e38).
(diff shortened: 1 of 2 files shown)


src/Midnight.sol
```diff
@@ -158,8 +158,8 @@ import {EventsLib} from "./libraries/EventsLib.sol";
 /// @dev When the claimer is set, the old claimer loses the unclaimed fees.
 ///
 /// MISC
-/// @dev creditOf, pendingFee, and lossFactor are not up to date. One must use updatePositionView to get the up to date
-/// values.
+/// @dev Credit, pending fee, and loss factor are not up to date. Use updatePositionView to get the up to date values.
```
QGarchery: Because this also includes the continuous fee credit this way

src/Midnight.sol
```diff
@@ -158,8 +158,8 @@ import {EventsLib} from "./libraries/EventsLib.sol";
...
+/// @dev Trading fees and the continuous fee are 0 until the obligation is created, then set to the default values.
```
QGarchery: Adding it because I feel like [this](https://github.com/morpho-org/midnight/blob/dc797418b586bbdd74f09094fd4001a7cf30269c/src/periphery/TakeBundler.sol#L164) is tricky, same reasoning for the comment that is above the `continuousFee` function
```solidity
// touchObligation to have the correct trading fees.
bytes32 id = IMidnight(midnight).touchObligation(takes[0].offer.obligation);
```

MathisGD: nb we added it then removed then re-added ^^

src/Midnight.sol
MathisGD: why does it include it

src/Midnight.sol
QGarchery: Notably the `require(_obligationState.created)` in `tradingFee()` was added in the [first bundler PR](https://github.com/morpho-org/midnight/pull/406/changes#diff-bc7f9788c87d9ce947db1da48e2c26ea767af169dda1aa6223af2a2b419a07eb), probably because we felt that it could be missed easily

src/Midnight.sol
```suggestion
/// @dev Credit, pending fee, and loss factor are not up to date. Use updatePositionView to get the up-to-date values.
```
nitpick cf [this](https://english.stackexchange.com/questions/180611/do-i-keep-myself-up-to-date-or-up-to-date-on-something)

QGarchery: because of the change `creditOf` -> `Credit`
but maybe it's not explicit enough

src/Midnight.sol
QGarchery: > nb we added it then removed then re-added ^^

The require or the comment ?

src/Midnight.sol
MathisGD: yes, let's make it more explicit

src/Midnight.sol
MathisGD: the comment

peyha (approved):

src/Midnight.sol
MathisGD: actually it doesn't really mean anything for the continuous fee credit to be updated (would need to touch all users). doesn't feel worth it

MathisGD (approved):

src/interfaces/IMidnight.sol
```diff
@@ -37,6 +37,7 @@ struct Offer {
     uint256 maxBuyerAssets;
 }
 
+/// @dev Trading fees are 0 until the obligation is created, then set to the default values.
```
adhusson:
```suggestion
/// @dev Trading fees and the continuous fee are 0 until the obligation is created, then set to the default values.
```

src/Midnight.sol
adhusson, line 891: /// @dev The trading fees are 0 until the obligation is created, then set to the default value.

adhusson, line 1: The "0 then default" behavior is mentioned in 3 places, that may be too many. I would put it only in front of the getters and above the struct.

MathisGD (approved):

adhusson (approved):

PR #825, opened 2026-05-07
calldata parameters

MathisGD:
https://morpholabs.slack.com/archives/C08A40KKN9W/p1778175387142719


src/Midnight.sol
```diff
@@ -737,7 +741,7 @@ contract Midnight is IMidnight {
     }
 
     /// @dev Returns the obligation id and creates the obligation if it doesn't exist yet.
-    function touchObligation(Obligation memory obligation) public returns (bytes32) {
+    function touchObligation(Obligation calldata obligation) public returns (bytes32) {
         bytes32 id = toId(obligation);
```
adhusson, line 745: if we don't make toId also take a `calldata` param, we lose the gains of making the obligation `calldata`.

adhusson: See #828

MathisGD:
closing as it cost more gas

PR #829, opened 2026-05-08
Optimize hash offer

adhusson:
* works with certora 
* saves 700-900 gas for 2-5 collats

### DIFF
--- a/src/ratifiers/HashLib.sol
+++ b/src/ratifiers/HashLib.sol
@@ -40,6 +40,7 @@ library HashLib {
     }
 
     /// @dev Computes the EIP-712 hash struct of an Offer.
+    /// @dev Same as kecccak256(abi.encode(OFFER_TYPEHASH, ...));
     function hashOffer(Offer memory offer) internal pure returns (bytes32 result) {
         bytes32[17] memory w;
         w[0] = OFFER_TYPEHASH;
--- a/src/ratifiers/HashLib.sol
+++ b/src/ratifiers/HashLib.sol
@@ -40,8 +40,8 @@ library HashLib {
     }
 
     /// @dev Computes the EIP-712 hash struct of an Offer.
-    /// @dev Same as kecccak256(abi.encode(OFFER_TYPEHASH, ...));
-    function hashOffer(Offer memory offer) internal pure returns (bytes32 result) {
+    /// @dev Same as keccak256(abi.encode(OFFER_TYPEHASH, ...));
+    function hashOffer(Offer memory offer) internal pure returns (bytes32) {
         bytes32[17] memory w;
         w[0] = OFFER_TYPEHASH;
         w[1] = hashObligation(offer.obligation);
@@ -60,8 +60,10 @@ library HashLib {
         w[14] = bytes32(offer.maxUnits);
         w[15] = bytes32(offer.maxSellerAssets);
         w[16] = bytes32(offer.maxBuyerAssets);
+        bytes32 result;
         assembly ("memory-safe") {
             result := keccak256(w, 0x220)
         }
+        return result;
     }
 }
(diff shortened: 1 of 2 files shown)


src/ratifiers/HashLib.sol
```diff
@@ -40,32 +40,27 @@ library HashLib {
     }
 
     /// @dev Computes the EIP-712 hash struct of an Offer.
```
MathisGD, line 42:
```suggestion
    /// @dev Computes the EIP-712 hash struct of an Offer.
    /// @dev Same as kecccak256(abi.encode(OFFER_TYPEHASH, ...));
```

src/ratifiers/HashLib.sol
```diff
@@ -40,32 +40,28 @@ library HashLib {
...
+        w[10] = keccak256(offer.callbackData);
+        w[11] = bytes32(uint256(uint160(offer.receiverIfMakerIsSeller)));
+        w[12] = bytes32(uint256(uint160(offer.ratifier)));
+        w[13] = bytes32(uint256(offer.reduceOnly ? 1 : 0));
+        w[14] = bytes32(offer.maxUnits);
+        w[15] = bytes32(offer.maxSellerAssets);
+        w[16] = bytes32(offer.maxBuyerAssets);
+        assembly ("memory-safe") {
+            result := keccak256(w, 0x220)
+        }
```
MathisGD, line 66:
```suggestion
        return keccak256(abi.encodePacked(w));
```

src/ratifiers/HashLib.sol
```diff
@@ -40,32 +40,28 @@ library HashLib {
...
-                    offer.receiverIfMakerIsSeller,
-                    offer.ratifier,
-                    offer.reduceOnly,
-                    offer.maxUnits,
-                    offer.maxSellerAssets,
-                    offer.maxBuyerAssets
-                )
-            )
-        );
+    /// @dev Same as kecccak256(abi.encode(OFFER_TYPEHASH, ...));
```
MathisGD:
```suggestion
    /// @dev Same as keccak256(abi.encode(OFFER_TYPEHASH, ...));
```

src/ratifiers/HashLib.sol
```diff
@@ -40,32 +40,28 @@ library HashLib {
...
+    function hashOffer(Offer memory offer) internal pure returns (bytes32 result) {
```
MathisGD:
```suggestion
    function hashOffer(Offer memory offer) internal pure returns (bytes32) {
```

MathisGD (approved):

src/ratifiers/HashLib.sol
MathisGD: actually we re-lose the gas...

src/ratifiers/HashLib.sol
```diff
@@ -40,32 +40,30 @@ library HashLib {
...
-                    offer.reduceOnly,
-                    offer.maxUnits,
-                    offer.maxSellerAssets,
-                    offer.maxBuyerAssets
-                )
-            )
-        );
+        bytes32[17] memory w;
+        w[0] = OFFER_TYPEHASH;
+        w[1] = hashObligation(offer.obligation);
```
peyha, line 47: would it be worth doing the same for `hashObligation` ?

peyha (approved):

src/ratifiers/HashLib.sol
adhusson: i tried and the gain was super low

QGarchery (approved):

src/ratifiers/HashLib.sol
adhusson: i think i measured wrong https://github.com/morpho-org/midnight/pull/843/changes

PR #836, opened 2026-05-11
show special consumed offer take case

adhusson:
see https://morpholabs.slack.com/archives/C08A40KKN9W/p1778491989906199

we could also fix


src/Midnight.sol
```diff
@@ -157,6 +157,7 @@ import {EventsLib} from "./libraries/EventsLib.sol";
 /// @dev Zero checks are not systematically performed.
 /// @dev No-ops are allowed. In particular, Midnight can call the callback of offers through a no-op take, even if those
 /// offers are "filled" (consumed=max).
+/// @dev It is possible to gift units to a a fully consumed buy offer with price < WAD.
```
MathisGD: it's symmetric 
```suggestion
/// @dev It is possible to gift units to a fully consumed buy offer with price < WAD.
/// @dev It is possible to gift assets to a fully consumed sell offer with price < WAD.
```

MathisGD (approved):

src/Midnight.sol
adhusson: i think this is only when sellerPrice==0, but then it seems natural, are you sure you want to include it?

MathisGD: >i think this is only when sellerPrice==0

I don't think so. If you have a price=0.5, a taker taking asset=1 won't receive any units, effectively making you an asset donation.

adhusson: with a sell offer, if the buyer sets an input such that `sellerAssets > 0` we get to

```
        if (offer.maxSellerAssets > 0) {
            newConsumed = consumed[offer.maker][offer.group] += sellerAssets;
            require(newConsumed <= offer.maxSellerAssets, ConsumedSellerAssets());
```

and the call reverts. It's not symmetric because a fully consumed offer that gives a `maxBuyerAssets` will not revert on `ConsumedBuyerAssets`, and there is no check that units is 0 there.

And actually when `sellerPrice == 0`, it's not a donation to the taker, it's a donaton to the fee recipient because we get `buyerAssets > 0` and `sellerAssets == 0`

MathisGD: no sorry that would be a unit based sell offer*

MathisGD: then:
```suggestion
/// @dev It is possible to gift units to a fully consumed buyerAssets based buy offer with price < WAD.
/// @dev It is possible to gift assets to a fully consumed units based sell offer with price < WAD.
```

adhusson:
```suggestion
/// @dev It is possible to gift units to a a fully consumed buyerAssets based offer with price < WAD.
```

wdyt (because with units limits, units must be zero, so buyer assets & seller assets must be zero)

MathisGD: right, mb

```suggestion
/// @dev It is possible to gift units to a fully consumed buyerAssets-based buy offer with price < WAD.
```

MathisGD (approved):

src/Midnight.sol
```diff
@@ -156,6 +156,7 @@ import {EventsLib} from "./libraries/EventsLib.sol";
 /// @dev Zero checks are not systematically performed.
 /// @dev No-ops are allowed. In particular, Midnight can call the callback of offers through a no-op take, even if those
 /// offers are "filled" (consumed=max).
+/// @dev It is possible to gift units to a fully consumed buyerAssets-based buy offer with price < WAD.
```
peyha: not a huge fan of the term "gift" here
wdyt of 

```suggestion
/// @dev It is possible to take a few units on a fully consumed buyerAssets-based buy offer with price < WAD.
```

adhusson: donate?

adhusson:
```suggestion
/// @dev It is possible to donate units to a fully consumed buyerAssets-based buy offer with price < WAD by taking it.
```

peyha: donate looks similar to gift, the thing that I don't like is that it makes me think of giving assets for free while the taker still has to collateralize it/repay it

MathisGD: don't like "donate"
```suggestion
/// @dev It is possible to give units to a fully consumed buyerAssets-based buy offer with price < WAD.
```

MathisGD: also it would be "by taking it with units small enough such that..."

adhusson: but do we need this precision? the important part is to watch out for it? but open to it

MathisGD (approved):

QGarchery (approved):

PR #837, opened 2026-05-11
Add comment for the same obligation

QGarchery:
Motivation is that the collateral supply/withdraw loop uses `takes[0].offer.obligation` as the obligation, but has specific inputs for the collateral index (eg `collateralSupplies[i].collateralIndex`). 
So it assumes that it always uses the same obligation. But in fact this is not always checked because the take loop could be stopped early (if the amount is filled)

Todo:
- [x] fix compilation (crazy that it does stack too deep, when the first commit of this PR only removes lines)
- [x] fix tests

### DIFF
--- a/src/periphery/MidnightBundles.sol
+++ b/src/periphery/MidnightBundles.sol
@@ -21,7 +21,7 @@ contract MidnightBundles is IMidnightBundles {
     /// @dev This function pulls maxBuyerAssets from the msg.sender and transfers back the remaining tokens at the end.
     /// @dev The msg.sender will pay at most maxBuyerAssets.
     /// @dev Total loan-token cost is filledBuyerAssets + filledBuyerAssets * pct / (WAD - pct).
-    function unitsTargetBuyAndWithdrawCollateral(
+    function buyWithUnitsTargetAndWithdrawCollateral(
         address midnight,
         uint256 targetUnits,
         uint256 maxBuyerAssets,
@@ -87,7 +87,7 @@ contract MidnightBundles is IMidnightBundles {
     /// @dev The msg.sender should have approved the bundler to transfer enough collateral.
     /// @dev The receiver will receive at least minSellerAssets.
     /// @dev Total receipt is filledSellerAssets - filledSellerAssets * pct / WAD.
-    function supplyCollateralAndUnitsTargetSell(
+    function supplyCollateralAndSellWithUnitsTarget(
         address midnight,
         uint256 targetUnits,
         uint256 minSellerAssets,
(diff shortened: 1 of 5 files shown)


src/periphery/MidnightBundles.sol
```diff
@@ -33,17 +34,14 @@ contract MidnightBundles is IMidnightBundles {
...
-        SafeTransferLib.safeTransferFrom(loanToken, msg.sender, address(this), maxBuyerAssets);
+        _forceApproveMax(takes[0].offer.obligation.loanToken, midnight);
+        SafeTransferLib.safeTransferFrom(takes[0].offer.obligation.loanToken, msg.sender, address(this), maxBuyerAssets);
 
         uint256 filledUnits;
         uint256 filledBuyerAssets;
         for (uint256 i; i < takes.length && filledUnits < targetUnits; i++) {
             require(!takes[i].offer.buy, InconsistentSide());
-            require(IMidnight(midnight).toId(takes[i].offer.obligation) == id, InconsistentObligation());
             try IMidnight(midnight)
```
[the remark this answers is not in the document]
QGarchery: This is documented at the top of the function

src/periphery/MidnightBundles.sol
adhusson, line 1: I'm not sure I understand the motivation. The collateral loop does not look at the rest of the takes array, only its first element, so I don't see the connection between "the loop could exit early" and the part about the withdrawals. 

And for the early exit, the `require` does enforce that every executed take is consistent with the the first element. So even if the non-executed tail of the array makes no sense, the `require` is useful.

QGarchery: I realize that the PR description is badly written: to be clear I don't think there is an security issue with the current code.
Still don't you think the code is a bit misleading and that callers might expect that different obligations can be passed ? Notably, it doesn't necessarily revert if a bundle has different obligations. 
Worst case:
- the caller thinks that each `CollateralTransfer` corresponds to the `Take` of the same index
- the caller accepts accepts a bundle based on this assumption, where a collateral index corresponds to another obligation
- the taker ends up supplying/withdrawing some collateral he didn't want want to supply/withdraw, on the "wrong" obligation

A comment might be enough though, so we could keep the `require`

adhusson: In favor of a comment (the one you added + maybe even another one about the fact that withdrawals are also on that same obligation), and still enforcing that executed takes all use the same obligations.

MathisGD (approved):

src/periphery/MidnightBundles.sol
MathisGD: in favor of removing the require. it will help with stack too deep too. but fine with both

QGarchery: @peyha helps us choose, I don't have a strong opinion either :)

peyha: tough choice, in my current mental model for bundles, we minimize the chances of users losing funds with conceivably correct input at a first sight. in this case, I don't see how the user would lose funds so a comment is fine by me

adhusson (approved):

peyha (approved):

PR #841, opened 2026-05-11
PoC rate ratifier

QGarchery:
Notes:
- implements a rate ratifier, where the tick is replaced by a limit rate, takers can choose a tick that leads to better rate for the maker than the limit rate (and a worse rate for them)
  - we could check that the taker chose the best tick easily, but we could not compute that tick easily (it requires `priceToTick`)
- the signed message is called a rate offer which, compared to an offer, replaces the tick by the limit price (happens to have the same `uint256` type)
- we could also have a Merkle tree of rate offers, I chose not to for simplicity in this POC
- see comments below for implementation details and points of discussion


src/ratifiers/RateRatifier.sol
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    function isRatified(Offer memory offer, bytes memory ratifierData) external view returns (bytes32) {
+        require(msg.sender == MIDNIGHT, NotMidnight());
+        (Signature memory sig, uint256 rate) = abi.decode(ratifierData, (Signature, uint256));
+        uint256 timeToMaturity = UtilsLib.zeroFloorSub(offer.obligation.maturity, block.timestamp);
+        uint256 offerPrice = TickLib.tickToPrice(offer.tick);
+        uint256 priceLimitUp = WAD.mulDivUp(WAD, WAD + rate * timeToMaturity);
+        uint256 priceLimitDown = WAD.mulDivDown(WAD, WAD + rate * timeToMaturity);
+        require(offer.buy ? offerPrice <= priceLimitDown : offerPrice >= priceLimitUp, WorsePrice());
```
QGarchery:
```suggestion
        uint256 newValue = WAD + rate * timeToMaturity;
        require(offer.buy ? offerPrice * newValue <= 1e36 : offerPrice * newValue >= 1e36, WorsePrice());
```
this does the same thing, but `newValue` is a bit less easy to grasp (makes sense when you think about it: it's your investment increased by the interest, compared to the base line)

src/ratifiers/HashLib.sol
```diff
@@ -66,4 +67,31 @@ library HashLib {
         }
         return result;
     }
+
+    /// @dev Computes the EIP-712 hash struct of a RateOffer (offer with `tick` replaced by `rate`).
+    function hashRateOffer(Offer memory offer, uint256 rate) internal pure returns (bytes32) {
```
QGarchery: We could also factorize with `hashOffer`

src/ratifiers/interfaces/IRateRatifier.sol
```diff
@@ -0,0 +1,21 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Copyright (c) 2025 Morpho Association
+pragma solidity ^0.8.0;
+
+import {IRatifier} from "../../interfaces/IRatifier.sol";
+
+bytes constant RATE_OFFER_TYPE =
+    "RateOffer(Obligation obligation,bool buy,address maker,uint256 start,uint256 expiry,uint256 rate,bytes32 group,bytes32 session,address callback,bytes callbackData,address receiverIfMakerIsSeller,address ratifier,bool reduceOnly,uint256 maxUnits,uint256 maxSellerAssets,uint256 maxBuyerAssets)";
+/// @dev keccak256(bytes.concat(RATE_OFFER_TYPE, COLLATERAL_PARAMS_TYPE, OBLIGATION_TYPE))
+bytes32 constant RATE_OFFER_TYPEHASH = 0x90e9f9e41e2cb91ddebe5f4e0924a3f0bc32d2c1048017a241339fabf364502c;
```
QGarchery: We could reuse `OFFER_TYPEHASH` and there would be no risk of collision, since `EcrecoverRatifier` always uses a Merkle tree. Also the current typehash doesn't correspond to a concrete struct.
Still prefer current code slightly, as it separates things even more explicitly. Minor benefit: a future ratifier could use a signature over a concrete Offer without risk of collision with this ratifier

src/ratifiers/RateRatifier.sol
```diff
@@ -0,0 +1,42 @@
...
+/// no longer valid.
+/// @dev This ratifier abstracts the offer in the price dimension. The maker signs every offer field but replacing tick
+/// with rate. At ratification time the rate is converted into a price limit using simple interest.
+contract RateRatifier is IRateRatifier {
+    using UtilsLib for uint256;
+
+    address public immutable MIDNIGHT;
+
+    constructor(address _midnight) {
+        MIDNIGHT = _midnight;
```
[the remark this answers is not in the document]
QGarchery: We don't do 0 checks (this is documented)

src/ratifiers/RateRatifier.sol
adhusson: wdyt


```suggestion
        if (offer.buy) {
          uint256 priceLimitDown = WAD.mulDivDown(WAD, WAD + rate * timeToMaturity);
          require(offerPrice <= priceLimitDown, WorsePrice());
        } else {
          uint256 priceLimitUp = WAD.mulDivUp(WAD, WAD + rate * timeToMaturity);
          require(offerPrice >= priceLimitUp, WorsePrice());
        }
```

src/ratifiers/interfaces/IRateRatifier.sol
adhusson: There is no collision anyway because EIP712 includes the contract address right?

src/ratifiers/RateRatifier.sol
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    function isRatified(Offer memory offer, bytes memory ratifierData) external view returns (bytes32) {
+        require(msg.sender == MIDNIGHT, NotMidnight());
+        (Signature memory sig, uint256 rate) = abi.decode(ratifierData, (Signature, uint256));
```
adhusson, line 28: it's not good for calldata but it would be cleaner to decode a full RateOffer

adhusson: Actually the midnight offer could be constructed inside the tx by a contract before calling take, thus avoiding calldata duplication.

Since the taker doesn't want to lose money due to tx delays (when selling) or getting his tx rejected due to delays (when buying), it's anyway good for the taker if a contract builds the Midnight offer on the fly.

Actually all ratifiers could return an offer instead of receiving an offer. It would make this ratifier simpler (although it would have to use tickToPrice).

In midnight you would start with
```
  Offer memory offer = IRatifier(ratifier).onRatify(ratifierData);
```

And then all the checks happen.

src/ratifiers/interfaces/IRateRatifier.sol
adhusson: not tested

src/ratifiers/RateRatifier.sol
adhusson: alternatively, this method could point us at: 
- moving things outside of the tree (if this ratifier starts accepting trees), 
- or replacing the price field of offers with a field of arbitrary bytes so we can generalize the "sign with actual data -> replace with price" technique?

src/ratifiers/interfaces/IRateRatifier.sol
QGarchery: right, no collision in what is signed since the ratifiers are the `verifyingContract` as defined in EIP712
Still could be a bit confusing to have a `tick` field hold a rate

adhusson: agree on field confusion

src/ratifiers/RateRatifier.sol
adhusson: abandoning these ideas because added complexity for a small benefit

QGarchery:
closing for now, will reopen when needed

PR #854, opened 2026-05-12
clean max assets

QGarchery:


test/FrontendSignatureTest.sol
QGarchery, line 1: I'm guessing an LLM modified this file to make it go through. It cheated because the offers were not sorted

src/Midnight.sol
```diff
@@ -348,8 +348,7 @@ contract Midnight is IMidnight {
         if (offer.maxAssets > 0) {
             uint256 makerAssets = offer.buy ? buyerAssets : sellerAssets;
             newConsumed = consumed[offer.maker][offer.group] += makerAssets;
```
```suggestion
newConsumed = consumed[offer.maker][offer.group] += offer.buy ? buyerAssets : sellerAssets;
```

PR #859, opened 2026-05-13
remove assembly in hashOffer

MathisGD:
works now that the offer is smaller (and we don't loose anymore because of the double abi.encode)


src/ratifiers/HashLib.sol
```diff
@@ -114,28 +114,25 @@ library HashLib {
...
+                offer.group,
+                offer.callback,
+                keccak256(offer.callbackData),
+                offer.receiverIfMakerIsSeller,
+                offer.ratifier,
+                offer.reduceOnly,
+                offer.maxUnits,
+                offer.maxAssets
+            )
+        );
```
QGarchery, line 136: No more gas gains then

QGarchery (approved):

adhusson (approved):

PR #867, opened 2026-05-13
Tick rounding 2

adhusson:
adjusts the rounding and the max tick so that
- tick 0 is the only tick that maps to price 0
- max_tick is the onlyu ticks that maps to price 1
- max_tick is divisible by 4


src/libraries/TickLib.sol
```diff
@@ -3,9 +3,8 @@
 pragma solidity ^0.8.0;
 
 int256 constant LN_ONE_PLUS_DELTA = 0.004987541511039073e18; // floor(ln(1.005) * 1e18)
-uint256 constant MAX_TICK = 5820;
-// Minimum representable price increment in WAD (1e-6 WAD). Tick prices are rounded to multiples of this value.
-uint256 constant PRICE_ROUNDING_STEP = 1e12;
+uint256 constant MAX_TICK = 5908;
```
adhusson, line 6:
```suggestion
uint256 constant MAX_TICK = 5908;
// Minimum representable price increment in WAD (1e-6 WAD). Tick prices are rounded to multiples of this value.
```

test/TickLibTest.sol
```diff
@@ -12,10 +12,12 @@ contract TickLibTest is BaseTest {
 
     // Tick to price
 
+    /// @dev With `testTickMonotonicity`, `tickToPrice(1) > 0` implies tick 0 is the unique zero-price tick, and
+    /// `tickToPrice(MAX_TICK - 1) < 1e18` implies `MAX_TICK` is the unique par-price tick.
```
adhusson:
```suggestion
```

test/TickLibTest.sol
```diff
@@ -14,8 +14,8 @@ contract TickLibTest is BaseTest {
 
     function testTickToPriceMinMax() public pure {
         assertEq(TickLib.tickToPrice(0), 0, "tick 0");
-        assertEq(TickLib.tickToPrice(2), 1e12, "first non-zero tick");
-        assertEq(TickLib.tickToPrice(MAX_TICK - 2), 1e18 - 1e12, "tick max - 2 just below par");
+        assertEq(TickLib.tickToPrice(1), 0.8e12, "first non-zero tick is at d (= 0.8e12)");
```
adhusson:
```suggestion
        assertEq(TickLib.tickToPrice(1), 0.8e12, "first non-zero tick");
```

PR #871, opened 2026-05-13
test: fix sig

peyha:


test/frontend/sign-root.ts
```diff
@@ -87,8 +87,8 @@ function defaultOffer(number: string) {
 // WARNING: The tree should be built by sorting the nodes in ascending order of their hash.
 function buildOfferTree() {
   return [
-    [defaultOffer("2"), defaultOffer("4")],
-    [defaultOffer("3"), defaultOffer("1")],
+    [defaultOffer("1"), defaultOffer("2")],
+    [defaultOffer("3"), defaultOffer("4")],
```
adhusson, line 91: note it already worked before

adhusson (approved):

MathisGD (approved):

test/FrontendSignatureTest.sol
QGarchery, line 48:
```suggestion
        bytes32 _root = HashLib.commutativeHash(left, right);

        require(uint256(h0) < uint256(h1), "h0 < h1");
        require(uint256(h2) < uint256(h3), "h2 < h3");
        require(uint256(left) < uint256(right), "left < right");
```
can you also add that so that we know more rapidly if it fails because the order is wrong

QGarchery (approved):

PR #872, opened 2026-05-13
document withdraw race (spearbit-8 and blackthorn-34)

MathisGD:


src/Midnight.sol
```diff
@@ -24,6 +24,9 @@ import {IMidnight, Market, Offer, CollateralParams, MarketState, Position} from
 /// particular, the RCF is "inactive", meaning liquidations can always liquidate everything.
 /// - It has no liquidation incentive, so liquidators repay at exactly the oracle price (plus roundings).
 /// @dev To check if a market has been touched, check if tickSpacing(marketId) > 0.
+/// @dev When some assets becomes withdrawable before the maturity (after a repayment or a liquidation), external 
+/// players will probably take resting sell offers with price<1 and withdraw instantly. Lenders (and the fee claimer)
```
```suggestion
/// @dev When some assets becomes withdrawable before the maturity (after a repayment or a liquidation), external
/// players will probably take resting sell offers with price<1 and withdraw instantly. Lenders (and the fee claimer)
```

adhusson (approved):

src/Midnight.sol
```diff
@@ -24,6 +24,9 @@ import {IMidnight, Market, Offer, CollateralParams, MarketState, Position} from
 /// particular, the RCF is "inactive", meaning liquidations can always liquidate everything.
 /// - It has no liquidation incentive, so liquidators repay at exactly the oracle price (plus roundings).
 /// @dev To check if a market has been touched, check if tickSpacing(marketId) > 0.
+/// @dev When some assets becomes withdrawable before the maturity (after a repayment or a liquidation), external
+/// players will probably take resting sell offers with price<1 and withdraw instantly. Lenders (and the fee claimer)
+/// might also race to withdraw first.
```
```suggestion
/// @dev When some assets become withdrawable before the maturity (after a repayment or a liquidation), there
/// is an incentive to take resting sell offers with price<1 and withdraw instantly. Lenders (and the fee claimer)
/// might also race to withdraw first.
```
"external players -> there is an incentive" because it applies to the user doing the interaction (repay or liquidate) as well

QGarchery (approved):

src/Midnight.sol
```diff
@@ -24,6 +24,9 @@ import {IMidnight, Market, Offer, CollateralParams, MarketState, Position} from
 /// particular, the RCF is "inactive", meaning liquidations can always liquidate everything.
 /// - It has no liquidation incentive, so liquidators repay at exactly the oracle price (plus roundings).
 /// @dev To check if a market has been touched, check if tickSpacing(marketId) > 0.
+/// @dev When some assets become withdrawable before the maturity (after a repayment or a liquidation), there
+/// is an incentive to take resting sell offers with price<1 and withdraw instantly. Lenders (and the fee claimer)
```
peyha, line 28: why mention the fee claimer ?

peyha (approved):

src/Midnight.sol
MathisGD: he has some credit indirectly, he might race too

src/Midnight.sol
```diff
@@ -24,6 +24,9 @@ import {IMidnight, Market, Offer, CollateralParams, MarketState, Position} from
...
+/// might also race to withdraw first.
```
adhusson, line 29:
```suggestion
/// @dev As it is usually profitable to withdraw assets before maturity, early repayments may trigger a race to withdraw.
```

QGarchery: It's not just withdraw, it's also taking sell offers. Important point to me is that it applies to liquidators, so that it documents that we didn't do pre-repay on purpose

adhusson: the comment is supposed to encompass taking sell offers without going into details, because the goal of taking the offer is to participate in the race to withdraw

MathisGD: I would rather be explicit about it, I think that the current form is fine

Saw-mon-and-Natalie:
Look good up to this point for [Cantina - Finding 8](https://cantina.xyz/code/45b68aa9-34ba-4eef-ba96-8fa07c0b240a/findings/8)

PR #880, opened 2026-05-15
[Certora] TickLib

QGarchery:
Notes:
- split the verification in 2 files, since the `priceToTick` spec uses a summarized version of `tickToPrice`
- the `tickToPrice` spec makes use of the fact that this function is increasing. This is proven by exhaustive testing, but could not be proven easily in CVL, because wExp is not actually an increasing function as is (see https://github.com/morpho-org/midnight/pull/882)

EDIT: after #882 we can improve (this will be done in another PR), and prove directly in CVL that wExp is non-decreasing, instead of relying on the fact that tickToPrice is non-decreasing is exhaustively tested

[corresponding thread](https://morpholabs.slack.com/archives/C0ACSA8TNMA/p1779281147970369)


certora/specs/TickToPrice.spec
```diff
@@ -0,0 +1,43 @@
...
+
+rule tickToPriceIsOneAtMaxTick() {
+    assert tickToPrice(maxTick()) == 10 ^ 18;
+}
+
+// Tick to price is at most 1e18.
+// This notably ensures that offer prices are at most 1e18.
+rule tickToPriceAtMostWad(uint256 tick) {
+    assert tickToPrice(tick) <= 10 ^ 18;
+}
```
MathisGD, line 50: can we prove that it's increasing?

certora/specs/PriceToTick.spec
```diff
@@ -0,0 +1,55 @@
...
+    return ghostTickToPrice(tick);
+}
+
+rule priceToTickIsMonotonic(uint256 price1, uint256 price2, uint256 spacing) {
+    assert price1 < price2 => priceToTick(price1, spacing) <= priceToTick(price2, spacing);
+}
+
+// For prices smaller than 1, priceToTick returns a tick with a price greater than or equal to the input price.
+rule priceToTickReturnsATickWithGreaterThanOrEqualPrice(uint256 price, uint256 spacing) {
+    require price <= 10 ^ 18, "assume that the price is smaller than 1";
```
MathisGD, line 34: why doesn't it work for 1?

certora/specs/PriceToTick.spec
```diff
@@ -0,0 +1,55 @@
...
+    }
+    return ghostTickToPrice(tick);
+}
+
+rule priceToTickIsMonotonic(uint256 price1, uint256 price2, uint256 spacing) {
+    assert price1 < price2 => priceToTick(price1, spacing) <= priceToTick(price2, spacing);
+}
+
+// For prices smaller than 1, priceToTick returns a tick with a price greater than or equal to the input price.
+rule priceToTickReturnsATickWithGreaterThanOrEqualPrice(uint256 price, uint256 spacing) {
```
MathisGD, line 33: not for this PR, but is that really the property that we want? or closest price?

certora/specs/TickToPrice.spec
QGarchery: I didn't find a clean way to do it, because `wExp` is not increasing (see #882)

certora/specs/PriceToTick.spec
QGarchery: `priceToTick` works by starting with the range of ticks [0, MAX_TICK] and reducing that range until it's a singleton. So it always returns a tick that is <= MAX_TICK, which means that the price of the returned tick is smaller than or equal to 1

certora/specs/PriceToTick.spec
QGarchery: not sure but it's nice to see on what side you will land, and you can find the closest tick easily from that (try `tick - spacing` and see if it has a more accurate price)

certora/specs/PriceToTick.spec
MathisGD: but if I pass price=1 it will return MAX_TICK (and price(MAX_TICK)=1) right? or maybe we need to assume that the price was "reachable"

certora/specs/TickToPrice.spec
MathisGD: after 882 it works?

certora/specs/PriceToTick.spec
QGarchery: price is 1 works (note that the require is that price <= 1 so it doesn't prevent it). What doesn't work, for this rule, is if price > 1. Then it will return MAX_TICK, and so it's not a tick with a price greater or equal

MathisGD: oh right I didn't read well

certora/specs/TickToPrice.spec
QGarchery: I have a proof for why wExp is increasing after 882, but I didn't check that the prover was able to find it

MathisGD (approved):

certora/specs/TickToPrice.spec
```diff
@@ -0,0 +1,43 @@
...
+
+rule cvlMaxTickIsMaxTick() {
+    assert cvlMaxTick() == maxTick();
+}
+
+// Check the casting assertions in the wExp function.
+rule wExpCasting(uint256 x) {
+    require x >= 0, "wExp calls wExp(-x) when x < 0";
+
+    mathint ln2 = 693147180559945309;
```
lilCertora, line 20: why not using the offset value of TickLib.sol `offset = 0.32261121498945987e18;` ? r ranges are different then no ? maybe i'm misunderstanding something

certora/specs/TickToPrice.spec
lilCertora, line 50: maybe we could add a rule to prove that `tickToPrice(tick) % PRICE_ROUNDING_STEP == 0` ?

lilCertora (approved):
LGTM with some comments

certora/specs/TickToPrice.spec
QGarchery: good catch, this was written before the #882 change and I forgot to update it in this PR

QGarchery:
@prd-carapulse[bot] fix both comments, in this PR (don't open a new one)

certora/specs/TickToPrice.spec
jhoenicke: Monotonicity is tricky.  First showing that r - (r*r)/2e18 is monotone for -ln2 < r < ln2 require reasoning about rounding.  Should be doable though.  Then this implies that  1e18 + (r - secondTerm) + thirdTerm is monotone.

Then the wrap around requires to show that wExp(r1) <= 2*wExp(r2) for all -offset <= r1, r2 < ln2 - offset.   Since we know it that wExp is monotone, a unit test for r1 =  ln2-offset-1 and r2=-offset would be enough.

Strict monotonicity doesn't hold for r < 1, since the derivative of wExp is < 1.

PR #885, opened 2026-05-18
[Certora] Document reentrancy

### LINKED ISSUE #522: verif: account for callbacks in slashing
Agree that it is too strong to not assume reentrancy. I still think that the current assumption is fine but it deserves a longer comment. The reasoning is the following:
- for checking if the balance can or cannot change (when checking for a different user, or a different obligation id), the property that is interesting is if the function itself can change the balance. So yes the callback of a take can change the balance of another random user, but it's still interesting to show that the body of `take` itself cannot change it
- for the positive/negative balances, the corresponding functions don't have callbacks except `liquidate`. For `liquidate` it's also not true that the balance of the borrower after is <= 0, but it's still interesting to show that the body of liquidate doesn't make the borrower a lender. So it's the same reason as the point above
- for exact balances change, for example `assert makerAfter == makerPostSlash + delta;` the important point is that the callback cannot affect the delta. So even with reentry and multiple interactions, the changes expected are the cumulative expected changes of all the interactions. To make that clearer we could also define the delta before the corresponding function is even called

_Originally posted by @QGarchery in https://github.com/morpho-org/midnight/pull/423#discussion_r2939976915_

QGarchery:
Fixes: https://github.com/morpho-org/midnight/issues/522


certora/specs/LossFactor.spec
```diff
@@ -21,7 +21,8 @@ methods {
     function SafeTransferLib.safeTransfer(address, address, uint256) internal => NONDET;
     function SafeTransferLib.safeTransferFrom(address, address, address, uint256) internal => NONDET;
 
-    // External calls are assumed non-reentrant.
+    // External calls are assumed non-reentrant: this is justified as we verify properties about the function's bodies.
+    // External calls are assumed non-reverting: we verify that reverts do not happen in the function's bodies.
```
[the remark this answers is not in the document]
QGarchery: there are some `@withrevert` occurrences

certora/specs/WithdrawableMonotonicity.spec
```diff
@@ -64,6 +64,18 @@ rule claimDecreasesClaimableTradingFee(env e, address token, uint256 amount, add
     assert claimableTradingFee(token) == before - amount;
 }
 
+rule takeIncreasesClaimableTradingFee(env e, uint256 units, address taker, address takerCallback, bytes takerCallbackData, address receiverIfTakerIsSeller, Midnight.Offer offer, bytes ratifierData, address anyToken) {
```
QGarchery: Adding this rule as:
- even though it is somewhat redundant with `takeInputOutputConsistency`, it fits nicely here
- it proves another assertion, that for different loan token in `take` it doesn't change the trading fee
- it is important to show why the hook is only updated for values that go up in Solvency.spec (see added comment there)

certora/specs/BalanceEffects.spec
```diff
@@ -19,8 +19,7 @@ methods {
     function TickLib.tickToPrice(uint256) internal returns (uint256) => NONDET;
 
     // Assume no reentrancy: callbacks and token transfers do not re-enter Midnight.
-    // This is justified because the properties we verify are about the effect of each function's own
-    // body on credit and debt, not the effect of the full transaction including callbacks.
+    // This is justified because the properties we verify are about the effect of each function's own body on credit and debt, not the effect of the full transaction including callbacks.
```
lilCertora, line 22: Minor nit: the assumption is even slightly stronger than reentrancy, NONDET on `_.transfer` also kills the external state effect (token balances), but the spec never reads token balances, so it's fine

lilCertora (approved):

certora/specs/BalanceEffects.spec
QGarchery: I think it's fine, especially since token contracts are not in the scene. So they are left abstract anyway

certora/specs/CreatedMarkets.spec
```diff
@@ -29,7 +29,7 @@ methods {
     // Sound because the protocol doesn't use toMarket.
     function IdLib.storeInCode(Midnight.Market memory, uint256) internal returns (address) => NONDET;
 
-    // Tokens are assumed to not reenter.
+    // Tokens are assumed to not reenter, for performance reasons.
```
MathisGD, line 32: doesn't pass if we remove?

MathisGD (approved):

bhargavbh (approved):

certora/specs/CreatedMarkets.spec
QGarchery: This spec is very heavy, so I prefer to leave it as is. But yes in theory it's not needed, that's why I added "for performance reasons". Looking into it [here](https://prover.certora.com/output/66603/c7764c3578164c428151a752891ff479?anonymousKey=f28c698cea1718751d8025e6029a16df7110bf18)

QGarchery: indeed it seems like this run is too slow

PR #888, opened 2026-05-19
docs: approval edge case

MathisGD:
fixes https://cantina.xyz/code/6b7e5600-4961-4cad-a9bd-4f952789d07f/findings/7

### DIFF
--- a/src/periphery/MidnightBundles.sol
+++ b/src/periphery/MidnightBundles.sol
@@ -20,6 +20,7 @@ import {TakeAmountsLib} from "./TakeAmountsLib.sol";
 import {ConsumableUnitsLib} from "./ConsumableUnitsLib.sol";
 import {WAD} from "../libraries/ConstantsLib.sol";
 
+/// @dev Unusable with tokens that do not allow approve(..., type(uint256).max) after approve(..., 0).
 contract MidnightBundles is IMidnightBundles {
     using UtilsLib for uint256;
 


src/periphery/MidnightBundles.sol
QGarchery, line 1: Should we have the same requirement on the tokens that we have on Midnight ? It seems free since we depend on Midnight, and it would invalidate similar issues in the future

MathisGD: yes see #890 for this

QGarchery: So then do we also need this PR ? Not sure, but I think this behavior is not ERC-20 compatible ?

QGarchery (approved):

src/periphery/MidnightBundles.sol
MathisGD: right, but doesn't cost much neither

adhusson (approved):

src/periphery/MidnightBundles.sol
Saw-mon-and-Natalie: > but I think this behavior is not ERC-20 compatible

I don't think the `EIP-20` specifically prevents such behaviours.

src/periphery/MidnightBundles.sol
```diff
@@ -20,6 +20,7 @@ import {TakeAmountsLib} from "./TakeAmountsLib.sol";
 import {ConsumableUnitsLib} from "./ConsumableUnitsLib.sol";
 import {WAD} from "../libraries/ConstantsLib.sol";
 
+/// @dev Unusable with tokens that do not allow approve(..., type(uint256).max) after approve(..., 0).
 contract MidnightBundles is IMidnightBundles {
```
Saw-mon-and-Natalie, line 24: some tokens don't even allow `approve(..., 0)`

PR #890, opened 2026-05-19
minor changes

MathisGD:
fixes https://cantina.xyz/code/6b7e5600-4961-4cad-a9bd-4f952789d07f/findings/1

### DIFF
--- a/src/periphery/interfaces/IEcrecoverAuthorizer.sol
+++ b/src/periphery/interfaces/IEcrecoverAuthorizer.sol
@@ -16,12 +16,10 @@ struct Authorization {
     uint256 deadline;
 }
 
-/// @dev keccak256("Authorization(address authorizer,address authorized,bool isAuthorized,uint256 nonce,uint256
-/// deadline)")
-bytes32 constant AUTHORIZATION_TYPEHASH = 0x81d0284fb0e2cde18d0553b06189d6f7613c96a01bb5b5e7828eade6a0dcac91;
+bytes32 constant AUTHORIZATION_TYPEHASH =
+    keccak256("Authorization(address authorizer,address authorized,bool isAuthorized,uint256 nonce,uint256 deadline)");
 
-/// @dev keccak256("EIP712Domain(uint256 chainId,address verifyingContract)")
-bytes32 constant EIP712_DOMAIN_TYPEHASH = 0x47e79534a245952e8b16893a336b85a3d9ea9fa8c573f3d803afb92a79469218;
+bytes32 constant EIP712_DOMAIN_TYPEHASH = keccak256("EIP712Domain(uint256 chainId,address verifyingContract)");
 
 interface IEcrecoverAuthorizer {
     /// ERRORS ///
--- a/src/periphery/interfaces/IEcrecoverAuthorizer.sol
+++ b/src/periphery/interfaces/IEcrecoverAuthorizer.sol
@@ -16,16 +16,18 @@ struct Authorization {
     uint256 deadline;
 }
 
-bytes32 constant AUTHORIZATION_TYPEHASH =
-    keccak256("Authorization(address authorizer,address authorized,bool isAuthorized,uint256 nonce,uint256 deadline)");
+/// @dev keccak256("Authorization(address authorizer,address authorized,bool isAuthorized,uint256 nonce,uint256 deadline)")
+bytes32 constant AUTHORIZATION_TYPEHASH = 0x81d0284fb0e2cde18d0553b06189d6f7613c96a01bb5b5e7828eade6a0dcac91;
 
-bytes32 constant EIP712_DOMAIN_TYPEHASH = keccak256("EIP712Domain(uint256 chainId,address verifyingContract)");
+/// @dev keccak256("EIP712Domain(uint256 chainId,address verifyingContract)")
+bytes32 constant EIP712_DOMAIN_TYPEHASH = 0x47e79534a245952e8b16893a336b85a3d9ea9fa8c573f3d803afb92a79469218;
 
 interface IEcrecoverAuthorizer {
     /// ERRORS ///
     error Expired();
     error InvalidNonce();
     error InvalidSignature();
+    error Unauthorized();
 
     /// EVENTS ///
     event SetIsAuthorized(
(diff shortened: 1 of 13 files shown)


src/periphery/interfaces/IEcrecoverAuthorizer.sol
```diff
@@ -16,16 +16,18 @@ struct Authorization {
     uint256 deadline;
 }
 
-bytes32 constant AUTHORIZATION_TYPEHASH =
-    keccak256("Authorization(address authorizer,address authorized,bool isAuthorized,uint256 nonce,uint256 deadline)");
+/// @dev keccak256("Authorization(address authorizer,address authorized,bool isAuthorized,uint256 nonce,uint256 deadline)")
+bytes32 constant AUTHORIZATION_TYPEHASH = 0x81d0284fb0e2cde18d0553b06189d6f7613c96a01bb5b5e7828eade6a0dcac91;
```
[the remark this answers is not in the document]
MathisGD: @AikidoSec ignore: it's not a secret

src/periphery/interfaces/IEcrecoverAuthorizer.sol
```diff
@@ -16,16 +16,19 @@ struct Authorization {
...
 
-bytes32 constant AUTHORIZATION_TYPEHASH =
-    keccak256("Authorization(address authorizer,address authorized,bool isAuthorized,uint256 nonce,uint256 deadline)");
+/// @dev keccak256("Authorization(address authorizer,address authorized,bool isAuthorized,uint256 nonce,uint256
+/// deadline)")
+bytes32 constant AUTHORIZATION_TYPEHASH = 0x81d0284fb0e2cde18d0553b06189d6f7613c96a01bb5b5e7828eade6a0dcac91;
 
-bytes32 constant EIP712_DOMAIN_TYPEHASH = keccak256("EIP712Domain(uint256 chainId,address verifyingContract)");
+/// @dev keccak256("EIP712Domain(uint256 chainId,address verifyingContract)")
+bytes32 constant EIP712_DOMAIN_TYPEHASH = 0x47e79534a245952e8b16893a336b85a3d9ea9fa8c573f3d803afb92a79469218;
```
QGarchery, line 24: I assume this type change is there for consistency, but I think it's not worth it: now we need to make sure that the given hash is what is advertised. If we still do it, then we should also have a test for it

src/ratifiers/interfaces/ISetterRatifier.sol
```diff
@@ -12,7 +12,7 @@ interface ISetterRatifier is IRatifier {
     error NotRatified();
 
     /// EVENTS ///
-    event SetIsRootRatified(address indexed maker, bytes32 indexed root, bool newIsRootRatified);
+    event SetIsRootRatified(address caller, address indexed maker, bytes32 indexed root, bool newIsRootRatified);
```
peyha: with the same idea, should ECRecoverRatifier's "CancelRoot" also log the caller ?

src/periphery/interfaces/IEcrecoverAuthorizer.sol
MathisGD: applied

QGarchery (approved):

peyha (approved):

src/ratifiers/libraries/HashLib.sol
```diff
@@ -5,16 +5,13 @@ pragma solidity ^0.8.0;
 import {Offer, Market, CollateralParams} from "../../interfaces/IMidnight.sol";
 
 bytes constant COLLATERAL_PARAMS_TYPE = "CollateralParams(address token,uint256 lltv,uint256 maxLif,address oracle)";
-/// @dev keccak256(COLLATERAL_PARAMS_TYPE)
-bytes32 constant COLLATERAL_PARAMS_TYPEHASH = 0xaf44a88eb50ebdbbebd980e5a23045c44f61ece5f80ab708a1bbe8718102e6af;
+bytes32 constant COLLATERAL_PARAMS_TYPEHASH = keccak256(COLLATERAL_PARAMS_TYPE);
```
adhusson: Constants in solidity are [recomputed every time they are used](https://docs.soliditylang.org/en/v0.8.35/contracts.html#constants:~:text=For%20a%20constant%20variable%2C%20the%20expression%20assigned%20to%20it%20is%20copied%20to%20all%20the%20places%20where%20it%20is%20accessed%20and%20also%20re%2Devaluated%20each%20time.) so this will increase the gas cost.

Note if we do 

```
keccak256("CollateralParams(address token,uint256 lltv,uint256 maxLif,address oracle)");
```

then it becomes a compile time evaluated constant and there is no re-evaluation.

MathisGD: >then it becomes a compile time evaluated constant and there is no re-evaluation.

why? not sure to understand that from the solidity docs

adhusson: I don't either, i just tried it.

adhusson: Could make uniform by inlining more here

https://github.com/morpho-org/midnight/blob/8de7076fb1edd52c39694ba09c65f6bbef2a9c25/src/ratifiers/interfaces/IEcrecoverRatifier.sol#L13

https://github.com/morpho-org/midnight/blob/8de7076fb1edd52c39694ba09c65f6bbef2a9c25/src/periphery/interfaces/IEcrecoverAuthorizer.sol#L19-L23

MathisGD: at least if we hardcode, it's clear that we won't have such issues

src/ratifiers/interfaces/IEcrecoverRatifier.sol
```diff
@@ -21,7 +22,7 @@ interface IEcrecoverRatifier is IRatifier {
     error Unauthorized();
 
     /// EVENTS ///
-    event CancelRoot(address indexed maker, bytes32 indexed root);
+    event CancelRoot(address caller, address indexed maker, bytes32 indexed root);
```
MathisGD:
```suggestion
    event CancelRoot(address indexed caller, address indexed maker, bytes32 indexed root);
```

src/ratifiers/interfaces/ISetterRatifier.sol
MathisGD:
```suggestion
    event SetIsRootRatified(address indexed caller, address indexed maker, bytes32 indexed root, bool newIsRootRatified);
```

adhusson (approved):

PR #891, opened 2026-05-19
[Certora] no debt without collateral improvement

lilCertora:
wishlist item  : "no position can have collateral = 0 everywhere and some debt, see https://github.com/morpho-org/midnight/pull/764#discussion_r3187642960"

Comes in complementary if the invariant in collateralBitmap

[corresponding thread](https://morpholabs.slack.com/archives/C0ACSA8TNMA/p1779295287758879)


certora/confs/NoDebtWithoutCollateral.conf
```diff
@@ -0,0 +1,24 @@
...
+  "loop_iter": 10,
+  "optimistic_hashing": true,
+  "hashing_length_bound": 2048,
+  "smt_timeout": 7200,
+  "prover_args": [
+    "-timeout 7200",
+    "-depth 5",
+    "-mediumTimeout 20",
+    "-destructiveOptimizations twostage",
+    "-dontStopAtFirstSplitTimeout true",
```
QGarchery:
```suggestion
```
only useful in the if there are counter-examples

certora/confs/NoDebtWithoutCollateral.conf
```diff
@@ -0,0 +1,24 @@
...
+  "solc": "solc-0.8.34",
+  "solc_via_ir": true,
+  "solc_evm_version": "osaka",
+  "optimistic_loop": true,
+  "loop_iter": 10,
+  "optimistic_hashing": true,
+  "hashing_length_bound": 2048,
+  "smt_timeout": 7200,
+  "prover_args": [
+    "-timeout 7200",
```
QGarchery:
```suggestion
```
this is useless with "smt_timeout" above, right ?

certora/specs/NoDebtWithoutCollateral.spec
```diff
@@ -0,0 +1,87 @@
...
+    // Implicit (default AUTO → HAVOC_ECF on external calls): callbacks, gates,
+    // ratifiers, oracles, and token transfers are assumed not to re-enter Midnight.
+}
+
+function summaryMulDivDown(uint256 a, uint256 b, uint256 d) returns uint256 {
+    bool overflow;
+    if (overflow || d == 0) {
+        revert();
+    }
+    return require_uint256(summaryMulDivDownM(a, b, d));
```
QGarchery:
```suggestion
    return summaryMulDivDownM(a, b, d);
```
by using `persistent ghost summaryMulDivDownM(uint256, uint256, uint256) returns uint256` instead of `persistent ghost summaryMulDivDownM(mathint, mathint, mathint) returns mathint`. Same fore mulDivUp

certora/specs/NoDebtWithoutCollateral.spec
```diff
@@ -0,0 +1,87 @@
...
+    }
+    return require_uint256(summaryMulDivUpM(a, b, d));
+}
+
+// Mirrors ConstantsLib.sol; same pattern as CollateralBitmap.spec.
+definition MAX_COLLATERALS_PER_BORROWER() returns uint256 = 10;
+
+/// MULDIV GHOSTS ///
+
+persistent ghost summaryMulDivDownM(mathint, mathint, mathint) returns mathint {
```
QGarchery:
```suggestion
persistent ghost ghostMulDivDown(mathint, mathint, mathint) returns mathint {
```
since there are not used as summaries

certora/specs/NoDebtWithoutCollateral.spec
```diff
@@ -0,0 +1,87 @@
...
+function summaryMulDivUp(uint256 a, uint256 b, uint256 d) returns uint256 {
+    bool overflow;
+    if (overflow || d == 0) {
+        revert();
+    }
+    return require_uint256(summaryMulDivUpM(a, b, d));
+}
+
+// Mirrors ConstantsLib.sol; same pattern as CollateralBitmap.spec.
+definition MAX_COLLATERALS_PER_BORROWER() returns uint256 = 10;
```
QGarchery: Let's change it (here and in `CollateralBitmap.spec`) to adding it to `Utils` instead of redefining it: more robust to changes, which might actually happen to this variable

certora/specs/NoDebtWithoutCollateral.spec
```diff
@@ -0,0 +1,87 @@
...
+
+/// MULDIV GHOSTS ///
+
+persistent ghost summaryMulDivDownM(mathint, mathint, mathint) returns mathint {
+    /* proved in mulDivZero in MulDiv.spec */
+    axiom forall uint256 b. forall uint256 d. d > 0 => summaryMulDivDownM(0, b, d) == 0;
```
QGarchery:
```suggestion
    axiom forall uint256 b. forall uint256 d. summaryMulDivDownM(0, b, d) == 0;
```
I think this is possible, because we revert for `d == 0` anyway. Motivation is that the axiom is closer to the rule that prove it in `MulDiv.spec`. 
This also appears in other specs, but it may not have the same reasoning. Just wanted to flag in case you have more input on this, but let's tackle that in a separate PR

certora/specs/NoDebtWithoutCollateral.spec
QGarchery: this will also simplify the other axioms (no need to specify that an argument is non-negative)

certora/specs/NoDebtWithoutCollateral.spec
```diff
@@ -0,0 +1,87 @@
...
+/* Axioms proved in MulDiv.spec. */
+
+/* proved in mulDivMonotoneA */
+definition axiomUpMonotoneA(mathint a1, mathint a2, mathint b, mathint d) returns bool = 0 <= a1 && a1 <= a2 && 0 <= b && 0 < d => summaryMulDivUpM(a1, b, d) <= summaryMulDivUpM(a2, b, d);
+
+/* proved in mulDivMonotoneD */
+definition axiomUpMonotoneD(mathint a, mathint b, mathint d1, mathint d2) returns bool = 0 <= a && 0 <= b && 0 < d1 && d1 <= d2 => summaryMulDivUpM(a, b, d1) >= summaryMulDivUpM(a, b, d2);
+
+/* proved in mulDivInverseUpDown */
+definition axiomInverseUpDown(mathint a, mathint b, mathint d) returns bool = a >= 0 && b > 0 && d > 0 => summaryMulDivUpM(summaryMulDivDownM(a, b, d), d, b) <= a;
```
QGarchery: Let's inline those definitions, they are only used once

certora/specs/NoDebtWithoutCollateral.spec
```diff
@@ -0,0 +1,87 @@
...
+
+// Minimal mulDiv axioms needed for liquidate's bad-debt + repayment cancellation argument.
+function requireLiquidateMulDivAxioms() {
+    require forall mathint a1. forall mathint a2. forall mathint b. forall mathint d. axiomUpMonotoneA(a1, a2, b, d);
+    require forall mathint a. forall mathint b. forall mathint d1. forall mathint d2. axiomUpMonotoneD(a, b, d1, d2);
+    require forall mathint a. forall mathint b. forall mathint d. axiomInverseUpDown(a, b, d);
+}
```
QGarchery: let's inline this function, it's only used once

certora/specs/NoDebtWithoutCollateral.spec
```diff
@@ -0,0 +1,87 @@
...
+    {
+        preserved take(Midnight.Offer offer, uint256 units, address taker, address receiverIfTakerIsSeller, address takerCallback, bytes takerCallbackData, bytes ratifierData) with (env e) {
+            // Transient storage is logically zero at the start of an externally-initiated tx.
+            require !liquidationLocked(id, taker), "transient lock zero at tx start";
+            require !liquidationLocked(id, offer.maker), "transient lock zero at tx start";
+        }
+        preserved liquidate(Midnight.Market market, uint256 collateralIndex, uint256 seizedAssets, uint256 repaidUnits, address borrower, bool healthyPath, address receiver, address callback, bytes data) with (env e) {
+            // Liquidate's debt-zeroing math (badDebt absorbed + repayment) needs mulDiv monotonicity + inverse axioms 
+            // (proved in MulDiv.spec) so the prover can derive repaidUnits >= debt_after_badDebt when the last bitmap bit is cleared.
+            require market.collateralParams.length <= MAX_COLLATERALS_PER_BORROWER(), "restrict to MAX_COLLATERALS_PER_BORROWER collaterals";
```
QGarchery: Note that this is not sound: a market can have up to 128 collaterals, but a borrower can have at most 10 of them activated at any time. I understand that we may need a lower number of loop iterations for performance reasons, but I think it's better to be explicit about it

certora/specs/NoDebtWithoutCollateral.spec
```diff
@@ -0,0 +1,87 @@
...
+    currentContract.position[id][user].collateralBitmap == 0 => currentContract.position[id][user].debt == 0
+    {
+        preserved take(Midnight.Offer offer, uint256 units, address taker, address receiverIfTakerIsSeller, address takerCallback, bytes takerCallbackData, bytes ratifierData) with (env e) {
+            // Transient storage is logically zero at the start of an externally-initiated tx.
+            require !liquidationLocked(id, taker), "transient lock zero at tx start";
+            require !liquidationLocked(id, offer.maker), "transient lock zero at tx start";
+        }
+        preserved liquidate(Midnight.Market market, uint256 collateralIndex, uint256 seizedAssets, uint256 repaidUnits, address borrower, bool healthyPath, address receiver, address callback, bytes data) with (env e) {
+            // Liquidate's debt-zeroing math (badDebt absorbed + repayment) needs mulDiv monotonicity + inverse axioms 
+            // (proved in MulDiv.spec) so the prover can derive repaidUnits >= debt_after_badDebt when the last bitmap bit is cleared.
```
QGarchery:
```suggestion
            // To derive repaidUnits >= debtAfterBadDebt when the last bitmap bit is cleared, the prover requires inverse axioms and mulDiv monotonicity (using lif <= maxLif).
```

certora/specs/NoDebtWithoutCollateral.spec
```diff
@@ -0,0 +1,87 @@
...
+
+/// INVARIANT ///
+
+weak invariant noDebtWithoutCollateral(bytes32 id, address user)
+    currentContract.position[id][user].collateralBitmap == 0 => currentContract.position[id][user].debt == 0
+    {
+        preserved take(Midnight.Offer offer, uint256 units, address taker, address receiverIfTakerIsSeller, address takerCallback, bytes takerCallbackData, bytes ratifierData) with (env e) {
+            // Transient storage is logically zero at the start of an externally-initiated tx.
+            require !liquidationLocked(id, taker), "transient lock zero at tx start";
+            require !liquidationLocked(id, offer.maker), "transient lock zero at tx start";
```
QGarchery:
```suggestion
```
and prove the following strong invariant instead: `liquidationLocked(id, user) || currentContract.position[id][user].collateralBitmap == 0 => currentContract.position[id][user].debt == 0`. Two potential benefits:
- then we could also prove the weak invariant `!liquidationLocked(id, user)` and use that to derive the weak invariant we have currently in this spec
- then we could not have to rely on "Implicit (default AUTO → HAVOC_ECF on external calls): callbacks, gates, ratifiers, oracles, and token transfers are assumed not to re-enter Midnight."

certora/confs/NoDebtWithoutCollateral.conf
lilCertora: right

certora/specs/NoDebtWithoutCollateral.spec
lilCertora: yes that's a good idea! 
on point 2 you mean adding the `havocAllByDefault` prover arg instead of relying on default HAVOC_ECF ?

QGarchery: yes exactly

certora/specs/NoDebtWithoutCollateral.spec
```diff
@@ -0,0 +1,123 @@
...
+    function _.isRatified(Midnight.Offer, bytes) external => NONDET;
+    function _.canIncreaseCredit(address) external => NONDET;
+    function _.canIncreaseDebt(address) external => NONDET;
+    function _.canLiquidate(address) external => NONDET;
+    function _.price() external => NONDET;
+
+    // NOTE: stateful external calls (`on*` callbacks and ERC20 `transfer`/`transferFrom`) are
+    // intentionally left unresolved. Combined with `-havocAllByDefault true`, this forces the
+    // strong-invariant boundary check on `lockedOrNoDebtWithoutCollateral` to fire at every such
+    // call site.
```
QGarchery:
```suggestion
```
this is usually what strong invariants are meant to check

certora/specs/NoDebtWithoutCollateral.spec
```diff
@@ -0,0 +1,123 @@
...
+    function liquidationLocked(bytes32 id, address user) external returns (bool) envfree;
+    function Utils.maxCollateralsPerBorrower() external returns (uint256) envfree;
+
+    // Internal library summaries.
+    function UtilsLib.mulDivDown(uint256 x, uint256 y, uint256 d) internal returns (uint256) => summaryMulDivDown(x, y, d);
+    function UtilsLib.mulDivUp(uint256 x, uint256 y, uint256 d) internal returns (uint256) => summaryMulDivUp(x, y, d);
+
+    // Mirror the transient liquidation-lock slot into `ghostLocked`.
+    // This makes the persistent ghost the source of truth and sidesteps CVL's caveat that transient storage
+    // is only nullified in the base step of invariants (see https://docs.certora.com/en/latest/docs/cvl/transient.html).
```
QGarchery: I don't understand this, why is this modeling an issue ?
The problem with those summaries is that they use a persistent ghost, so that HAVOC_ALL does not impact the lock (as it should)

certora/specs/NoDebtWithoutCollateral.spec
```diff
@@ -0,0 +1,123 @@
...
+        }
+    }
+
+weak invariant liquidationLockClearedAtBoundary(bytes32 id, address user)
+    ghostLocked[id][user] == false;
+
+// Derived from `lockedOrNoDebtWithoutCollateral` and `liquidationLockClearedAtBoundary`:
+//   (ghostLocked[id][user] || (bitmap == 0 => debt == 0)) ∧ !ghostLocked[id][user] ⇒ (bitmap == 0 => debt == 0)
+// at every method boundary. The preserved block injects both invariants at the start of the
+// inductive step; the heavy mulDiv axioms and loop bound live on `lockedOrNoDebtWithoutCollateral`.
```
QGarchery:
```suggestion
```

certora/specs/NoDebtWithoutCollateral.spec
```diff
@@ -0,0 +1,123 @@
...
+/// INVARIANT ///
+
+strong invariant lockedOrNoDebtWithoutCollateral(bytes32 id, address user)
+    ghostLocked[id][user] || (currentContract.position[id][user].collateralBitmap == 0 => currentContract.position[id][user].debt == 0)
+    {
+        preserved liquidate(Midnight.Market market, uint256 collateralIndex, uint256 seizedAssets, uint256 repaidUnits, address borrower, bool healthyPath, address receiver, address callback, bytes data) with (env e) {
+            // To derive repaidUnits >= debtAfterBadDebt when the last bitmap bit is cleared, the prover requires inverse axioms and mulDiv monotonicity (using lif <= maxLif).
+            // Scope restriction (not CVL-enforced): contract allows up to MAX_COLLATERALS (128) collateralParams, but a borrower has at most
+            // MAX_COLLATERALS_PER_BORROWER (10) activated (see `nonZeroCollateralsAreActivated`); we narrow the market shape to match, for loop-unrolling tractability.
+            require market.collateralParams.length <= Utils.maxCollateralsPerBorrower(), "scope restriction: market shape narrowed to <= MAX_COLLATERALS_PER_BORROWER";
```
QGarchery:
```suggestion
            require market.collateralParams.length <= Utils.maxCollateralsPerBorrower(), "assume less than 10 collaterals so that the loop can be bounded";
```
`MAX_COLLATERALS_PER_BORROWER` changed to 16 actually, which is another reason to unlink the tow concepts

certora/specs/NoDebtWithoutCollateral.spec
```diff
@@ -0,0 +1,123 @@
...
+        
+            // Inlined axioms (proved in MulDiv.spec): mulDivUp monotonicity in a and d, and the up/down inverse.
+            // mulDivMonotoneA
+            require forall uint256 a1. forall uint256 a2. forall uint256 b. forall uint256 d. a1 <= a2 && d > 0 => ghostMulDivUp(a1, b, d) <= ghostMulDivUp(a2, b, d);
```
QGarchery:
```suggestion
            require forall uint256 a1. forall uint256 a2. forall uint256 b. forall uint256 d. a1 <= a2 && d > 0 => ghostMulDivUp(a1, b, d) <= ghostMulDivUp(a2, b, d), "see mulDivMonotoneA";
```
and same for the `require`s below

certora/specs/NoDebtWithoutCollateral.spec
```diff
@@ -0,0 +1,123 @@
...
+
+    // Pure-view external callbacks: summarized as NONDET on the return value, no storage effect.
+    // NONDET is sound here because all five interfaces declare these methods `external view`
+    // We make the summaries explicit so they are not affected by `-havocAllByDefault true`, which would model these reads as HAVOC_ALL.
+    function _.isRatified(Midnight.Offer, bytes) external => NONDET;
+    function _.canIncreaseCredit(address) external => NONDET;
+    function _.canIncreaseDebt(address) external => NONDET;
+    function _.canLiquidate(address) external => NONDET;
+    function _.price() external => NONDET;
+
```
QGarchery:
```suggestion
```
this does not seem to be needed, see [this run](https://prover.certora.com/output/66603/4c3bd78f93d64a1687fec24c522c3cbd?anonymousKey=3f693a87ff7120b84b1e6dfd99f4b70eccb4de7d)

certora/specs/NoDebtWithoutCollateral.spec
lilCertora: Wording was a bit off that's right, there are 2 reasons for the ghosts being persistent , 1[st is that the strong invariant's "Reset transient storage" check (transaction boundary) clears a real transient lock while the persistent debt/bitmap state persists which leads to a counter example on `lockedOrNoDebtWithoutCollateral` (lock=true, debt=1, bitmap=0, then reset → false).](https://prover.certora.com/output/7508195/d463ef1530d64885ad995b168f4a2028/?anonymousKey=6542e881e471779f6b69a5abacd9b2dde2f3a17b&params=%7B%2212%22%3A%7B%22index%22%3A0%2C%22ruleCounterExamples%22%3A%5B%7B%22name%22%3A%22rule_output_5.json%22%2C%22selectedRepresentation%22%3A%7B%22label%22%3A%22PRETTY%22%2C%22value%22%3A0%7D%2C%22callResolutionSingleFilter%22%3A%22%22%2C%22variablesFilter%22%3A%22%22%2C%22callTraceFilter%22%3A%22%22%2C%22variablesOpenItems%22%3A%5Btrue%2Ctrue%5D%2C%22callTraceCollapsed%22%3Atrue%2C%22rightSidePanelCollapsed%22%3Afalse%2C%22rightSideTab%22%3A%22%22%2C%22callResolutionSingleCollapsed%22%3Atrue%2C%22viewStorage%22%3Atrue%2C%22variablesExpandedArray%22%3A%22%22%2C%22expandedArray%22%3A%22%22%2C%22orderVars%22%3A%5B%22%22%2C%22%22%2C0%5D%2C%22orderParams%22%3A%5B%22%22%2C%22%22%2C0%5D%2C%22scrollNode%22%3A0%2C%22currentPoint%22%3A0%2C%22trackingChildren%22%3A%5B%5D%2C%22trackingParents%22%3A%5B%5D%2C%22trackingOnly%22%3Afalse%2C%22highlightOnly%22%3Afalse%2C%22filterPosition%22%3A0%2C%22singleCallResolutionOpen%22%3A%5B%5D%2C%22snap_drop_1%22%3Anull%2C%22snap_drop_2%22%3Anull%2C%22snap_filter%22%3A%22%22%7D%5D%7D%7D&generalState=%7B%22fileViewOpen%22%3Afalse%2C%22fileViewCollapsed%22%3Atrue%2C%22mainTreeViewCollapsed%22%3Atrue%2C%22callTraceClosed%22%3Afalse%2C%22mainSideNavItem%22%3A%22rules%22%2C%22globalResSelected%22%3Afalse%2C%22isSideBarCollapsed%22%3Atrue%2C%22isRightSideBarCollapsed%22%3Afalse%2C%22selectedFile%22%3A%22%22%2C%22fileViewFilter%22%3A%22%22%2C%22mainTreeViewFilter%22%3A%22%22%2C%22contractsFilter%22%3A%22%22%2C%22globalCallResolutionFilter%22%3A%22%22%2C%22currentRuleUiId%22%3A12%2C%22counterExamplePos%22%3A1%2C%22expandedKeysState%22%3A%2220-10-00_2-02_4-05_6-1-09_10-011_12-013_14-1-016_18-018_20%22%2C%22expandedFilesState%22%3A%5B%5D%2C%22outlinedfilterShared%22%3A%22000000000%22%7D) and the second is that `-havocAllByDefault` would havoc the non persistent ghost and during take the bitmap==0 => debt==0 can't hold at the callback boundary , the seller's debt is bumped at L414, before the callbacks, and solvency is only enforced at L476 so this would create a counter example, maybe that's not what you meant ?

QGarchery: I meant why can't we use the native transient storage value (no ghost at all):
- first reason is weak invariant that should hold (there is a weak invariant about every transient variable, that this variable is 0). Maybe it's not possible to do in CVL but we could just document that it holds, instead of an actual `weak invariant`
- For the second reason, you would still need to state the strong invariant as  `!liquidationLocked || (bitmap == 0 => debt == 0)`, but using the getter for the transient value `liquidationLocked`

certora/specs/NoDebtWithoutCollateral.spec
```diff
@@ -0,0 +1,108 @@
...
+persistent ghost ghostMulDivUp(uint256, uint256, uint256) returns uint256;
+
+/// INVARIANT ///
+
+strong invariant lockedOrNoDebtWithoutCollateral(bytes32 id, address user)
+    ghostLocked[id][user] || (currentContract.position[id][user].collateralBitmap == 0 => currentContract.position[id][user].debt == 0)
+    {
+        preserved liquidate(Midnight.Market market, uint256 collateralIndex, uint256 seizedAssets, uint256 repaidUnits, address borrower, bool healthyPath, address receiver, address callback, bytes data) with (env e) {
+            // To derive repaidUnits >= debtAfterBadDebt when the last bitmap bit is cleared, the prover requires inverse axioms and mulDiv monotonicity (using lif <= maxLif).
+            require market.collateralParams.length <= Utils.maxCollateralsPerBorrower(), "assume less than 10 collaterals so that the loop can be bounded";
```
QGarchery:
```suggestion
            require market.collateralParams.length <= 10, "assume less than 10 collaterals so that the loop can be bounded";
```

certora/specs/NoDebtWithoutCollateral.spec
lilCertora: but the strong invariant `liquidationLocked || (bitmap == 0 => debt == 0)` doesn't hold if `HAVOC_ALL` impact the lock and reset it to false while the debt has increased before the callback , and in reality the callback can't write the transient lock so we would over assume and get a spurious CEX here

QGarchery: > but the strong invariant liquidationLocked || (bitmap == 0 => debt == 0) doesn't hold if HAVOC_ALL impact the lock and reset it to false

why not ? Isn't it the same thing as `HAVOC_ALL` changing the `bitmap` or the `debt` variables ?

> and in reality the callback can't write the transient lock so we would over assume and get a spurious CEX here

curious to see the CEX

lilCertora: > why not ? Isn't it the same thing as HAVOC_ALL changing the bitmap or the debt variables ?

I mean `debt`/`bitmap` are regular storage and a reentrant call into Midnight (take, liquidate, …) really can change them via the public entrypoints, so `HAVOC_ALL` havocing them is faithful and pinning them would be unsound but the lock is transient storage so not modifiable via reentrancy.

[Here is the CEX](https://prover.certora.com/output/7508195/90bf4b929f2e49d1bd3b6aa89f451c36?anonymousKey=9934922e06101b704c9682c9364f39492c5e23ea) without the ghost : 
- `weak invariant liquidationLocked == false` fails after every method with an external call because `-havocAllByDefault` havocs the transient slot at the unresolved call, so the lock is nonzero when the method returns
- the `strong invariant liquidationLocked || (bitmap==0 ⇒ debt==0)` fails at the "Reset transient storage" step: the prover nullifies the lock at the tx boundary → `liquidationLocked` becomes false while the persistent `debt/bitmap` still hold a mid-take state (`debt>0, bitmap=0`) → violation.

QGarchery: > the lock is transient storage so not modifiable via reentrancy.

I don't think this holds: I can create a function that changes the transient storage (`take` does it) and that function could change it in arbitrary ways. This means that a reentrancy could change the transient storage in arbitrary ways. So, without further analysis, the safe modeling is that external calls havoc the transient storage.


> `weak invariant liquidationLocked == false` fails after every method with an external call because -`havocAllByDefault` havocs the transient slot at the unresolved call, so the lock is nonzero when the method returns

For this counter-example, it's showing that we are missing a lemma to make the proof go through, not that we need to constrain the behavior of the transient storage.
here we are missing the property that each external call leaves the lock unchanged. This property can first be proven by assuming that external calls leave the lock unchanged as well, and by induction it would hold in the general case

> the `strong invariant liquidationLocked || (bitmap==0 ⇒ debt==0)` fails at the "Reset transient storage" step: the prover nullifies the lock at the tx boundary → `liquidationLocked` becomes false while the persistent `debt/bitmap` still hold a mid-take state (`debt>0, bitmap=0`) → violation.

this explains the ghost, if there is no way to explain to the prover that `weak invariant liquidationLocked == false` is enough to rule out this counter-example. But we could still use a non-persistent ghost I think

certora/confs/NoDebtWithoutCollateralNativeECF.conf
```diff
@@ -0,0 +1,25 @@
+{
+  "files": [
+    "src/Midnight.sol",
+    "certora/helpers/Utils.sol"
+  ],
+  "verify": "Midnight:certora/specs/NoDebtWithoutCollateral.spec",
```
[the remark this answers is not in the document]
QGarchery: see #985

certora/specs/NoDebtWithoutCollateral.spec
```diff
@@ -0,0 +1,97 @@
...
+            requireInvariant liquidationLockClearedAtBoundary(id, user);
+        }
+    }
+
+weak invariant liquidationLockClearedAtBoundary(bytes32 id, address user)
+    !liquidationLocked(id, user);
+
+// Lemma justifying that external calls leave the liquidation lock unchanged, where callbacks are HAVOC_ECF and so cannot touch currentContract's transient storage). 
+// This is the induction step that makes the HAVOC_ECF modeling of the lock (liquidationLockClearedAtBoundary) faithful, including under reentrancy.
+rule liquidationLockNeutral(method f, env e, calldataarg args, bytes32 id, address user) filtered { f -> f.selector != sig:multicall(bytes[]).selector && f.contract == currentContract } {
```
QGarchery:
```suggestion
rule liquidationLockNeutral(method f, env e, calldataarg args, bytes32 id, address user) {
```
because `multicall` is DELETED, and we restrict the parametric contracts to Midnight only

certora/specs/NoDebtWithoutCollateral.spec
```diff
@@ -0,0 +1,97 @@
...
+
+strong invariant lockedOrNoDebtWithoutCollateral(bytes32 id, address user)
+    liquidationLocked(id, user) || (currentContract.position[id][user].collateralBitmap == 0 => currentContract.position[id][user].debt == 0)
+    {
+        preserved liquidate(Midnight.Market market, uint256 collateralIndex, uint256 seizedAssets, uint256 repaidUnits, address borrower, bool postMaturityMode, address receiver, address callback, bytes data) with (env e) {
+            // To derive repaidUnits >= debtAfterBadDebt when the last bitmap bit is cleared, the prover requires inverse axioms and mulDiv monotonicity (using lif <= maxLif).
+            require market.collateralParams.length <= 10, "assume less than 10 collaterals so that the loop can be bounded";
+        
+            // Inlined axioms (proved in MulDiv.spec): mulDivUp monotonicity in a and d, and the up/down inverse.
+            // mulDivMonotoneA
```
QGarchery:
```suggestion
```
we already have the reason string

certora/specs/NoDebtWithoutCollateral.spec
```diff
@@ -0,0 +1,97 @@
...
+            require forall uint256 a1. forall uint256 a2. forall uint256 b. forall uint256 d. a1 <= a2 && d > 0 => ghostMulDivUp(a1, b, d) <= ghostMulDivUp(a2, b, d), "see mulDivMonotoneA";
+        
+            // mulDivMonotoneD
```
QGarchery:
```suggestion
```

certora/specs/NoDebtWithoutCollateral.spec
```diff
@@ -0,0 +1,97 @@
...
+            require forall uint256 a. forall uint256 b. forall uint256 d1. forall uint256 d2. d1 > 0 && d1 <= d2 => ghostMulDivUp(a, b, d1) >= ghostMulDivUp(a, b, d2), "see mulDivMonotoneD";
+        
+            // mulDivInverseUpDown
```
QGarchery:
```suggestion
```

QGarchery (approved):

jhoenicke (approved):
While the Healthiness rules are stronger than this rule (isHealthy and collateral==0 implies debt==0), this is still useful, as it can be proved with more collaterals and larger unrolling depth.

The HAVOC_ECF trick for the liquidationLockNeutral and llClearedAtBoundary is a bit hand-weavy: it assumes that no variable changes, but since this part only needs to reason about liquidationLocked itself, it should still be sound.    The alternative would be the explicit genericSummary like in Healthiness.spec where you require that liquidationLocked is unchanged by the subcall, that is inductively guaraneteed by the Neutral rule.

MathisGD (approved):

PR #894, opened 2026-05-19
pass more args to callback

MathisGD:
- receiver is useful
- msg.sender is very useful in liquidate and flashLoans because they are not authenticated


adhusson:
Why not msg.sender in liquidation? Right now a liquidator needs to use a 2 step system with some transient variable if he wants to completely control his liquidations.

adhusson (approved):

QGarchery (approved):

PR #911, opened 2026-05-20
rcf for unhealthy liq post maturity

MathisGD:
- do not deactivate the RCF at the end, as it is bad for borrowers (for no real reason), and it create a weird incentive for liquidators to leave position slightly unhealthy to try to liquidate them at maxLIF just at maturity.
- add a unhealthyPath boolean for liquidators to be able to choose if they want to liquidate in unhealthyMode (maxLIF but RCF) or not (ramped LIF but no cap). Prevents some issues were an account is just slightly unhealthy and it makes liquidating it hard.

full context: https://morpholabs.slack.com/archives/C02N7CZ088N/p1779266663065369

### DIFF
--- a/certora/specs/LiquidationProfitability.spec
+++ b/certora/specs/LiquidationProfitability.spec
@@ -78,8 +78,8 @@ rule liquidationLifRepaidUnits(env e, Midnight.Market market, uint256 collateral
 
     uint256 seizedResult;
     uint256 repaidResult;
-    bool _unhealthy;
-    seizedResult, repaidResult = liquidate(e, market, collateralIndex, 0, repaidUnits, borrower, _unhealthy, receiver, callback, data);
+    bool _unhealthyPath;
+    seizedResult, repaidResult = liquidate(e, market, collateralIndex, 0, repaidUnits, borrower, _unhealthyPath, receiver, callback, data);
 
     mathint price = summaryPrice(market.collateralParams[collateralIndex].oracle);
 
@@ -100,8 +100,8 @@ rule liquidationLifSeizedAssets(env e, Midnight.Market market, uint256 collatera
 
     uint256 seizedResult;
     uint256 repaidResult;
-    bool _unhealthy;
-    seizedResult, repaidResult = liquidate(e, market, collateralIndex, seizedAssets, 0, borrower, _unhealthy, receiver, callback, data);
+    bool _unhealthyPath;
+    seizedResult, repaidResult = liquidate(e, market, collateralIndex, seizedAssets, 0, borrower, _unhealthyPath, receiver, callback, data);
 
     mathint price = summaryPrice(market.collateralParams[collateralIndex].oracle);
 
--- a/certora/specs/LiquidationProfitability.spec
+++ b/certora/specs/LiquidationProfitability.spec
@@ -5,7 +5,6 @@ using Utils as Utils;
 methods {
     function multicall(bytes[]) external => HAVOC_ALL DELETE;
 
-    function isHealthy(Midnight.Market market, bytes32 id, address borrower) external returns (bool) envfree;
     function Utils.hashMarket(Midnight.Market) external returns (bytes32) envfree;
 
     // Summary to capture the oracle price so the spec can reference it in assertions.
@@ -68,13 +67,12 @@ function summaryMulDivUp(uint256 a, uint256 b, uint256 d) returns uint256 {
 
 /// LIF CHARACTERIZATION ///
 
-/// For repaidUnits input: lif >= WAD (solvency), and lif == maxLif when borrower is unhealthy or >= 15 min post-maturity (profitability).
+/// For repaidUnits input: lif >= WAD (solvency), and lif == maxLif when the unhealthy path is taken or the call is >= 15 min post-maturity (profitability).
 rule liquidationLifRepaidUnits(env e, Midnight.Market market, uint256 collateralIndex, uint256 repaidUnits, address borrower, address receiver, address callback, bytes data, bool unhealthyPath) {
     uint256 maxLif = market.collateralParams[collateralIndex].maxLif;
     require maxLif >= WAD(), "see the rule maxLifIsAtLeastWad";
 
-    bytes32 id = toId(e, market);
-    bool maxLifReached = !isHealthy(market, id, borrower) || e.block.timestamp >= require_uint256(market.maturity + TIME_TO_MAX_LIF());
+    bool maxLifReached = unhealthyPath || e.block.timestamp >= require_uint256(market.maturity + TIME_TO_MAX_LIF());
 
     uint256 seizedResult;
     uint256 repaidResult;
@@ -85,17 +83,16 @@ rule liquidationLifRepaidUnits(env e, Midnight.Market market, uint256 collateral
     // lif >= WAD: liquidator receives collateral worth at least the repaid debt (up to 1 unit floor rounding on seizedAssets) at the oracle price.
     assert (seizedResult + 1) * price >= repaidResult * ORACLE_PRICE_SCALE();
 
-    // lif == maxLif when borrower is unhealthy or >= 15 min post-maturity: full liquidation incentive factor applies.
+    // lif == maxLif when the unhealthy path is taken or >= 15 min post-maturity: full liquidation incentive factor applies.
     assert maxLifReached => (seizedResult + 1) * price * WAD() + ORACLE_PRICE_SCALE() * WAD() > repaidResult * maxLif * ORACLE_PRICE_SCALE();
 }
 
-/// For seizedAssets input: lif >= WAD (solvency), and lif == maxLif when borrower is unhealthy or >= 15 min post-maturity (profitability).
+/// For seizedAssets input: lif >= WAD (solvency), and lif == maxLif when the unhealthy path is taken or the call is >= 15 min post-maturity (profitability).
 rule liquidationLifSeizedAssets(env e, Midnight.Market market, uint256 collateralIndex, uint256 seizedAssets, address borrower, address receiver, address callback, bytes data, bool unhealthyPath) {
     uint256 maxLif = market.collateralParams[collateralIndex].maxLif;
     require maxLif >= WAD(), "see the rule maxLifIsAtLeastWad";
 
-    bytes32 id = toId(e, market);
-    bool maxLifReached = !isHealthy(market, id, borrower) || e.block.timestamp >= require_uint256(market.maturity + TIME_TO_MAX_LIF());
+    bool maxLifReached = unhealthyPath || e.block.timestamp >= require_uint256(market.maturity + TIME_TO_MAX_LIF());
 
     uint256 seizedResult;
     uint256 repaidResult;
@@ -106,6 +103,6 @@ rule liquidationLifSeizedAssets(env e, Midnight.Market market, uint256 collatera
     // lif >= WAD: liquidator receives collateral worth at least the repaid debt (up to 1 unit ceil rounding on repaidUnits) at the oracle price.
     assert seizedResult * price > (repaidResult - 1) * ORACLE_PRICE_SCALE();
 
-    // lif == maxLif when borrower is unhealthy or >= 15 min post-maturity: full liquidation incentive factor applies.
+    // lif == maxLif when the unhealthy path is taken or >= 15 min post-maturity: full liquidation incentive factor applies.
     assert maxLifReached => seizedResult * price * WAD() + ORACLE_PRICE_SCALE() * WAD() > (repaidResult - 1) * maxLif * ORACLE_PRICE_SCALE();
 }
--- a/certora/specs/LiquidationProfitability.spec
+++ b/certora/specs/LiquidationProfitability.spec
@@ -68,15 +68,15 @@ function summaryMulDivUp(uint256 a, uint256 b, uint256 d) returns uint256 {
 /// LIF CHARACTERIZATION ///
 
 /// For repaidUnits input: lif >= WAD (solvency), and lif == maxLif when the unhealthy path is taken or the call is >= 15 min post-maturity (profitability).
-rule liquidationLifRepaidUnits(env e, Midnight.Market market, uint256 collateralIndex, uint256 repaidUnits, address borrower, address receiver, address callback, bytes data, bool unhealthyPath) {
+rule liquidationLifRepaidUnits(env e, Midnight.Market market, uint256 collateralIndex, uint256 repaidUnits, address borrower, address receiver, address callback, bytes data, bool healthyPath) {
     uint256 maxLif = market.collateralParams[collateralIndex].maxLif;
     require maxLif >= WAD(), "see the rule maxLifIsAtLeastWad";
 
-    bool maxLifReached = unhealthyPath || e.block.timestamp >= require_uint256(market.maturity + TIME_TO_MAX_LIF());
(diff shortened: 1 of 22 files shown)


src/Midnight.sol
```diff
@@ -646,7 +647,7 @@ contract Midnight is IMidnight {
                 seizedAssets = repaidUnits.mulDivDown(lif, WAD).mulDivDown(ORACLE_PRICE_SCALE, liquidatedCollatPrice);
             }
 
-            if (block.timestamp <= market.maturity) {
+            if (originalDebt > maxDebt) {
                 uint256 lltv = market.collateralParams[collateralIndex].lltv;
                 // Note that debt >= maxDebt in this branch.
```
adhusson, line 653:
```suggestion
```

not needed anymore

src/Midnight.sol
```diff
@@ -61,15 +61,16 @@ import {IMidnight, Market, Offer, CollateralParams, MarketState, Position} from
...
+/// factor), unless the liquidation could leave a collateral with a value that would not be enough to repay rcfThreshold
+/// units.
 /// @dev The "recovery close factor" (RCF) limits the amount that can be liquidated. In particular, it prevents the
 /// liquidation from putting the borrower back into health. Which means (omitting scaling and roundings):
 ///   newDebt >= newMaxDebt <=> debtOf - repaidUnits >= maxDebt - repaidUnits*LIF*LLTV
 ///                         <=> repaidUnits <= (debtOf-maxDebt) / (1 - LIF*LLTV).
 /// The maxRepaid computation is rounded up to avoid consecutive max liquidations, so the position could be slightly
 /// healthy after a liquidation.
-/// @dev The RCF is deactivated after the maturity.
+/// @dev The RCF is deactivated when the borrower is healthy (which can only happen after the maturity).
```
```suggestion
/// @dev The RCF is deactivated for post-maturity healthy borrowers.
```

src/Midnight.sol
MathisGD: it's debt not maxDebt, and it's not that clear

src/Midnight.sol
adhusson: i would try splitting it in 2 cases, here there are 2 independent originalDebt>maxDebt checks and it makes the code less clear. Logically we have

```
if debt>max
  liquidate with rcf and maxlif
otherwise
  liquidate without rcf and with time based lif
```

src/Midnight.sol
adhusson: before you had to infer `now <= maturity` -> unhealthy, now you just need to know badDebt deduction brings you back still >= maxDebt.

anyway it does not hurt.

src/Midnight.sol
MathisGD: as discussed, we need a way to bypass the RCF at the end

src/Midnight.sol
```diff
@@ -613,7 +618,7 @@ contract Midnight is IMidnight {
 
         require(
             originalDebt > 0 && !liquidationLocked(id, borrower)
-                && (block.timestamp > market.maturity || originalDebt > maxDebt),
+                && (unhealthyPath ? originalDebt > maxDebt : block.timestamp > market.maturity),
```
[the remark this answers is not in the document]
MathisGD: wanted

src/Midnight.sol
```diff
@@ -61,15 +61,19 @@ import {IMidnight, Market, Offer, CollateralParams, MarketState, Position} from
...
+/// factor), unless the liquidation could leave a collateral with a value that would not be enough to repay rcfThreshold
+/// units.
 /// @dev The "recovery close factor" (RCF) limits the amount that can be liquidated. In particular, it prevents the
 /// liquidation from putting the borrower back into health. Which means (omitting scaling and roundings):
 ///   newDebt >= newMaxDebt <=> debtOf - repaidUnits >= maxDebt - repaidUnits*LIF*LLTV
 ///                         <=> repaidUnits <= (debtOf-maxDebt) / (1 - LIF*LLTV).
 /// The maxRepaid computation is rounded up to avoid consecutive max liquidations, so the position could be slightly
 /// healthy after a liquidation.
-/// @dev The RCF is deactivated after the maturity.
+/// @dev The RCF is deactivated for post-maturity healthy borrowers.
```
qdrn: I think slightly hard to read as is.

qdrn:
```suggestion
/// @dev Post-maturity the liquidator chooses between two liquidation paths via the `unhealthyPath` flag:
```
Works fine I think.

MathisGD: improved the docs

src/Midnight.sol
adhusson: its unrelated, now there are multiple `if unhealthyPath`, and it feels path dependent. But I tried doing it in 2 larger blocks and the result is not satisfying.

src/Midnight.sol
```diff
@@ -59,23 +59,25 @@ import {IMidnight, Market, Offer, CollateralParams, MarketState, Position} from
...
-/// @dev Before or at maturity, the liquidation cannot put the borrower back into health (recovery close factor), unless
-/// the liquidation could leave a collateral with a value that would not be enough to repay rcfThreshold units.
-/// @dev The "recovery close factor" (RCF) limits the amount that can be liquidated. In particular, it prevents the
-/// liquidation from putting the borrower back into health. Which means (omitting scaling and roundings):
+/// @dev There are two liquidations modes: The "unhealthy path", available if the borrower is unhealthy and the
+/// "post-maturity path", available after the market's maturity. For a unhealthy borrower after the maturity, the
+/// liquidator can choose between both modes.
+/// @dev In the "unhealthy path", the liquidation incentive factor (LIF) is maxLif and the liquidation amount is capped
+/// by what is needed to put back the position into health ("recovery close factor", or "RCF"). The RCF is deactivated
+/// if the liquidation could leave a collateral with a value that would not be enough to repay rcfThreshold.
```
adhusson: there is a repetition of this idea now (l66-67 and l74-75)

src/Midnight.sol
```diff
@@ -59,23 +59,25 @@ import {IMidnight, Market, Offer, CollateralParams, MarketState, Position} from
 /// shouldn't be locked either.
 /// @dev Liquidations are locked for the seller during the callbacks of take.
 /// @dev Liquidations can revert for other reasons, see LIVENESS.
-/// @dev If an account is healthy, the LIF (liquidation incentive factor) grows linearly from 1 at maturity to maxLif at
-/// maturity + TIME_TO_MAX_LIF.
-/// @dev Before or at maturity, the liquidation cannot put the borrower back into health (recovery close factor), unless
-/// the liquidation could leave a collateral with a value that would not be enough to repay rcfThreshold units.
-/// @dev The "recovery close factor" (RCF) limits the amount that can be liquidated. In particular, it prevents the
-/// liquidation from putting the borrower back into health. Which means (omitting scaling and roundings):
+/// @dev There are two liquidations modes: The "unhealthy path", available if the borrower is unhealthy and the
```
```suggestion
/// @dev There are two liquidation modes: The "unhealthy path", available if the borrower is unhealthy and the
```

src/Midnight.sol
```diff
@@ -59,23 +59,25 @@ import {IMidnight, Market, Offer, CollateralParams, MarketState, Position} from
...
+/// "post-maturity path", available after the market's maturity. For a unhealthy borrower after the maturity, the
```
```suggestion
/// "post-maturity path", available after the market's maturity. For an unhealthy borrower after the maturity, the
```

src/Midnight.sol
MathisGD: you think that we should remove completely from here?

src/Midnight.sol
peyha, line 676: should probably log the bool to make the lif re-computable

adhusson (approved):

src/Midnight.sol
```diff
@@ -59,23 +59,24 @@ import {IMidnight, Market, Offer, CollateralParams, MarketState, Position} from
...
-/// @dev Before or at maturity, the liquidation cannot put the borrower back into health (recovery close factor), unless
-/// the liquidation could leave a collateral with a value that would not be enough to repay rcfThreshold units.
-/// @dev The "recovery close factor" (RCF) limits the amount that can be liquidated. In particular, it prevents the
-/// liquidation from putting the borrower back into health. Which means (omitting scaling and roundings):
+/// @dev There are two liquidation modes: The "unhealthy path", available if the borrower is unhealthy and the
+/// "post-maturity path", available after the market's maturity. For an unhealthy borrower after the maturity, the
+/// liquidator can choose between both modes.
+/// @dev In the "unhealthy path", the liquidation incentive factor (LIF) is maxLif and the liquidation amount is capped
+/// by what is needed to put back the position into health ("recovery close factor", or "RCF").
+/// @dev The RCF means (omitting scaling and roundings):
```
```suggestion
/// @dev The RCF condition is (omitting scaling and roundings):
```
nitpick

src/Midnight.sol
```diff
@@ -580,6 +581,7 @@ contract Midnight is IMidnight {
         uint256 seizedAssets,
         uint256 repaidUnits,
         address borrower,
+        bool unhealthyPath,
```
```suggestion
        bool rcfPath,
```
wdyt ? looks more accurate because an unhealthy position could be liquidated with `!unhealthyPath`

peyha (approved):
lgtm, pending small naming issues
The change is small & way cleaner than I expected

src/Midnight.sol
QGarchery: good catch. Other naming option is to use `healthyPath`: it's the one with no RCF but LIF that grows after maturity

MathisGD: oh didn't think about healthyPath, that's great!

src/Midnight.sol
```diff
@@ -613,7 +615,7 @@ contract Midnight is IMidnight {
 
         require(
             originalDebt > 0 && !liquidationLocked(id, borrower)
-                && (block.timestamp > market.maturity || originalDebt > maxDebt),
+                && (unhealthyPath ? originalDebt > maxDebt : block.timestamp > market.maturity),
             NotLiquidatable()
         );
```
MathisGD, line 620: arguably the way this is written is not the most readable. more expensive, but clearer:
```solidity
        require(
            originalDebt > 0 && !liquidationLocked(id, borrower)
                && (originalDebt > maxDebt || block.timestamp > market.maturity),
            NotLiquidatable()
        );
        require(!unHealthy || originalDebt > maxDebt);
```

certora/specs/LiquidationProfitability.spec
```diff
@@ -68,44 +67,42 @@ function summaryMulDivUp(uint256 a, uint256 b, uint256 d) returns uint256 {
...
-/// For seizedAssets input: lif >= WAD (solvency), and lif == maxLif when borrower is unhealthy or >= 15 min post-maturity (profitability).
-rule liquidationLifSeizedAssets(env e, Midnight.Market market, uint256 collateralIndex, uint256 seizedAssets, address borrower, address receiver, address callback, bytes data) {
+/// For seizedAssets input: lif >= WAD (solvency), and lif == maxLif when the unhealthy path is taken or the call is >= 15 min post-maturity (profitability).
+rule liquidationLifSeizedAssets(env e, Midnight.Market market, uint256 collateralIndex, uint256 seizedAssets, address borrower, address receiver, address callback, bytes data, bool unhealthyPath) {
     uint256 maxLif = market.collateralParams[collateralIndex].maxLif;
     require maxLif >= WAD(), "see the rule maxLifIsAtLeastWad";
 
-    bytes32 id = toId(e, market);
-    bool maxLifReached = !isHealthy(market, id, borrower) || e.block.timestamp >= require_uint256(market.maturity + TIME_TO_MAX_LIF());
+    bool maxLifReached = unhealthyPath || e.block.timestamp >= require_uint256(market.maturity + TIME_TO_MAX_LIF());
```
QGarchery:
```suggestion
    bool maxLifReached = e.block.timestamp <= market.maturity || e.block.timestamp >= require_uint256(market.maturity + TIME_TO_MAX_LIF());
```
I think this spec is weaker without the suggested change: the result depends on a parameter that the liquidator passes. What we want is a property about the profitability of the liquidation, and maybe that argument constrains too much the liquidations

bhargavbh (commented):
it is not very clear how this PR interacts with the multicollat feature. 
Consider the unhealthy post-maturity case. The liquidator profits for the two modes are:
- unhealthyMode: `(debt-maxDebt)*(maxLif-1)/(1-lltv.maxLif)`. here maxLif and lltv are of the collateral being liquidated. 
- postMaturity mode: `CollatValue*((maxLif-1)*t)/ (TIME_TO_MAXLIF + (maxLif-1)*t)`, where collatValue is the available amount of collat being liquidated and t=block.timestamp - market.maturity

If you observe, the profits don't depend on the collatValue in unhealthyMode but is dependent (non-linearly) in postMaturityMode. As a result, the mode chosen is dependent on the collatValue. Higher the collatValue, earlier the time-point at which post-maturity mode is more profitable than unhealthyMode. This is luckily good for us, because we want collats with higher collatValue to be liquidated early. All to say that the liquidators optimisation problem is not straight-froward and the dynamics are hazy. 

This was not the case earlier because unhealthy post-maturity meant all collats could be fully liquidated at maxLif.

src/Midnight.sol
QGarchery, line 1: Reasons that make it better than the previous code is not so much the fact that it's nicer for borrowers, but more:
- it's clearer, it looks more like 2 functions, which makes sense.
- the [remark](https://morpholabs.slack.com/archives/C02N7CZ088N/p1779290108275409?thread_ts=1779266663.065369&cid=C02N7CZ088N) from @bhargavbh about the continuity of liquidator incentive
- the fact that it composes well. Meaning that it looks like you can have a bot for the healthy path + a bot for the unhealthy path. It was not the case before (basically re-framing the issue "Prevents some issues were an account is just slightly unhealthy and it makes liquidating it hard")

Still not sure if it makes sense to do it, because we are actually relaxing the liquidation and the optimal path (without considering gas) is not as easy

QGarchery: optimal path without gas is better described [here](https://github.com/morpho-org/midnight/pull/911#pullrequestreview-4330590349)

certora/specs/LiquidationProfitability.spec
MathisGD: the spec won't pass with this, I think. we need the maxLif to be reached, and it's not reached if isHealthy=true but healthyPass=false, before the 15 min

QGarchery: My reasoning is: either you are before the maturity, and the max lif is necessarily reached (only the unhealthy path is available). Or you are more than 15 minutes after the maturity, and no matter the path it will use the max lif

MathisGD: ok, but then it doesn't cover unhealthy in the first 15 min after the maturity

QGarchery: right, this is arguably a bigger issue than what I found. I don't see how to fix it easily, closing this comment for now

src/Midnight.sol
QGarchery: this new way of writing allows to use the healthy mode before maturity (because `originalDebt > maxDebt` works for both `require` independently of `unhealthyPath`)

QGarchery (approved):

MathisGD:
discussed live: Actually, the optimal strategy is always trying to do both liquidations. Which means that there is no new timing game that is introduced by this change.

src/Midnight.sol
MathisGD: yes it's broken actually. let's not change.

PR #912, opened 2026-05-20
[Certora] Optional fix lossFactor and createdMarket timeouts

lilCertora:


lilCertora:
checkout"ed" from bad branch that's why we have the long list of fix

PR #916, opened 2026-05-22
[Certora] lib/bundler assets/debts reachability

lilCertora:
[corresponding thread](https://morpholabs.slack.com/archives/C0ACSA8TNMA/p1779727229055869)


certora/specs/BundlerRepayInvertibility.spec
```diff
@@ -0,0 +1,83 @@
...
+// with assets = floor(U * WAD / (WAD - pct)) repays exactly U units on Midnight.
+// Composes with repayUnitsFormula: units(assets, pct) == U, hence debt decreases by U.
+rule repayAndWithdrawCollateralRepaysTargetUnits(env e, Midnight.Market market, address onBehalf, address collateralReceiver, address referralFeeRecipient, uint256 referralFeePct, uint256 U) {
+    require referralFeePct < WAD(), "PctExceeded";
+
+    bytes32 id = summaryToId(market);
+    uint256 debtBefore = midnight.debtOf(id, onBehalf);
+    require U > 0 && U <= debtBefore, "target units within current debt";
+
+    require midnight.isAuthorized(onBehalf, currentContract), "bundler must be authorized on Midnight for repay";
```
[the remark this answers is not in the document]
lilCertora: that's the idea of CVL

certora/specs/TakeAmountsLibInvertibility.spec
```diff
@@ -0,0 +1,134 @@
...
+    requireMulDivAxioms();
+
+    bytes32 id = summaryToId(offer.market);
+    require tickSpacing(id) > 0, "market is already created";
+
+    uint256 units = TakeAmountsLibHarness.buyerAssetsToUnits(e, currentContract, id, offer, targetBuyerAssets);
+
+    uint256 buyerAssets;
+    uint256 sellerAssets;
+    buyerAssets, sellerAssets = take(e, offer, ratifierData, units, taker, receiverIfTakerIsSeller, takerCallback, takerCallbackData);
```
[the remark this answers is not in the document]
lilCertora: that's the idea of CVL

certora/specs/TakeAmountsLibInvertibility.spec
```diff
@@ -0,0 +1,134 @@
...
+    bytes32 id = summaryToId(offer.market);
+    require tickSpacing(id) > 0, "market is already created";
+
+    // Mirror the library's sellerPrice computation in CVL to express the WAD precondition.
+    uint256 offerPrice = summaryTickToPrice(offer.tick);
+    uint256 timeToMaturity = e.block.timestamp <= offer.market.maturity ? assert_uint256(offer.market.maturity - e.block.timestamp) : 0;
+    uint256 fee = settlementFee(id, timeToMaturity);
+    require !offer.buy || fee <= offerPrice, "library reverts otherwise";
+    uint256 sellerPrice = offer.buy ? assert_uint256(offerPrice - fee) : offerPrice;
+    require sellerPrice <= WAD(), "invertibility requires sellerPrice <= WAD";
```
bhargavbh: this assumption is not necessary as it follows from tickToPriceAtMostWad in TickToPrice.spec. We can just add a require in the summaryTickToPrice (and justify by linking to the  tickToPriceAtMostWad rule). 
this way, we also avoid recomputing sellerPrice in CVL. 
[Prover Link](https://prover.certora.com/output/1438622/ad19dbd612d14b4c81101793f2ecdec6?anonymousKey=66bc1b00c3f4d2dad8a393cc17109b8d2579423c&params=%7B%7D&generalState=%7B%22fileViewOpen%22%3Atrue%2C%22fileViewCollapsed%22%3Atrue%2C%22mainTreeViewCollapsed%22%3Atrue%2C%22callTraceClosed%22%3Atrue%2C%22mainSideNavItem%22%3A%22rules%22%2C%22globalResSelected%22%3Afalse%2C%22isSideBarCollapsed%22%3Afalse%2C%22isRightSideBarCollapsed%22%3Atrue%2C%22selectedFile%22%3A%7B%22uiID%22%3A%22e82682%22%2C%22output%22%3A%22.certora_sources%2Fcertora%2Fspecs%2FTakeAmountsLibInvertibility.spec%22%2C%22name%22%3A%22TakeAmountsLibInvertibility.spec%22%7D%2C%22fileViewFilter%22%3A%22%22%2C%22mainTreeViewFilter%22%3A%22%22%2C%22contractsFilter%22%3A%22%22%2C%22globalCallResolutionFilter%22%3A%22%22%2C%22currentRuleUiId%22%3Anull%2C%22counterExamplePos%22%3A1%2C%22expandedKeysState%22%3A%229-10-1-02-1-04-1-1-1-08-09-1%22%2C%22expandedFilesState%22%3A%5B%5D%2C%22outlinedfilterShared%22%3A%22000000000%22%7D) with the changes

certora/specs/BundlerRepayInvertibility.spec
```diff
@@ -0,0 +1,83 @@
...
+    }
+    return require_uint256((to_mathint(a) * to_mathint(b)) / to_mathint(d));
+}
+
+definition WAD() returns uint256 = 10 ^ 18;
+
+// assets = floor(D * W / (W - p)), r = floor(assets * p / W), units = assets - r.
+// Show units = D for all D > 0 and 0 <= p < W (bundler requires p < W).
+//
+// Proof sketch Euclidean division on Dw = k(W-p)+ell and kp = k'W+ell'
```
bhargavbh: the comment here is not very clear: what does `a` stand for?
IMO its an independent property of mulDivDown and maybe we do not need a proof sketch here.

certora/specs/BundlerRepayInvertibility.spec
```diff
@@ -0,0 +1,83 @@
...
+    uint256 assets = summaryMulDivDown(D, WAD(), wMinusP);
+    uint256 r = summaryMulDivDown(assets, referralFeePct, WAD());
+    uint256 units = assert_uint256(assets - r);
+
+    assert units == D;
+}
+
+// End-to-end: for any target units U <= debtBefore, calling repayAndWithdrawCollateral
+// with assets = floor(U * WAD / (WAD - pct)) repays exactly U units on Midnight.
+// Composes with repayUnitsFormula: units(assets, pct) == U, hence debt decreases by U.
```
bhargavbh: Can we use `repayUnitsFormula` as an axiom in this rule? removes the need for relying on comments to connect the rules, similar to axioms on muldivs.

bhargavbh (changes requested):
very nice roundtrip properties to have for bundler's liveness guarantees.

certora/specs/BundlerRepayInvertibility.spec
lilCertora: a is meant to stand for `assets` but i agree it's ok without, i'm removing it

certora/specs/BundlerRepayInvertibility.spec
lilCertora: Using a ghost axiom would weaken the spec , don't you think ? it's nice to use axiom when they are proved somewhere else and not used as an assumption if possible.

certora/specs/TakeAmountsLibInvertibility.spec
lilCertora: yes I agree, it's a nice idea, thank you

certora/specs/TakeAmountsLibInvertibility.spec
```diff
@@ -0,0 +1,136 @@
...
+    function tickSpacing(bytes32) external returns (uint8) envfree;
+
+    function TakeAmountsLibHarness.buyerAssetsToUnits(address, bytes32, Midnight.Offer, uint256) external returns (uint256);
+    function TakeAmountsLibHarness.sellerAssetsToUnits(address, bytes32, Midnight.Offer, uint256) external returns (uint256);
+
+    // Deterministic id: same Market => same id. Matches what take() and TakeAmountsLib observe (both reach IdLib.toId via Midnight.toId / Midnight.touchMarket).
+    function IdLib.toId(Midnight.Market memory market, uint256, address) internal returns (bytes32) => summaryToId(market);
+
+    // Deterministic price: the same tick yields the same price for both the library and take().
+    function TickLib.tickToPrice(uint256 tick) internal returns (uint256) => summaryTickToPrice(tick);
```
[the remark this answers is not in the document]
lilCertora: it would be cleaner but doesn't add any value to the rule here, we would need to add maxTick() and tickLibWrapper to the scene which i don't think is needed

certora/specs/BundlerRepayInvertibility.spec
bhargavbh: my bad, units(assets,pct) == U was a bit unclear. Worth clarifying:
```suggestion
// Composes with repayUnitsFormula: assets - floor(assets*pct/WAD) == U, hence debt decreases by U.
```

bhargavbh (approved):
LGTM. Please update the README with the rules description.

certora/helpers/TakeAmountsLibHarness.sol
```diff
@@ -0,0 +1,25 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Copyright (c) 2025 Morpho Association
+pragma solidity ^0.8.0;
+
+import {Offer} from "../../src/interfaces/IMidnight.sol";
+import {TakeAmountsLib} from "../../src/periphery/TakeAmountsLib.sol";
+
+/// @dev Thin harness exposing TakeAmountsLib's `internal` library functions to CVL.
```
QGarchery:
```suggestion
```

certora/specs/BundlerRepayInvertibility.spec
```diff
@@ -0,0 +1,80 @@
...
+methods {
+    function Utils.hashMarket(Midnight.Market) external returns (bytes32) envfree;
+    function midnight.debtOf(bytes32 id, address user) external returns (uint128) envfree;
+    function midnight.isAuthorized(address authorizer, address authorized) external returns (bool) envfree;
+    function midnight.tickSpacing(bytes32 id) external returns (uint8) envfree;
+
+    // Deterministic market id (same pattern as Midnight.spec / TakeAmountsLibInvertibility.spec).
+    function IdLib.toId(Midnight.Market memory market, uint256, address) internal returns (bytes32) => summaryToId(market);
+
+    // Deterministic floor division; must not call MulDiv (it delegates back to UtilsLib.mulDivDown).
```
QGarchery:
```suggestion
    // Deterministic mulDivDown.
```
I don't understand the rest

certora/specs/BundlerRepayInvertibility.spec
```diff
@@ -0,0 +1,80 @@
...
+}
+
+/// HELPERS ///
+
+function summaryToId(Midnight.Market market) returns bytes32 {
+    return Utils.hashMarket(market);
+}
+
+function summaryMulDivDown(uint256 a, uint256 b, uint256 d) returns uint256 {
+    if (d == 0) {
```
QGarchery:
```suggestion
    bool overflow;
    if (d == 0 || overflow) {
```

certora/specs/BundlerRepayInvertibility.spec
```diff
@@ -0,0 +1,80 @@
...
+        revert();
+    }
+    return require_uint256((to_mathint(a) * to_mathint(b)) / to_mathint(d));
```
QGarchery:
```suggestion
    return require_uint256(a * b / d);
```

certora/specs/BundlerRepayInvertibility.spec
```diff
@@ -0,0 +1,80 @@
...
+    function midnight.isAuthorized(address authorizer, address authorized) external returns (bool) envfree;
+    function midnight.tickSpacing(bytes32 id) external returns (uint8) envfree;
+
+    // Deterministic market id (same pattern as Midnight.spec / TakeAmountsLibInvertibility.spec).
+    function IdLib.toId(Midnight.Market memory market, uint256, address) internal returns (bytes32) => summaryToId(market);
+
+    // Deterministic floor division; must not call MulDiv (it delegates back to UtilsLib.mulDivDown).
+    function UtilsLib.mulDivDown(uint256 x, uint256 y, uint256 d) internal returns (uint256) => summaryMulDivDown(x, y, d);
+
+    // Token / permit functions are irrelevant to the units formula but NONDET prevent havoc calls.
```
QGarchery:
```suggestion
    // Token / permit functions are irrelevant to the units formula, use a NONDET summary to prevent calls that havoc Midnight.
```

certora/specs/BundlerRepayInvertibility.spec
```diff
@@ -0,0 +1,80 @@
...
+    function _.allowance(address, address) external => NONDET;
```
QGarchery:
```suggestion
```
since it's view

certora/specs/BundlerRepayInvertibility.spec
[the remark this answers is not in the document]
QGarchery: codex is suggesting to add a `require`, but do we even need the one that is currently written ?

certora/specs/BundlerRepayInvertibility.spec
```diff
@@ -0,0 +1,80 @@
...
+
+// End-to-end: for any target units U <= debtBefore, calling repayAndWithdrawCollateral
+// with assets = floor(U * WAD / (WAD - pct)) repays exactly U units on Midnight.
+// Composes with repayUnitsFormula: assets - floor(assets*pct/WAD) == U, hence debt decreases by U.
+rule repayAndWithdrawCollateralRepaysTargetUnits(env e, Midnight.Market market, address onBehalf, address collateralReceiver, address referralFeeRecipient, uint256 referralFeePct, uint256 U) {
+    require referralFeePct < WAD(), "PctExceeded";
+
+    bytes32 id = summaryToId(market);
+    uint256 debtBefore = midnight.debtOf(id, onBehalf);
+    require U > 0 && U <= debtBefore, "target units within current debt";
```
QGarchery:
```suggestion
```
why is this needed ?

certora/specs/BundlerRepayInvertibility.spec
```diff
@@ -0,0 +1,80 @@
...
+    require referralFeePct < WAD(), "PctExceeded: bundler reverts otherwise";
+
+    uint256 wMinusP = assert_uint256(WAD() - referralFeePct);
+    uint256 assets = summaryMulDivDown(D, WAD(), wMinusP);
+    uint256 r = summaryMulDivDown(assets, referralFeePct, WAD());
+    uint256 units = assert_uint256(assets - r);
+
+    assert units == D;
+}
+
```
QGarchery: a bit redundant with the next rule, I would remove it

certora/specs/BundlerRepayInvertibility.spec
```diff
@@ -0,0 +1,80 @@
...
+// End-to-end: for any target units U <= debtBefore, calling repayAndWithdrawCollateral
+// with assets = floor(U * WAD / (WAD - pct)) repays exactly U units on Midnight.
+// Composes with repayUnitsFormula: assets - floor(assets*pct/WAD) == U, hence debt decreases by U.
```
QGarchery: I don't understand the last line, why do we need `repayUnitsFormula` for this ? It seems like it's proven on its own

certora/specs/BundlerRepayInvertibility.spec
```diff
@@ -0,0 +1,80 @@
...
+
+    require midnight.isAuthorized(onBehalf, currentContract), "bundler must be authorized on Midnight for repay";
+
+    uint256 wMinusP = assert_uint256(WAD() - referralFeePct);
+    uint256 assets = summaryMulDivDown(U, WAD(), wMinusP);
+
+    MidnightBundles.TokenPermit loanTokenPermit;
+
+    // PermitKind.None — avoid ERC2612 / Permit2 paths (irrelevant to units, havoc external calls).
+    require assert_uint8(loanTokenPermit.kind) == 0, "PermitKind.None == 0.";
```
QGarchery:
```suggestion
    require assert_uint8(loanTokenPermit.kind) == 0, "paths irrelevant to units, havoc external calls";
```

certora/specs/TakeAmountsLibInvertibility.spec
```diff
@@ -0,0 +1,136 @@
...
+// equality reached at MAX_TICK via tickToPriceIsOneAtMaxTick).
+persistent ghost summaryTickToPrice(uint256) returns uint256 {
+    axiom forall uint256 tick. summaryTickToPrice(tick) <= 10 ^ 18;
+}
+
+definition WAD() returns uint256 = 10 ^ 18;
+
+/// MULDIV GHOST SUMMARIES (mirrors AccrualRounding.spec) ///
+
+persistent ghost summaryMulDivDownM(mathint, mathint, mathint) returns mathint {
```
QGarchery:
```suggestion
persistent ghost ghostMulDivDown(mathint, mathint, mathint) returns mathint {
```

certora/specs/TakeAmountsLibInvertibility.spec
```diff
@@ -0,0 +1,136 @@
...
+definition axiomUpLowerBound(mathint a, mathint b, mathint d) returns bool = 0 <= a && 0 <= b && 0 < d => summaryMulDivUpM(a, b, d) * d >= a * b;
+
+/* proved in mulDivUpUpperBound */
+definition axiomUpUpperBound(mathint a, mathint b, mathint d) returns bool = 0 <= a && 0 <= b && 0 < d => summaryMulDivUpM(a, b, d) * d <= a * b + d - 1;
+
+/* proved in mulDivDownRoundsDown */
+definition axiomDownUpperBound(mathint a, mathint b, mathint d) returns bool = 0 <= a && 0 <= b && 0 < d => summaryMulDivDownM(a, b, d) * d <= a * b;
+
+/* proved in mulDivDownTightBound */
+definition axiomDownTightBound(mathint a, mathint b, mathint d) returns bool = 0 <= a && 0 <= b && 0 < d => (summaryMulDivDownM(a, b, d) + 1) * d > a * b;
```
QGarchery: let's put them as axioms in the ghost variables, that avoids defining the `requireMulDivAxioms` function

certora/specs/TakeAmountsLibInvertibility.spec
```diff
@@ -0,0 +1,136 @@
...
+/// TAKE AMOUNTS LIB INVERTIBILITY ///
+
+// Invertibility of TakeAmountsLib.buyerAssetsToUnits.
+//
+// If buyerAssetsToUnits(midnight, id, offer, T) returns u, then take(u, ...) produces buyerAssets == T.
+rule buyerAssetsReachable(env e, Midnight.Offer offer, uint256 targetBuyerAssets, address taker, address takerCallback, bytes takerCallbackData, address receiverIfTakerIsSeller, bytes ratifierData) {
+    requireMulDivAxioms();
+
+    bytes32 id = summaryToId(offer.market);
+    require tickSpacing(id) > 0, "market is already created";
```
QGarchery:
```suggestion
```
it should not be needed because `settlementFee()` reverts if the market is not created

certora/specs/TakeAmountsLibInvertibility.spec
```diff
@@ -0,0 +1,136 @@
...
+}
+
+// Invertibility of TakeAmountsLib.sellerAssetsToUnits.
+//
+// If sellerAssetsToUnits(midnight, id, offer, T) returns u, then take(u, ...) produces sellerAssets == T.
+rule sellerAssetsReachable(env e, Midnight.Offer offer, uint256 targetSellerAssets, address taker, address takerCallback, bytes takerCallbackData, address receiverIfTakerIsSeller, bytes ratifierData) {
+    requireMulDivAxioms();
+
+    bytes32 id = summaryToId(offer.market);
+    require tickSpacing(id) > 0, "market is already created";
```
QGarchery:
```suggestion
```

certora/specs/TakeAmountsLibInvertibility.spec
```diff
@@ -0,0 +1,136 @@
...
+
+    // Mirror the library's sellerPrice computation in CVL to express the WAD precondition.
+    uint256 offerPrice = summaryTickToPrice(offer.tick);
+    uint256 timeToMaturity = e.block.timestamp <= offer.market.maturity ? assert_uint256(offer.market.maturity - e.block.timestamp) : 0;
+    uint256 fee = settlementFee(id, timeToMaturity);
+    require !offer.buy || fee <= offerPrice, "library reverts otherwise";
+
```
QGarchery:
```suggestion
```
same reasoning

certora/specs/BundlerRepayInvertibility.spec
QGarchery: or maybe be more specific in the overflow condition, which would allow to user `assert_uint256` instead of `require_uint256`

certora/specs/BundlerRepayInvertibility.spec
lilCertora: fair enough

certora/specs/BundlerRepayInvertibility.spec
lilCertora: yes it is, i removed the rule

certora/specs/BundlerRepayInvertibility.spec
lilCertora: yes good idea

certora/specs/BundlerRepayInvertibility.spec
lilCertora: we don't actually, thanks

certora/specs/TakeAmountsLibInvertibility.spec
```diff
@@ -0,0 +1,136 @@
...
+// equality reached at MAX_TICK via tickToPriceIsOneAtMaxTick).
+persistent ghost summaryTickToPrice(uint256) returns uint256 {
+    axiom forall uint256 tick. summaryTickToPrice(tick) <= 10 ^ 18;
+}
+
+definition WAD() returns uint256 = 10 ^ 18;
+
+/// MULDIV GHOST SUMMARIES (mirrors AccrualRounding.spec) ///
+
+persistent ghost ghostMulDivDown(mathint, mathint, mathint) returns mathint {
```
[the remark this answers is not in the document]
lilCertora: thanks

certora/specs/BundlerRepayInvertibility.spec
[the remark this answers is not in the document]
lilCertora: no in fact

certora/specs/BundlerRepayInvertibility.spec
```diff
@@ -29,47 +28,43 @@ function summaryToId(Midnight.Market market) returns bytes32 {
...
+    // proved in mulDivUpRoundsUp
+    axiom forall uint256 b. forall uint256 d. d > 0 => ghostMulDivDown(0, b, d) == 0;
+
+    // proved in mulDivDownRoundsDown
+    axiom forall mathint a. forall mathint b. forall mathint d. 0 <= a && 0 <= b && 0 < d => ghostMulDivDown(a, b, d) * d <= a * b;
+
+    // proved in mulDivDownTightBound
+    axiom forall mathint a. forall mathint b. forall mathint d. 0 <= a && 0 <= b && 0 < d => (ghostMulDivDown(a, b, d) + 1) * d > a * b;
+}
+
```
QGarchery: Why do we have this now ? It wasn't needed in my last review

certora/specs/BundlerRepayInvertibility.spec
```diff
@@ -0,0 +1,75 @@
...
+function summaryToId(Midnight.Market market) returns bytes32 {
+    return Utils.hashMarket(market);
+}
+
+/// MULDIV GHOST SUMMARIES (mirrors TakeAmountsLibInvertibility.spec) ///
+
+// MulDiv rounding axioms (each proved in MulDiv.spec).
+persistent ghost ghostMulDivDown(mathint, mathint, mathint) returns mathint {
+    // proved in mulDivUpRoundsUp
```
QGarchery:
```suggestion
    // proved in mulDivZero
```

certora/specs/BundlerRepayInvertibility.spec
QGarchery: difference from last review: https://github.com/morpho-org/midnight/pull/916/changes/a31b6f211154421fd37c58bee4b9bba35220a684..db15b44bc66a3597e20963499d63bb2032a2355b

passing run from a31b6f2: https://prover.certora.com/output/21317/f05efe5a6b5b40b0beeb0274c479b47f?anonymousKey=80040a02be07668836cb65096eb7710cf7928769

lilCertora: I introduced them by error but due to this https://prover.certora.com/output/7508195/f0ed1006a3ad4a19a80c44279b3737b2/ , if we `assert_uint256` there is a spurious case bound on the division , I will add again the `require_uint256(a*b/d)`, this is better than the ghost version, thanks for catching

QGarchery (approved):

certora/specs/BundlerRepayInvertibility.spec
```diff
@@ -0,0 +1,63 @@
...
+    return Utils.hashMarket(market);
+}
+
+function summaryMulDivDown(uint256 a, uint256 b, uint256 d) returns uint256 {
+    if (d == 0 || a * b > max_uint256) {
+        revert();
+    }
+
+    // a*b <= max_uint256 and d >= 1  =>  a*b/d <= a*b <= max_uint256, so the cast never truncates.
+    return require_uint256(a * b / d);
```
jhoenicke: Can we use an assert_uint256 here?

```suggestion
    return assert_uint256(a * b / d);
```

certora/specs/BundlerRepayInvertibility.spec
```diff
@@ -0,0 +1,63 @@
...
+
+    bytes32 id = summaryToId(market);
+    uint256 debtBefore = midnight.debtOf(id, onBehalf);
+
+    uint256 wMinusP = assert_uint256(WAD() - referralFeePct);
+    uint256 assets = summaryMulDivDown(U, WAD(), wMinusP);
+
+    MidnightBundles.TokenPermit loanTokenPermit;
+
+    require assert_uint8(loanTokenPermit.kind) == 0, "paths irrelevant to units, havoc external calls";
```
jhoenicke:
```suggestion
    require assert_uint8(loanTokenPermit.kind) == 0, "ignore irrelevant Permit2 and ERC2612 paths to avoid summarizing external calls";
```

jhoenicke (changes requested):

certora/specs/BundlerRepayInvertibility.spec
lilCertora: i had problems doing it : https://prover.certora.com/output/7508195/f0ed1006a3ad4a19a80c44279b3737b2/

jhoenicke (approved):

certora/specs/BundlerRepayInvertibility.spec
QGarchery: can't access this link, but I also had problems in [this run](https://prover.certora.com/output/66603/30ee46747ebd48a3a14901e69fae72d8?anonymousKey=79069427593dfc91ba137216340494aae4ea05f5). Any idea what that might be ?

lilCertora: looks like an internal bug before reaching the assert_uint256 : 
```
ERROR COMMON - Had an exception while getting object to verify for rule
repayAndWithdrawCollateralRepaysTargetUnits,
java.lang.IllegalArgumentException: Failed requirement.
    at verifier.PatchingProgramWrapper.updateAssertSnippetsOf(PatchingProgramWrapper.kt:149)
 ```

File: certora/specs/TakeAmountsLibInvertibility.spec [hunk 27]
```diff
@@ -0,0 +1,116 @@
...
+
+/// TAKE AMOUNTS LIB INVERTIBILITY ///
+
+// Invertibility of TakeAmountsLib.buyerAssetsToUnits.
+//
+// If buyerAssetsToUnits(midnight, id, offer, T) returns u, then take(u, ...) produces buyerAssets == T.
+rule buyerAssetsReachable(env e, Midnight.Offer offer, uint256 targetBuyerAssets, address taker, address takerCallback, bytes takerCallbackData, address receiverIfTakerIsSeller, bytes ratifierData) {
+    bytes32 id = summaryToId(offer.market);
+
+    uint256 units = TakeAmountsLibHarness.buyerAssetsToUnits(e, currentContract, id, offer, targetBuyerAssets);
```
MathisGD, line 94: the fact that revert cases are ignored hides the fact that actually not all buyerAssets are reachable all the time. maybe at least:

```suggestion
    // reverts if the offer's price is above one (in that case not all buyerAssets are reachable.
    uint256 units = TakeAmountsLibHarness.buyerAssetsToUnits(e, currentContract, id, offer, targetBuyerAssets);
```

or a more advanced verification with @withrevert

certora/specs/BundlerRepayInvertibility.spec
MathisGD, line 1: I would rename this file, as it's not really checking invertibility. btw why don't we do the same thing for other bundles? not sure that it's really interesting though, we could also remove completely imo

MathisGD (approved):

certora/specs/BundlerRepayInvertibility.spec
QGarchery: > I would rename this file, as it's not really checking invertibility

How would you rename it ? The function `repayAndWithdrawCollateral`, from its `assets` input, computes `units(assets)`. We show that this computation is invertible, in the sense that for a target `targetUnits` you can find `assetsInput` such that `targetUnits == units(assetsInputs)`

> btw why don't we do the same thing for other bundles?

I think that it doesn't make sense for other bundles. Take `supplyCollateralAndSellWithUnitsTarget` for example:
- the fee applies to `filledSellerAssets` but this is not an input of the function. `filledSellerAssets` is the result of multiple `take`, so it is subject to asynchrony, which means that you can't easily invert it by passing the a different input
- inverting `filledSellerAssets` would serve no purpose because the taker cares about total seller assets. The `minSellerAssets` does a better job for that purpose, it allows to make sure that you receive at least a minimum of assets

 But it is interesting for `repayAndWithdrawCollateral` (and documented above the function), to be able to repay a debt fully

MathisGD: >The function repayAndWithdrawCollateral, from its assets input, computes units(assets)

mb, I forgot about the referral fee actually

certora/specs/BundlerRepayInvertibility.spec
```diff
@@ -0,0 +1,63 @@
...
+    return Utils.hashMarket(market);
+}
+
+function summaryMulDivDown(uint256 a, uint256 b, uint256 d) returns uint256 {
+    if (d == 0 || a * b > max_uint256) {
+        revert();
+    }
+    mathint temp = a * b / d;
+    assert temp >= 0 && temp <= 2 ^ 256;
+    return require_uint256(temp);
```
QGarchery:
```suggestion
    mathint result = a * b / d;
    assert result >= 0 && result <= 2 ^ 256;
    return require_uint256(result);
```

QGarchery: curious to know why this trick works 👀

lilCertora: R&D said it was some malformed assert annotations on the backend side, offered this fix.It will be deeply analyzed but it's not a soundness issue like this

QGarchery (approved):

PR #917, opened 2026-05-24
[Certora] wExp increasing

QGarchery:
- show that `wExp` is monotonic
- in turns it allows to prove that `tickToPrice` is monotonic, instead of relying on exhaustive testing

[corresponding thread](https://morpholabs.slack.com/archives/C0ACSA8TNMA/p1780569405369519)


certora/specs/TickToPriceIsMonotonic.spec
```diff
@@ -0,0 +1,17 @@
...
+
+import "TickToPrice.spec";
+
+methods {
+    function TickLib.wExp(int256 x) internal returns (uint256) => summaryWExp(x);
+}
+
+persistent ghost summaryWExp(int256) returns uint256 {
+    // The rule wExpIsMonotonicOnPositiveRange and wExpIsMonotonicOnNegativeRange in WExp.spec prove that wExp is non-decreasing.
+    axiom forall int256 x. forall int256 y. x <= y => summaryWExp(x) <= summaryWExp(y);
```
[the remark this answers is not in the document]
QGarchery: this is true by induction, which is not available in CVL

certora/specs/TickToPriceIsMonotonic.spec
```diff
@@ -0,0 +1,17 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+import "TickToPrice.spec";
+
+methods {
+    function TickLib.wExp(int256 x) internal returns (uint256) => summaryWExp(x);
+}
+
+persistent ghost summaryWExp(int256) returns uint256 {
+    // The rule wExpIsMonotonicOnPositiveRange and wExpIsMonotonicOnNegativeRange in WExp.spec prove that wExp is non-decreasing.
```
bhargavbh:
```suggestion
    // The rule wExpIsMonotonicOnPositiveRange and wExpIsMonotonicOnNegativeRange in WExp.spec prove that wExp is non-decreasing. 
```
    // The rule wExpIsMonotonicOnPositiveRange and wExpIsMonotonicOnNegativeRange in WExp.spec prove that wExp is non-decreasing. qIncreasesByZeroOrOne shows that the above two rules capture all cases exhaustively.

bhargavbh (approved):
nitpick: justify the assumption on summaryWExp clearer.

certora/specs/TickToPriceIsMonotonic.spec
QGarchery: in the end I made it more accurate by mentioning the whole section

jhoenicke:
Just a warning:  CVL division is currently not rounding the same way as solidity division for negative numbers.  There is a plan to change this in the next release, so that it also rounds towards 0 instead of towards -infinity.

This doesn't affect correctness of the rules, but adding the monotonicity of rExp alone will currently not guarantee the monotonicity of wExp for same q. I'm not sure if you noticed that verification took longer than expected.  If yes, then this may be the reason.

jhoenicke (approved):
Looks good, the case distinction is clear.

I find it a bit strange that some cases are easy for the solver, e.g. negative case goes through without case distinction and some are harder like the positive case.  It doesn't reuse the verification of the positive case, right?  There is no summary for wExp in the spec.

I did investigate a bit further, but as far as I could tell, the solver is checking with the right solidity division rounding and also handles the recursion of wExp, so I don't think there is a problem.

MathisGD (approved):

QGarchery:
Indeed it doesn't the negative case doesn't reuse the positive case

PR #918, opened 2026-05-24
[Certora] liquidation profitability fix timeout

lilCertora:


certora/specs/LiquidationProfitability.spec
```diff
@@ -23,6 +23,11 @@ methods {
     // Token transfers happen after return values are computed; irrelevant to the assertion.
     function SafeTransferLib.safeTransfer(address, address, uint256) internal => NONDET;
     function SafeTransferLib.safeTransferFrom(address, address, address, uint256) internal => NONDET;
+
+    // Sound over approximations, returns value are subset of prover paths.
+    function Midnight.liquidationLocked(bytes32, address) internal returns (bool) => NONDET;
+    function _.onLiquidate(bytes32 id, Midnight.Market market, uint256 collateralIndex, uint256 seizedAssets, uint256 repaidUnits, uint256 badDebt, address liquidator, address borrower, address receiver, bytes data) external => NONDET;
```
MathisGD: this one is not obviously sound. we need to say "we care about the function's body..." like in #885

MathisGD (approved):
see [this](https://morpholabs.slack.com/archives/C0ACSA8TNMA/p1779698612228919?thread_ts=1779453119.694479&cid=C0ACSA8TNMA)

certora/confs/LiquidationProfitability.conf
```diff
@@ -13,9 +13,14 @@
   "hashing_length_bound": 2048,
   "smt_timeout": 7200,
   "prover_args": [
-    "-timeout 3600",
-    "-split false",
-    "-destructiveOptimizations twostage -backendStrategy singleRace -smt_useLIA false -smt_useNIA true -depth 0 -s [z3:def{randomSeed=11},z3:def{randomSeed=12},z3:def{randomSeed=13},z3:def{randomSeed=14},z3:def{randomSeed=15},z3:def{randomSeed=16},z3:def{randomSeed=17},z3:def{randomSeed=18},z3:def{randomSeed=19},z3:def{randomSeed=20}]"
+    "-destructiveOptimizations",
+    "twostage",
```
QGarchery:
```suggestion
    "-destructiveOptimizations twostage",
```

certora/confs/LiquidationProfitability.conf
```diff
@@ -13,9 +13,14 @@
...
+    "-s",
+    "[z3:def{randomSeed=1},z3:def{randomSeed=2},z3:def{randomSeed=3},z3:def{randomSeed=4},z3:def{randomSeed=5},z3:def{randomSeed=6},z3:def{randomSeed=7},z3:def{randomSeed=8},z3:def{randomSeed=9},z3:def{randomSeed=10}]",
+    "-split false"
```
QGarchery, line 19: curious about what this does, I don't find it in the docs. Why did you add it in this conf ?

certora/specs/LiquidationProfitability.spec
```diff
@@ -23,6 +23,14 @@ methods {
...
     function SafeTransferLib.safeTransfer(address, address, uint256) internal => NONDET;
     function SafeTransferLib.safeTransferFrom(address, address, address, uint256) internal => NONDET;
+
+    // Sound over-approximation: NONDET return values are a superset of the prover paths.
+    function Midnight.liquidationLocked(bytes32, address) internal returns (bool) => NONDET;
+    function _.canLiquidate(address) external => NONDET;
+
+    // Assume no reentrancy: the onLiquidate callback does not re-enter Midnight.
+    // The properties we verify are about the effect of liquidate's own body on the returned (seizedAssets, repaidUnits), not the effect of the full transaction including callbacks.
+    function _.onLiquidate(bytes32 id, Midnight.Market market, uint256 collateralIndex, uint256 seizedAssets, uint256 repaidUnits, uint256 badDebt, address liquidator, address borrower, address receiver, bytes data) external => NONDET;
```
QGarchery: I'm surprised that this helps: we have only persistent ghosts, and Utils has no state. I'm thinking that we could even use `-havocAllByDefault true` here

certora/confs/LiquidationProfitability.conf
QGarchery: and same below, group "-s" with the different prover args

lilCertora: done

certora/confs/LiquidationProfitability.conf
lilCertora: `-split false` is the same as `-depth 0` basically , this is something I usually try when dealing with heavy nonlinear assertions and multiple z3 solvers

certora/specs/LiquidationProfitability.spec
lilCertora: yes that's right , we can remove that and add this prover arg

QGarchery (approved):

PR #921, opened 2026-05-26
justify 2^95-1 re-approval (Spearbit 1)

MathisGD:
fixes point 16 in https://cantina.xyz/code/6b7e5600-4961-4cad-a9bd-4f952789d07f/findings/1

### DIFF
--- a/src/periphery/MidnightBundles.sol
+++ b/src/periphery/MidnightBundles.sol
@@ -357,7 +357,7 @@ contract MidnightBundles is IMidnightBundles {
         require(returndata.length == 0 || abi.decode(returndata, (bool)));
     }
 
-    /// @dev Skips the approval entirely to save gas when the current allowance is already 2^95 - 1 (value chosen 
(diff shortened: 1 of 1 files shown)


src/periphery/MidnightBundles.sol
```diff
@@ -357,7 +357,8 @@ contract MidnightBundles is IMidnightBundles {
         require(returndata.length == 0 || abi.decode(returndata, (bool)));
     }
 
-    /// @dev Skips the approval entirely when the current allowance is already 2^95 - 1.
+    /// @dev Skips the approval entirely to save gas when the current allowance is already 2^95 - 1 (value chosen 
+    /// because some token like COMP and UNI on ethereum have a max allowance of type(uint96).max).
```
MathisGD:
```suggestion
    /// @dev Skips the approval entirely to save gas when the current allowance is already 2^95 - 1 (value chosen
    /// because some token like COMP and UNI on ethereum have a max allowance of type(uint96).max).
```

adhusson (approved):

peyha (approved):

QGarchery (approved):

PR #922, opened 2026-05-26
test: events & TickSpacingSetter

adhusson:
- add missing events test
- fix others
- add tick spacing setter success test


MathisGD (approved):

peyha (commented):
The Liquidate event is not tested with `healthyPath=true`
Take event is only tested once in ContinuousFeeTest, would be worse adding a simple test in TakeTest as well

PR #923, opened 2026-05-26
docs: same offer.buy in same group (Spearbit 1)

MathisGD:
fixed issue 18: https://cantina.xyz/code/6b7e5600-4961-4cad-a9bd-4f952789d07f/findings/1

### DIFF
--- a/src/Midnight.sol
+++ b/src/Midnight.sol
@@ -84,7 +84,8 @@ import {IMidnight, Market, Offer, CollateralParams, MarketState, Position} from 
 ///
 /// GROUPS
 /// @dev Groups are useful to have a global offered amount shared across multiple offers ("One cancels the other").
-/// @dev To work as expected, all offers in the same group should have the same max values and loan token.
+/// @dev To work as expected, all offers in the same group should have the same direction (offer.buy), max values and
+/// loan token.
 ///
 /// OFFER CAPS
 /// @dev At most one of maxAssets or maxUnits can be nonzero per offer.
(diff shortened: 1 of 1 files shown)


src/Midnight.sol
```diff
@@ -84,7 +84,8 @@ import {IMidnight, Market, Offer, CollateralParams, MarketState, Position} from
 ///
 /// GROUPS
 /// @dev Groups are useful to have a global offered amount shared across multiple offers ("One cancels the other").
-/// @dev To work as expected, all offers in the same group should have the same max values and loan token.
+/// @dev To work as expected, all offers in the same group should have the same direction (offer.buy), max values and
```
```suggestion
/// @dev To work as expected, all offers in the same group should have the same direction (offer.buy), max values and
```

src/Midnight.sol
```diff
@@ -84,7 +84,8 @@ import {IMidnight, Market, Offer, CollateralParams, MarketState, Position} from
 ///
 /// GROUPS
 /// @dev Groups are useful to have a global offered amount shared across multiple offers ("One cancels the other").
-/// @dev To work as expected, all offers in the same group should have the same max values and loan token.
+/// @dev To work as expected, all offers in the same group should have the same direction (offer.buy), max values and
```
adhusson, line 87: 'max values' can sound like it must be the same actual value.

```suggestion
/// @dev To work as expected, all offers in the same group should have the same direction (offer.buy), max value unit and
```

MathisGD: yes it must be the same actual value

adhusson: right my bad

peyha (approved):

QGarchery (approved):

PR #927, opened 2026-05-27
refactor: healthy path -> post maturity path

peyha:
To stay closer to `healthyPath`

### DIFF
--- a/src/Midnight.sol
+++ b/src/Midnight.sol
@@ -59,12 +59,12 @@ import {IMidnight, Market, Offer, CollateralParams, MarketState, Position} from 
 /// shouldn't be locked either.
 /// @dev Liquidations are locked for the seller during the callbacks of take.
 /// @dev Liquidations can revert for other reasons, see LIVENESS.
-/// @dev There are two liquidation paths: The "healthy path", available after the market's maturity and the
-/// "unhealthy path", available if the borrower is unhealthy. For an unhealthy borrower after the maturity, the
+/// @dev There are two liquidation modes: The "post-maturity mode", available after the market's maturity and the
+/// "unhealthy mode", available if the borrower is unhealthy. For an unhealthy borrower after the maturity, the
 /// liquidator can choose between both modes.
-/// @dev In the "healthy path", the LIF (liquidation incentive factor) grows linearly from 1 at maturity to maxLif
+/// @dev In the "post-maturity mode", the LIF (liquidation incentive factor) grows linearly from 1 at maturity to maxLif
 /// at maturity + TIME_TO_MAX_LIF, and the RCF is deactivated.
-/// @dev In the "unhealthy path", the liquidation incentive factor (LIF) is maxLif and the liquidation amount is capped
+/// @dev In the "unhealthy mode", the liquidation incentive factor (LIF) is maxLif and the liquidation amount is capped
 /// by what is needed to put back the position into health ("recovery close factor", or "RCF").
 /// @dev The RCF condition is (omitting scaling and roundings):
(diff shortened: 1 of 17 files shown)


src/Midnight.sol
```diff
@@ -59,8 +59,8 @@ import {IMidnight, Market, Offer, CollateralParams, MarketState, Position} from
 /// shouldn't be locked either.
 /// @dev Liquidations are locked for the seller during the callbacks of take.
 /// @dev Liquidations can revert for other reasons, see LIVENESS.
-/// @dev There are two liquidation modes: The "unhealthy path", available if the borrower is unhealthy and the
-/// "post-maturity path", available after the market's maturity. For an unhealthy borrower after the maturity, the
+/// @dev There are two liquidation modes: The "healthy path", available after the market's maturity and the
+/// "unhealthy path", available if the borrower is unhealthy. For an unhealthy borrower after the maturity, the
```
MathisGD: we want to try to avoid saying "unhealthy path", right? because you can be unhealthy yet someone might choose the healthy path

I'm not sure what to recommend... what about changing for "linear LIF path". At the end you are choosing the LIF (and RCF)

src/Midnight.sol
```diff
@@ -59,8 +59,8 @@ import {IMidnight, Market, Offer, CollateralParams, MarketState, Position} from
 /// shouldn't be locked either.
 /// @dev Liquidations are locked for the seller during the callbacks of take.
 /// @dev Liquidations can revert for other reasons, see LIVENESS.
-/// @dev There are two liquidation modes: The "unhealthy path", available if the borrower is unhealthy and the
-/// "post-maturity path", available after the market's maturity. For an unhealthy borrower after the maturity, the
+/// @dev There are two liquidation modes: The "healthy path", available after the market's maturity and the
```
```suggestion
/// @dev There are two liquidation paths: The "healthy path", available after the market's maturity and the
```

or rename the variable `healthyMode`?

src/Midnight.sol
adhusson: The variable name could be `postMaturityMode`, then the text here is


```
/// @dev There are two liquidation modes: The post-maturity mode, available after the market's maturity, and the
/// normal mode, available only if the borrower is unhealthy. For an unhealthy borrower after the maturity, the
```

src/Midnight.sol
```diff
@@ -75,7 +75,7 @@ import {IMidnight, Market, Offer, CollateralParams, MarketState, Position} from
 ///   minNewCollateral * liquidatedCollatPrice / LIF < rcfThreshold
 ///     <=> (collateral - maxRepaid * LIF / liquidatedCollatPrice) * liquidatedCollatPrice / LIF < rcfThreshold
 ///     <=> collateral * liquidatedCollatPrice / LIF - maxRepaid < rcfThreshold
-/// @dev In the "post-maturity path", the LIF (liquidation incentive factor) grows linearly from 1 at maturity to maxLif
+/// @dev In the "healthy path", the LIF (liquidation incentive factor) grows linearly from 1 at maturity to maxLif
```
adhusson: Since the post maturity / healthy / linear lif path is introduced first in the intro, I would also fully describe it first (to respect the order), and instead of saying "the rcf is deactivated" just "the liquidator can liquidate as much of the position as he wants" or similar.

src/Midnight.sol
peyha: I think rcfPath & healthyPath would work fine, postMaturity is also ok

MathisGD: ~~let's change for postMaturityMode~~

peyha:
closed after discussion (keep healthy path and !healthy will be the normal path)

PR #934, opened 2026-05-27
Update comment to clarify debt condition

MathisGD:
credit to @StErMi 

it's not trivial btw, we might want to verify this (in CVL)

### DIFF
--- a/src/Midnight.sol
+++ b/src/Midnight.sol
@@ -651,7 +651,7 @@ contract Midnight is IMidnight {
 
             if (!healthyPath) {
                 uint256 lltv = market.collateralParams[collateralIndex].lltv;
-                // Note that debt >= maxDebt in this branch.
+                // Note that debt > maxDebt in this branch.
                 uint256 maxRepaid = lltv < WAD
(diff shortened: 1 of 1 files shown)


StErMi:
Ref in cantina: https://cantina.xyz/code/81abfa1d-eee2-4afd-9c3b-8134f4d0254b/src/Midnight.sol?preview=code&comment_id=88a56e9a-b7e9-4d9e-8421-884ab4f8c129

When we are in the scenario `LLTV=WAD` and the liquidation is executed on a borrower **with** bad debt, we **can** reach the condition `_position.debt == maxDebt`. That's why on cantina I asked if the comment was referencing the "special" scenario with the collateral configured with `LLTV=WAD`.

So the comment "// Note that debt >= maxDebt in this branch." **can** be true but only for **`LLTV==WAD`**, which I think must be clarified.

I don't think that it's correct (without any clarification) to change it from `>=` to `>` given that it's above the Conditional (ternary) operator. It's confusing and misleading IMHO.

MathisGD:
indeed, the proposed comment is false. thus we don't change anything

PR #939, opened 2026-05-28
Update liquidation modes documentation in Midnight.sol

MathisGD:


src/Midnight.sol
```diff
@@ -59,10 +59,10 @@ import {IMidnight, Market, Offer, CollateralParams, MarketState, Position} from
 /// shouldn't be locked either.
 /// @dev Liquidations are locked for the seller during the callbacks of take.
 /// @dev Liquidations can revert for other reasons, see LIVENESS.
-/// @dev There are two liquidation modes: The "unhealthy path", available if the borrower is unhealthy and the
-/// "post-maturity path", available after the market's maturity. For an unhealthy borrower after the maturity, the
-/// liquidator can choose between both modes.
-/// @dev In the "unhealthy path", the liquidation incentive factor (LIF) is maxLif and the liquidation amount is capped
+/// @dev There are two liquidation modes: The "healthy path", available after the market's maturity and the "normal
```
```suggestion
/// @dev There are two liquidation paths: The "healthy path", available after the market's maturity and the "normal
```

adhusson (approved):

src/Midnight.sol
```diff
@@ -59,10 +59,10 @@ import {IMidnight, Market, Offer, CollateralParams, MarketState, Position} from
 /// shouldn't be locked either.
 /// @dev Liquidations are locked for the seller during the callbacks of take.
 /// @dev Liquidations can revert for other reasons, see LIVENESS.
-/// @dev There are two liquidation modes: The "unhealthy path", available if the borrower is unhealthy and the
-/// "post-maturity path", available after the market's maturity. For an unhealthy borrower after the maturity, the
-/// liquidator can choose between both modes.
-/// @dev In the "unhealthy path", the liquidation incentive factor (LIF) is maxLif and the liquidation amount is capped
+/// @dev There are two liquidation paths: The "healthy path", available after the market's maturity and the "normal
+/// path", available if the borrower is unhealthy. For an unhealthy borrower after the maturity, the liquidator can
+/// choose between both modes.
```
```suggestion
/// choose between both paths.
```
even if I thought we were going for mode instead of path but it touches the code so it could be done in another PR

peyha (approved):

QGarchery (approved):

PR #940, opened 2026-05-28
fix: prevent debt increase after maturity more strictly

MathisGD:
context:https://morpholabs.slack.com/archives/C08A40KKN9W/p1779721801233629

### DIFF
--- a/src/Midnight.sol
+++ b/src/Midnight.sol
@@ -387,6 +387,7 @@ contract Midnight is IMidnight {
             ? UtilsLib.toUint128(sellerPos.pendingFee.mulDivUp(sellerCreditDecrease, sellerPos.credit))
             : 0;
 
+        require(sellerDebtIncrease == 0 || block.timestamp <= offer.market.maturity, CannotIncreaseDebtPostMaturity());
         require(
             !offer.reduceOnly || (offer.buy ? buyerCreditIncrease == 0 : sellerDebtIncrease == 0),
             MakerCreditOrDebtIncreased()
@@ -472,8 +473,7 @@ contract Midnight is IMidnight {
         }
         if (!wasLocked) UtilsLib.tExchange(LIQUIDATION_LOCK_SLOT, id, seller, false);
         require(
-            position[id][seller].debt == 0 || liquidationLocked(id, seller)
-                || (block.timestamp <= offer.market.maturity && isHealthy(offer.market, id, seller)),
+            position[id][seller].debt == 0 || liquidationLocked(id, seller) || isHealthy(offer.market, id, seller),
             SellerIsLiquidatable()
         );
 
--- a/src/Midnight.sol
+++ b/src/Midnight.sol
@@ -473,7 +473,7 @@ contract Midnight is IMidnight {
         }
         if (!wasLocked) UtilsLib.tExchange(LIQUIDATION_LOCK_SLOT, id, seller, false);
         require(
-            position[id][seller].debt == 0 || liquidationLocked(id, seller) || isHealthy(offer.market, id, seller),
+            liquidationLocked(id, seller) || position[id][seller].debt == 0 || isHealthy(offer.market, id, seller),
             SellerIsLiquidatable()
         );
 
--- a/src/Midnight.sol
+++ b/src/Midnight.sol
@@ -472,10 +472,7 @@ contract Midnight is IMidnight {
             );
         }
         if (!wasLocked) UtilsLib.tExchange(LIQUIDATION_LOCK_SLOT, id, seller, false);
-        require(
-            liquidationLocked(id, seller) || isHealthy(offer.market, id, seller),
-            SellerIsLiquidatable()
-        );
+        require(liquidationLocked(id, seller) || isHealthy(offer.market, id, seller), SellerIsLiquidatable());
 
         return (buyerAssets, sellerAssets);
     }
--- a/src/Midnight.sol
+++ b/src/Midnight.sol
@@ -473,7 +473,7 @@ contract Midnight is IMidnight {
         }
         if (!wasLocked) UtilsLib.tExchange(LIQUIDATION_LOCK_SLOT, id, seller, false);
         require(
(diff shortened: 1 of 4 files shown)


src/Midnight.sol
```diff
@@ -472,8 +473,7 @@ contract Midnight is IMidnight {
         }
         if (!wasLocked) UtilsLib.tExchange(LIQUIDATION_LOCK_SLOT, id, seller, false);
         require(
-            position[id][seller].debt == 0 || liquidationLocked(id, seller)
-                || (block.timestamp <= offer.market.maturity && isHealthy(offer.market, id, seller)),
+            position[id][seller].debt == 0 || liquidationLocked(id, seller) || isHealthy(offer.market, id, seller),
```
```suggestion
            liquidationLocked(id, seller) || position[id][seller].debt == 0 || isHealthy(offer.market, id, seller),
```

test/TakeTest.sol
```diff
@@ -456,6 +456,22 @@ contract TakeTest is BaseTest {
...
+        vm.warp(timestamp);
+        otherLenderOffer.expiry = timestamp;
+        otherLenderOffer.maxUnits = units;
+        deal(address(loanToken), lender, units);
+        collateralize(market, lender, units);
+
+        vm.expectRevert(IMidnight.CannotIncreaseDebtPostMaturity.selector);
+        take(units, lender, otherLenderOffer);
+    }
+
```
MathisGD: redundant
```suggestion
```

src/Midnight.sol
```diff
@@ -472,8 +473,7 @@ contract Midnight is IMidnight {
         }
         if (!wasLocked) UtilsLib.tExchange(LIQUIDATION_LOCK_SLOT, id, seller, false);
         require(
-            position[id][seller].debt == 0 || liquidationLocked(id, seller)
-                || (block.timestamp <= offer.market.maturity && isHealthy(offer.market, id, seller)),
+            liquidationLocked(id, seller) || position[id][seller].debt == 0 || isHealthy(offer.market, id, seller),
```
MathisGD: actually isHealthy does it
```suggestion
            liquidationLocked(id, seller) || isHealthy(offer.market, id, seller),
```

src/Midnight.sol
```diff
@@ -471,11 +472,7 @@ contract Midnight is IMidnight {
             );
         }
         if (!wasLocked) UtilsLib.tExchange(LIQUIDATION_LOCK_SLOT, id, seller, false);
-        require(
-            position[id][seller].debt == 0 || liquidationLocked(id, seller)
-                || (block.timestamp <= offer.market.maturity && isHealthy(offer.market, id, seller)),
-            SellerIsLiquidatable()
-        );
+        require(liquidationLocked(id, seller) || isHealthy(offer.market, id, seller), SellerIsLiquidatable());
```
adhusson, line 476: Note that before this PR, when we had `seller.debt == 0 || liquidationLocked(id,seller) || (now <= market.maturity && healthy(id,seller))`, if we removed `seller.debt == 0`, then a postmaturity sell would always fail, but a lender could want to sell postmaturity.

adhusson, line 476: The only new take that gets allowed by this PR is a noop (0 unit) sell by someone with debt and healthy. Before it would fail because `debt > 0 && !lock &&  now > maturity`, but now it passes.

adhusson (approved):

src/Midnight.sol
peyha: unhealthy borrower cannot do the no-op after maturity while healthy borrowers can. does not seem like a huge deal but feels a bit weird conceptually

peyha (approved):
A few remarks about this change
- without the change, pre-maturity, seller with 0 debt (lender withdrawing) were skipping the liquidation lock check and health check but it's not the case anymore (which is fine because the health check will skip the oracle calls in this case) so this path will be slightly more costly in gas (one more warm tload & warm sload). The execution is essentially equivalent from a logical perspective before maturity
- without the change, after maturity, seller could increase their debt in a nested take as long as the debt was 0 at the end of the nested calls but now it's no longer possible. Nonetheless sellers can always buy before selling  in the nested calls so it should theoretically not be an issue but if the same counterpart is taken both as a seller and a buyer it might revert whenever the counterpart has increased debt at some point. All in all it might make some arbitrage impossible in very specific situations but I think it's safe because 1) we're post maturity 2) the same user would have sell/buy offers valid after maturity which is a very edge situation (and I don't see why the user would do that)

A quick example on the second remark (because it's not so clear)
Suppose we're post-maturity and Bob still has 2 offers. Bob also has collateral supplied and 5 credit 
- offer A: sell 10 unit at price 1
- offer B: buy 10 unit at price 1

In the previous situation, Alice could take buy 5 from A (triggering the liquidation lock), then re-buy 5 from A before selling 10 unit from B. In the new situation, taking both offers is impossible because either bob or alice will increase their debt at some point

src/Midnight.sol
```diff
@@ -387,6 +387,7 @@ contract Midnight is IMidnight {
             ? UtilsLib.toUint128(sellerPos.pendingFee.mulDivUp(sellerCreditDecrease, sellerPos.credit))
             : 0;
 
+        require(sellerDebtIncrease == 0 || block.timestamp <= offer.market.maturity, CannotIncreaseDebtPostMaturity());
```
```suggestion
        require(block.timestamp <= offer.market.maturity || sellerDebtIncrease == 0, CannotIncreaseDebtPostMaturity());
```
(optional) closer to `block.timestamp > offer.market.maturity => sellerDebtIncrease == 0` which is what `CannotIncreaseDebtPostMaturity` says

QGarchery (approved):

Saw-mon-and-Natalie:
> * without the change, after maturity, seller could increase their debt in a nested take as long as the debt was 0 at the end of the nested calls but now it's no longer possible.

this is true. Now the `take(...)` route is more strict. I think it would be great to have these remarks as CHANGELOGs or appended to the PR description since the linked morpho slack threads are gated to Morpho org only.

PR #943, opened 2026-05-28
document max lif in bad debt

QGarchery:
https://morpholabs.slack.com/archives/C08A40KKN9W/p1779437409778259

### DIFF
--- a/src/Midnight.sol
+++ b/src/Midnight.sol
@@ -75,8 +75,7 @@ import {IMidnight, Market, Offer, CollateralParams, MarketState, Position} from 
 ///     <=> collateral * liquidatedCollatPrice / LIF - maxRepaid < rcfThreshold
 /// @dev In the "post-maturity path", the LIF (liquidation incentive factor) grows linearly from 1 at maturity to maxLif
 /// at maturity + TIME_TO_MAX_LIF, and the RCF is deactivated.
-/// @dev In both paths, maxLif is used to determine if the account has some bad debt, so that the bad debt solvency
-/// guarantees are the same before and after maturity.
+/// @dev In both paths, maxLif is used to determine if the account has some bad debt, to always assume the worst case.
 ///
 /// SLASHING
 /// @dev When a borrower's bad debt is realized, it is socialized among lenders in this market.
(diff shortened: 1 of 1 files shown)


src/Midnight.sol
```diff
@@ -75,6 +75,8 @@ import {IMidnight, Market, Offer, CollateralParams, MarketState, Position} from
 ///     <=> collateral * liquidatedCollatPrice / LIF - maxRepaid < rcfThreshold
 /// @dev In the "post-maturity path", the LIF (liquidation incentive factor) grows linearly from 1 at maturity to maxLif
 /// at maturity + TIME_TO_MAX_LIF, and the RCF is deactivated.
+/// @dev In both paths, maxLif is used to determine if the account has some bad debt, so that the solvency guarantees
```
```suggestion
/// @dev In both paths, maxLif is used to determine if the account has some bad debt, so that bad debt solvency guarantees
```
nitpick, to not be misleading with the usual notion of solvency (healthiness)

peyha (approved):

MathisGD (approved):

src/Midnight.sol
```diff
@@ -75,6 +75,8 @@ import {IMidnight, Market, Offer, CollateralParams, MarketState, Position} from
 ///     <=> collateral * liquidatedCollatPrice / LIF - maxRepaid < rcfThreshold
 /// @dev In the "post-maturity path", the LIF (liquidation incentive factor) grows linearly from 1 at maturity to maxLif
 /// at maturity + TIME_TO_MAX_LIF, and the RCF is deactivated.
+/// @dev In both paths, maxLif is used to determine if the account has some bad debt, so that the bad debt solvency
+/// guarantees are the same before and after maturity.
```
MathisGD: maybe a bit more intuitive 
```suggestion
/// @dev In both paths, maxLif is used to determine if the account has some bad debt, to be always the worst case.
```

adhusson:
```suggestion
/// @dev In both paths, maxLif is used to determine if the account has some bad debt, to always assume the worst case.
```

src/Midnight.sol
```diff
@@ -75,6 +75,7 @@ import {IMidnight, Market, Offer, CollateralParams, MarketState, Position} from
 ///     <=> collateral * liquidatedCollatPrice / LIF - maxRepaid < rcfThreshold
 /// @dev In the "post-maturity path", the LIF (liquidation incentive factor) grows linearly from 1 at maturity to maxLif
 /// at maturity + TIME_TO_MAX_LIF, and the RCF is deactivated.
+/// @dev In both paths, maxLif is used to determine if the account has some bad debt, to always assume the worst case.
```
```suggestion
/// @dev In both modes, maxLif is used to determine if the account has some bad debt, to always assume the worst case.
```
to change if #945 is merged

MathisGD (approved):

PR #944, opened 2026-05-28
More precise `maxRepaid` computation

QGarchery:
Resources:
- [Cantina finding](https://cantina.xyz/code/6b7e5600-4961-4cad-a9bd-4f952789d07f/findings/11)
- [Corresponding thread](https://morpholabs.slack.com/archives/C08A40KKN9W/p1779347911234319)

Goals (borrowed from [this message](https://morpholabs.slack.com/archives/C08A40KKN9W/p1779873776746999?thread_ts=1779347911.234319&cid=C08A40KKN9W)):
> maxRepaid has to satisfy two properties:
> - Don't restore borrower too much into health: the position after liquidation should be only slightly healthy and not too far from health boundary.
> - Avoid consecutive maxLiquidations: its deliberately rounded up so that there is no immediate consecutive liquidation.

Call R the perfect computation of the bound of the code. `R = (debt - maxDebt) / (1 - lif * lltv)`

From the goals above, we want the max repaid to be as close to the bound of the finding:
Screenshot 2026-05-28 at 18 28 12
This bound is different from `R` by the rounding term `X`.

This means that option 2 (adding 200) and option 3 (adding `(WAD + lltv).mulDivUp(WAD, WAD * WAD - lif * lltv)`) are not really helping: they help with one property wanted, at the expense of the other.

The current code (before this PR), has the additional error term `Y` compared to `R`, defined as:
Screenshot 2026-05-28 at 18 31 11
This can be mitigated with option 1 (what this PR does), where `Y` becomes 1 at most.

We still would be off from the bound of the finding by `X`. In absolute value, `X` is less than `A / (1 - lif * lltv) + B * price / (1 - lif * lltv)` with `A` and `B` small integers constants (could be negative). Note that I didn't add a term of the form `C * lltv / (1 - lif * lltv)`  nor of the form `D * lif * lltv / (1 - lif * lltv)` because you can bound `lltv` and `lif * lltv` by 1.
Then we conclude by the following interpretation: we are at most hundreds of WEI of debt (the term `A / (1 - lif * lltv)`) and of collateral (the term `B * price / (1 - lif * lltv)`) off from being exactly at the healthy threshold. See the corresponding added test

Finally: this change cannot overflow, because we multiply `debt - maxDebt <= debt` with `1e36`, and both are below `type(uint128).max`. It cannot do a division by 0 either (there is a spec verifying that)

### DIFF
--- a/src/Midnight.sol
+++ b/src/Midnight.sol
@@ -657,8 +657,9 @@ contract Midnight is IMidnight {
                 uint256 lltv = market.collateralParams[collateralIndex].lltv;
                 // Note that debt >= maxDebt in this branch.
                 // The imprecision in this computation is at most a few hundreds WEI of collateral or loan token.
-                uint256 maxRepaid =
-                    lltv < WAD ? (_position.debt - maxDebt).mulDivUp(WAD * WAD, WAD - lif * lltv) : type(uint256).max;
+                uint256 maxRepaid = lltv < WAD
+                    ? (_position.debt - maxDebt).mulDivUp(WAD * WAD, WAD * WAD - lif * lltv)
+                    : type(uint256).max;
                 require(
                     repaidUnits <= maxRepaid
                         || _position.collateral[collateralIndex].mulDivDown(liquidatedCollatPrice, ORACLE_PRICE_SCALE)
--- a/src/Midnight.sol
+++ b/src/Midnight.sol
@@ -656,7 +656,7 @@ contract Midnight is IMidnight {
             if (!healthyPath) {
                 uint256 lltv = market.collateralParams[collateralIndex].lltv;
                 // Note that debt >= maxDebt in this branch.
-                // The imprecision in this computation is at most a few hundreds WEI of collateral or loan token.
+                // The imprecision in this computation is at most a few hundreds collateral or loan token assets.
                 uint256 maxRepaid = lltv < WAD
                     ? (_position.debt - maxDebt).mulDivUp(WAD * WAD, WAD * WAD - lif * lltv)
                     : type(uint256).max;
--- a/src/Midnight.sol
+++ b/src/Midnight.sol
@@ -656,9 +656,9 @@ contract Midnight is IMidnight {
             if (!healthyPath) {
                 uint256 lltv = market.collateralParams[collateralIndex].lltv;
                 // Note that debt >= maxDebt in this branch.
-                uint256 maxRepaid = lltv < WAD
-                    ? (_position.debt - maxDebt).mulDivUp(WAD, WAD - lif.mulDivUp(lltv, WAD))
-                    : type(uint256).max;
+                // The imprecision in this computation is at most a few hundreds WEI of collateral or loan token.
+                uint256 maxRepaid =
+                    lltv < WAD ? (_position.debt - maxDebt).mulDivUp(WAD * WAD, WAD - lif * lltv) : type(uint256).max;
                 require(
                     repaidUnits <= maxRepaid
                         || _position.collateral[collateralIndex].mulDivDown(liquidatedCollatPrice, ORACLE_PRICE_SCALE)
--- a/test/UtilsLibTest.sol
+++ b/test/UtilsLibTest.sol
@@ -4,6 +4,20 @@ pragma solidity ^0.8.0;
 import {Test, stdError} from "../lib/forge-std/src/Test.sol";
 import {UtilsLib} from "../src/libraries/UtilsLib.sol";
 import {LN_ONE_PLUS_DELTA, MAX_TICK, TickLib} from "../src/libraries/TickLib.sol";
+import {
+    WAD,
+    LIQUIDATION_CURSOR_LOW,
+    LIQUIDATION_CURSOR_HIGH,
+    LLTV_0,
+    LLTV_1,
+    LLTV_2,
+    LLTV_3,
+    LLTV_4,
+    LLTV_5,
+    LLTV_6,
+    LLTV_7,
+    maxLif
+} from "../src/libraries/ConstantsLib.sol";
 
 contract UtilsLibTest is Test {
     int256 internal constant WEXP_LN2 = 0.693147180559945309e18;
@@ -183,5 +197,22 @@ contract UtilsLibTest is Test {
         assertApproxEqRel(TickLib.wExp(18 ether), 65659969.137330511139838976 ether, 0.001 ether, "exp(18)");
         assertApproxEqRel(TickLib.wExp(19 ether), 178482300.96318726092869632 ether, 0.001 ether, "exp(19)");
         assertApproxEqRel(TickLib.wExp(20 ether), 485165195.409790277969936384 ether, 0.001 ether, "exp(20)");
+    }
+
+    // This test makes sure that the computation of maxRepaid (for RCF) is not too imprecise.
+    function testRcfBound() public pure {
+        uint256[8] memory lltvs = [LLTV_0, LLTV_1, LLTV_2, LLTV_3, LLTV_4, LLTV_5, LLTV_6, LLTV_7];
+        uint256[2] memory cursors = [LIQUIDATION_CURSOR_LOW, LIQUIDATION_CURSOR_HIGH];
+        for (uint256 i = 0; i < lltvs.length; i++) {
+            uint256 lltv = lltvs[i];
+            for (uint256 j = 0; j < cursors.length; j++) {
+                uint256 lif = maxLif(lltv, cursors[j]);
+                assertLt(lif * lltv, WAD * WAD);
+                assertLt(WAD * WAD / (WAD * WAD - lif * lltv), 100);
+            }
+        }
+        // Also ensure that the bound is close to the max bound.
+        uint256 lifHigh7 = maxLif(LLTV_7, LIQUIDATION_CURSOR_HIGH);
+        assertGt(WAD * WAD / (WAD * WAD - lifHigh7 * LLTV_7), 90);
     }
 }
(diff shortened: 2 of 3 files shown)


src/Midnight.sol
```diff
@@ -656,8 +656,9 @@ contract Midnight is IMidnight {
             if (!healthyPath) {
                 uint256 lltv = market.collateralParams[collateralIndex].lltv;
                 // Note that debt >= maxDebt in this branch.
+                // The imprecision in this computation is at most a few hundreds WEI of collateral or loan token.
```
```suggestion
                // The imprecision in this computation is at most a few hundreds collateral or loan token assets.
```

src/Midnight.sol
```diff
@@ -656,8 +656,9 @@ contract Midnight is IMidnight {
...
                 uint256 maxRepaid = lltv < WAD
-                    ? (_position.debt - maxDebt).mulDivUp(WAD, WAD - lif.mulDivUp(lltv, WAD))
+                    ? (_position.debt - maxDebt).mulDivUp(WAD * WAD, WAD * WAD - lif * lltv)
```
Saw-mon-and-Natalie, line 661: This is the 1st suggestion:

$$
\left\lceil    \frac{        D_{\mathcal{O},u}^{capped} - D_{\mathcal{O},u}^{max}    }{        \left(            1 - \frac{\texttt{lif}_{\mathcal{O}, T_j}^{max}}{10^{18}} \cdot             \frac{\texttt{lltv}_{\mathcal{O}, T_j}}{10^{18}}         \right)    }\right\rceil=\frac{    D_{\mathcal{O},u}^{capped} - D_{\mathcal{O},u}^{max}}{    \left(        1 - \frac{\texttt{lif}_{\mathcal{O}, T_j}^{max}}{10^{18}} \cdot         \frac{\texttt{lltv}_{\mathcal{O}, T_j}}{10^{18}}     \right)}+ \epsilon_{10}
$$

any reason, you didn't go with 2nd or 3rd suggestion?

Saw-mon-and-Natalie:
> Call R the perfect computation of the bound of the code. `R = 1 / (1 - lif * lltv)`
> 
> From the goals above, we want the max repaid to be as close to the bound of the finding: Screenshot 2026-05-28 at 18 28 12 This bound is different from `R` by the rounding term `X`.

Either $$R$$ is definied incorrectly or the last sentence does not make sense. Since in the above screenshot one can see that the first term is not $$R$$ but the differences in debts times $$R$$ (assuming $$R$$ is defined using scaled parameters).

src/Midnight.sol
```diff
@@ -656,9 +656,9 @@ contract Midnight is IMidnight {
             if (!healthyPath) {
                 uint256 lltv = market.collateralParams[collateralIndex].lltv;
                 // Note that debt >= maxDebt in this branch.
-                uint256 maxRepaid = lltv < WAD
-                    ? (_position.debt - maxDebt).mulDivUp(WAD, WAD - lif.mulDivUp(lltv, WAD))
-                    : type(uint256).max;
+                // The imprecision in this computation is at most a few hundreds WEI of collateral or loan token.
+                uint256 maxRepaid =
+                    lltv < WAD ? (_position.debt - maxDebt).mulDivUp(WAD * WAD, WAD - lif * lltv) : type(uint256).max;
```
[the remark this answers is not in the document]
Saw-mon-and-Natalie: What 🤖 corrected was the original suggestion in the finding ✅ The finding has the `WAD * WAD - lif * lltv` term.

Saw-mon-and-Natalie:
> This means that option 2 (adding 200) and option 3 (adding `(WAD + lltv).mulDivUp(WAD, WAD * WAD - lif * lltv)`) are not really helping: they help with one property wanted, at the expense of the other.

both of options 2 and 3 help with:

- Avoid consecutive maxLiquidations: its deliberately rounded up so that there is no immediate consecutive liquidation.

And possibly the 1st property.

Saw-mon-and-Natalie:
> We still would be off from the bound of the finding by `X`. In absolute value, `X` is less than `A / (1 - lif * lltv) + B * price / (1 - lif * lltv)` with `A` and `B` small integers constants (could be negative). Note that I didn't add a term of the form `C * lltv / (1 - lif * lltv)`  nor of the form `D * lif * lltv / (1 - lif * lltv)` because you can bound `lltv` and `lif * lltv` by 1.
> Then we conclude by the following interpretation: we are at most hundreds of WEI of debt (the term `A / (1 - lif * lltv)`) and of collateral (the term `B * price / (1 - lif * lltv)`) off from being exactly at the healthy threshold. See the corresponding added test

These two paragraphs don't sound so rigourous. More rigorous arguments in the finding and the comment there.

src/Midnight.sol
Saw-mon-and-Natalie: Note that in some cases this first option can break the desired property:

- Avoid consecutive maxLiquidations: its deliberately rounded up so that there is no immediate consecutive liquidation.

whereas options 2. and 3. would preserve the above. I think it would be best to go with **option 3**.

QGarchery: > any reason, you didn't go with 2nd or 3rd suggestion?

I added the reason in the description of the PR. I see that you left other comments about it below. I'm answering them here, but let me know if you have any other:
- [comment1](https://github.com/morpho-org/midnight/pull/944#issuecomment-4567840569). Thanks, my mistake it's a typo, I will fix the description. The numerator should be `debt - maxDebt` (written $D_{\mathcal{O},u}^{capped} - D_{\mathcal{O},u}^{max}$ in the finding)
- [comment2](https://github.com/morpho-org/midnight/pull/944#issuecomment-4567874282) I think they don't help with the first property. Adding `200` for example allow repay even more, which would make the position potentially more healthy than what is necessary
- [comment3](https://github.com/morpho-org/midnight/pull/944#issuecomment-4567919326). What do you find not satisfying ? It can be proven that the difference between specification (which is the value of eq.1 in the finding) and the code after this PR, is of the form `A / (1 - lif * lltv) + B * price / (1 - lif * lltv)` with `A` and `B` small integers constants. Then we can compute the higher values that `1 / (1 - lif * lltv)` can take and conclude with the interpretation. Let me know if I went too fast on a given step, and what can be improved

bhargavbh:
> both of options 2 and 3 help with:
> 
> Avoid consecutive maxLiquidations: its deliberately rounded up so that there is no immediate consecutive liquidation.
> And possibly the 1st property.

option-2 and option-3 go in the opposite direction for property-1. they do help in property-2, but IMO option-1 already guarantees that maxRepaid (as computed) would be greater than what is required to put position at exact health boundary. In @QGarchery notation, this should mean `A / (1 - lif * lltv) + B * price / (1 - lif * lltv)` should be positive, correct?

bhargavbh:
from a practical viewpoint, there are other aspects that help with property-2. 
lets say the property is not satisfied because of rounding, the gas costs are presumably high enough that a consecutive liquidation just to capture the small rounding difference in not profitable. If we are worried of dusts accumulating, then rcfThreshold mitigates it.

src/Midnight.sol
```diff
@@ -656,8 +656,9 @@ contract Midnight is IMidnight {
             if (!healthyPath) {
                 uint256 lltv = market.collateralParams[collateralIndex].lltv;
                 // Note that debt >= maxDebt in this branch.
+                // The imprecision in this computation is at most a few hundreds collateral or loan token assets.
```
bhargavbh, line 659: can we say here that it is higher than the spec?

QGarchery: I don't think it's necessarily higher, because the imprecision goes in both ways. Only thing we know is that it's bounded

bhargavbh (approved):

test/UtilsLibTest.sol
```diff
@@ -184,4 +198,21 @@ contract UtilsLibTest is Test {
...
+    // This test makes sure that the computation of maxRepaid (for RCF) is not too imprecise.
+    function testRcfBound() public pure {
+        uint256[8] memory lltvs = [LLTV_0, LLTV_1, LLTV_2, LLTV_3, LLTV_4, LLTV_5, LLTV_6, LLTV_7];
+        uint256[2] memory cursors = [LIQUIDATION_CURSOR_LOW, LIQUIDATION_CURSOR_HIGH];
+        for (uint256 i = 0; i < lltvs.length; i++) {
+            uint256 lltv = lltvs[i];
+            for (uint256 j = 0; j < cursors.length; j++) {
+                uint256 lif = maxLif(lltv, cursors[j]);
+                assertLt(lif * lltv, WAD * WAD);
+                assertLt(WAD * WAD / (WAD * WAD - lif * lltv), 100);
```
peyha, line 211: could you please expand on how this assertion gives guarantees that the comp of `maxRepaid` is accurate ?

adhusson (approved):
approving as it improves precision, and i couldn't find a scenario that makes "maxRepaid is rounded up to avoid consecutive max liquidations" wrong (ie a way to max liquidate then immediately liquidate again)

peyha (approved):

test/UtilsLibTest.sol
QGarchery: yes, so the rounding term is of the form `A / (1 - lif * lltv) + B * price / (1 - lif * lltv)` where `A` and `B` are small integer constants, with small absolute values (let's say something like 5 max). This bound on `A` & bounding `1 / (1 - lif * lltv)` is enough to bound `A / (1 - lif * lltv)` which represents a number of loan assets. The same bound on `B` & bounding `1 / (1 - lif * lltv)` is enough to bound `B * price / (1 - lif * lltv)` which represents a number of collateral assets. This test is about bounding `1 / (1 - lif * lltv)`

MathisGD (approved):

PR #951, opened 2026-05-29
bundles: custom error in safeApprove

MathisGD:
cc @StErMi 

https://cantina.xyz//code/81abfa1d-eee2-4afd-9c3b-8134f4d0254b/src/periphery/MidnightBundles.sol?preview=code#comment-ef962209-b072-4b52-be76-5ae5e9ddbeea

### DIFF
--- a/src/periphery/MidnightBundles.sol
+++ b/src/periphery/MidnightBundles.sol
@@ -25,13 +25,14 @@ import {WAD} from "../libraries/ConstantsLib.sol";
 contract MidnightBundles is IMidnightBundles {
     using UtilsLib for uint256;
 
-    address internal constant PERMIT2 = 0x000000000022D473030F116dDEE9F6B43aC78BA3;
-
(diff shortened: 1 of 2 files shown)


adhusson (approved):

src/periphery/MidnightBundles.sol
```diff
@@ -25,14 +25,15 @@ import {WAD} from "../libraries/ConstantsLib.sol";
 contract MidnightBundles is IMidnightBundles {
     using UtilsLib for uint256;
 
-    address internal constant PERMIT2 = 0x000000000022D473030F116dDEE9F6B43aC78BA3;
-
+    address public constant PERMIT2 = 0x000000000022D473030F116dDEE9F6B43aC78BA3;
```
QGarchery, line 30: So we want to add it to `IMidnightBundles` as well now that it is public. This makes me realize that we are also missing the `MIDNIGHT` function in `IMidnightBundles`

peyha (approved):

PR #962, opened 2026-06-02
Stermi informationals (no. 2)

MathisGD:
https://cantina.xyz/code/81abfa1d-eee2-4afd-9c3b-8134f4d0254b/findings/2

### DIFF
--- a/src/Midnight.sol
+++ b/src/Midnight.sol
@@ -87,11 +87,12 @@ import {IMidnight, Market, Offer, CollateralParams, MarketState, Position} from 
 /// loan token.
 ///
 /// OFFER CAPS
-/// @dev At most one of maxAssets or maxUnits can be nonzero per offer.
+/// @dev Exactly one of maxAssets or maxUnits must be nonzero per offer (take reverts otherwise).
 /// @dev maxAssets caps max buyer assets if offer.buy is true, and caps max seller assets otherwise.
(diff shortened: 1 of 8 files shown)


src/Midnight.sol
```diff
@@ -347,11 +348,15 @@ contract Midnight is IMidnight {
         bytes32 id = touchMarket(offer.market);
         MarketState storage _marketState = marketState[id];
         require(_marketState.lossFactor < type(uint128).max, MarketLossFactorMaxedOut());
-        require(UtilsLib.atMostOneNonZero(offer.maxAssets, offer.maxUnits), MultipleNonZero());
+        require((offer.maxAssets == 0) != (offer.maxUnits == 0), InvalidOfferCaps());
```
QGarchery, line 351: maybe slightly inconsistent now to have `liquidate` use a lib function, and `take` that does not. 
For `liquidate` it could be simply `repaidUnits == 0 || seizedAssets == 0`

QGarchery (approved):

src/Midnight.sol
adhusson: The gain of functions like `min` and `zeroFloorSub` is mostly readability really. I would do `exactlyOneNonZero` here and your suggestion `repaidUnits == 0 || seizedAssets == 0` in liquidate.

MathisGD: went for full inlined

peyha (approved):

adhusson (approved):

PR #963, opened 2026-06-02
Stermi natspecs (no. 3)

MathisGD:
https://cantina.xyz/code/81abfa1d-eee2-4afd-9c3b-8134f4d0254b/findings/3

### DIFF
--- a/src/Midnight.sol
+++ b/src/Midnight.sol
@@ -110,8 +110,9 @@ import {IMidnight, Market, Offer, CollateralParams, MarketState, Position} from 
 /// @dev updatePosition and liquidate (for liquidatable users) also impact the position and are permissionless.
 ///
 /// ROUNDINGS
-/// @dev assets are rounded against the taker and in favor of the maker in take. Therefore, the settlement fee has no
-/// defined rounding direction, which could lead to fees manipulations on chains with very cheap gas.
+/// @dev assets are rounded against the taker and in favor of the maker in take (in particular a take with non-zero
+/// units could end up with buyerAssets or sellerAssets equal to zero). Therefore, the settlement fee has no defined
+/// rounding direction, which could lead to fees manipulations on chains with very cheap gas.
 /// @dev pendingFee updates are rounded in favor of the user. It could lead to fees manipulations too.
 /// @dev maxDebt is rounded down in isHealthy and liquidate.
 /// @dev lossFactor is rounded up so lenders collectively lose a bit more than badDebt on each bad debt realization.
--- a/src/Midnight.sol
+++ b/src/Midnight.sol
@@ -25,7 +25,7 @@ import {IMidnight, Market, Offer, CollateralParams, MarketState, Position} from 
 /// - It has no liquidation incentive, so liquidators repay at exactly the oracle price (plus roundings).
 /// @dev To check if a market has been touched, check if tickSpacing(marketId) > 0.
 /// @dev When some assets become withdrawable before maturity (after a repayment or a liquidation), there
-/// is an incentive to take resting sell offers with price < 1. (more precisely price < 1 - settlementFee) and withdraw
+/// is an incentive to take resting sell offers with price < 1 (more precisely price < 1 - settlementFee) and withdraw
 /// instantly. Lenders (and the fee claimer) might also race to withdraw first.
 ///
 /// MULTI-COLLATERAL MARKETS
--- a/src/Midnight.sol
+++ b/src/Midnight.sol
@@ -87,11 +87,12 @@ import {IMidnight, Market, Offer, CollateralParams, MarketState, Position} from 
 /// loan token.
 ///
 /// OFFER CAPS
-/// @dev At most one of maxAssets or maxUnits can be nonzero per offer.
+/// @dev Exactly one of maxAssets or maxUnits must be nonzero per offer (take reverts otherwise).
 /// @dev maxAssets caps max buyer assets if offer.buy is true, and caps max seller assets otherwise.
 /// @dev If maxAssets > 0, assets are capped to maxAssets, otherwise units are capped to maxUnits.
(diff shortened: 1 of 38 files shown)


src/Midnight.sol
```diff
@@ -25,8 +25,8 @@ import {IMidnight, Market, Offer, CollateralParams, MarketState, Position} from
 /// - It has no liquidation incentive, so liquidators repay at exactly the oracle price (plus roundings).
 /// @dev To check if a market has been touched, check if tickSpacing(marketId) > 0.
 /// @dev When some assets become withdrawable before maturity (after a repayment or a liquidation), there
-/// is an incentive to take resting sell offers with price < 1 and withdraw instantly. Lenders (and the fee claimer)
-/// might also race to withdraw first.
+/// is an incentive to take resting sell offers with price < 1. (more precisely price < 1 - settlementFee) and withdraw
```
```suggestion
/// is an incentive to take resting sell offers with price < 1 (more precisely price < 1 - settlementFee) and withdraw
```

QGarchery (approved):

src/periphery/interfaces/IMidnightBundles.sol
```diff
@@ -5,9 +5,9 @@ pragma solidity >=0.5.0;
 import {Offer, Market} from "../../interfaces/IMidnight.sol";
 
 struct Take {
-    uint256 units;
     Offer offer;
     bytes ratifierData;
+    uint256 units;
```
adhusson, line 10: what is this addressing?

src/periphery/MidnightBundles.sol
```diff
@@ -21,6 +21,9 @@ import {ConsumableUnitsLib} from "./ConsumableUnitsLib.sol";
 import {WAD} from "../libraries/ConstantsLib.sol";
 
 /// @dev Inherits the token safety requirements of Midnight (see Midnight.sol).
+/// @dev Offers are taken in the order they are passed. One sensible strategy is to sort them by price (increasing to
+/// buy, decreasing to sell).
+/// @dev takes.units should prevent to take more than what is takeable w.r.t. the callback / the balances / the health.
```
adhusson:
```suggestion
/// @dev takes.units should prevent taking more than what is takeable w.r.t. the callback / the balances / the health.
```

src/ratifiers/EcrecoverRatifier.sol
```diff
@@ -34,6 +34,7 @@ contract EcrecoverRatifier is IEcrecoverRatifier {
     function isRatified(Offer memory offer, bytes memory ratifierData) external view returns (bytes32) {
         (Signature memory sig, bytes32 root, uint256 leafIndex, bytes32[] memory proof) =
             abi.decode(ratifierData, (Signature, bytes32, uint256, bytes32[]));
+        require(proof.length <= 20, HashLib.TreeTooHigh());
```
adhusson: Strange since this reverts later anyway.

src/ratifiers/SetterRatifier.sol
```diff
@@ -31,6 +31,7 @@ contract SetterRatifier is ISetterRatifier {
     function isRatified(Offer memory offer, bytes memory ratifierData) external view returns (bytes32) {
         (bytes32 root, uint256 leafIndex, bytes32[] memory proof) =
             abi.decode(ratifierData, (bytes32, uint256, bytes32[]));
+        require(proof.length <= 20, HashLib.TreeTooHigh());
```
adhusson: Will we limit length to 20 in all future ratifiers? Wdyt moving the proof size limit comment just to ecrecover ratifier and removing the proof length checks.

src/Midnight.sol
adhusson, line 120: The following @StErMi comment (13) is marked as addressed in the current PR but the lines have not changed:

> [Midnight.sol?lines=111,118](https://cantina.xyz/code/81abfa1d-eee2-4afd-9c3b-8134f4d0254b/src/Midnight.sol?lines=111,118): there are scenarios where the `Minight.take` operation could end up (depending on the `tick`, `price`, `_tradingFee` and rounding) with `buyerAssets` or `sellerAssets` equal to zero. This means that (depending on the `offer` configuration) one of the parts will lose something and the other will gain something. The "ROUNDING" section of the natspec should be updated to cover these scenarios.

src/periphery/interfaces/IMidnightBundles.sol
MathisGD: nothing particular, I just thought that the order was weird

src/ratifiers/SetterRatifier.sol
MathisGD: it's 1M offers so I don't see any use-case for going above

src/Midnight.sol
MathisGD: not exactly but it's implied by the newest comments, right?

```suggestion
/// @dev assets are rounded against the taker and in favor of the maker in take. In particular, a take with units>0 could end up with buyerAssets or sellerAssets equal to zero. 
/// The settlement fee has no defined rounding direction, which could lead to fees manipulations on chains with very cheap gas.
```

MathisGD: wdyt @adhusson

src/ratifiers/SetterRatifier.sol
adhusson: Agree we don't care it's just that if we enforce consistency here we need to remember to be consistent in the future ones (otherwise it doesn't make sense to add the require in the setter ratifier here)

src/Midnight.sol
adhusson: for it

adhusson: i like this https://github.com/morpho-org/midnight/pull/963#discussion_r3355311539 !

adhusson (approved):

QGarchery (approved):

PR #967, opened 2026-06-02
[certora] Post Maturity Debt check

lilCertora:
Post maturity the liquidation the debt cannot increase 

[corresponding thread](https://morpholabs.slack.com/archives/C0ACSA8TNMA/p1780516456760859)


certora/specs/PostMaturityDebt.spec
```diff
@@ -0,0 +1,34 @@
...
+
+methods {
+    function multicall(bytes[]) external => HAVOC_ALL DELETE;
+
+    function debtOf(bytes32 id, address user) external returns (uint128) envfree;
+    function liquidationLocked(bytes32 id, address user) external returns (bool) envfree;
+    function Utils.hashMarket(Midnight.Market) external returns (bytes32) envfree;
+
+    // Deterministic, injective id linking a call-site market to its maturity.
+    function IdLib.toId(Midnight.Market memory market, uint256, address) internal returns (bytes32) => summaryToId(market);
```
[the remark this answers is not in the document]
lilCertora: that wouldn't be better and would weaken the proof

certora/specs/PostMaturityDebt.spec
```diff
@@ -0,0 +1,39 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+// Property: "Post maturity, the liquidation is locked or the debt cannot increase."
+
```
QGarchery: let's put this on top of the rule

certora/specs/PostMaturityDebt.spec
```diff
@@ -0,0 +1,39 @@
...
+using Utils as Utils;
+
+methods {
+    function multicall(bytes[]) external => HAVOC_ALL DELETE;
+
+    function debtOf(bytes32 id, address user) external returns (uint128) envfree;
+    function liquidationLocked(bytes32 id, address user) external returns (bool) envfree;
+    function Utils.hashMarket(Midnight.Market) external returns (bytes32) envfree;
+
+    // Deterministic, injective id linking a call-site market to its maturity.
```
QGarchery:
```suggestion
    // Deterministic toId summary.
```

certora/specs/PostMaturityDebt.spec
```diff
@@ -0,0 +1,39 @@
...
+/// RULE ///
+
+rule lockedOrDebtCannotIncreasePostMaturity(env e, method f, calldataarg args, Midnight.Market market, address user) filtered { f -> !f.isView } {
+    bytes32 id = summaryToId(market);
+
+    mathint debtBefore = debtOf(id, user);
+
+    f(e, args);
+
+    assert e.block.timestamp > to_mathint(market.maturity) => (liquidationLocked(id, user) || debtOf(id, user) <= debtBefore);
```
QGarchery:
```suggestion
    assert e.block.timestamp > market.maturity => (liquidationLocked(id, user) || debtOf(id, user) <= debtBefore);
```

QGarchery:
```suggestion
    assert e.block.timestamp > to_mathint(market.maturity) => debtOf(id, user) <= debtBefore;
```
we even have this in the new code

QGarchery (changes requested):

certora/specs/PostMaturityDebt.spec
```diff
@@ -0,0 +1,39 @@
...
+    function liquidationLocked(bytes32 id, address user) external returns (bool) envfree;
+    function Utils.hashMarket(Midnight.Market) external returns (bytes32) envfree;
+
+    // Deterministic, injective id linking a call-site market to its maturity.
+    function IdLib.toId(Midnight.Market memory market, uint256, address) internal returns (bytes32) => summaryToId(market);
+
+    // No explicit summaries for the callbacks and token transfers reached by take/repay/liquidate/flashLoan:
+    // we rely on the default AUTO summary (HAVOC_ECF for these non-view external calls).
+    // This encodes a no-reentrancy assumption, justified because the property is about the effect of each
+    // function's own body on debt, not the full transaction including (re-entrant) callbacks.
```
QGarchery: Actually the interesting case was in callbacks in the previous code: we want to make sure that the lock is properly set there. 

Now, it's not that it's "justified because the property is about the effect of each function's own body on debt, not the full transaction including (re-entrant) callbacks." We do care about the full transaction including callbacks. The reason why it's ok is because if you prove that no function can increase the debt without callbacks, then by induction you also prove that no function can increase the debt with callbacks (the inner-most call doesn't increase debt, and by the induction hypothesis the whole call without the inner-most call doesn't increase the debt)

certora/specs/PostMaturityDebt.spec
```diff
@@ -0,0 +1,40 @@
...
+
+/// HELPERS ///
+
+function summaryToId(Midnight.Market market) returns bytes32 {
+    return Utils.hashMarket(market);
+}
+
+/// RULE ///
+
+// Post maturity, the liquidation is locked or the debt cannot increase
```
jhoenicke: The rule itself doesn't check the liquidationLock any more, right?  I guess this is because the property is checked before and after a complete call.

I think we should drop the liquidation is locked from the description.

jhoenicke (changes requested):
I added the suggestion to change the description of the rule because this rule doesn't check liquidationLocked.

Otherwise this looks good.  Using HAVOC_ECF makes sense here.

certora/specs/PostMaturityDebt.spec
lilCertora: yes good catch i agree

certora/specs/PostMaturityDebt.spec
```diff
@@ -0,0 +1,40 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+using Utils as Utils;
+
+methods {
+    function multicall(bytes[]) external => HAVOC_ALL DELETE;
+
+    function debtOf(bytes32 id, address user) external returns (uint128) envfree;
+    function liquidationLocked(bytes32 id, address user) external returns (bool) envfree;
```
QGarchery:
```suggestion
```

certora/specs/PostMaturityDebt.spec
```diff
@@ -0,0 +1,40 @@
...
+/// HELPERS ///
+
+function summaryToId(Midnight.Market market) returns bytes32 {
+    return Utils.hashMarket(market);
+}
+
+/// RULE ///
+
+// Post maturity, the debt cannot increase
+rule lockedOrDebtCannotIncreasePostMaturity(env e, method f, calldataarg args, Midnight.Market market, address user) filtered { f -> !f.isView } {
```
QGarchery:
```suggestion
rule debtCannotIncreasePostMaturity(env e, method f, calldataarg args, Midnight.Market market, address user) filtered { f -> !f.isView } {
```

QGarchery (approved):

certora/specs/PostMaturityDebt.spec
MathisGD, line 1: all good except the PR description is outdated

MathisGD (approved):

certora/specs/PostMaturityDebt.spec
QGarchery: updated the PR description

jhoenicke (approved):

PR #542, opened 2026-03-19
[Certora] Split Properties 

lilCertora:
- offer can be split : `splitPreservesPrincipalAccounting`: position-level accounting (`creditOf`, `debtOf`, `totalUnits`, `lastLossFactor`, `lastAccrual`, `continuousFeeCredit`) is bitwise equal across the two paths.
- Splitting an offer will not punish the maker : `splitDoesNotPunishMakerOrFavorTaker`: taking A vs taking B then C (with A = B + C) differs by at most 1 wei on buyer/seller assets, fee delta, `consumed`, and `claimableTradingFee` (exact for `maxUnits` mode).

[corresponding thread](https://morpholabs.slack.com/archives/C0ACSA8TNMA/p1778243486980189)

PR #970, opened 2026-06-02
[Certora] update position missing rules

QGarchery:
Proves the following items in the wish list:
- similarly use updatePositionView(affectedUser) instead of creditOf(affectedUser) for the credit after the op or prove that both functions return the same.
- updatePosition and updatePositionView give the same result

Also proves `updatePositionPreservesContinuousFeeCreditWhenLastAccrualIsUpToDate`, see [this message](https://morpholabs.slack.com/archives/C0ACSA8TNMA/p1780568737842119?thread_ts=1780297229.558389&cid=C0ACSA8TNMA)

[corresponding thread](https://morpholabs.slack.com/archives/C0ACSA8TNMA/p1780568988249209)


certora/specs/OnlyAuthorizedCanChangeUpdatedValues.spec
```diff
@@ -0,0 +1,87 @@
...
+    function _.isRatified(Midnight.Offer offer, bytes) external => CVL_isRatified(offer) expect(bytes32);
+    function _.onFlashLoan(address, address[], uint256[], bytes) external => NONDET;
+    function SafeTransferLib.safeTransferFrom(address, address, address, uint256) internal => NONDET;
+    function SafeTransferLib.safeTransfer(address, address, uint256) internal => NONDET;
+}
+
+ghost ghostMulDivDown(uint256, uint256, uint256) returns uint256 {
+    // Axioms proven in MulDiv.spec.
+    axiom forall uint256 y. forall uint256 z. ghostMulDivDown(0, y, z) == 0;
+    axiom forall uint256 x. forall uint256 z. ghostMulDivDown(x, 0, z) == 0;
```
lilCertora, line 52: Axiom 2 is not actually proven in Muldiv.spec no ? we should extend mulDivZero to prove that `assert mulDivDown(a, 0, d) == 0;` and ` assert mulDivUp(a, 0, d) == 0;` i think.
Axiom 3 is ok as it is derivable

certora/specs/OnlyAuthorizedCanChangeUpdatedCredit.spec
```diff
@@ -0,0 +1,87 @@
...
+
+function summaryToId(Midnight.Market market) returns (bytes32) {
+    return Utils.hashMarket(market);
+}
+
+/// UPDATED CREDIT CHANGE RULES ///
+
+/// An unauthorized caller cannot change a user's updated credit except via liquidate.
+/// Assumes no reentrancy: callbacks and token transfers are not modeled as re-entering Midnight, so re-entrant collateral changes are not covered.
+rule onlyAuthorizedCanChangeUpdatedCreditExceptLiquidate(env e, method f, calldataarg args, Midnight.Market market, address user) filtered { f -> f.selector != sig:liquidate(Midnight.Market, uint256, uint256, uint256, address, bool, address, address, bytes).selector } {
```
lilCertora: we can also prove `onlyAuthorizedCanChangeUpdatedPendingFeeExceptLiquidate` as face vzalue of a lender is actually credit - pendingFee maybe it's interesting and a natural companion to this rule , wdyt ?

lilCertora (approved):

certora/specs/OnlyAuthorizedCanChangeUpdatedValues.spec
QGarchery: I thought it was added already, but actually it will be the case only after #542

QGarchery: now this PR is on top of #542

certora/specs/OnlyAuthorizedCanChangeUpdatedCredit.spec
QGarchery: done

certora/specs/OnlyAuthorizedCanChangeUpdatedValues.spec
```diff
@@ -0,0 +1,118 @@
...
+
+    uint128 updatedCreditBefore;
+    uint128 updatedPendingFeeBefore;
+    updatedCreditBefore, updatedPendingFeeBefore, _ = updatePositionView(e, market, id, user);
+    f(e, args);
+    uint128 updatedCreditAfter;
+    uint128 updatedPendingFeeAfter;
+    updatedCreditAfter, updatedPendingFeeAfter, _ = updatePositionView(e, market, id, user);
+
+    assert (updatedCreditAfter == updatedCreditBefore && updatedPendingFeeAfter == updatedPendingFeeBefore) || userIsAuthorized || makerRatified[user];
```
MathisGD, line 117: why does this prove that updatePosition and updatePositionView return the same thing?

MathisGD (approved):

lilCertora (approved):

certora/specs/OnlyAuthorizedCanChangeUpdatedValues.spec
QGarchery: actually it's the change in `BalanceEffects` that proves this property. This rule is to prove
- "similarly use updatePositionView(affectedUser) instead of creditOf(affectedUser) for the credit after the op"
- extended with [this idea](https://github.com/morpho-org/midnight/pull/970#discussion_r3343443186)

PR #973, opened 2026-06-03
log more

adhusson:
in Take, log 
- offerHash because it allows to recover the exact offer that was taken 
- ratifier and ratifierData because the ratifier call is view so it cannot log anything

Also, stop logging credit deltas so certora can compile, and log total units delta instead. User credit/debt can still be reconstructed with the logged `units` (and other events) and total units with the delta.


src/libraries/EventsLib.sol
```diff
@@ -20,7 +20,7 @@ library EventsLib {
     event SetDefaultContinuousFee(address indexed loanToken, uint256 newContinuousFee);
     event UpdatePosition(bytes32 indexed id_, address indexed user, uint256 creditDecrease, uint256 pendingFeeDecrease, uint256 accruedFee);
     event MarketCreated(Market market, bytes32 indexed id_);
-    event Take(address caller, bytes32 indexed id_, uint256 units, address indexed taker, address indexed maker, bool offerIsBuy, bytes32 group, uint256 buyerAssets, uint256 sellerAssets, uint256 consumed, uint256 buyerPendingFeeIncrease, uint256 sellerPendingFeeDecrease, uint256 buyerCreditIncrease, uint256 sellerCreditDecrease, address receiver, address payer);
+    event Take(address caller, bytes32 indexed id_, uint256 units, address indexed taker, Offer offer, uint256 buyerAssets, uint256 sellerAssets, uint256 consumed, uint256 buyerPendingFeeIncrease, uint256 sellerPendingFeeDecrease, uint256 buyerCreditIncrease, uint256 sellerCreditDecrease, address receiver, address payer, address takerCallback);
```
MathisGD: why do you add takerCallback?

MathisGD: I would log ratifier and ratifierData instead (see #966)

adhusson: it's in the context of [this discussion](https://morpholabs.slack.com/archives/C08A40KKN9W/p1781619863191509?thread_ts=1781616123.064059&cid=C08A40KKN9W) where the model is "reconstruct function calls from events".

Btw there are 2 ways to go
- before i updated this PR: log everything except `bytes` because they can be long, but then i had to have 2 events because of certora compilation.
- after i updated this PR: just log everything. So 1 event now works.

adhusson: (but all this logging seems to be hard for certora)

adhusson: > I would log ratifier and ratifierData instead (see https://github.com/morpho-org/midnight/pull/966)

agree, i forgot about `ratifierData` (ratifier is logged)

certora/confs/OnlyExplicitPayerCanLoseTokens.conf
adhusson, line 22: probably not needed anymore

certora/confs/Reentrancy.conf
```diff
@@ -4,7 +4,8 @@
   ],
   "prover_args": [
     "-enableStorageSplitting false",
-    "-destructiveOptimizations twostage"
+    "-destructiveOptimizations twostage",
+    "-sanitySolverTimeout 3600"
```
adhusson: probably not needd anymore

src/Midnight.sol
```diff
@@ -452,9 +452,8 @@ contract Midnight is IMidnight {
             id,
             units,
             taker,
-            offer.maker,
-            offer.buy,
-            offer.group,
+            keccak256(abi.encode(offer)),
```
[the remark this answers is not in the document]
adhusson: btw we could use hashOffer as long as the codesize is OK?

src/libraries/EventsLib.sol
```diff
@@ -20,7 +20,7 @@ library EventsLib {
     event SetDefaultContinuousFee(address indexed loanToken, uint256 newContinuousFee);
     event UpdatePosition(bytes32 indexed id_, address indexed user, uint256 creditDecrease, uint256 pendingFeeDecrease, uint256 accruedFee);
     event MarketCreated(Market market, bytes32 indexed id_);
-    event Take(address caller, bytes32 indexed id_, uint256 units, address indexed taker, address indexed maker, bool offerIsBuy, bytes32 group, uint256 buyerAssets, uint256 sellerAssets, uint256 consumed, uint256 buyerPendingFeeIncrease, uint256 sellerPendingFeeDecrease, uint256 buyerCreditIncrease, uint256 sellerCreditDecrease, address receiver, address payer);
+    event Take(address caller, bytes32 indexed id_, uint256 units, address indexed taker, bytes32 indexed offerHash, address maker, bool offerIsBuy, bytes32 group, bytes ratifierData, uint256 buyerAssets, uint256 sellerAssets, uint256 consumed, uint256 buyerPendingFeeIncrease, uint256 sellerPendingFeeDecrease, uint256 buyerCreditIncrease, uint256 sellerCreditDecrease, address receiver, address payer);
```
MathisGD: not sure
```suggestion
    event Take(address caller, bytes32 indexed id_, uint256 units, address indexed taker, bytes32 offerHash, address indexed maker, bool offerIsBuy, bytes32 group, bytes ratifierData, uint256 buyerAssets, uint256 sellerAssets, uint256 consumed, uint256 buyerPendingFeeIncrease, uint256 sellerPendingFeeDecrease, uint256 buyerCreditIncrease, uint256 sellerCreditDecrease, address receiver, address payer);
```

src/Midnight.sol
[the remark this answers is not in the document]
MathisGD: I think it's ok.

MathisGD (approved):

src/libraries/EventsLib.sol
adhusson: will change

src/libraries/EventsLib.sol
```diff
@@ -20,7 +20,7 @@ library EventsLib {
     event SetDefaultContinuousFee(address indexed loanToken, uint256 newContinuousFee);
     event UpdatePosition(bytes32 indexed id_, address indexed user, uint256 creditDecrease, uint256 pendingFeeDecrease, uint256 accruedFee);
     event MarketCreated(Market market, bytes32 indexed id_);
-    event Take(address caller, bytes32 indexed id_, uint256 units, address indexed taker, address indexed maker, bool offerIsBuy, bytes32 group, uint256 buyerAssets, uint256 sellerAssets, uint256 consumed, uint256 buyerPendingFeeIncrease, uint256 sellerPendingFeeDecrease, uint256 buyerCreditIncrease, uint256 sellerCreditDecrease, address receiver, address payer);
+    event Take(address caller, bytes32 indexed id_, uint256 units, address indexed taker, bytes32 offerHash, address indexed maker, bool offerIsBuy, bytes32 group, bytes ratifierData, uint256 buyerAssets, uint256 sellerAssets, uint256 consumed, uint256 buyerPendingFeeIncrease, uint256 sellerPendingFeeDecrease, uint256 buyerCreditIncrease, uint256 sellerCreditDecrease, address receiver, address payer);
```
MathisGD: we should log the ratifier and re-order closer to the function arguments

adhusson:
- added `offer.ratifier`
- reordered to match take arguments, with:
  * offer hash first (in place of offer)
  * then necessary offer fields, same order as in the offer struct, with `id` instead of `offer.market`
  * then the rest

src/libraries/EventsLib.sol
MathisGD: ratifier is still missing?

adhusson: did not push

src/libraries/EventsLib.sol
```diff
@@ -20,7 +20,7 @@ library EventsLib {
     event SetDefaultContinuousFee(address indexed loanToken, uint256 newContinuousFee);
     event UpdatePosition(bytes32 indexed id_, address indexed user, uint256 creditDecrease, uint256 pendingFeeDecrease, uint256 accruedFee);
     event MarketCreated(Market market, bytes32 indexed id_);
-    event Take(address caller, bytes32 indexed id_, uint256 units, address indexed taker, address indexed maker, bool offerIsBuy, bytes32 group, uint256 buyerAssets, uint256 sellerAssets, uint256 consumed, uint256 buyerPendingFeeIncrease, uint256 sellerPendingFeeDecrease, uint256 buyerCreditIncrease, uint256 sellerCreditDecrease, address receiver, address payer);
+    event Take(bytes32 offerHash, bytes32 indexed id_, bool offerIsBuy, address indexed maker, bytes32 group, address ratifier, bytes ratifierData, uint256 units, address indexed taker, address receiver, uint256 buyerAssets, uint256 sellerAssets, uint256 consumed, uint256 buyerPendingFeeIncrease, uint256 sellerPendingFeeDecrease, address payer, address caller);
```
MathisGD: caller is always at the beginning

MathisGD: the receiver should be later, as it's not always the offer parameter (I wouldn't move it basically)

MathisGD (approved):

QGarchery (approved):

PR #976, opened 2026-06-04
round up collat value in rcf check

adhusson:
Motivation:
- the value `collateral.mulDivUp(price, 1e36).mulDivUp(WAD, lif)` can be interpreted as "how much would be repaid if all the collateral was seized", and the RCF check as "if all the collat is seized, is the excess repaid amount small enough?“"
- simplifying the computation of the deactivation in case of a full seize
- `<=` instead of `<` makes it more obvious that using type(uint256).max as rcfThreshold deactivates it
- the difference with the previous, conservative rounding direction can be proven to be only a few WEI, because lif is bounded by 2

[corresponding thread](https://morpholabs.slack.com/archives/C08A40KKN9W/p1780536108185089):


adhusson:
closed to bias on the side of lender protection in all terms

PR #980, opened 2026-06-05
[Certora] remove some axioms in Split specs

QGarchery:


certora/specs/SplitDoesNotPunishMakerOrFavorTaker.spec
QGarchery, line 1: only axioms kept in this file are the super/sub-additivity which are central to the properties (what happens to rounding when an offer is split)

certora/specs/SplitPreservesAccounting.spec
QGarchery, line 47: needed because updatePosition could be called twice, and the second time it should do a no-op on the slashing:
```solidity
        uint256 postSlashCredit = _lastLossFactor < type(uint128).max
            ? credit.mulDivDown(type(uint128).max - marketState[id].lossFactor, type(uint128).max - _lastLossFactor)
            : 0;
```

certora/specs/SplitPreservesAccounting.spec
```diff
@@ -48,28 +48,10 @@ persistent ghost ghostMulDivDown(uint256, uint256, uint256) returns uint256 {
 
     // floor(a*0/c) = 0. Proven by mulDivZero in MulDiv.spec.
     axiom forall uint256 a. forall uint256 c. c != 0 => ghostMulDivDown(a, 0, c) == 0;
```
QGarchery, line 50: same as above but for the fee, where `accrualEnd - _lastAccrual == 0`:
```solidity
        uint128 fee = _lastAccrual < market.maturity
            ? uint128(postSlashPendingFee.mulDivDown(accrualEnd - _lastAccrual, market.maturity - _lastAccrual))
            : 0;
```

lilCertora (approved):

PR #989, opened 2026-06-15
[Certora] settlement fee spread assumption

### LINKED ISSUE #932: [Certora] side effect of calling settlementFee in a rule
> **<sub><sub>P2 Badge</sub></sub>  Cover market-creating takes in fee spread proof**
> 
> When `take` is the first interaction with an uncreated market, this pre-call to `settlementFee` reverts because `marketState[id].tickSpacing == 0`; in CVL, a plain reverting call prunes the path, so `settlementFeeSpreadBounds` never checks market-creating takes that initialize and charge the default settlement fee. Consider touching/creating the market in the rule before reading the fee, or otherwise adding a separate rule for the first-take path.
> 
> Useful? React with 👍 / 👎. 

 _Originally posted by @chatgpt-codex-connector[bot] in [#931](https://github.com/morpho-org/midnight/pull/931/changes#r3312618183)_

QGarchery:
Fixes:
- #932 

Also:
- removes occurrences of `onRatify` (now renamed `isRatified`)
- make settlementFeeSpread use havoc all by default

[corresponding thread](https://morpholabs.slack.com/archives/C0ACSA8TNMA/p1782236700834569)


certora/specs/SettlementFeeSpread.spec
```diff
@@ -49,8 +47,10 @@ rule makerFavorableRounding(env e, Midnight.Offer offer, bytes ratifierData, uin
 // The spread between what the buyer pays and what the seller receives is at least floor(units * fee / WAD) and at most ceil(units * fee / WAD).
 rule settlementFeeSpreadBounds(env e, Midnight.Offer offer, bytes ratifierData, uint256 units, address taker, address receiver, address takerCallback, bytes takerCallbackData) {
     uint256 timeToMaturity = e.block.timestamp <= offer.market.maturity ? assert_uint256(offer.market.maturity - e.block.timestamp) : 0;
-
     bytes32 id = summaryToId(offer.market);
+
+    // take calls touchMarket see rule takeCallsTouchMarket.
+    // Thus calling settlementFee (in particular checking if the market is touched) doesn't prune meaningful take paths.
     uint256 fee = settlementFee(id, timeToMaturity);
```
QGarchery: I tried this

```suggestion
    touchMarket(e, offer.market);
    uint256 fee = settlementFee@withrevert(id, timeToMaturity);
    assert !lastReverted;
```
but it was causing a long running time in [this run](https://prover.certora.com/output/66603/4b8b9371b6bf4fbb9134c9f29aedd611?anonymousKey=b5d010176080d98f5ea68aba33a151274a1fc205). Also it wasn't proving exactly that it's not pruning meaningful paths, because `storeInCode` is summarized to not revert in this file

QGarchery: as is, it feels like `TouchMarketIsCalled.spec` is a bit redundant with the `marketIsCreatedAfter*` rules, but it's actually not proving exactly the same thing

jhoenicke: would it work to call settlementFee after `take` or does the take change the settlementFee?

After take it should not revert, right?

jhoenicke: I tried calling settlementFee after and the HAVOC_ALL summary of the callbacks set tickSpacing to 0 :frowning:

With HAVOC_ECF it works, though.

jhoenicke (approved):
The TouchMarketIsCalled looks fine, the rest is just a comment update.

certora/specs/SettlementFeeSpread.spec
QGarchery: Yes basically we can either:
- have it before which uses havoc all by default, but there is this question of whether it prunes a path
- have it after which requires to have havoc ecf for all callbacks

I like the first solution better, since it has no unrelated assumptions about callbacks (you need to prove something about the settlement fee, which makes sense since this rule is about settlement fee)

certora/specs/SettlementFeeSpread.spec
```diff
@@ -49,8 +47,10 @@ rule makerFavorableRounding(env e, Midnight.Offer offer, bytes ratifierData, uin
 // The spread between what the buyer pays and what the seller receives is at least floor(units * fee / WAD) and at most ceil(units * fee / WAD).
 rule settlementFeeSpreadBounds(env e, Midnight.Offer offer, bytes ratifierData, uint256 units, address taker, address receiver, address takerCallback, bytes takerCallbackData) {
     uint256 timeToMaturity = e.block.timestamp <= offer.market.maturity ? assert_uint256(offer.market.maturity - e.block.timestamp) : 0;
-
     bytes32 id = summaryToId(offer.market);
+
+    // take calls touchMarket see rule takeCallsTouchMarket.
+    // Thus calling settlementFee (in particular checking if the market is touched) doesn't prune meaningful take paths.
```
MathisGD: why?

QGarchery: since `take` calls `touchMarket`, all meaningful paths call `touchMarket`. So it means that you could call `touchMarket` first and prove that the `settlementFee` call doesn't revert like [this](https://github.com/morpho-org/midnight/pull/989#discussion_r3415998884). 
But it's true that it doesn't mean that this verification checks that you can always call `settlementFee` and know what would be the spread. So if the market is not created one could use the default settlement fee instead

certora/specs/SettlementFeeSpread.spec
```diff
@@ -20,11 +22,17 @@ methods {
...
+    function _.defaultSettlementFeeCbp(address, uint256) external => DISPATCHER(true);
+
+    // function _updatePosition(Midnight.Market memory, bytes32, address) internal returns (uint128, uint128, uint128) => NONDET;
 }
 
+persistent ghost ghostToId(uint256, address, address, uint256, uint256, address, address) returns bytes32;
+
 function summaryToId(Midnight.Market market) returns (bytes32) {
-    return Utils.hashMarket(market);
+    return ghostToId(market.chainId, market.midnight, market.loanToken, market.maturity, market.rcfThreshold, market.enterGate, market.liquidatorGate);
```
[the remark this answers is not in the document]
QGarchery: fine since there is only one market in this spec file (it's documented)

certora/specs/SettlementFeeSpread.spec
QGarchery: irrelevant now in the latest code

certora/specs/SettlementFeeSpread.spec
QGarchery: The latest code is a complete specification

MathisGD (approved):

MathisGD:
@codex review

certora/confs/SettlementFeeSpread.conf
```diff
@@ -1,18 +1,23 @@
 {
   "files": [
     "certora/helpers/Utils.sol",
```
adhusson: not used anymore

adhusson (approved):

PR #992, opened 2026-06-16
modular lif

MathisGD:
modularize liquidation cursors.

PR #993, opened 2026-06-16
modular lltvs

MathisGD:
#992 will add useful bounds (after having added modular LIFs)

### DIFF
--- a/certora/specs/Role.spec
+++ b/certora/specs/Role.spec
@@ -9,6 +9,7 @@ methods {
     function feeSetter() external returns (address) envfree;
     function feeClaimer() external returns (address) envfree;
     function tickSpacingSetter() external returns (address) envfree;
+    function isLltvAllowed(uint256 lltv) external returns (bool) envfree;
     function tickSpacing(bytes32 id) external returns (uint8) envfree;
     function continuousFee(bytes32 id) external returns (uint32) envfree;
     function claimableSettlementFee(address token) external returns (uint256) envfree;
@@ -25,6 +26,8 @@ methods {
 }
 
 /// HELPERS ///
+
+definition WAD() returns uint256 = 10 ^ 18;
 
 definition CBP() returns uint256 = 10 ^ 12;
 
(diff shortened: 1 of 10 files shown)


src/Midnight.sol
```diff
@@ -163,7 +163,7 @@ import {IMidnight, Market, Offer, CollateralParams, MarketState, Position} from
 /// revert.
 ///
 /// ROLES
-/// @dev The role setter can set the role setter, fee setter, fee claimer, and tick spacing setter.
+/// @dev The role setter can set the role setter, fee setter, fee claimer, and tick spacing setter, and add LLTV tiers.
```
adhusson, line 166: should we rename roleSetter? eg configurator

certora/specs/Role.spec
```diff
@@ -85,6 +88,14 @@ rule roleSetterCanChangeTickSpacingSetter(env e, address newTickSpacingSetter) {
     assert !lastReverted => tickSpacingSetter() == newTickSpacingSetter;
 }
 
+rule roleSetterCanAddLltv(env e, uint256 lltv) {
+    address roleSetterBefore = roleSetter();
+
+    addLltv@withrevert(e, lltv);
+    assert !lastReverted <=> e.msg.sender == roleSetterBefore && e.msg.value == 0 && lltv <= WAD();
+    assert !lastReverted => isLltvAllowed(lltv);
```
adhusson, line 96: to show "roleSetterCanAddLltv" (`onlyRoleSetterCanAddLltv` is later) it's enough to say

```suggestion
    assert e.msg.sender == roleSetterBefore && e.msg.value == 0 && lltv <= WAD() => !lastReverted;
```

and `    assert !lastReverted => isLltvAllowed(lltv);` could be in a separate `allowedLltvLessThanOrEqualToOne` and the current `allowedLltvLessThanOrEqualToOne` could be renamd `marketLltvIsAllowedLltv`.

adhusson (approved):

certora/specs/Role.spec
MathisGD: > it's enough to say

yes and no, you also need to show that the LLTV is now allowed (the second line)

src/Midnight.sol
QGarchery: or should it set a LLTV setter ? Maybe not so useful since we don't have it in Blue

QGarchery (approved):

src/Midnight.sol
MathisGD: agree that it's not useful

PR #998, opened 2026-06-17
[Certora] Prove that take has no new reverts due to continuous fees

jhoenicke:
Compare a run of take with zero fees with a run of take with arbitrary fee.  Show that if the first does not revert, the second doesn't as well.

Requires a deterministic setting (all external gates and internal abstracted functions are modelled by a deterministic ghost).  This is why it is using its own separate spec file.


certora/specs/ContinuousFeeReverts.spec
```diff
@@ -0,0 +1,182 @@
...
+        revert();
+    }
+    return require_uint256(summaryMulDivDownGhost(x, y, d));
+}
+
+function summaryMulDivUpWithRevert(uint256 x, uint256 y, uint256 d) returns uint256 {
+    if (d == 0 || mulOverflow(x, y)) {
+        revert();
+    }
+    return require_uint256(summaryMulDivUpGhost(x, y, d));
```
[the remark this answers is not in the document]
jhoenicke: The old was overapproximating, but used the same for Up and Down, which is technically not correct.  I added the precise overflow condition instead.

certora/specs/ContinuousFeeReverts.spec
```diff
@@ -0,0 +1,159 @@
...
+    require postPendingBuyer <= postCreditBuyer, "pendingContinuousFeeBoundedByCredit, preserved by updatePosition";
+
+    storage initState = lastStorage;
+
+    // arbitrary pre-state before continuous fee is set.  Record its revert status.
+    take@withrevert(e, offer, ratifierData, units, taker, receiver, takerCallback, takerCallbackData);
+    bool revertedDisabled = lastReverted;
+
+    // now set the new continuous fee and check again.
+    setMarketContinuousFee(eSetter, id, newContinuousFee) at initState;
```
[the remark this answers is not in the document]
jhoenicke: Do I understand this correctly?  take can succeed on a not created market.  This rule would not revert in the first scenario, it reverts in the second scenario on setMarketContinousFee(), but that revert is not guarded by `@withrevert`, so the rule succeeds.  So in this case we don't compare the defaultContinousFee against the newContinuousFee case.   To do that we would need to create the market first, before calling setMarketContinousFee, but only if the market does not already exist?

certora/specs/ContinuousFeeReverts.spec
```diff
@@ -0,0 +1,159 @@
...
+// both runs revert anyway (e.g. the rate-independent toUint128(buyerCreditIncrease) at Midnight.sol:417).
+persistent ghost mulOverflow(mathint, mathint) returns bool {
+    axiom forall mathint x. forall mathint y. (0 <= x && x <= max_uint128 && 0 <= y && y <= max_uint128) => !mulOverflow(x, y);
+}
+
+function summaryMulDivDownWithRevert(uint256 x, uint256 y, uint256 d) returns uint256 {
+    if (d == 0 || x * y >= 2 ^ 256) {
+        revert();
+    }
+    return require_uint256(summaryMulDivDownGhost(x, y, d));
```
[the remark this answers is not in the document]
jhoenicke: Overapproximating the result is okay here.  The prover will try any value of the ghostMulDiv, even the one where it has the right behavior.  For these the require_uint256 will not prune the path.

certora/specs/ContinuousFeeReverts.spec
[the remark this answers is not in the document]
jhoenicke: Adding `@withrevert` to `setMarketContinuousFee` would also need more `require` to ensure the caller has the permission to set the fee.  This clutters the rule a lot.  I think it is okay to assume the setMarketContinuousFee will succeed.  We already test the scenario where the market is in the state "market is created with default fee".

certora/specs/ContinuousFeeReverts.spec
```diff
@@ -0,0 +1,159 @@
...
+
+    // Callbacks and ratifier: assumed to succeed deterministically. We verify take's own body, not the
+    // behavior of untrusted callbacks, and this spec (like ContinuousFee.spec) assumes no reentrancy.
+    function _.onBuy(bytes32, Midnight.Market, uint256, uint256, uint256, address, bytes) external => deterministicSuccess() expect(bytes32);
+    function _.onSell(bytes32, Midnight.Market, uint256, uint256, uint256, address, address, bytes) external => deterministicSuccess() expect(bytes32);
+    function _.isRatified(Midnight.Offer, bytes, address) external => deterministicSuccess() expect(bytes32);
+
+    // Token transfers: deterministic no-op (void => no havoc); identical across both runs.
+    function SafeTransferLib.safeTransferFrom(address, address, address, uint256) internal => NONDET;
+    function SafeTransferLib.safeTransfer(address, address, uint256) internal => NONDET;
```
MathisGD: not important, but to be 100% correct they should do deterministicSuccess as well no?

certora/specs/ContinuousFeeReverts.spec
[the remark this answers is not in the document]
MathisGD: the market has no fee if it's not created, so we don't need to prove that the continuous fee doesn't create revert on a non created market

MathisGD (approved):

certora/specs/ContinuousFeeReverts.spec
```diff
@@ -0,0 +1,159 @@
...
+// EQUIVALENCE OF REVERTS: activating the continuous fee never makes `take` revert in new ways, provided 
+// the fee doesn't exceed the offer's continuousFeeCap.
+
+using Utils as Utils;
+
+methods {
+    function multicall(bytes[]) external => HAVOC_ALL DELETE;
+
+    function continuousFee(bytes32 id) external returns (uint32) envfree;
+    function tickSpacing(bytes32 id) external returns (uint8) envfree;
```
QGarchery:
```suggestion
```

certora/specs/ContinuousFeeReverts.spec
```diff
@@ -0,0 +1,144 @@
...
+    function multicall(bytes[]) external => HAVOC_ALL DELETE;
+
+    function Utils.hashMarket(Midnight.Market) external returns (bytes32) envfree;
+    function Utils.callbackSuccess() external returns (bytes32) envfree;
+
+    // Deterministic id: links the market argument to stored state, identical across both runs.
+    function IdLib.toId(Midnight.Market memory market) internal returns (bytes32) => summaryToId(market);
+    function IdLib.storeInCode(Midnight.Market memory) internal returns (address) => NONDET;
+
+    // Deterministic ghost summaries of the rate-independent helpers (see header).
```
QGarchery: This comment is not very clear to me (what header, what are rate-independent helpers ?). 

Also those summaries hide a lot of potential reverts. For example `isHealthy` is assumed to never revert, and always return the same value independently of storage. But the storage changed between the 2 runs.
If we can't improve it, I think it should be documented as one of the assumptions of the rule.

certora/specs/ContinuousFeeReverts.spec
```diff
@@ -0,0 +1,144 @@
...
+// mulDivDown: deterministic, reverts on d == 0 and on x*y overflow (see mulOverflow below).
+// The axiom mulDivArgumentLesserThanDenominator, proved in MulDiv.spec, is needed:
+//     y <= d => result <= x.
+// This bounds the enabled-run fee increase by buyerCreditIncrease.
+persistent ghost summaryMulDivDownGhost(mathint, mathint, mathint) returns mathint {
+    axiom forall mathint x. forall mathint y. forall mathint d. (d > 0 && 0 <= y && y <= d) => summaryMulDivDownGhost(x, y, d) <= x;
+}
+
+// mulDivUp: deterministic, reverts on d == 0 and on x*y overflow.
+persistent ghost summaryMulDivUpGhost(mathint, mathint, mathint) returns mathint;
```
QGarchery: We could use the new `MulDivAxioms.spec` here

certora/specs/ContinuousFeeReverts.spec
```diff
@@ -0,0 +1,144 @@
...
+
+// mulDiv reverts when x*y overflows 256 bits.
+// We model that overflow as a deterministic uninterpreted predicate rather than the literal nonlinear `x * y >= 2^256`, for two reasons:
+//   - The many rate-independent mulDiv calls have identical (x,y) across both runs, so they share the same overflow outcome by congruence (an uninterpreted function), with no nonlinear arithmetic to discharge.
+//   - The one rate-dependent fee mulDiv (x = buyerCreditIncrease, y = continuousFee*ttm) is provably non-overflowing via the LINEAR safe-region axiom below instead of a product-of-bounds NIA argument.
+// Soundness: the axiom only asserts no-overflow where it genuinely holds (x,y <= max_uint128 => x*y < 2^256), so it never hides a real overflow revert. Outside that region the predicate is free, but in every such case both runs revert anyway (e.g. the rate-independent toUint128(buyerCreditIncrease) at Midnight.sol:417).
+persistent ghost mulOverflow(mathint, mathint) returns bool {
+    axiom forall mathint x. forall mathint y. (0 <= x && x <= max_uint128 && 0 <= y && y <= max_uint128) => !mulOverflow(x, y);
+}
+
```
QGarchery:
```suggestion
```
this is not used ?

certora/specs/ContinuousFeeReverts.spec
QGarchery: no need because their returned value is not checked like the callbacks are. The problem with using NONDET everywhere is that then there could be reverts due to a callback in one call (because it returns something different from CALLBACK_SUCCESS) and no reverts in another call (because the callback actually returns CALLBACK_SUCCESS)

certora/specs/ContinuousFeeReverts.spec
jhoenicke: Now I only summarize the deterministic math functions, not `isHealthy()`.

certora/specs/ContinuousFeeReverts.spec
jhoenicke: done

certora/specs/ContinuousFeeReverts.spec
QGarchery, line 1: We should be very clear about the assumptions we are making:
- `tickToPrice` & `settlementFee` are assumed to not revert, it's not just about having a deterministic summary for math functions
- similarly `price` and the transfers are assumed to not revert
- gates have a nice explanation already ("decision is identical across both runs"), even though it seems to be only about their return value but that's ok because reverting is the same as returning false for a gate

So I think we should: update the `methods` block to reflect those assumptions and/or have a recap above the rule

QGarchery (approved):

tarikbellamine:
@codex review

certora/specs/ContinuousFeeReverts.spec
```diff
@@ -0,0 +1,132 @@
...
+    uint128 postCreditBuyer;
+    uint128 postPendingBuyer;
+    postCreditBuyer, postPendingBuyer, _ = updatePositionView(e, offer.market, id, buyer);
+    require postPendingBuyer <= postCreditBuyer, "pendingContinuousFeeBoundedByCredit, preserved by updatePosition";
+
+    storage initState = lastStorage;
+
+    // arbitrary pre-state before continuous fee is set.  Record its revert status.
+    take@withrevert(e, offer, ratifierData, units, taker, receiver, takerCallback, takerCallbackData);
+    bool revertedDisabled = lastReverted;
```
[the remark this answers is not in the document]
QGarchery: it's ok to keep the stronger spec

PR #1012, opened 2026-06-23
gas opti: no call toId

MathisGD:
saves ~1000 gas per loop, not sure if worth it


src/periphery/MidnightBundles.sol
```diff
@@ -43,9 +44,11 @@ contract MidnightBundles is IMidnightBundles {
 
     address public constant PERMIT2 = 0x000000000022D473030F116dDEE9F6B43aC78BA3;
     address public immutable MIDNIGHT;
+    uint256 public immutable INITIAL_CHAIN_ID;
```
QGarchery, line 47:
```suggestion
    uint256 public immutable MIDNIGHT_INITIAL_CHAIN_ID;
```
(optional)

QGarchery (approved):

adhusson (approved):

MathisGD:
at the end judged not worth it

PR #1014, opened 2026-06-23
[Certora] No subtraction underflow (WIP)

QGarchery:


QGarchery:
Abandoned because it requires a huge setup

PR #1018, opened 2026-06-24
nit: re-order functions

MathisGD:


src/Midnight.sol
```diff
@@ -815,8 +815,6 @@ contract Midnight is IMidnight {
         return id;
     }
 
-    /// SLASHING AND CONTINUOUS FEE ACCRUAL ///
-
```
QGarchery, line 813: now `updatePositionView` and some other view functions are in the `ENTRY-POINTS` section. It makes sense to have `updatePosition` in the entry points though

src/Midnight.sol
```diff
@@ -875,11 +873,52 @@ contract Midnight is IMidnight {
...
+
+        return (feeLower * (end - timeToMaturity) + feeUpper * (timeToMaturity - start)) / (end - start);
+    }
+
     function hasCredit(bytes32 id, address user) internal view returns (bool) {
         return position[id][user].credit > 0;
     }
 
-    /// OTHER VIEW FUNCTIONS ///
+    /// EXTERNAL GETTERS ///
```
adhusson: for max usefulness add `isHealthy` `settlementFee` `toId` (even if it will be gone soon) `liquidationLocked` `updatePositionView` and rename


```suggestion
    /// EXTERNAL AND PUBLIC GETTERS ///
```

?

adhusson (approved):

adhusson (approved):

QGarchery (approved):

PR #1026, opened 2026-06-24
Refactor modular lif

QGarchery:


certora/specs/ExactMath.spec
```diff
@@ -24,3 +24,10 @@ rule mulDivLifLLTV(uint256 a, uint256 lif, uint256 lltv) {
     // lif > 0, see rule maxLifIsAtLeastWad.
     assert lltv * lif <= WAD() * WAD() => MulDiv.mulDivUp(a, lltv, WAD()) <= MulDiv.mulDivUp(a, WAD(), lif);
 }
+
+/// Check that for created markets with LLTV < WAD, the imprecision factor of the maxRepaid computation is bounded by 1000.
+rule maxRepaidImprecisionFactorIsAtMostOneThousand(uint256 lltv, uint256 lif) {
```
QGarchery, line 30: to recover the `testRcfBound` that was removed

certora/specs/Midnight.spec
```diff
@@ -165,5 +165,5 @@ strong invariant noCreditAndDebt(bytes32 id, address user)
 strong invariant enabledLltvIsLessThanOrEqualToOne(uint256 lltv)
     currentContract.isLltvEnabled[lltv] => lltv <= WAD();
 
-strong invariant enabledLiquidationCursorsAreBounded(uint256 liquidationCursor)
+strong invariant enabledLiquidationCursorsIsLessThanOrEqualToOne(uint256 liquidationCursor)
```
QGarchery, line 168: consistency with the above invariant

certora/specs/ExactMath.spec
```diff
@@ -24,3 +24,10 @@ rule mulDivLifLLTV(uint256 a, uint256 lif, uint256 lltv) {
...
+    require lltv < WAD(), "assume that lltv < WAD";
+    require lltv == WAD() || lltv * lif <= 999 * 10 ^ 15 * WAD(), "see createdMarketsRespectMaxLifBound in CreatedMarkets.spec";
+    assert WAD() * WAD() / (WAD() * WAD() - lif * lltv) <= 1000;
```
MathisGD, line 33: why is that the rounding? we could write it

MathisGD (approved):

src/Midnight.sol
```diff
@@ -684,13 +684,11 @@ contract Midnight is IMidnight {
                 seizedAssets = repaidUnits.mulDivDown(lif, WAD).mulDivDown(ORACLE_PRICE_SCALE, liquidatedCollatPrice);
             }
 
-            if (!postMaturityMode) {
+            if (!postMaturityMode && lltv < WAD) {
                 // Note that debt >= maxDebt in this branch.
                 // For lltv == WAD, the RCF is inactive (see LIQUIDATIONS).
```
adhusson, line 689: could move this

certora/specs/ExactMath.spec
QGarchery: added the definition of the imprecision, but not the details as they are a bit too much (it's [this issue](https://cantina.xyz/code/6b7e5600-4961-4cad-a9bd-4f952789d07f/findings?finding=11))

src/Midnight.sol
QGarchery: thought about it, but it's nice to have it close to the next line

adhusson: not a strong take, ok with both

certora/specs/ExactMath.spec
```diff
@@ -24,3 +24,11 @@ rule mulDivLifLLTV(uint256 a, uint256 lif, uint256 lltv) {
     // lif > 0, see rule maxLifIsAtLeastWad.
     assert lltv * lif <= WAD() * WAD() => MulDiv.mulDivUp(a, lltv, WAD()) <= MulDiv.mulDivUp(a, WAD(), lif);
 }
+
+/// Check that for created markets with LLTV < WAD, the imprecision of rcfThreshold check is bounded by a factor of 1000.
+/// The imprecision is the difference between what is checked and the requirement wanted: minNewCollateral * liquidatedCollatPrice / LIF < rcfThreshold.
```
adhusson:
```suggestion
/// The imprecision is the difference between what is checked and the requirement wanted: minRemainingCollateralValue / LIF < rcfThreshold.
```

(to update in midnight too)

adhusson (approved):

certora/specs/ExactMath.spec
QGarchery: I got confused, sorry: the imprecision comes from the `maxRepaid` computation only, not the `rcfThreshold`. 

In Midnight we need `liquidatedCollatPrice` for the following proof that it's the same as the check we do in the code

QGarchery: The imprecision comes in part because this:
Screenshot 2026-06-24 at 19 07 15
omits the way max debt changes: you remove the old collateral contribution of the collateral token being liquidated, and add another collateral contribution for the same token but with a different amount

(and all the rounding errors associated to it)

MathisGD: maybe we should apply the change to the midnight natspec?

MathisGD: it's still not clear to me why this spec checks what we need

adhusson: `maxRepaid` is `ceil((debt - maxDebt) * WAD^2 / (WAD^2 - lif*lltv))` (`maxDebt` can be underestimated by < 2 per collateral).

Ideally from this we want to compute `maxRepaid = ceil((debt-maxDebt)/(1 - LIF*LLTV))` where LIF and LLTV are the real LIF and real LLTV.

If LIF*LLTV <= 0.999 then maxRepaid increases by at most 1000 for every increase in `debt-maxDebt`. So the bound shows how much the maxDebt error can be amplified into the maxRepaid error.

QGarchery: > Ideally from this we want to compute maxRepaid = ceil((debt-maxDebt)/(1 - LIF*LLTV)) where LIF and LLTV are the real LIF and real LLTV.

ideally we want `maxRepaid` to be the biggest value such that the position would still be unhealthy. So we want the max `repaidUnit` value such that `newDebt >= newMaxDebt`. We have `newDebt = debt - repaidUnit`, but the `newMaxDebt` is a bit more tricky: `newMaxDebt = maxDebt - collateral *↓ price *↓ lltv + (collateral - seizedAssets) *↓ price *↓ lltv` where `seizedAssets` is defined as usual. If there was no rounding errors, this would give `repaidUnits <= (debt - maxDebt) / (1 - lif * lltv)`

QGarchery: full proof of why `newDebt >= newMaxDebt` leads to `repaidUnits <= (debt - maxDebt) / (1 - lif * lltv) + X` in this file
[Midnight_rounding (1).pdf](https://github.com/user-attachments/files/29326810/Midnight_rounding.1.pdf)

MathisGD (approved):

certora/specs/ExactMath.spec
MathisGD: what's important here is that the rounding is X / (1-lif*lltv), and X<=2

adhusson: btw initially i thought the imprecision came from wad scaling in `... * WAD^2 / (WAD^2 - lif*lltv)` but no, it's just about the value of `LIF*LLTV`, and with wad scaling the last value before `LIF*LLTV = 1` is `LIF*LLTV = 1-1e-36`.

PR #1028, opened 2026-06-24
remove bundles

### LINKED ISSUE #678: move Bundles to external repos
- [ ] rename "V1"
- [ ] move to new repo

MathisGD:
fixes #678


qdrn (approved):

QGarchery (approved):

test/MidnightBundlesTest.sol
QGarchery, line 1: we can also remove `test/vendor/Permit2.sol`

src/periphery/MidnightBundles.sol
adhusson, line 1: Should remove src/periphery/ConsumableUnitsLib.sol as well

MathisGD: I think it's useful for all future bundlers so I would keep (same as takeAmountsLib)

adhusson: It's not tested at all now, we should add tests. (made an issue)

adhusson (approved):

PR #1034, opened 2026-06-25
[Certora] Performance fix for Role specifiction.

jhoenicke:
No need to model mulDiv and msb but use non-deterministic summary. Also model reverting behavior.

### DIFF
--- a/certora/specs/Role.spec
+++ b/certora/specs/Role.spec
@@ -21,6 +21,13 @@ methods {
     // This function is over-approximated, except for the reverting behavior. This is still sound as it is only used inside take but we don't look at the reverting behavior of take in this file.
     function TickLib.tickToPrice(uint256) internal returns (uint256) => NONDET;
 
+    // Summarize mulDivDown and mulDivUp by deterministic ghost functions (also modelling reverts).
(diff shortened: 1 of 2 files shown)


MathisGD (approved):

certora/specs/Role.spec
```diff
@@ -40,6 +47,26 @@ definition marketSettlementFee(bytes32 id, uint256 index) returns uint256 = asse
 
 definition defaultSettlementFee(address loanToken, uint256 index) returns uint256 = assert_uint256(currentContract.defaultSettlementFeeCbp[loanToken][index] * CBP());
 
+persistent ghost ghostMulDivDown(mathint, mathint, mathint) returns mathint;
+
+persistent ghost ghostMulDivUp(mathint, mathint, mathint) returns mathint;
```
QGarchery:
```suggestion
persistent ghost ghostMulDivDown(uint256, uint256, uint256) returns uint256;

persistent ghost ghostMulDivUp(uint256, uint256, uint256) returns uint256;
```
to avoid the `require_uint256` below

QGarchery (approved):

certora/specs/Role.spec
jhoenicke: done; also removed require_uint256.

QGarchery (approved):

PR #1037, opened 2026-06-26
docs: update comments on asset based offers behavior

MathisGD:
trust-R-14


src/Midnight.sol
adhusson, line 1: https://github.com/morpho-org/midnight/pull/1039

adhusson: It's about R-15?

adhusson (approved):

src/Midnight.sol
QGarchery: not sure either what it fixes exactly

QGarchery (approved):

src/Midnight.sol
QGarchery: Found it in R-15:
> The Midnight offer-cap documentation states:
`/// @dev It is possible to give units to a fully consumed assets-based buy offer with
price < 1.`
This can also apply outside the stated buy-offer case, namely for sell offers with zero price. It
is recommended to replace the comment with:
`/// @dev Fully consumed assets-based offers can still be taken for positive units
when the consumed asset amount rounds to zero.`

PR #1038, opened 2026-06-27
docs: maxRepaid rounding

MathisGD:
trust-R-15.


src/Midnight.sol
```diff
@@ -70,6 +70,7 @@ import {IMidnight, Market, Offer, CollateralParams, MarketState, Position} from
 /// @dev The RCF condition is (omitting scaling and roundings):
 ///   newDebt >= newMaxDebt <=> debt - repaidUnits >= maxDebt - repaidUnits*LIF*LLTV
 ///                         <=> repaidUnits <= (debt-maxDebt) / (1 - LIF*LLTV).
+/// @dev maxRepaid is rounded up such that it's always possible to liquidate enough to put back the account into health.
```
adhusson: there are some conditions to "always possible" (like there is enough collat), maybe should be reworded slightly less strongly

MathisGD: @adhusson wdyt
```suggestion
/// @dev maxRepaid is rounded up such that it doesn't prevent to liquidate enough to put back the account into health.
```

QGarchery (approved):

adhusson (approved):

PR #1039, opened 2026-06-29
[1037 review] update comment and add matching test

adhusson:


test/TakeTest.sol
```diff
@@ -1062,6 +1062,40 @@ contract TakeTest is BaseTest {
         assertGt(midnight.totalUnits(id), totalUnitsBefore);
     }
 
+    // Show that a sell offer with offerPrice = 0 can be taken with units > 0
+    function testBugSellMaxAssetsBypass() public {
```
adhusson, line 1066: matches the test just above `testBugBuyMaxAssetsBypass`

MathisGD (approved):

PR #1047, opened 2026-06-30
explain multi maker tree effect on cancel

adhusson:


src/Midnight.sol
```diff
@@ -91,8 +91,8 @@ import {IMidnight, Market, Offer, CollateralParams, MarketState, Position} from
 ///
 /// GROUPS
 /// @dev Groups are useful to have a global offered amount shared across multiple offers ("One cancels the other").
-/// @dev To work as expected, all offers in the same group should have the same direction (offer.buy), max values and
-/// loan token.
+/// @dev To work as expected, all offers in the same group should have the same maker, direction (offer.buy), max values
+/// and loan token.
```
QGarchery, line 95: A tree can contain multiple groups, so this comment doesn't prevent a tree from having different `maker` values

src/ratifiers/EcrecoverRatifier.sol
MathisGD, line 1: setterRatifier has the same issue right?

MathisGD, line 1: alternative would be to have cancelSig (which checks the sig (and authorization) to know if you can cancel or not)

src/ratifiers/EcrecoverRatifier.sol
```diff
@@ -25,6 +25,8 @@ contract EcrecoverRatifier is IEcrecoverRatifier {
         MIDNIGHT = _midnight;
     }
 
+    /// @dev All offers in a tree are expected to share the same maker and ratifier. Otherwise all offers in a
+    /// tree might not be cancelled by a single call to this function.
```
MathisGD, line 29:
```suggestion
```

src/Midnight.sol
adhusson: yes, the comment is about the group since the consumed is also keyed by the maker. So "to work as expected" this is necessary.

Having multiple makers per tree by itself does not break anything.

src/ratifiers/EcrecoverRatifier.sol
adhusson: note this is the "restrict per tree" part of the docs (the comment in midnight is about restricting per group). without it [this message is no longer true](https://morpholabs.slack.com/archives/C08A40KKN9W/p1782817814279369?thread_ts=1782808061.630369&cid=C08A40KKN9W)

src/ratifiers/EcrecoverRatifier.sol
adhusson: not great, also will have to check malleability

QGarchery (approved):

MathisGD (approved):

PR #1048, opened 2026-06-30
collat activation revert on reverting oracle

MathisGD:
the scenario we are trying to prevent is the one that was documented:
```
/// @dev Note that a borrower can activate a collateral once its oracle is reverting because the oracle is not called in supplyCollateral.
```
at first we didn't implement a fix because we thought that it would make the liveness spec much more complex, but we went back on that. Actually it's ok. so we decided to implement the fix.

### DIFF
--- a/certora/specs/BalanceEffects.spec
+++ b/certora/specs/BalanceEffects.spec
@@ -21,9 +21,6 @@ methods {
     function UtilsLib.msb(uint128) internal returns (uint256) => NONDET;
     function TickLib.tickToPrice(uint256) internal returns (uint256) => NONDET;
 
-    // The oracle price is irrelevant to credit, debt, and collateral tracking.
-    function _.price() external => NONDET;
-
     // Assume no reentrancy: callbacks and token transfers do not re-enter Midnight.
     // This is justified because the properties we verify are about the effect of each function's own body on credit and debt, not the effect of the full transaction including callbacks.
     function _.onBuy(bytes32, Midnight.Market, uint256, uint256, uint256, address, bytes) external => NONDET;
--- a/certora/specs/BalanceEffects.spec
+++ b/certora/specs/BalanceEffects.spec
@@ -21,6 +21,9 @@ methods {
     function UtilsLib.msb(uint128) internal returns (uint256) => NONDET;
     function TickLib.tickToPrice(uint256) internal returns (uint256) => NONDET;
 
+    // The oracle price is irrelevant to credit, debt, and collateral tracking.
+    function _.price() external => NONDET;
+
     // Assume no reentrancy: callbacks and token transfers do not re-enter Midnight.
     // This is justified because the properties we verify are about the effect of each function's own body on credit and debt, not the effect of the full transaction including callbacks.
     function _.onBuy(bytes32, Midnight.Market, uint256, uint256, uint256, address, bytes) external => NONDET;
--- a/certora/specs/CreatedMarkets.spec
+++ b/certora/specs/CreatedMarkets.spec
@@ -31,6 +31,9 @@ methods {
     // Tokens are assumed to not reenter, for performance reasons.
     function SafeTransferLib.safeTransferFrom(address, address, address, uint256) internal => NONDET;
     function SafeTransferLib.safeTransfer(address, address, uint256) internal => NONDET;
+
+    // The oracle price is irrelevant to market creation.
+    function _.price() external => NONDET;
 }
 
 /// HELPERS ///
--- a/certora/specs/CreatedMarkets.spec
+++ b/certora/specs/CreatedMarkets.spec
@@ -31,9 +31,6 @@ methods {
     // Tokens are assumed to not reenter, for performance reasons.
     function SafeTransferLib.safeTransferFrom(address, address, address, uint256) internal => NONDET;
     function SafeTransferLib.safeTransfer(address, address, uint256) internal => NONDET;
-
-    // The oracle price is irrelevant to market creation.
-    function _.price() external => NONDET;
 }
 
 /// HELPERS ///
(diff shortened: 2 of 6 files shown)


certora/specs/CreatedMarkets.spec
```diff
@@ -31,6 +31,9 @@ methods {
     // Tokens are assumed to not reenter, for performance reasons.
     function SafeTransferLib.safeTransferFrom(address, address, address, uint256) internal => NONDET;
     function SafeTransferLib.safeTransfer(address, address, uint256) internal => NONDET;
+
+    // The oracle price is irrelevant to market creation.
+    function _.price() external => NONDET;
```
MathisGD:
```suggestion
```

certora/specs/BalanceEffects.spec
```diff
@@ -21,6 +21,9 @@ methods {
     function UtilsLib.msb(uint128) internal returns (uint256) => NONDET;
     function TickLib.tickToPrice(uint256) internal returns (uint256) => NONDET;
 
+    // The oracle price is irrelevant to credit, debt, and collateral tracking.
+    function _.price() external => NONDET;
+
```
MathisGD:
```suggestion
```

src/Midnight.sol
```diff
@@ -576,6 +574,9 @@ contract Midnight is IMidnight {
             );
         }
 
+        // Calling the oracle prevents activating a collateral whose oracle is reverting.
+        IOracle(market.collateralParams[collateralIndex].oracle).price();
```
```suggestion
            // Calling the oracle prevents activating a collateral whose oracle is reverting.
                        IOracle(market.collateralParams[collateralIndex].oracle).price();
        }
```

src/Midnight.sol
```diff
@@ -152,6 +149,7 @@ import {IMidnight, Market, Offer, CollateralParams, MarketState, Position} from
 ///
 /// LIVENESS
 /// @dev If an activated collateral oracle reverts on price, liquidate reverts.
+/// @dev If the supplied collateral oracle reverts on price, supplyCollateral reverts.
```
```suggestion
/// @dev If the supplied collateral oracle reverts on price, the activation of that collat through supplyCollateral reverts.
```

certora/specs/Reverts.spec
```diff
@@ -192,6 +192,15 @@ rule oracleRevertCausesLiquidateRevert(env e, Midnight.Market market, uint256 co
     assert lastReverted;
 }
 
+/// If the supplied collateral oracle reverts on price, supplyCollateral reverts.
+rule oracleRevertCausesSupplyCollateralRevert(env e, Midnight.Market market, uint256 collateralIndex, uint256 assets, address onBehalf) {
+    require singleRevertingOracle == market.collateralParams[collateralIndex].oracle, "oracle is reverting";
+
+    supplyCollateral@withrevert(e, market, collateralIndex, assets, onBehalf);
+
+    assert lastReverted;
```
MathisGD:
```suggestion
    assert assets > 0 && !summaryGetBit(bitmap, revertingCollateralIndex) => lastReverted;
```

certora/specs/Reverts.spec
```diff
@@ -192,6 +192,15 @@ rule oracleRevertCausesLiquidateRevert(env e, Midnight.Market market, uint256 co
     assert lastReverted;
 }
 
+/// If the supplied collateral oracle reverts on price, supplyCollateral reverts.
+rule oracleRevertCausesSupplyCollateralRevert(env e, Midnight.Market market, uint256 collateralIndex, uint256 assets, address onBehalf) {
```
MathisGD:
```suggestion
rule oracleRevertCausesCollatActivationRevert(env e, Midnight.Market market, uint256 collateralIndex, uint256 assets, address onBehalf) {
```

src/Midnight.sol
MathisGD: the Solvency spec fails because there is indeed a transient solvency issue at the moment the oracle is called. we could simply move all this block above the collateral update, wdyt?

adhusson: moving above the block is OK (only difference is a weird reentering token could change the collateral set and make the supply possible but it's not bad)

src/Midnight.sol
```diff
@@ -563,19 +563,21 @@ contract Midnight is IMidnight {
...
+            // Calling the oracle prevents activating a collateral whose oracle is reverting.
+            IOracle(market.collateralParams[collateralIndex].oracle).price();
             uint128 newCollateralBitmap = _position.collateralBitmap.setBit(collateralIndex);
             _position.collateralBitmap = newCollateralBitmap;
             require(
                 UtilsLib.countBits(newCollateralBitmap) <= MAX_COLLATERALS_PER_BORROWER, TooManyActivatedCollaterals()
             );
         }
 
+        _position.collateral[collateralIndex] = UtilsLib.toUint128(oldCollateral + assets);
```
MathisGD, line 579: moved below to fix the solvency rule

src/Midnight.sol
```diff
@@ -152,6 +150,8 @@ import {IMidnight, Market, Offer, CollateralParams, MarketState, Position} from
 ///
 /// LIVENESS
 /// @dev If an activated collateral oracle reverts on price, liquidate reverts.
+/// @dev If the supplied collateral oracle reverts on price, the activation of that collat through supplyCollateral
```
```suggestion
/// @dev If the supplied collateral oracle reverts on price, the activation of that collateral through supplyCollateral
```

adhusson (approved):

QGarchery (approved):

src/Midnight.sol
QGarchery: see https://morpholabs.slack.com/archives/C0ACSA8TNMA/p1782908148328449

PR #1049, opened 2026-07-01
re-add buyerCreditIncrease and sellerCreditDecrease

MathisGD:


src/Midnight.sol
```diff
@@ -929,10 +929,6 @@ contract Midnight is IMidnight {
         return (feeLower * (end - timeToMaturity) + feeUpper * (timeToMaturity - start)) / (end - start);
     }
 
-    function hasCredit(bytes32 id, address user) internal view returns (bool) {
-        return position[id][user].credit > 0;
-    }
```
MathisGD, line 937: why not now that we have a munging :-)

adhusson: i'm not sure it's worth it

munging is scary soundness wise

src/Midnight.sol
```diff
@@ -407,8 +407,8 @@ contract Midnight is IMidnight {
         Position storage buyerPos = position[id][buyer];
         Position storage sellerPos = position[id][seller];
 
-        if (hasCredit(id, buyer) || units > buyerPos.debt) _updatePosition(offer.market, id, buyer);
-        if (hasCredit(id, seller)) _updatePosition(offer.market, id, seller);
+        if (position[id][buyer].credit > 0 || units > buyerPos.debt) _updatePosition(offer.market, id, buyer);
+        if (position[id][seller].credit > 0) _updatePosition(offer.market, id, seller);
```
```suggestion
        if (buyerPos.credit > 0 || units > buyerPos.debt) _updatePosition(offer.market, id, buyer);
        if (sellerPos.credit > 0) _updatePosition(offer.market, id, seller);
```

src/Midnight.sol
adhusson: ok with this one

adhusson (approved):

PR #1055, opened 2026-07-01
add onchain-mempool

adhusson:


MathisGD (approved):

src/periphery/Log.sol
```diff
@@ -0,0 +1,18 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Copyright (c) 2025 Morpho Association
```
QGarchery:
```suggestion
// Copyright (c) 2026 Morpho Association
```

src/periphery/Log.sol
QGarchery, line 1: Maybe worth to have more `optimizer_runs` for this contract (using compiler restrictions/profiles)

adhusson: agree i was doing exactlythis in the deployment repo

QGarchery (approved):

MathisGD (approved):

PR #1067, opened 2026-07-17
blue buy callback

MathisGD:
important choices:
- the callback holds the position on Blue. it acts as an escrow. the ux is better and it's simpler to think about (everywhere in the stack).
- it's a per user contract instead of a singleton. this allows to rely on authorizations, and keeps the callback super simple. so it comes with a factory as well.
- the event pattern is different than usual, because we try to leverage the blue events for authorizations. we just need to log the msg.sender and nonce in setAuthorizedWithSig.


src/periphery/BlueBuyCallbackFactory.sol
```diff
@@ -0,0 +1,27 @@
...
+    address public immutable BLUE;
+
+    mapping(address owner => address) public callbackOf;
+
+    constructor(address _midnight, address _blue) {
+        MIDNIGHT = _midnight;
+        BLUE = _blue;
+    }
+
+    function createBlueBuyCallback() external {
```
peyha: might be worth to be able to create a callback on behalf of another contract

src/periphery/BlueBuyCallback.sol
```diff
@@ -0,0 +1,53 @@
...
+        require(blueMarketParams.loanToken == market.loanToken, InconsistentLoanToken());
+
+        IBlue(BLUE).withdraw(blueMarketParams, buyerAssets, 0, address(this), address(this));
+        safeApprove(market.loanToken, MIDNIGHT, buyerAssets);
+
+        return CALLBACK_SUCCESS;
+    }
+
+    function safeApprove(address token, address spender, uint256 value) internal {
+        (bool success, bytes memory returndata) = token.call(abi.encodeWithSelector(0x095ea7b3, spender, value));
```
peyha: nitpick but i tend to prefer `abi.encodeWithSignature`

src/periphery/BlueBuyCallback.sol
```diff
@@ -0,0 +1,53 @@
...
+        bytes memory data
+    ) external returns (bytes32) {
+        require(msg.sender == MIDNIGHT, NotMidnight());
+        require(buyer == OWNER, NotOwnerBuyer());
+
+        BlueMarketParams memory blueMarketParams = abi.decode(data, (BlueMarketParams));
+        require(blueMarketParams.loanToken == market.loanToken, InconsistentLoanToken());
+
+        IBlue(BLUE).withdraw(blueMarketParams, buyerAssets, 0, address(this), address(this));
+        safeApprove(market.loanToken, MIDNIGHT, buyerAssets);
```
peyha: in practice there should only be very few loan tokens so could this callback contract only handle a single loan token ? (and max approve it on midnight in the constructor)

src/periphery/BlueBuyCallback.sol
MathisGD: the thing is we don't have IERC20 with approve available, but I can add it

src/periphery/BlueBuyCallback.sol
MathisGD: we don't gain much no?

peyha: it's still ~45k gas saved but it adds a few constraint to the design so not sure

src/periphery/BlueBuyCallback.sol
peyha, line 1: might be worth implementing a `withdraw` function to let the owner withdraw from blue if he ends up not using midnight (maybe he can also self match on midnight but that's fairly complex)

src/periphery/BlueBuyCallbackFactory.sol
```diff
@@ -0,0 +1,29 @@
...
+        MIDNIGHT = _midnight;
+        BLUE = _blue;
+    }
+
+    function createBlueBuyCallback(address owner) external {
+        require(callbackOf[owner] == address(0), AlreadyCreated());
+
+        address callback = address(new BlueBuyCallback(owner, MIDNIGHT, BLUE));
+        callbackOf[owner] = callback;
+        isBlueCallback[callback] = true;
```
peyha, line 23: not mandatory, if we use create2 to create the contract, we could just check that there is some code at the corresponding address

src/periphery/BlueBuyCallback.sol
peyha: oh wait it's not necessary the callback authorizes the owner on blue at construction so the owner can withdraw safely

src/periphery/BlueBuyCallbackFactory.sol
MathisGD: indeed, but we usually still do it as it's more practical and it's used a lot

src/periphery/BlueBuyCallback.sol
MathisGD: or we can do the same design as in bundles?

src/periphery/BlueBuyCallback.sol
peyha: probably worth having a test for this

adhusson, line 1: A factory costs more with [upcoming state creation cost increases](https://eips.ethereum.org/EIPS/eip-8037). Is there a strong reason to avoid having 1 contract with a large position?

src/periphery/BlueBuyCallback.sol
```diff
@@ -0,0 +1,53 @@
...
+
+    constructor(address _owner, address _midnight, address _blue) {
+        OWNER = _owner;
+        MIDNIGHT = _midnight;
+        BLUE = _blue;
+
+        IMorpho(BLUE).setAuthorization(OWNER, true);
+    }
+
+    function onBuy(
```
adhusson, line 57:
```suggestion
    /// @dev Anyone authorized by the owner on Midnight can pull from the Blue position held by this callback contract by making the owner buy dummy credit on Midnight.
    function onBuy(
```

adhusson: (could also add a note about offers reverting if the pool does not have buyerAssets but less improtant)

src/periphery/BlueBuyCallbackFactory.sol
```diff
@@ -0,0 +1,29 @@
...
+
+        emit CreateBlueBuyCallback(owner, callback);
```
adhusson, line 25:
```suggestion
        emit CreateBlueBuyCallback(owner, callback);
        return callback;
```

src/periphery/BlueBuyCallback.sol
peyha: you mean approving max only once, yes I would think it's better

src/periphery/BlueBuyCallbackFactory.sol
adhusson: agree with @peyha, create2 would be nicer for offchain, no need to take asynchrony into account

src/periphery/BlueBuyCallback.sol
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    function safeApprove(address token, address spender, uint256 value) internal {
+        (bool success, bytes memory returndata) = token.call(abi.encodeWithSelector(0x095ea7b3, spender, value));
+        if (!success) {
+            assembly ("memory-safe") {
+                revert(add(returndata, 0x20), mload(returndata))
+            }
+        }
+        require(returndata.length == 0 || abi.decode(returndata, (bool)), ApproveReturnedFalse());
```
adhusson, line 94: not tested

src/periphery/BlueBuyCallback.sol
```diff
@@ -0,0 +1,53 @@
...
+        bytes memory data
+    ) external returns (bytes32) {
+        require(msg.sender == MIDNIGHT, NotMidnight());
+        require(buyer == OWNER, NotOwnerBuyer());
+
+        MarketParams memory blueMarketParams = abi.decode(data, (MarketParams));
+        require(blueMarketParams.loanToken == market.loanToken, InconsistentLoanToken());
+
+        if (buyerAssets > 0) IMorpho(BLUE).withdraw(blueMarketParams, buyerAssets, 0, address(this), address(this));
+        safeApprove(market.loanToken, MIDNIGHT, buyerAssets);
```
adhusson: why approve if there are 0 buyer assets?

src/periphery/BlueBuyCallbackFactory.sol
MathisGD: will create2 anyway, like in [vaultv2](https://github.com/morpho-org/vault-v2/blob/main/src/VaultV2Factory.sol#L14)

src/periphery/BlueBuyCallback.sol
MathisGD: My initial motivation was analytics (it will appear everywhere, you need to disambiguate by hand...). also note that it will need to hold mappings of what position is to who, which will take some state too

MathisGD: otherwise we abandon the idea of the escrow and the callback withdraws on behalf the user..

src/periphery/BlueBuyCallback.sol
MathisGD: will switch to something more efficient

https://github.com/morpho-org/midnight/pull/1067#discussion_r3612835183

src/periphery/BlueBuyCallback.sol
MathisGD: solved now, I needed the allowance function anyway

src/periphery/BlueBuyCallback.sol
adhusson: > which will take some state too

I don't think it's different N slots in 1 contract vs N slots in N contracts though (not sure i don't know EIP-8037 well at all).

It depends on the pros/cons of escrow, I'm not clear on that we can discuss live.

lib/forge-std
QGarchery, line 1: is updating needed here ? it causes a mismatch with what's in foundry.lock

lib/morpho-blue
QGarchery, line 1: but here it would make sense, v1.0.0 is very old

test/imports/MorphoImport.sol
QGarchery, line 1: we can move it to the `test` folder (should change the `foundry.toml` compiler restrictions as well)

src/periphery/BlueBuyCallback.sol
```diff
@@ -0,0 +1,71 @@
...
+
+import {IMorpho, MarketParams} from "morpho-blue/src/interfaces/IMorpho.sol";
+import {Market} from "../interfaces/IMidnight.sol";
+import {CALLBACK_SUCCESS} from "../libraries/ConstantsLib.sol";
+import {IBlueBuyCallback} from "./interfaces/IBlueBuyCallback.sol";
+
+interface IERC20 {
+    function allowance(address owner, address spender) external view returns (uint256);
+    function approve(address spender, uint256 value) external returns (bool);
+}
```
QGarchery, line 14: to put in its own file ? annoying that we already have an IERC20 file though

test/UtilsLibTest.sol
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: GPL-2.0-or-later
-pragma solidity ^0.8.0;
+pragma solidity ^0.8.34;
```
QGarchery: to revert

src/periphery/BlueBuyCallback.sol
QGarchery, line 1: May be worth to add it to the README. It seems that the whole periphery folder is missing from it actually, but the rest is ok since it appears in the whitepaper

src/periphery/BlueBuyCallback.sol
```diff
@@ -0,0 +1,71 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Copyright (c) 2026 Morpho Association
+pragma solidity 0.8.34;
+
+import {IMorpho, MarketParams} from "morpho-blue/src/interfaces/IMorpho.sol";
```
QGarchery:
```suggestion
import {IMorpho, MarketParams} from "../../lib/morpho-blue/src/interfaces/IMorpho.sol";
```

test/BlueBuyCallbackFactoryTest.sol
```diff
@@ -0,0 +1,96 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Copyright (c) 2026 Morpho Association
+pragma solidity ^0.8.0;
+
+import {Test} from "../lib/forge-std/src/Test.sol";
+import {Vm} from "../lib/forge-std/src/Vm.sol";
+import {IMorpho} from "morpho-blue/src/interfaces/IMorpho.sol";
```
QGarchery:
```suggestion
import {IMorpho} from "../lib/morpho-blue/src/interfaces/IMorpho.sol";
```

test/BlueBuyCallbackTest.sol
```diff
@@ -0,0 +1,84 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Copyright (c) 2026 Morpho Association
+pragma solidity ^0.8.0;
+
+import {Test} from "../lib/forge-std/src/Test.sol";
+import {IMorpho, MarketParams} from "morpho-blue/src/interfaces/IMorpho.sol";
```
QGarchery:
```suggestion
import {IMorpho, MarketParams} from "../lib/morpho-blue/src/interfaces/IMorpho.sol";
```

src/periphery/interfaces/IBlueBuyCallbackFactory.sol
```diff
@@ -0,0 +1,20 @@
...
+    event CreateBlueBuyCallback(address indexed owner, address callback);
+
+    /// STORAGE GETTERS ///
+    function MIDNIGHT() external view returns (address);
+    function BLUE() external view returns (address);
+    function callbackOf(address owner) external view returns (address);
+    function isBlueCallback(address callback) external view returns (bool);
+
+    /// FUNCTIONS ///
+    function createBlueBuyCallback(address owner) external returns (address callback);
```
QGarchery:
```suggestion
    function createBlueBuyCallback(address owner) external returns (address);
```

src/periphery/BlueBuyCallback.sol
```diff
@@ -0,0 +1,71 @@
...
+
+interface IERC20 {
+    function allowance(address owner, address spender) external view returns (uint256);
+    function approve(address spender, uint256 value) external returns (bool);
+}
+
+/// @dev Anyone authorized by the owner on Midnight can pull from the Blue position held by this callback contract by
+/// making the owner buy dummy credit on Midnight.
+/// @dev Reverts if the owner position on the requested market is too small or if the liquidity on that market is too
+/// small.
```
QGarchery:
```suggestion
/// small.
/// @dev No-ops are not systematically prevented.
/// @dev Zero checks are not systematically performed.
```

test/UtilsLibTest.sol
MathisGD: I did that to fix the CI (but it's not enough apparently)

foundry.toml
QGarchery, line 7:
```suggestion
fs_permissions = [
    { access = "read", path = "test/ticks_exact.json" },
    { access = "read", path = "out/Morpho.sol/Morpho.json" },
]
```
this seems to allow`forge t --mc Blue` to work

src/periphery/BlueBuyCallbackFactory.sol
```diff
@@ -0,0 +1,30 @@
...
+    mapping(address callback => bool) public isBlueCallback;
+
+    constructor(address _midnight, address _blue) {
+        MIDNIGHT = _midnight;
+        BLUE = _blue;
+    }
+
+    function createBlueBuyCallback(address owner) external returns (address) {
+        require(callbackOf[owner] == address(0), AlreadyCreated());
+
```
QGarchery:
```suggestion
```
no need because it will revert because of collision

test/BlueBuyCallbackFactoryTest.sol
```diff
@@ -0,0 +1,96 @@
...
+        vm.prank(otherOwner);
+        factory.createBlueBuyCallback(otherOwner);
+
+        address callback = factory.callbackOf(owner);
+        address otherCallback = factory.callbackOf(otherOwner);
+
+        assertTrue(callback != otherCallback);
+        assertEq(factory.callbackOf(owner), callback);
+        assertEq(factory.callbackOf(otherOwner), otherCallback);
+    }
```
peyha: not sure this test is particularly relevant

src/periphery/interfaces/IBlueBuyCallbackFactory.sol
MathisGD: actually it's ambiguous no?

src/periphery/BlueBuyCallback.sol
MathisGD: it risks to be confusing if we have two IERC20 imo

src/periphery/interfaces/IBlueBuyCallbackFactory.sol
QGarchery: I don't see it, what else could it be ?

lib/forge-std
QGarchery: I think there is still a mismatch with what's in foundry.lock

test/imports/MorphoImport.sol
QGarchery: I think this wasn't changed, but I find it cleaner than having it in `src`. I tried locally and it seems to work

foundry.toml
MathisGD: works without

src/periphery/BlueBuyCallback.sol
MathisGD: wrote a readme in the periphery folder to not pollute the other one

src/periphery/BlueBuyCallback.sol
```diff
@@ -0,0 +1,78 @@
...
+        MIDNIGHT = _midnight;
+        BLUE = _blue;
+
+        IMorpho(BLUE).setAuthorization(OWNER, true);
+    }
+
+    function setAuthorization(address authorized, bool newIsAuthorized) external {
+        require(msg.sender == OWNER, NotOwner());
+        IMorpho(BLUE).setAuthorization(authorized, newIsAuthorized);
+    }
```
QGarchery, line 37: it is possible to authorize too many users, and then it is not possible to create another blue callback to start fresh (it's one per owner). Maybe worth a comment

src/periphery/BlueBuyCallback.sol
```diff
@@ -0,0 +1,78 @@
...
+
+    constructor(address _owner, address _midnight, address _blue) {
+        OWNER = _owner;
+        MIDNIGHT = _midnight;
+        BLUE = _blue;
+
+        IMorpho(BLUE).setAuthorization(OWNER, true);
+    }
+
+    function setAuthorization(address authorized, bool newIsAuthorized) external {
```
QGarchery, line 32: Shouldn't we do something like `setAuthorizationWithSig` as well ? Use case: you want to do a bundle that involves a withdraw from the blue callback, and so you want to authorize the bundle contract, and it's best if it's only done in one tx

QGarchery (approved):

src/periphery/BlueBuyCallback.sol
MathisGD: same as authorizing too much people and having to start a fresh EOA?

test/imports/MorphoImport.sol
MathisGD: sorry I misunderstood your request. you want it to put it where?

lib/forge-std
MathisGD: sorry I'm submodules noob. should be good now

src/periphery/BlueBuyCallback.sol
QGarchery: yes but here you can't start a fresh blue callback. Or you mean that you should also start a new EOA, not just a new callback ?

MathisGD: yes

src/periphery/BlueBuyCallback.sol
```diff
@@ -0,0 +1,97 @@
...
+        require(authorization.nonce == nonce++, InvalidNonce());
+
+        bytes32 hashStruct = keccak256(abi.encode(AUTHORIZATION_TYPEHASH, authorization));
+        bytes32 domainSeparator = keccak256(abi.encode(DOMAIN_TYPEHASH, block.chainid, address(this)));
+        bytes32 digest = keccak256(bytes.concat("\x19\x01", domainSeparator, hashStruct));
+        address signer = ecrecover(digest, signature.v, signature.r, signature.s);
+        require(signer != address(0) && signer == authorization.authorizer && signer == OWNER, InvalidSignature());
+
+        emit SetAuthorizationWithSig(
+            msg.sender, authorization.authorized, authorization.isAuthorized, authorization.nonce
```
peyha: we could also split this event in 2 distinct events `SetAuthorization` and `IncrementNonce` (like on blue)
it's a bit more expensive to interact but the indexing of authorizations is slightly more efficient

src/periphery/BlueBuyCallback.sol
```diff
@@ -0,0 +1,97 @@
...
+
+    function setAuthorizationWithSig(Authorization memory authorization, Signature calldata signature) external {
+        require(block.timestamp <= authorization.deadline, Expired());
+        require(authorization.nonce == nonce++, InvalidNonce());
+
+        bytes32 hashStruct = keccak256(abi.encode(AUTHORIZATION_TYPEHASH, authorization));
+        bytes32 domainSeparator = keccak256(abi.encode(DOMAIN_TYPEHASH, block.chainid, address(this)));
+        bytes32 digest = keccak256(bytes.concat("\x19\x01", domainSeparator, hashStruct));
+        address signer = ecrecover(digest, signature.v, signature.r, signature.s);
+        require(signer != address(0) && signer == authorization.authorizer && signer == OWNER, InvalidSignature());
```
peyha, line 47: do we need `authorizer` in `Authorization` as we know that it should be the owner ? we could use a new lighter `Authorization` custom struct

src/periphery/BlueBuyCallback.sol
MathisGD: feels simpler like that, feel free to re-open if you feel strongly about it

src/periphery/BlueBuyCallback.sol
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    function setAuthorization(address authorized, bool newIsAuthorized) external {
+        require(msg.sender == OWNER, NotOwner());
+        emit SetAuthorization(msg.sender, authorized, newIsAuthorized);
+        IMorpho(BLUE).setAuthorization(authorized, newIsAuthorized);
+    }
+
+    function setAuthorizationWithSig(Authorization memory authorization, Signature calldata signature) external {
+        require(block.timestamp <= authorization.deadline, Expired());
```
peyha:
```suggestion
        require(block.timestamp <= authorization.deadline, AuthorizationExpired());
```

src/periphery/BlueBuyCallback.sol
peyha: not strong at all, both choices look relevant

test/UtilsLibTest.sol
MathisGD:
```suggestion
pragma solidity ^0.8.0;
```

src/periphery/BlueBuyCallback.sol
QGarchery: I think it's not clear that it's what you should do: throwing your account because some other contract is not working anymore. So maybe best to still have a comment about it

MathisGD: unsure how to frame it, do you have an idea

src/periphery/BlueBuyCallback.sol
```diff
@@ -0,0 +1,97 @@
...
+        OWNER = _owner;
+        MIDNIGHT = _midnight;
+        BLUE = _blue;
+
+        IMorpho(BLUE).setAuthorization(OWNER, true);
+    }
+
+    function setAuthorization(address authorized, bool newIsAuthorized) external {
+        require(msg.sender == OWNER, NotOwner());
+        emit SetAuthorization(msg.sender, authorized, newIsAuthorized);
```
QGarchery:
```suggestion
```
not needed because the state is on Blue

src/periphery/BlueBuyCallback.sol
```diff
@@ -0,0 +1,97 @@
...
+
+        bytes32 hashStruct = keccak256(abi.encode(AUTHORIZATION_TYPEHASH, authorization));
+        bytes32 domainSeparator = keccak256(abi.encode(DOMAIN_TYPEHASH, block.chainid, address(this)));
+        bytes32 digest = keccak256(bytes.concat("\x19\x01", domainSeparator, hashStruct));
+        address signer = ecrecover(digest, signature.v, signature.r, signature.s);
+        require(signer != address(0) && signer == authorization.authorizer && signer == OWNER, InvalidSignature());
+
+        emit SetAuthorizationWithSig(
+            msg.sender, authorization.authorized, authorization.isAuthorized, authorization.nonce
+        );
```
QGarchery: same, unclear why we log that when the state in in Blue, except for the nonce. Other pattern is to log every function arg

src/periphery/BlueBuyCallback.sol
```diff
@@ -0,0 +1,97 @@
...
+        MIDNIGHT = _midnight;
+        BLUE = _blue;
+
+        IMorpho(BLUE).setAuthorization(OWNER, true);
+    }
+
+    function setAuthorization(address authorized, bool newIsAuthorized) external {
+        require(msg.sender == OWNER, NotOwner());
+        emit SetAuthorization(msg.sender, authorized, newIsAuthorized);
+        IMorpho(BLUE).setAuthorization(authorized, newIsAuthorized);
```
peyha:
```suggestion
        IMorpho(BLUE).setAuthorization(authorized, newIsAuthorized);
        emit SetAuthorization(msg.sender, authorized, newIsAuthorized);
```
nitpick but we usually order it this way (in blue & midnight)

src/periphery/BlueBuyCallback.sol
```diff
@@ -0,0 +1,97 @@
...
+        bytes32 hashStruct = keccak256(abi.encode(AUTHORIZATION_TYPEHASH, authorization));
+        bytes32 domainSeparator = keccak256(abi.encode(DOMAIN_TYPEHASH, block.chainid, address(this)));
+        bytes32 digest = keccak256(bytes.concat("\x19\x01", domainSeparator, hashStruct));
+        address signer = ecrecover(digest, signature.v, signature.r, signature.s);
+        require(signer != address(0) && signer == authorization.authorizer && signer == OWNER, InvalidSignature());
+
+        emit SetAuthorizationWithSig(
+            msg.sender, authorization.authorized, authorization.isAuthorized, authorization.nonce
+        );
+        IMorpho(BLUE).setAuthorization(authorization.authorized, authorization.isAuthorized);
```
QGarchery: don't have time to do dig into it now, will do later, but this is not a no-op on Blue if `BLUE.isAuthorized(OWNER, authorization.authorized) == authorization.isAuthorized`. Maybe for the use case of cancelling a signature, this would cause an issue ?

src/periphery/BlueBuyCallback.sol
```diff
@@ -0,0 +1,97 @@
...
+    constructor(address _owner, address _midnight, address _blue) {
+        OWNER = _owner;
+        MIDNIGHT = _midnight;
+        BLUE = _blue;
+
+        IMorpho(BLUE).setAuthorization(OWNER, true);
+    }
+
+    function setAuthorization(address authorized, bool newIsAuthorized) external {
+        require(msg.sender == OWNER, NotOwner());
```
peyha, line 33: should we add a check to prevent `authorized==OWNER` ? the owner could unauthorize himself and lose his position (as well as his ability to use the callback)

src/periphery/BlueBuyCallback.sol
MathisGD: but here it's a call that will itself have its event

src/periphery/BlueBuyCallback.sol
MathisGD: yes we need it for the nonce

src/periphery/BlueBuyCallback.sol
peyha: on second thought not sure it's worth a require maybe just a comment

src/periphery/BlueBuyCallback.sol
MathisGD: so we would log in `withSig` but not here?

src/periphery/BlueBuyCallback.sol
MathisGD: he can always re-authorize himself. still we might want to have it to have a nice invariant?

src/periphery/BlueBuyCallback.sol
MathisGD: not sure to understand, what can go wrong actually?

src/periphery/BlueBuyCallback.sol
peyha: I hallucinated a bit, worst case would be that the unauthorize himself so withdrawing his position from his EOA reverts

test/imports/MorphoImport.sol
QGarchery: my mistake sorry

src/periphery/BlueBuyCallback.sol
MathisGD: let's allow no-ops

src/periphery/BlueBuyCallback.sol
QGarchery: it would be more confusing to add a comment about it, and after reflection it doesn't seem like a big issue: same problem can happen with approving an ERC20 too much

src/periphery/BlueBuyCallback.sol
```diff
@@ -0,0 +1,96 @@
...
+import {IBlueBuyCallback} from "./interfaces/IBlueBuyCallback.sol";
+
+interface IERC20 {
+    function allowance(address owner, address spender) external view returns (uint256);
+    function approve(address spender, uint256 value) external returns (bool);
+}
+
+/// @dev Anyone authorized by the owner on Midnight can pull from the Blue position held by this callback contract by
+/// making the owner buy dummy credit on Midnight.
+/// @dev Reverts if the owner position on the requested market is too small or if the liquidity on that market is too
```
peyha: shouldn't this be either rephrased to "using this contract as midnight callback reverts if ..." or moved to `onBuy` ? (slight pref for the former)

src/periphery/BlueBuyCallbackFactory.sol
```diff
@@ -0,0 +1,28 @@
...
+contract BlueBuyCallbackFactory is IBlueBuyCallbackFactory {
+    address public immutable MIDNIGHT;
+    address public immutable BLUE;
+
+    mapping(address owner => address) public callbackOf;
+    mapping(address callback => bool) public isBlueCallback;
+
+    constructor(address _midnight, address _blue) {
+        MIDNIGHT = _midnight;
+        BLUE = _blue;
```
peyha, line 17: `MIDNIGHT` and `BLUE` are never logged, would it be worth having a `Constructor` event ?

QGarchery (approved):

src/periphery/BlueBuyCallbackFactory.sol
QGarchery: I don't think so because they have public getters

peyha: this should be fine as we don't intend to re-deploy neither blue nor midnight

peyha (approved):

src/periphery/BlueBuyCallback.sol
```diff
@@ -0,0 +1,96 @@
...
+    constructor(address _owner, address _midnight, address _blue) {
+        OWNER = _owner;
+        MIDNIGHT = _midnight;
+        BLUE = _blue;
+
+        IMorpho(BLUE).setAuthorization(OWNER, true);
+    }
+
+    function setAuthorization(address authorized, bool newIsAuthorized) external {
+        require(msg.sender == OWNER, NotOwner());
```
adhusson, line 33: Could do the midnight pattern and check `msg.sender == OWNER || IMorpho(BLUE).isAuthorized(address(this), msg.sender)`, inconsistent with Blue but consistent with Midnight.

adhusson (commented):
LGTM

src/periphery/BlueBuyCallback.sol
MathisGD: good point, I think that we are ok as is though

PR #1075, opened 2026-07-23
review buyer assets flashloan

QGarchery:


foundry.toml
```diff
@@ -4,7 +4,10 @@ optimizer = true
 optimizer_runs = 466
 bytecode_hash = "none"
 evm_version = "osaka"
-fs_permissions = [{ access = "read", path = "test/ticks_exact.json" }]
+fs_permissions = [
+    { access = "read", path = "test/ticks_exact.json" },
+    { access = "read", path = "out/Morpho.sol/Morpho.json" },
+]
```
MathisGD, line 10: why do you need this?

MathisGD (approved):

src/periphery/BlueBuyCallback.sol
```diff
@@ -96,9 +96,11 @@ contract BlueBuyCallback is IBlueBuyCallback {
...
 
-        return supplyAssets < liquidity
-            ? (supplyAssets < blueBalance ? supplyAssets : blueBalance)
-            : (liquidity < blueBalance ? liquidity : blueBalance);
+        return min(min(supplyAssets, liquidity), blueBalance);
+    }
+
+    function min(uint256 a, uint256 b) internal pure returns (uint256) {
+        return a < b ? a : b;
     }
```
MathisGD, line 104: could import UtilsLib

peyha (approved):

foundry.toml
QGarchery: it allows to do `forge test --mc Blue` for example. Otherwise it can't find the artifact

PR #1081, opened 2026-07-24
blue fallback rolling

MathisGD:
_fallback rolling_ is here for midnight borrowers who might forget or not be able to repay their debt before the maturity, rolling their loan to avoid the post-maturity liquidation ("auto-deleverage").

Here it rolls to a Blue variable rate market. The borrower must have a single collateral activated as Blue markets are only single collateral.

When opening their loan, the borrower configures a Blue fallback setting: the time at which the fallback starts, and the incentive paid to the account who will do it. The incentive is paid by taking a loan on Blue slightly bigger than the current loan on Midnight. The loan on Midnight is repaid at par, so it shouldn't be used too far from the maturity.

Rolling loans through this contracts is permissionless. MEV should take the opportunity when the underlying Blue markets are liquid. But when they are illiquid, lenders of the Midnight market could roll themselves from Midnight to Blue to facilitate borrowers' fallback rolling and pocket the incentive.


src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
```diff
@@ -0,0 +1,96 @@
...
+        require(UtilsLib.countBits(collateralBitmap) == 1, IncorrectActivatedCollateral());
+        uint256 collateralIndex = UtilsLib.msb(collateralBitmap);
+        require(
+            blueMarketParams.collateralToken == midnightMarket.collateralParams[collateralIndex].token,
+            InconsistentCollateralToken()
+        );
+
+        // Round in favor of the Midnight position.
+        uint256 collateralAssets = IMidnight(MIDNIGHT).collateral(midnightMarketId, user, collateralIndex)
+            .mulDivDown(assets, IMidnight(MIDNIGHT).debt(midnightMarketId, user));
```
MathisGD: technically there is a super edge attack here... not sure how to solve though. random rounding?

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
```diff
@@ -0,0 +1,94 @@
...
+        (
+            Market memory midnightMarket,
+            MarketParams memory blueMarketParams,
+            uint256 collateralIndex,
+            uint256 assets,
+            uint256 incentiveAssets,
+            address user
+        ) = abi.decode(data, (Market, MarketParams, uint256, uint256, uint256, address));
+
+        IMorpho(BLUE).borrow(blueMarketParams, assets + incentiveAssets, 0, user, address(this));
```
peyha, line 128: If the user is close to unhealthiness, liquidators will batch roll and liquidate to provoke a liquidation of the position. we might want to add a quick health check (giving sufficient margin to the borrower)

MathisGD: it means that they are very close to unhealtyness on midnight as well right? (considering they use the same LLTV and oracle). you think that it's still worth it because of the blue interest?

MathisGD: note also that if the rolling reverts, you get liquidated on midnight anyway (as it is a fallback rolling, we are close to the maturity)

peyha: yes you have to be close to liquidation but because of the RCF mechanism, the profit difference for the liquidator is significant

peyha: I don't have any idea of what would be a relevant incentive value but if it is big compared to the rate on blue then it would still be worth it for the liquidator

MathisGD: but the liquidation that you are trying to avoid is a full liquidation (with the dutch auction but still)

peyha: Ah yes I assumed that the profit/loss on midnight both with the RCF and with the dutch auction would be way lower than on Blue but that's not necessarily true

benson-budiman-cb (commented):
Hello! Looks good to me at a glance. I would like to ensure that we are careful around edge-cases around partial rolls and with dust/small amounts, since I do see some `mulDivDown`. Also, wondering if there is a possible race condition between setting a configuration as false once `start` has passed.

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
```diff
@@ -0,0 +1,94 @@
...
+        uint256 collateralAssets = IMidnight(MIDNIGHT).collateral(midnightId, user, collateralIndex)
+            .mulDivDown(assets, IMidnight(MIDNIGHT).debt(midnightId, user));
+        // Round in favor of the borrower.
+        uint256 incentiveAssets = UtilsLib.mulDivDown(assets, incentive, WAD);
+
+        bytes memory data = abi.encode(midnightMarket, blueMarketParams, collateralIndex, assets, incentiveAssets, user);
+        IMorpho(BLUE).supplyCollateral(blueMarketParams, collateralAssets, user, data);
+        if (incentiveAssets > 0) SafeTransferLib.safeTransfer(midnightMarket.loanToken, msg.sender, incentiveAssets);
+
+        emit Roll(msg.sender, user, midnightId, blueId, assets, collateralAssets, incentiveAssets);
```
MathisGD: usually we put the event above the transfers/calls/callbacks

peyha: hesitated to put the event just after `supplyCollateral` because I wasn't really seeing it as a callback (I rather see it as a call to blue eventually calling back)
will revert to the previous version

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
```diff
@@ -0,0 +1,96 @@
...
+    mapping(address user => mapping(bytes32 configId => bool)) public override isConfig;
+
+    constructor(address _midnight, address _blue) {
+        MIDNIGHT = _midnight;
+        BLUE = _blue;
+    }
+
+    /// @param start The start time of the rolling period.
+    /// @param incentive The caller incentive as a WAD-scaled percentage of the debt rolled.
+    /// @dev The LLTV of the Blue market must be greater than or equal to the LLTV of the Midnight market.
```
chrmatt: Should this be checked somewhere (maybe in roll())?

peyha: done in the [upcoming auction PR](https://github.com/morpho-org/midnight/pull/1127/changes/1fc423430284a5c0115c433ef88f00abc7f96ba3)

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
```diff
@@ -0,0 +1,111 @@
...
+    address public immutable override BLUE;
+
+    mapping(address user => mapping(bytes32 configId => bool)) public override isConfig;
+
+    constructor(address _midnight, address _blue) {
+        MIDNIGHT = _midnight;
+        BLUE = _blue;
+    }
+
+    /// @dev The LLTV of the Blue market must be greater than or equal to the LLTV of the Midnight market.
```
peyha: in the same vein, should we also require the oracles of both market's collateral to be equal ?
not sure because we could have an hardcoded exchange rate on one protocol and a liquidity oracle on another one

chrmatt: Not sure it's a good idea. This would mean if markets on Blue and Midnight all have different oracles, e.g., because the Midnights one were created later with a new oracle provider, one cannot enable the fallback?

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
```diff
@@ -0,0 +1,114 @@
...
+    function setConfig(
+        bytes32 midnightId,
+        bytes32 blueId,
+        uint64 start,
+        uint64 end,
+        uint64 incentiveAtStart,
+        uint64 incentiveAtEnd,
+        bool enabled
+    ) external override {
+        require(start < end, EndNotAfterStart());
```
peyha, line 45:
```suggestion
        require(start <= end, EndNotAfterStart());
```
does not change much, might simplify a bit

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
peyha: solved with `minRollableAssets`

MathisGD: fixed by #1141

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
peyha: ended up implementing it, simpler to reason about and a similar check is used in midnight (`block.timestamp <= offer.market.maturity`) so we can use end = `maturity`

chrmatt (approved):
Looks good to me!

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
MathisGD, line 106: reverts when start=end

peyha: ah yes start = end introduces some edge case when incentiveAtEnd != incentiveAtStart

peyha: it would be quite ugly to add a new ternary operator for `start == end` will revert

chrmatt: `start=end` doesn't make much sense, does it? We used to have `require(start < end)`.

MathisGD: then let's make it a prevented config

test/BlueFallbackRollingTest.sol
```diff
@@ -0,0 +1,1234 @@
...
+    IMorpho internal blue;
+    BlueFallbackRolling internal fallbackContract;
+    Market internal midnightMarket;
+    MarketParams internal blueMarketParams;
+    uint256 internal blueCollateralIndex;
+    uint64 internal start;
+    uint64 internal end;
+
+    function setUp() public override {
+        super.setUp();
```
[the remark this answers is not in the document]
MathisGD: @claude open a separate PR (targeting main) to fix that

src/periphery/blue-fallback-rolling/interfaces/IBlueFallbackRolling.sol
```diff
@@ -0,0 +1,74 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Copyright (c) 2026 Morpho Association
+pragma solidity >=0.5.0;
+
+import {IMorphoSupplyCollateralCallback} from "../../../../lib/morpho-blue/src/interfaces/IMorphoCallbacks.sol";
+import {MarketParams} from "../../../../lib/morpho-blue/src/interfaces/IMorpho.sol";
+import {Market} from "../../../interfaces/IMidnight.sol";
+
+interface IBlueFallbackRolling is IMorphoSupplyCollateralCallback {
```
MathisGD, line 9: not mandatory but we can disable fmt to online events and functions definition

src/periphery/README.md
```diff
@@ -21,6 +21,20 @@ The owner is included in the creation code and combined with the salt, so differ
...
+reaches the configured start timestamp. It supplies the activated collateral to Blue before, through a callback, borrowing against
+it on Blue to reward the caller, repay the debt and withdraw the collateral on Midnight.
+
+Users and accounts they authorize on Midnight may enable multiple fallback configurations per Midnight market. Users
+must authorize the contract on both Midnight and Blue. Each configuration selects a Blue market, start timestamp, caller
+incentive, and minimum rollable amount, and
+can later be disabled. The caller incentive is a percentage of the debt and is capped at 100%. The minimum rollable
+amount is the smallest debt a single roll can migrate, and is bypassed by a roll migrating the whole remaining Midnight
+debt, so a position can always be closed out. A roll requires the borrower to have exactly one activated Midnight
+collateral, matching the collateral token of the configured Blue market.
```
MathisGD: let's do one sentence per line + not wrap

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
peyha: does not work

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
```diff
@@ -0,0 +1,134 @@
...
+        // Round in favor of the Midnight position. Thus, splitting the rolls can lower the blue final LTV. This is
+        // mitigated by the min rollable debt.
+        uint256 collateralAssets =
+            IMidnight(MIDNIGHT).collateral(midnightId, user, collateralIndex).mulDivDown(assets, debtAssets);
+        // Round against the roller.
+        uint256 incentiveFactor = incentiveAtEnd >= incentiveAtStart
+            ? incentiveAtStart
+                + UtilsLib.mulDivDown(incentiveAtEnd - incentiveAtStart, block.timestamp - start, end - start)
+            : incentiveAtStart
+                - UtilsLib.mulDivUp(incentiveAtStart - incentiveAtEnd, block.timestamp - start, end - start);
```
MathisGD, line 106: shouldn't we always round down if we want to round "against the roller"?

peyha: isn't it already rounded against the roller ? we substract the muldivup

MathisGD: yes mb

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
```diff
@@ -0,0 +1,134 @@
...
+        require(
+            blueMarketParams.collateralToken == midnightMarket.collateralParams[collateralIndex].token,
+            InconsistentCollateralToken()
+        );
+
+        uint256 debtAssets = IMidnight(MIDNIGHT).debt(midnightId, user);
+        require(assets >= minRollableAssets || assets == debtAssets, RollableAssetsTooLow());
+
+        // Round in favor of the Midnight position. Thus, splitting the rolls can lower the blue final LTV. This is
+        // mitigated by the min rollable debt.
```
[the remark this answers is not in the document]
```suggestion
        // Round in favor of the Midnight position. Thus, splitting the rolls can raise the blue final LTV. This is
        // mitigated by the min rollable debt.
```

src/periphery/blue-fallback-rolling/interfaces/IBlueFallbackRolling.sol
peyha: let's do it for consistency

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
```diff
@@ -0,0 +1,134 @@
...
+        uint128 collateralBitmap = IMidnight(MIDNIGHT).collateralBitmap(midnightId, user);
+        require(UtilsLib.countBits(collateralBitmap) == 1, IncorrectActivatedCollateral());
+        uint256 collateralIndex = UtilsLib.msb(collateralBitmap);
+        require(
+            blueMarketParams.collateralToken == midnightMarket.collateralParams[collateralIndex].token,
+            InconsistentCollateralToken()
+        );
+
+        uint256 debtAssets = IMidnight(MIDNIGHT).debt(midnightId, user);
+        require(assets >= minRollableAssets || assets == debtAssets, RollableAssetsTooLow());
```
peyha:
```suggestion
        require(assets >= minRollableAssets || assets == debtAssets, RolledAssetsTooLow());
```
more accurate

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
peyha: note that it also means that we can't set a config available for a single timestamp/block, pretty sure audits will flag this but no big deal

MathisGD (commented):
LGTM

peyha (approved):

adhusson (commented):
LGTM

PR #1085, opened 2026-07-27
[blue buy callback] docs: fee recipient edge case

MathisGD:
https://cantina.xyz//code/0cbb7a34-a9ad-4b58-8d87-4c6c85fdd6cc/src/periphery/BlueBuyCallback.sol?preview=code#comment-3a609153-75a9-4935-af2e-9a967c279900

### DIFF
--- a/src/periphery/BlueBuyCallback.sol
+++ b/src/periphery/BlueBuyCallback.sol
@@ -87,6 +87,7 @@ contract BlueBuyCallback is IBlueBuyCallback {
     /// @dev This function is useful for bundles to query how much is available at the moment.
     /// @dev Other buy callbacks might not take all constraints into account to provide their bound, but this is fine,
     /// if the routing layer can take into account the other reasons.
+    /// @dev Under-estimates the real bound if the callback is the fee recipient of the blue market.
     function buyerAssetsBound(bytes32, Market memory, address, bytes memory data) external view returns (uint256) {
         MarketParams memory marketParams = abi.decode(data, (MarketParams));
 


src/periphery/BlueBuyCallback.sol
```diff
@@ -87,6 +87,7 @@ contract BlueBuyCallback is IBlueBuyCallback {
     /// @dev This function is useful for bundles to query how much is available at the moment.
     /// @dev Other buy callbacks might not take all constraints into account to provide their bound, but this is fine,
     /// if the routing layer can take into account the other reasons.
+    /// @dev Under-estimates the real bound if the callback is the fee recipient of the blue market.
```
peyha, line 90:
```suggestion
    /// @dev Under-estimates the real bound if the callback is the fee recipient of the blue market if interests have not been accrued in the block.
```
prop wdyt ?

peyha: oh wait it's not necessary, having the callback being the fee recipient of the underlying market should not happen so it's not really useful to suggest a way to handle it

peyha (approved):

src/periphery/BlueBuyCallback.sol
peyha, line 90: should we write other conditions for the bound to work ? for instance the IRM has to be the adaptivecurve

Saw-mon-and-Natalie: related to [Cantina - F1](https://cantina.xyz/code/0cbb7a34-a9ad-4b58-8d87-4c6c85fdd6cc/findings/1)'s footnote.

MathisGD: >IRM has to be the adaptivecurve 

why?

MathisGD: @Saw-mon-and-Natalie related PR #1095

Saw-mon-and-Natalie: > > IRM has to be the adaptivecurve
> why ?

@MathisGD where are the assumptions around the IRM stated?

MathisGD: I don't think that we need any assumption about the IRM other than the ones that Blue expect (non reverting etc)

peyha: indeed there is no IRM condition (thought `MorphoBalancesLib` was AdaptiveCurve specific but  it's not)

QGarchery (approved):

src/periphery/BlueBuyCallback.sol
Saw-mon-and-Natalie: My comment is about some custom implementation where the `borrowRate` and `borrowRateView` have potentially some diverging side-effects.

adhusson (commented):
LGTM (would have detaild a bit more but is ok)

PR #1088, opened 2026-07-27
[blue buy callback] docs: refine documentation for buyerAssetsBound function

MathisGD:
it is useless to talk about not taking into account all the capping factor in this callback that takes into account all the capping factors. in a future callback who might not take everything into account we will document it and it will be easier to explain and more concrete.

### DIFF
--- a/src/periphery/BlueBuyCallback.sol
+++ b/src/periphery/BlueBuyCallback.sol
@@ -86,7 +86,7 @@ contract BlueBuyCallback is IBlueBuyCallback {
     /// @dev Max buyerAssets amount that the callback can handle.
     /// @dev Bundles receive the amount to take per offer from the routing layer. But the routing layer is
     /// asynchronous/offchain, and might not be up to date on the chain's latest state. To counter this, bundles can
-    /// query this function to cap their take.
+    /// query atomically this function to cap their take.
     function buyerAssetsBound(bytes32, Market memory, address, bytes memory data) external view returns (uint256) {
         MarketParams memory marketParams = abi.decode(data, (MarketParams));
 
--- a/src/periphery/BlueBuyCallback.sol
+++ b/src/periphery/BlueBuyCallback.sol
@@ -84,7 +84,7 @@ contract BlueBuyCallback is IBlueBuyCallback {
     }
 
     /// @dev Max buyerAssets amount that the callback can handle.
-    /// @dev Bundles receive the amount to take per offer from the routing layer. But the routing layer is
+    /// @dev Takers receive the amount to take per offer from the routing layer. But the routing layer is
     /// asynchronous/offchain, and might not be up to date on the chain's latest state. To counter this, bundles can
     /// query atomically this function to cap their take.
     /// @dev Ignores some static reasons why the bound might be smaller, such as wrong loan token, wrong owner... But it is easy for the routing layer to take that into account.
--- a/src/periphery/BlueBuyCallback.sol
+++ b/src/periphery/BlueBuyCallback.sol
@@ -84,9 +84,7 @@ contract BlueBuyCallback is IBlueBuyCallback {
     }
 
     /// @dev Max buyerAssets amount that the callback can handle.
-    /// @dev This function is useful for bundles to query how much is available at the moment.
-    /// @dev Other buy callbacks might not take all constraints into account to provide their bound, but this is fine,
-    /// if the routing layer can take into account the other reasons.
+    /// @dev Bundles receive the amount to take per offer from the routing layer. But the routing layer is asynchronous/offchain, and might not be up to date on the chain's latest state. To counter this, bundles can querty this function to cap their take.
     function buyerAssetsBound(bytes32, Market memory, address, bytes memory data) external view returns (uint256) {
         MarketParams memory marketParams = abi.decode(data, (MarketParams));
 
--- a/src/periphery/BlueBuyCallback.sol
+++ b/src/periphery/BlueBuyCallback.sol
@@ -87,7 +87,8 @@ contract BlueBuyCallback is IBlueBuyCallback {
     /// @dev Takers receive the amount to take per offer from the routing layer. But the routing layer is
     /// asynchronous/offchain, and might not be up to date on the chain's latest state. To counter this, takers can
     /// query atomically this function to cap their take.
-    /// @dev Ignores some static reasons why the bound might be smaller, such as wrong loan token, wrong owner... But it is easy for the routing layer to take that into account.
+    /// @dev Ignores some static reasons why the bound might be smaller, such as wrong loan token, wrong owner... But it
+    /// is easy for the routing layer to take that into account.
     /// @dev Reverts if data is not well formed.
     /// @dev Under-estimates the real bound if the callback is the fee recipient of the blue market.
     function buyerAssetsBound(bytes32, Market memory, address, bytes memory data) external view returns (uint256) {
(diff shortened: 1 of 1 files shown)


src/periphery/BlueBuyCallback.sol
```diff
@@ -84,9 +84,7 @@ contract BlueBuyCallback is IBlueBuyCallback {
     }
 
     /// @dev Max buyerAssets amount that the callback can handle.
-    /// @dev This function is useful for bundles to query how much is available at the moment.
-    /// @dev Other buy callbacks might not take all constraints into account to provide their bound, but this is fine,
-    /// if the routing layer can take into account the other reasons.
+    /// @dev Bundles receive the amount to take per offer from the routing layer. But the routing layer is asynchronous/offchain, and might not be up to date on the chain's latest state. To counter this, bundles can querty this function to cap their take.
```
Saw-mon-and-Natalie:
```suggestion
    /// @dev Bundles receive the amount to take per offer from the routing layer. But the routing layer is asynchronous/offchain, and might not be up to date on the chain's latest state. To counter this, bundles can query this function to cap their take.
```

src/periphery/BlueBuyCallback.sol
```diff
@@ -84,9 +84,7 @@ contract BlueBuyCallback is IBlueBuyCallback {
     }
 
     /// @dev Max buyerAssets amount that the callback can handle.
-    /// @dev This function is useful for bundles to query how much is available at the moment.
-    /// @dev Other buy callbacks might not take all constraints into account to provide their bound, but this is fine,
-    /// if the routing layer can take into account the other reasons.
+    /// @dev Bundles receive the amount to take per offer from the routing layer. But the routing layer is asynchronous/offchain, and might not be up to date on the chain's latest state. To counter this, bundles can query this function to cap their take.
```
dmitriia: Also the bound can be manipulated within a block (e.g. via a large Blue borrow), so bundles need to query this atomically with execution.

MathisGD: you think that we should mention the malicious case basically?

MathisGD: because the comment already captures the fact that it is async so bundles should query this

dmitriia: I would say something like "bundles can query this function atomically to cap their take"

MathisGD: ok so basically add "atomically"?

dmitriia: Yes, to outline that the only correct reading is on execution

MathisGD:
```suggestion
    /// @dev Bundles receive the amount to take per offer from the routing layer. But the routing layer is
    /// asynchronous/offchain, and might not be up to date on the chain's latest state. To counter this, bundles can
    /// query this function to cap their take.
```

src/periphery/BlueBuyCallback.sol
```diff
@@ -84,9 +84,9 @@ contract BlueBuyCallback is IBlueBuyCallback {
     }
 
     /// @dev Max buyerAssets amount that the callback can handle.
-    /// @dev This function is useful for bundles to query how much is available at the moment.
-    /// @dev Other buy callbacks might not take all constraints into account to provide their bound, but this is fine,
-    /// if the routing layer can take into account the other reasons.
+    /// @dev Bundles receive the amount to take per offer from the routing layer. But the routing layer is
+    /// asynchronous/offchain, and might not be up to date on the chain's latest state. To counter this, bundles can
+    /// query this function to cap their take.
```
MathisGD:
```suggestion
    /// @dev Bundles receive the amount to take per offer from the routing layer. But the routing layer is
    /// asynchronous/offchain, and might not be up to date on the chain's latest state. To counter this, bundles can
    /// query atomically this function to cap their take.
```

src/periphery/BlueBuyCallback.sol
```diff
@@ -84,9 +84,9 @@ contract BlueBuyCallback is IBlueBuyCallback {
     }
 
     /// @dev Max buyerAssets amount that the callback can handle.
-    /// @dev This function is useful for bundles to query how much is available at the moment.
-    /// @dev Other buy callbacks might not take all constraints into account to provide their bound, but this is fine,
-    /// if the routing layer can take into account the other reasons.
+    /// @dev Bundles receive the amount to take per offer from the routing layer. But the routing layer is
+    /// asynchronous/offchain, and might not be up to date on the chain's latest state. To counter this, bundles can
+    /// query atomically this function to cap their take.
     function buyerAssetsBound(bytes32, Market memory, address, bytes memory data) external view returns (uint256) {
```
MathisGD, line 94: fixes https://cantina.xyz/code/0cbb7a34-a9ad-4b58-8d87-4c6c85fdd6cc/findings/13 and https://cantina.xyz/code/0cbb7a34-a9ad-4b58-8d87-4c6c85fdd6cc/findings/2
```suggestion
    /// @dev Ignores some static reasons why the bound might be smaller, such as wrong loan token, wrong owner... But it is easy for the routing layer to take that into account.
    /// @dev Reverts if data is not well formed.
    function buyerAssetsBound(bytes32, Market memory, address, bytes memory data) external view returns (uint256) {
```

src/periphery/BlueBuyCallback.sol
```diff
@@ -84,9 +84,9 @@ contract BlueBuyCallback is IBlueBuyCallback {
     }
 
     /// @dev Max buyerAssets amount that the callback can handle.
-    /// @dev This function is useful for bundles to query how much is available at the moment.
-    /// @dev Other buy callbacks might not take all constraints into account to provide their bound, but this is fine,
-    /// if the routing layer can take into account the other reasons.
+    /// @dev Bundles receive the amount to take per offer from the routing layer. But the routing layer is
```
QGarchery:
```suggestion
    /// @dev Takers receive the amount to take per offer from the routing layer. But the routing layer is
```
this is agnostic of the bundles contract, that is not define nor linked before

peyha (approved):

src/periphery/BlueBuyCallback.sol
```diff
@@ -84,9 +84,11 @@ contract BlueBuyCallback is IBlueBuyCallback {
     }
 
     /// @dev Max buyerAssets amount that the callback can handle.
-    /// @dev This function is useful for bundles to query how much is available at the moment.
-    /// @dev Other buy callbacks might not take all constraints into account to provide their bound, but this is fine,
-    /// if the routing layer can take into account the other reasons.
+    /// @dev Takers receive the amount to take per offer from the routing layer. But the routing layer is
+    /// asynchronous/offchain, and might not be up to date on the chain's latest state. To counter this, bundles can
```
QGarchery:
```suggestion
    /// asynchronous/offchain, and might not be up to date on the chain's latest state. To counter this, takers can
```

src/periphery/BlueBuyCallback.sol
```diff
@@ -84,9 +84,11 @@ contract BlueBuyCallback is IBlueBuyCallback {
     }
 
     /// @dev Max buyerAssets amount that the callback can handle.
-    /// @dev This function is useful for bundles to query how much is available at the moment.
-    /// @dev Other buy callbacks might not take all constraints into account to provide their bound, but this is fine,
-    /// if the routing layer can take into account the other reasons.
+    /// @dev Takers receive the amount to take per offer from the routing layer. But the routing layer is
+    /// asynchronous/offchain, and might not be up to date on the chain's latest state. To counter this, takers can
+    /// query atomically this function to cap their take.
+    /// @dev Ignores some static reasons why the bound might be smaller, such as wrong loan token, wrong owner... But it is easy for the routing layer to take that into account.
```
MathisGD:
```suggestion
    /// @dev Ignores some static reasons why the bound might be smaller, such as wrong loan token, wrong owner... But it
    /// is easy for the routing layer to take that into account.
```

QGarchery (approved):

PR #1104, opened 2026-07-31
[fallback rolling] refactor: config struct

peyha:
looks cleaner and would be better to handle some kind of auction

### DIFF
--- a/src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
+++ b/src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
@@ -9,7 +9,7 @@ import {WAD} from "../../libraries/ConstantsLib.sol";
 import {IdLib} from "../../libraries/IdLib.sol";
 import {SafeTransferLib} from "../../libraries/SafeTransferLib.sol";
 import {UtilsLib} from "../../libraries/UtilsLib.sol";
-import {IBlueFallbackRolling} from "./IBlueFallbackRolling.sol";
+import {IBlueFallbackRolling, Config} from "./IBlueFallbackRolling.sol";
 import {SafeApproveLib} from "../libraries/SafeApproveLib.sol";
 
 /// @dev Users must authorize this contract on both Midnight and Blue before their debt can be rolled.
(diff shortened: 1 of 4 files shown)


src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
```diff
@@ -27,32 +27,30 @@ contract BlueFallbackRolling is IBlueFallbackRolling {
         BLUE = _blue;
     }
 
-    /// @param incentive The caller incentive as a WAD-scaled percentage of the debt rolled.
-    function setConfig(bytes32 midnightId, bytes32 blueId, uint64 start, uint64 incentive, bool enabled)
-        external
-        override
-    {
-        require(incentive <= WAD, IncentiveTooHigh());
+    /// @param config.incentive The caller incentive as a WAD-scaled percentage of the debt rolled.
```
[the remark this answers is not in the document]
```suggestion
```

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
```diff
@@ -27,32 +27,30 @@ contract BlueFallbackRolling is IBlueFallbackRolling {
...
+        emit SetConfig(msg.sender, config.midnightId, config.blueId, config.start, config.incentive, enabled);
     }
 
     function roll(
         Market memory midnightMarket,
         MarketParams memory blueMarketParams,
         address user,
-        uint64 start,
-        uint64 incentive,
+        Config memory config,
```
MathisGD, line 42: this is actually more complex no? you have to check that the ids match

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
MathisGD, line 1: >would be better to handle some kind of auction

curious about this

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
peyha: yes you have two extra check (not sure if removing the check and fetching the markets from their respective contract would be better)

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
peyha: not yet sure but was thinking about having `startIncentive` and `endIncentive` that would grow linearly (and that could be equal to have a constant incentive) just like on preliquidations to implement some kind of dutch auction

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
MathisGD: that would be very expensive! I don't think that's a good idea

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
MathisGD: yes I get that but how is the struct helping?

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
peyha: we could also remove `midnightId` and `blueId` from `Config` to prevent this check (but then the struct becomes useless except if we add more auction parameters)

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
peyha: would make the auction parameters more explicit but not much value added in the end

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
```diff
@@ -27,32 +27,29 @@ contract BlueFallbackRolling is IBlueFallbackRolling {
         BLUE = _blue;
     }
 
-    /// @param incentive The caller incentive as a WAD-scaled percentage of the debt rolled.
-    function setConfig(bytes32 midnightId, bytes32 blueId, uint64 start, uint64 incentive, bool enabled)
-        external
-        override
-    {
-        require(incentive <= WAD, IncentiveTooHigh());
+    function setConfig(Config memory config, bool enabled) external override {
```
MathisGD: here it's great but I would rollback in roll

peyha: wdyt now ? kept midnight and blue id out of config and changed `start` for `rollWindow`. Config can now be re-used across markets

peyha: we could even go further and allow only one config per user. Users would then just have to enableRolling on a given pair (midnight id, blue id) and then keeper would be able to roll on this specific user config

MathisGD (commented):
not fan tbh..

I found the previous version simpler and not sure about the benefit of the approach here

peyha (commented):
benefit would be mostly operational, users would have an easier time using the same config everywhere but it's not such an issue

peyha:
not convinced in the end

PR #1114, opened 2026-08-03
[blue buy callback] document skim

MathisGD:


src/periphery/blue-buy-callback/BlueBuyCallback.sol
```diff
@@ -64,6 +64,7 @@ contract BlueBuyCallback is IBlueBuyCallback {
         }
     }
 
+    /// @dev Useful to handle offchain rewards that the callback earned through its Blue positions.
```
peyha:
```suggestion
    /// @dev Useful to handle rewards that the callback earned through its Blue positions.
```
not sure offchain is relevant here (it's a bit ambiguous because rewards are sent onchain)

StErMi (approved):

peyha (approved):

adhusson (commented):
LGTM

PR #1115, opened 2026-08-03
[fallback rolling] max LTV

peyha:
Similarly as for Bundles, implement a max LTV. Rationale being

- the incentive is taken as additional debt for the borrower so the rolling deteriorates the LTV of the position
- liquidation is much harsher on blue than on midnight so a borrower close to insolvency would prefer staying on midnight instead of rolling

### DIFF
--- a/src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
+++ b/src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
@@ -58,7 +58,7 @@ contract BlueFallbackRolling is IBlueFallbackRolling {
         uint256 maxLtv,
         uint256 assets
     ) external override {
-        require(maxLtv < blueMarketParams.lltv, InvalidMaxLtv());
+        require(maxLtv <= blueMarketParams.lltv, InvalidMaxLtv());
         bytes32 midnightId = IdLib.toId(midnightMarket);
         bytes32 blueId = Id.unwrap(blueMarketParams.id());
         require(isConfig[user][keccak256(abi.encode(midnightId, blueId, start, incentive, maxLtv))], NotConfigured());
--- a/src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
+++ b/src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
@@ -52,6 +52,7 @@ contract BlueFallbackRolling is IBlueFallbackRolling {
         uint256 maxLtv,
         uint256 assets
     ) external override {
+        require(maxLtv < blueMarketParams.lltv, InvalidMaxLtv());
         bytes32 midnightId = IdLib.toId(midnightMarket);
         bytes32 blueId = Id.unwrap(blueMarketParams.id());
         require(isConfig[user][keccak256(abi.encode(midnightId, blueId, start, incentive, maxLtv))], NotConfigured());
@@ -100,7 +101,6 @@ contract BlueFallbackRolling is IBlueFallbackRolling {
     }
 
     function requireMaxLtv(MarketParams memory marketParams, address sender, uint256 maxLtv) internal view {
-        if (maxLtv >= marketParams.lltv) return;
         Position memory position = IMorpho(BLUE).position(marketParams.id(), sender);
         if (position.borrowShares == 0) return;
         BlueMarket memory market = IMorpho(BLUE).market(marketParams.id());
--- a/src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
+++ b/src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
@@ -2,10 +2,12 @@
 // Copyright (c) 2026 Morpho Association
 pragma solidity 0.8.34;
 
-import {IMorpho, Id, MarketParams} from "../../../lib/morpho-blue/src/interfaces/IMorpho.sol";
+import {IMorpho, Id, MarketParams, Market as BlueMarket, Position} from "../../../lib/morpho-blue/src/interfaces/IMorpho.sol";
 import {MarketParamsLib} from "../../../lib/morpho-blue/src/libraries/MarketParamsLib.sol";
+import {SharesMathLib} from "../../../lib/morpho-blue/src/libraries/SharesMathLib.sol";
 import {IMidnight, Market} from "../../interfaces/IMidnight.sol";
-import {WAD} from "../../libraries/ConstantsLib.sol";
+import {IOracle} from "../../interfaces/IOracle.sol";
+import {WAD, ORACLE_PRICE_SCALE} from "../../libraries/ConstantsLib.sol";
 import {IdLib} from "../../libraries/IdLib.sol";
 import {SafeTransferLib} from "../../libraries/SafeTransferLib.sol";
 import {UtilsLib} from "../../libraries/UtilsLib.sol";
@@ -15,7 +17,9 @@ import {SafeApproveLib} from "../libraries/SafeApproveLib.sol";
 /// @dev Users must authorize this contract on both Midnight and Blue before their debt can be rolled.
 contract BlueFallbackRolling is IBlueFallbackRolling {
     using MarketParamsLib for MarketParams;
+    using SharesMathLib for uint256;
     using UtilsLib for uint128;
+    using UtilsLib for uint256;
 
     address public immutable override MIDNIGHT;
     address public immutable override BLUE;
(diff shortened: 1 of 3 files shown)


src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
MathisGD, line 1: wrong PR title right?

MathisGD, line 1: If we say that one must use the fallback rolling only with a Blue market that has an LLTV equal or lower than the Midnight's LLTV (reasonable constrain imo), it means that they health is only improved by the rolling (minus the rolling incentive, but this is predictable and set by them). what is a situation where you would rather prevent the rolling (with `requireMaxLTV`) and be liquidated on Midnight post-maturity? because the LIF will be smaller on Midnight? the rolling incentive might change things as well (if you get rolled then liquidated on Blue, you have a implied LIF higher because of the incentive).

peyha: Was expecting the Blue market LLTV to be always equal to Midnight's LLTV thus the health would be (predictably) slightly worsen by the rolling and would continue worsening through time because of the interest accrual. Thought that we could have a LTV margin that would include these two elements (interest at the expected rate over some short period such as a month) and eventually the volatility of the collateral.

MathisGD: the question we should answer is: is there a setup in which you prefer to be liquidated on midnight's post maturity liquidation instead of being rolled on Blue? Maybe when 1) your LTV is so high that you will likely be liquidated soon on Blue, so you would rather be liquidated at a soft LIF on Midnight or 2) when the rolling incentive makes you close to be liquidated on Blue so same you prefer a soft liquidation?

peyha: I would expect 2) to be way less likely than 1) because the incentive should be computed such that it's minimal (just enough to cover the gas cost)

MathisGD: about 2): interestingly, the rolling incentive should be in the order of magnitude of the midnight post-maturity effective LIF (assuming gas to liquidate ~= gas to roll on blue). Meaning that they cancel out => you want fallback rolling to keep your position open (in order to not pay tax for example), not to avoid paying the midnight LIF.

then it kind of invalidates 1), no? you would rather try it on blue, hope that you will not get liquidated, rather than staying on midnight and get liquidated?

peyha: > in the order of magnitude of the midnight post-maturity effective LIF

I don't think that it would be the case for large positions because the post-maturity effecitve LIF would take the swapping cost into account (which is usually a few bps in pool fees and possibly a few extra bps in price impact) while rolling is just the gas cost (which would be the same for all position size).
That's one of the reason why I think having an auction could be beneficial here, the incentive is currently in proportion of the debt but the gas cost isn't (and can be very volatile) so an auction could capture the relevant incentive more accurately

>no? you would rather try it on blue, hope that you will not get liquidated

That's an interesting point and I'm honestly not sure how borrowers would behave. Suppose my LTV on midnight is quite high and I'm close to maturity, there are three possible paths
1) I don't roll and I get liquidated post-maturity (small loss, taxable event trigger, will happen surely)
2) I roll and I get liquidated on blue (larger loss, taxable event trigger, more likely to happen if my LTV is very close to LLTV)
3) I roll and I don't get liquidated on blue
I would guess I absolutely don't want 2) and 1) but if I know I am likely to have 2) I would prefer 1) and the max LTV could be set to ensure that

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
```diff
@@ -28,15 +38,15 @@ contract BlueFallbackRolling is IBlueFallbackRolling {
     }
 
     /// @param incentive The caller incentive as a WAD-scaled percentage of the debt rolled.
```
chrmatt, line 41: What about the other parameters?

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
chrmatt, line 77: Why not round in favor of Blue? Liquidations are worse on Blue, so it seems better to have more collateral there to avoid liquidation.

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
```diff
@@ -45,11 +55,13 @@ contract BlueFallbackRolling is IBlueFallbackRolling {
         address user,
         uint64 start,
         uint64 incentive,
+        uint256 maxLtv,
         uint256 assets
     ) external override {
+        require(maxLtv < blueMarketParams.lltv, InvalidMaxLtv());
```
chrmatt: Why require strictly less? If I want to roll no matter what, it seems sensible to set them equal (or even `maxLtv = 1`?).

test/BlueFallbackRollingTest.sol
```diff
@@ -178,45 +185,75 @@ contract BlueFallbackRollingTest is BaseTest {
...
-            toId(midnightMarket), Id.unwrap(blueMarketParams.id()), otherStart, MAX_INCENTIVE, true
+            toId(midnightMarket), Id.unwrap(blueMarketParams.id()), otherStart, MAX_INCENTIVE, MAX_LTV, true
+        );
+
+        assertTrue(fallbackContract.isConfig(borrower, configId(start, INCENTIVE, MAX_LTV)));
+        assertTrue(fallbackContract.isConfig(borrower, configId(otherStart, MAX_INCENTIVE, MAX_LTV)));
+    }
+
+    function testRollRevertsWhenMaxLtvExceeded() public {
+        uint256 tightMaxLtv = BLUE_LLTV / 2;
```
chrmatt, line 233: Not sure I understand the logic here. This might not always be tight if other numbers change? Wouldn't it be better to retrieve the actual LTV and then pick something tight based on the actual numbers?

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
```diff
@@ -28,15 +38,15 @@ contract BlueFallbackRolling is IBlueFallbackRolling {
     }
 
     /// @param incentive The caller incentive as a WAD-scaled percentage of the debt rolled.
-    function setConfig(bytes32 midnightId, bytes32 blueId, uint64 start, uint64 incentive, bool enabled)
+    function setConfig(bytes32 midnightId, bytes32 blueId, uint64 start, uint64 incentive, uint256 maxLtv, bool enabled)
```
chrmatt, line 43: I don't think I understand how this works. Can there be multiple configs be valid at once? If so, can the one rolling a position choose which one to use? So if I want to set a stricter maxLtv, can the roller still use the old value unless I disable it (how?)? It would be more intuitive for me to just have a single config at every point in time.

chrmatt (commented):
Left some comments. Also, README probably should be updated?

How relevant is this PR after our discussion this morning?

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
peyha: mistake here, should be less than equal. It would allow to have `maxLtv = LLTV` which would deactivate the `maxLtv` feature

test/BlueFallbackRollingTest.sol
peyha: the position is collateralized very close to LLTV and then a smaller tightMaxLtv is used to raise the `MaxLtvExceeded`. It's just a small test for this error. We could also use actual numbers yes

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
peyha: In the current setting yes nothing prevents the user from setting multiple configs. The roller can use any enabled config but the user can also deactivate a config with setConfig setting `enabled=False`

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
MathisGD: https://morpholabs.slack.com/archives/C02N7CZ088N/p1785835573039849?thread_ts=1785834049.461249&cid=C02N7CZ088N

peyha:
Not planned

PR #1127, opened 2026-08-05
[rolling fallback] incentive dutch auction

peyha:


src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
```diff
@@ -28,45 +28,80 @@ contract BlueFallbackRolling is IBlueFallbackRolling {
...
         require(block.timestamp >= start, NotStarted());
+        require(block.timestamp <= end, Ended());
         uint128 collateralBitmap = IMidnight(MIDNIGHT).collateralBitmap(midnightId, user);
         require(UtilsLib.countBits(collateralBitmap) == 1, IncorrectActivatedCollateral());
         uint256 collateralIndex = UtilsLib.msb(collateralBitmap);
         require(
             blueMarketParams.collateralToken == midnightMarket.collateralParams[collateralIndex].token,
             InconsistentCollateralToken()
         );
+        require(midnightMarket.collateralParams[collateralIndex].lltv <= blueMarketParams.lltv, BlueLltvTooLow());
```
peyha, line 76: these checks (as well as the loan token check) are not perfect because the users could set non-valid configs

peyha: potential solution, check these in `setConfig` and use a midnight and blue market instead of their respective ids (it would be more expensive in calldata though)

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
```diff
@@ -28,45 +28,80 @@ contract BlueFallbackRolling is IBlueFallbackRolling {
...
+        bytes32 midnightId,
+        bytes32 blueId,
+        uint64 start,
+        uint64 end,
+        uint64 incentiveAtStart,
+        uint64 incentiveAtEnd,
+        bool enabled
+    ) external override {
+        require(start <= end, EndBeforeStart());
+        require(incentiveAtStart <= incentiveAtEnd, IncentiveNotIncreasing());
```
peyha: is this check really necessary ? would a decreasing incentive make sense ? my gut feeling is that it does not

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
```diff
@@ -28,45 +28,80 @@ contract BlueFallbackRolling is IBlueFallbackRolling {
...
+            enabled;
 
-        emit SetConfig(msg.sender, midnightId, blueId, start, incentive, enabled);
+        emit SetConfig(msg.sender, midnightId, blueId, start, end, incentiveAtStart, incentiveAtEnd, enabled);
+    }
+
+    /// @notice The caller incentive at the current timestamp, as a WAD-scaled percentage of the debt rolled.
+    /// @dev The incentive is auctioned off: it grows linearly from `incentiveAtStart` at `start` to `incentiveAtEnd` at
+    /// `end`, and stays at `incentiveAtEnd` afterwards.
+    function incentive(uint64 start, uint64 end, uint64 incentiveAtStart, uint64 incentiveAtEnd)
```
MathisGD: except is there is a good specific reason to not do it, I would inline

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
```diff
@@ -28,45 +28,80 @@ contract BlueFallbackRolling is IBlueFallbackRolling {
...
         require(
             blueMarketParams.collateralToken == midnightMarket.collateralParams[collateralIndex].token,
             InconsistentCollateralToken()
         );
+        require(midnightMarket.collateralParams[collateralIndex].lltv <= blueMarketParams.lltv, BlueLltvTooLow());
 
         // Round in favor of the Midnight position.
         uint256 collateralAssets = IMidnight(MIDNIGHT).collateral(midnightId, user, collateralIndex)
             .mulDivDown(assets, IMidnight(MIDNIGHT).debt(midnightId, user));
         // Round in favor of the borrower.
```
MathisGD: I'm realizing that this is not really well said. Less incentive might not always be good for the borrower. In doubt I would remove since this is not very important (unlike the one above). same in `incentive()` btw
```suggestion
```

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
MathisGD: but at the same time nothing bad happens. we can keep or remove for me, not a big deal

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
```diff
@@ -28,45 +28,80 @@ contract BlueFallbackRolling is IBlueFallbackRolling {
...
+    /// @dev The incentive is auctioned off: it grows linearly from `incentiveAtStart` at `start` to `incentiveAtEnd` at
+    /// `end`, and stays at `incentiveAtEnd` afterwards.
+    function incentive(uint64 start, uint64 end, uint64 incentiveAtStart, uint64 incentiveAtEnd)
+        public
+        view
+        override
+        returns (uint256)
+    {
+        if (block.timestamp >= end) return incentiveAtEnd;
+        if (block.timestamp <= start) return incentiveAtStart;
```
MathisGD: useless since there is this right?

```solidity
        require(block.timestamp >= start, NotStarted());
        require(block.timestamp <= end, Ended());
```

btw good reason to inline (not have to define incentive out of the start and end)

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
```diff
@@ -28,45 +28,80 @@ contract BlueFallbackRolling is IBlueFallbackRolling {
...
+        // Round in favor of the borrower.
```
```suggestion
```

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
MathisGD: it's not too bad that one can set an invalid config, but IMO we should do either all checks in setConfig (meaning you can't enable an invalid config) or all checks in roll

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
```diff
@@ -28,45 +28,64 @@ contract BlueFallbackRolling is IBlueFallbackRolling {
...
         uint128 collateralBitmap = IMidnight(MIDNIGHT).collateralBitmap(midnightId, user);
         require(UtilsLib.countBits(collateralBitmap) == 1, IncorrectActivatedCollateral());
         uint256 collateralIndex = UtilsLib.msb(collateralBitmap);
         require(
             blueMarketParams.collateralToken == midnightMarket.collateralParams[collateralIndex].token,
             InconsistentCollateralToken()
         );
+        require(midnightMarket.collateralParams[collateralIndex].lltv <= blueMarketParams.lltv, BlueLltvTooLow());
 
         // Round in favor of the Midnight position.
```
peyha, line 78: the rounding direction has been questioned by @chrmatt and I currently don't have a compelling reason for choosing this direction

MathisGD: yes, see [this](https://github.com/morpho-org/midnight/pull/1081#discussion_r3646318772)

MathisGD: let's not tackle that in this PR

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
```diff
@@ -28,45 +28,64 @@ contract BlueFallbackRolling is IBlueFallbackRolling {
     }
 
     /// @param start The start time of the rolling period.
-    /// @param incentive The caller incentive as a WAD-scaled percentage of the debt rolled.
+    /// @param end The end time of the rolling period.
+    /// @param incentiveAtStart The caller incentive at `start`, as a WAD-scaled percentage of the debt rolled.
+    /// @param incentiveAtEnd The caller incentive at `end`, as a WAD-scaled percentage of the debt rolled. The
```
MathisGD: clear enough we don't need these

```suggestion
```

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
chrmatt: I'd remove this check. I don't see why decreasing incentive would make sense, but if someone wants to do it, why prevent it? It's also extra gas cost to check.

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
chrmatt: Imho it's better to do all check in `setConfig` to prevent a user from setting something invalid by accident, feeling safe, and then getting liquidated because rolling was not possible.

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
```diff
@@ -28,45 +28,64 @@ contract BlueFallbackRolling is IBlueFallbackRolling {
...
             InconsistentCollateralToken()
         );
+        require(midnightMarket.collateralParams[collateralIndex].lltv <= blueMarketParams.lltv, BlueLltvTooLow());
 
         // Round in favor of the Midnight position.
         uint256 collateralAssets = IMidnight(MIDNIGHT).collateral(midnightId, user, collateralIndex)
             .mulDivDown(assets, IMidnight(MIDNIGHT).debt(midnightId, user));
-        // Round in favor of the borrower.
-        uint256 incentiveAssets = UtilsLib.mulDivDown(assets, incentive, WAD);
+        // The incentive grows linearly from `incentiveAtStart` at `start` to `incentiveAtEnd` at `end`.
```
chrmatt: Rounding in favor of borrower comment is gone. Maybe keep it? Or was it removed on purpose?

chrmatt (commented):
Seems there is no test case to check that having a LLTV on Blue that is lower than the one on Midnight reverts.

Also added some comments. Otherwise looks good.

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
MathisGD: it was removed on purpose because this rounding is not so important + what is "in favor of the borrower" exactly? one could say that with less incentive they might get liquidated (not very realistic that the rounding causes this, but still worth removing the comment imo)?

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
peyha: removed

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
```diff
@@ -28,45 +28,61 @@ contract BlueFallbackRolling is IBlueFallbackRolling {
...
         require(block.timestamp >= start, NotStarted());
+        require(block.timestamp <= end, Ended());
         uint128 collateralBitmap = IMidnight(MIDNIGHT).collateralBitmap(midnightId, user);
         require(UtilsLib.countBits(collateralBitmap) == 1, IncorrectActivatedCollateral());
         uint256 collateralIndex = UtilsLib.msb(collateralBitmap);
         require(
             blueMarketParams.collateralToken == midnightMarket.collateralParams[collateralIndex].token,
             InconsistentCollateralToken()
         );
+        require(midnightMarket.collateralParams[collateralIndex].lltv <= blueMarketParams.lltv, BlueLltvTooLow());
```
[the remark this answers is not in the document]
peyha: we said we don't care about the health of the user

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
peyha: there are currently 4 checks on a given config: start <= end, incentives <= WAD, loan token match and collateral token match. We can't really check the collateral tokens because the config might be set before the user supply any collateral into midnight. It's also more expensive in gas to check the loan token in `setConfig` because in the current version we only use the market Ids. So I slightly favor the current version

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
peyha: wrote that the rounding was against the keeper

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
```diff
@@ -28,45 +28,62 @@ contract BlueFallbackRolling is IBlueFallbackRolling {
...
 
         // Round in favor of the Midnight position.
         uint256 collateralAssets = IMidnight(MIDNIGHT).collateral(midnightId, user, collateralIndex)
             .mulDivDown(assets, IMidnight(MIDNIGHT).debt(midnightId, user));
-        // Round in favor of the borrower.
-        uint256 incentiveAssets = UtilsLib.mulDivDown(assets, incentive, WAD);
+        // Round against the keeper.
```
```suggestion
        // Round against the roller.
```

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
```diff
@@ -28,45 +28,62 @@ contract BlueFallbackRolling is IBlueFallbackRolling {
     }
 
     /// @param start The start time of the rolling period.
-    /// @param incentive The caller incentive as a WAD-scaled percentage of the debt rolled.
+    /// @param end The end time of the rolling period.
```
MathisGD: we can remove these natspec, clear enough
```suggestion
```

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
```diff
@@ -28,45 +28,62 @@ contract BlueFallbackRolling is IBlueFallbackRolling {
     }
 
     /// @param start The start time of the rolling period.
```
```suggestion
```

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
```suggestion
```

MathisGD (approved):

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
peyha: hesitated between these 2 terms. Roller looks a bit ambiguous, it could refer to both the borrower getting rolled the keeper making the transaction

peyha: thinking a bit more about it it's probably ok, it's similar to the liquidator naming

PR #1128, opened 2026-08-07
Blue fallback rolling auction negative

peyha:


peyha:
Not planned after [this ](https://morpholabs.slack.com/archives/C02N7CZ088N/p1786094484001899?thread_ts=1786093278.520899&cid=C02N7CZ088N)

PR #1141, opened 2026-08-12
[rolling fallback] min rollable debt

peyha:


src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
```diff
@@ -75,9 +81,12 @@ contract BlueFallbackRolling is IBlueFallbackRolling {
         );
         require(midnightMarket.collateralParams[collateralIndex].lltv <= blueMarketParams.lltv, BlueLltvTooLow());
 
+        uint256 debtAssets = IMidnight(MIDNIGHT).debt(midnightId, user);
+        require(assets >= minRollableAssets || assets == debtAssets, RollableAssetsTooLow());
```
peyha, line 83: can be this or 

```suggestion
        require(assets >= minRollableAssets || debtAssets <= minRollableAssets, RollableAssetsTooLow());
```
to allow partial rolls when the debt is lower than minRollableAssets

test/BlueFallbackRollingTest.sol
```diff
@@ -19,6 +19,7 @@ contract BlueFallbackRollingTest is BaseTest {
     uint64 internal constant INCENTIVE_AT_END = 0.001e18;
     uint64 internal constant MAX_INCENTIVE = 1e18;
     uint256 internal constant DEBT = 10_000e18;
+    uint128 internal constant MIN_ROLLABLE_ASSETS = 2_500e18;
```
peyha, line 22: might add some fuzzing

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
chrmatt: Why should we allow partial rolls in that case? Seems cleaner to only allow moving all of it.

chrmatt (commented):
We also discussed requiring to leave less remaining than the minimum. Not required if we allow full roll always, but would fix issue that it's not profitable to roll s small remainder. What was the conclusion on this?

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
peyha: not 100% convinced but I prefer the partial roll version because
- once debt is under min rollable assets it should be very unlikely that the dust rounding attack can be done profitably
- it's consistent with the design of rcf threshold, once we're below the threshold, it's possible to liquidate any partial amount
- it slightly mitigates a few front running attack where an attacker could roll 1 wei just before a full rolling transaction (though this could be fixed by using a smart contract to compute the remaining debt on the fly)

src/periphery/README.md
chrmatt: I think this is not accurate. Isn't the borrowing and repay is done through a callback after supplying collateral? Also it's technically not the "selected collateral", but the assets are provided as input and the collateral is computed from it.

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
chrmatt: The new rule seems less intuitive to me. If I set a minimum amount as a user to avoid small rolls, it's unclear why it's suddenly allowed to move even less than what is left. The rounding attack could still annoy the victim if it allows the attacker to liquidate (at a loss).

The front running attack appears interesting! But I don't see how this mitigates it. It should revert anyway when you try to repay more than the full amount?

MathisGD: imo the first version assets == debt was better

peyha: > this mitigates it

this mitigates it because I should be able to use `assets / 2` (unless the frontrunner has repaid more than half the remaining debt but that's still not quite satisfying

peyha: i'll revert back to the previous version then

chrmatt: > > this mitigates it
> 
> this mitigates it because I should be able to use `assets / 2` (unless the frontrunner has repaid more than half the remaining debt but that's still not quite satisfying

But then I can never roll the full amount?

Another approach would be to clamp assets to debt. Then it would be fine to supply a too large amount of assets and one can always safely roll the full amount.

peyha: I'm not sure clamping assets would be nice UX, especially if the roller is also a lender

src/periphery/README.md
peyha: rephrased, tell me if it's clearer now

chrmatt: The supply itself is in a direct call, not the callback I think?

This is what Claude suggests to write:
```It supplies collateral to Blue on the borrower's behalf and, from within Blue's supply callback, borrows against that collateral, repays the Midnight debt, and withdraws the matching collateral from Midnight to settle the supply. The caller is rewarded out of the additional Blue borrow. Rolls may migrate all debt and collateral, or a proportional partial amount.```

File: src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol (hunk 1 above)
Comment (peyha) [reply src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol:83]: closing this issue as I think the current version is fine

File: src/periphery/README.md
Comment (peyha) [reply src/periphery/README.md]: indeed, the flow is not explained well enough, should be good now

File: src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol [hunk 3]
```diff
@@ -75,9 +82,12 @@ contract BlueFallbackRolling is IBlueFallbackRolling {
...
         require(midnightMarket.collateralParams[collateralIndex].lltv <= blueMarketParams.lltv, BlueLltvTooLow());
 
+        uint256 debtAssets = IMidnight(MIDNIGHT).debt(midnightId, user);
+        require(assets >= minRollableAssets || debtAssets <= minRollableAssets, RollableAssetsTooLow());
+
         // Round in favor of the Midnight position.
-        uint256 collateralAssets = IMidnight(MIDNIGHT).collateral(midnightId, user, collateralIndex)
-            .mulDivDown(assets, IMidnight(MIDNIGHT).debt(midnightId, user));
+        uint256 collateralAssets =
+            IMidnight(MIDNIGHT).collateral(midnightId, user, collateralIndex).mulDivDown(assets, debtAssets);
```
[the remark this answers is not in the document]
peyha: that's valid, I don't think it's worth a comment though

test/BlueFallbackRollingTest.sol
```diff
@@ -506,18 +703,405 @@ contract BlueFallbackRollingTest is BaseTest {
...
+            INCENTIVE_AT_START,
+            INCENTIVE_AT_END,
+            minRollableAssets,
+            assets
+        );
+    }
+
+    /// @dev One wei of debt above `minRollableAssets` is enough for the constraint to apply.
+    function testRollRevertsWhenDebtIsJustAboveMinRollableAssets() public {
+        // forge-lint: disable-next-line(unsafe-typecast) as DEBT - 1 < type(uint128).max
```
chrmatt, line 791: How could DEBT - 1 be larger? Is this copy paste artifact?

chrmatt (approved):
Looks good to me. Approved with minor comment.

test/BlueFallbackRollingTest.sol
peyha: it's to prevent a foundry lint error

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
```diff
@@ -75,9 +79,12 @@ contract BlueFallbackRolling is IBlueFallbackRolling {
             InconsistentCollateralToken()
         );
 
+        uint256 debtAssets = IMidnight(MIDNIGHT).debt(midnightId, user);
+        require(assets >= minRollableAssets || assets == debtAssets, RollableAssetsTooLow());
+
         // Round in favor of the Midnight position.
```
MathisGD: let's append: "Thus, splitting the rolls can lower the blue final LTV. This is mitigated by the min rollable debt."

MathisGD (approved):

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
MathisGD: @claude please apply

MathisGD: @claude please apply

MathisGD: @claude please apply

PR #1151, opened 2026-08-17
[Certora] Fix healthiness check for token transfer.

jhoenicke:
The previous simplification for token transfer added an unsoundness: It didn't check the healthiness invariant before token transfer, but still assumed it afterwards.  This way it would miss bugs that violate the invariant, because the token transfer itself repairs the invariant.

This pull request adds checking the invariant before token transfer again.  The axioms were optimized a bit (just include the correct axioms) to avoid timeouts.  The changes are based on the pull request that adds mulDiv axioms for mathint.

[corresponding thread](https://morpholabs.slack.com/archives/C0ACSA8TNMA/p1786369811444439)

### DIFF
--- a/certora/confs/Healthiness.conf
+++ b/certora/confs/Healthiness.conf
@@ -23,7 +23,6 @@
     "-mediumTimeout 60",
     "-timeout 7200"
   ],
-  "rule_sanity": "basic",
   "smt_timeout": 7200,
   "msg": "Midnight Healthiness"
 }
--- a/certora/confs/Healthiness.conf
+++ b/certora/confs/Healthiness.conf
@@ -23,6 +23,7 @@
     "-mediumTimeout 60",
     "-timeout 7200"
   ],
+  "rule_sanity": "basic",
   "smt_timeout": 7200,
   "msg": "Midnight Healthiness"
 }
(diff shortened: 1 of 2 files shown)


certora/confs/Healthiness.conf
```diff
@@ -21,9 +21,9 @@
     "-smt_useNIA true",
     "-depth 0",
     "-mediumTimeout 60",
-    "-timeout 7200",
-    "-s [z3:def{randomSeed=1},z3:def{randomSeed=2},z3:def{randomSeed=3},z3:def{randomSeed=4},z3:def{randomSeed=5},z3:def{randomSeed=6},z3:def{randomSeed=7},z3:def{randomSeed=8},z3:def{randomSeed=9},z3:def{randomSeed=10}]"
+    "-timeout 7200"
   ],
+  "rule_sanity": "basic",
```
QGarchery:
```suggestion
```

QGarchery (approved):

certora/specs/Healthiness.spec
QGarchery, line 1: we can remove the `transferCallback` function

MathisGD (approved):

PR #1160, opened 2026-08-24
review fallback rolling

QGarchery:


src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
QGarchery, line 25:
```suggestion
    mapping(bytes32 configId => bool) public override isConfig;
```
this is also possible, by putting the `user` in the config id. What's the rationale for extracting `user` ?

QGarchery:
```suggestion
uint256 start,
uint256 end,
uint256 incentiveAtStart,
uint256 incentiveAtEnd,
uint256 minRollableAssets,
```
there is no cost to do this (change everything to `uint256`) ?

QGarchery, line 104:
```suggestion
        uint256 incentiveAssets = UtilsLib.mulDivDown(assets, incentiveFactor, WAD - incentiveFactor);
```
otherwise the incentive is bounded to 50% of the total borrowed (also it aligns with how the referral fee is taken in the bundles)

QGarchery: We probably want to also bound the incentive to be strictly less than `WAD` if we do this change

peyha: not sure we challenged it, it just felt more natural to have a per user config to be able to re-use some configIds for many users for clarity (we only expect a few relevant config for a given midnight/blue market pair)

peyha: The benefit I see from using `uint64` is that it's quite natural for timestamps and percentages in WAD as the value is capped by ~18e18. We could also have a small gas improvement by using `encodePacked` but it's fairly negligible (~20 gas)
I don't see any strong reason to go for `uint256` (maybe normalizing types but that's not too convincing)

peyha: Effectively we could tweak the formula so that it corresponds to a % of the total debt. Note that the current version is consistent with the public allocator fee while your proposition is consistent with the referral fee from the bundles. The current version is also naturally capped by `LLTV/LTV-1`

peyha: I feel like the current version is simpler to think about, it should be set mostly to reimburse gas fees so it's easier to come up with an incentive value that will cover it

QGarchery: > The current version is also naturally capped by LLTV/LTV-1

I didn't get that, isn't it capped by 50% ? The incentive assets are at most equal to the repaid assets, and the total borrowed assets is the sum of the incentive assets plus the repaid assets.

To me capping at 50% is a downside, as it's not very clear that `incentiveFactor = WAD` means 50%

QGarchery: Reasons for switching to `uint256` are usually:
- no need for an integrator to cast the variable. Even if they use a shorter type as you mention, like `uint64` for timestamps, they can use that easily (in solidity it actually requires 0 code)
- no risk of introducing unexpected overflow. Doing `x + y` on `uint64` variables is not the same as doing it on `uint256` variables

QGarchery: Not really convinced since `start` and `end` are hashed as well, so there will be plenty of different config ids

chrmatt: The incentive factor is a fraction of the rolled assets. If I understand correctly, it is bounded by 100% of the rolled assets. It can then borrow more on Blue because it needs to borrow the initial debt plus the incentive. But I don't see why it would make sense to pay an incentive that is higher than 100% of what was borrowed on Midnight.

peyha: Sorry I was not clear, I didn't mean a fixed config id for all markets but a fixed config id for a given midnight market - blue market pair which mean that many users would share same config ids so it would be easier for rollers/interface to reason with this definition of config id imo. I think it's relevant to separate the user from the configuration in this abstraction

adhusson: from discussion with mathis a few days ago: the calls will not be from contracts (actually debatable, and even from the outside it could avoid a cast depending on the library), and putting the arg in the packed struct would need a cast otherwise, so being uint64 avoids having to justify it.

QGarchery: Agree that most of the time it wouldn't make sense, but I would ask the opposite question: why having this bound ? If this bound is not needed, let's get rid of it, the spec is simpler. If this bound is needed, then let's document it, because now it's not very clear clear that the incentive is capped

chrmatt: It's capped at 100% which imo is natural. But anyway I think the cap doesn't come from the formula but from `require(incentiveAtEnd <= WAD)`. We could remove this if we want to allow larger incentives.

QGarchery: ok, and it's very low cost as well

QGarchery: > putting the arg in the packed struct would need a cast otherwise, so being uint64 avoids having to justify it.

I did not follow this, what do you need to justify ?

peyha: Your point is quite convincing, I'll open a PR

adhusson: > I did not follow this, what do you need to justify ?

Actually it would not introduce a cast to widen any of those, my bad.

MathisGD: in the public allocator, passing arguments [directly as uint64](https://github.com/morpho-org/vault-v2/blob/80e1fead368e0626ec84c1c6245e84f6e0f01fef/src/periphery/blue-public-allocator/BluePublicAllocator.sol#L104-L110) avoids casting and justifying why unsafe casting is ok. Here there is no real benefit though.

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
```diff
@@ -90,12 +89,13 @@ contract BlueFallbackRolling is IBlueFallbackRolling {
...
             InconsistentCollateralToken()
         );
+        require(blueMarketParams.loanToken == midnightMarket.loanToken, InconsistentLoanToken());
 
         uint256 debtAssets = IMidnight(MIDNIGHT).debt(midnightId, user);
         require(assets >= minRollableAssets || assets == debtAssets, RolledAssetsTooLow());
 
-        // Round in favor of the Midnight position. Thus, splitting the rolls can raise the blue final LTV. This is
-        // mitigated by the min rollable debt.
+        // Round in favor of making the Midnight position safer. Thus, splitting the rolls can raise the blue final LTV.
```
```suggestion
        // Round in favor of making the Midnight position healthier. Thus, splitting the rolls can raise the blue final LTV.
```

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
MathisGD: > otherwise the incentive is bounded to 50% of the total borrowed

@QGarchery naturally I think as the fee as A. "% borrowed in addition to what was repaid", not as B. "% borrowed going to the fee". When thinking about your cost, I think that you want to reason about A not B.

agree though that the 100% (A.) (50% if you see it as B.) is somewhat arbitrary but it's clearly an upper bound of any sensible use-case.

>(also it aligns with how the referral fee is taken in the bundles)

which bundle are you making a parallel with?

MathisGD: #1161

MathisGD: discussed live: we keep this model because it's more natural for the borrower to think about. we will add a comment about this in an other PR

MathisGD (approved):

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
QGarchery: keeping it as it is currently, because the incentive corresponds to rate (of interest), so it's simpler to think about

adhusson (approved):

MathisGD (approved):

PR #1163, opened 2026-08-26
[blue rolling fallback] incentive doc

peyha:


src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
```diff
@@ -15,6 +15,9 @@ import {SafeApproveLib} from "../libraries/SafeApproveLib.sol";
 /// @dev Users must authorize this contract on both Midnight and Blue before their debt can be rolled.
 /// @dev Users must make sure that the oracle and the LLTV of the Blue market are appropriate; otherwise, their
 /// position on Blue could be left close to liquidation.
+/// @dev The rolling incentive corresponds to the percentage of the rolled debt being paid as incentive which is
+/// equivalent to a proportion of added interest.
+/// @dev The rolling incentive is capped by 100% of the rolled debt.
```
adhusson:
```suggestion
/// @dev The rolling incentive is capped by 100% of the rolled debt on Midnight, and is thus at most 50% of the debt created on Blue.
```

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
```diff
@@ -15,6 +15,9 @@ import {SafeApproveLib} from "../libraries/SafeApproveLib.sol";
 /// @dev Users must authorize this contract on both Midnight and Blue before their debt can be rolled.
 /// @dev Users must make sure that the oracle and the LLTV of the Blue market are appropriate; otherwise, their
 /// position on Blue could be left close to liquidation.
+/// @dev The rolling incentive corresponds to the percentage of the rolled debt being paid as incentive which is
```
adhusson:
```suggestion
/// @dev The rolling incentive corresponds to the percentage of the debt repaid on Midnight that is given as incentive.
```

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
```diff
@@ -15,6 +15,9 @@ import {SafeApproveLib} from "../libraries/SafeApproveLib.sol";
...
+/// equivalent to a proportion of added interest.
```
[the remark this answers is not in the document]
peyha: The interest equivalence helps understand the concept

chrmatt (approved):

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
```diff
@@ -15,6 +15,10 @@ import {SafeApproveLib} from "../libraries/SafeApproveLib.sol";
 /// @dev Users must authorize this contract on both Midnight and Blue before their debt can be rolled.
 /// @dev Users must make sure that the oracle and the LLTV of the Blue market are appropriate; otherwise, their
 /// position on Blue could be left close to liquidation.
+/// @dev The rolling incentive corresponds to the percentage of the debt repaid on Midnight that is given as incentive
+/// equivalent to added interest on Blue.
+/// @dev The rolling incentive is capped by 100% of the rolled debt on Midnight, and is thus at most 50% of the debt
+/// created on Blue.
```
```suggestion
/// @dev The rolling incentive cap at 100% is arbitrary from a technical POV.
```

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
```diff
@@ -15,6 +15,10 @@ import {SafeApproveLib} from "../libraries/SafeApproveLib.sol";
 /// @dev Users must authorize this contract on both Midnight and Blue before their debt can be rolled.
 /// @dev Users must make sure that the oracle and the LLTV of the Blue market are appropriate; otherwise, their
 /// position on Blue could be left close to liquidation.
+/// @dev The rolling incentive corresponds to the percentage of the debt repaid on Midnight that is given as incentive
+/// equivalent to added interest on Blue.
```
QGarchery, line 19:
```suggestion
/// @dev The rolling incentive factor corresponds to the percentage of the debt repaid on Midnight that is given as incentive.
```
wondering if we should align "incentive factor" everywhere. For now we have the local variable `incentiveFactor` but `incentiveAtStart` and `incentiveAtEnd` are also "factors"

QGarchery (approved):

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
chrmatt: Strictly speaking, it's also not a percentage (i.e., a value of 10 does not mean 10%). The scaling could be made more explicit.

MathisGD: @qgarchery incentive works as well if we call the amount incentiveAssets

@chrmatt indeed but wad scaling is everywhere, people are used to it imo

chrmatt: Putting WAD scaling aside, a factor is not the same as a percentage: A factor of 0.5 corresponds to a percentage of 50 (at least that's how I think about it).

peyha: Renaming `incentiveFactor` to `incentivePct` could indeed be an idea.
Although `factor` is consistent with the liquidation incentive factor (which is also a percentage) and I don't think there were any misunderstanding so far so I think it's fine to keep it

adhusson (approved):

MathisGD (approved):

PR #1167, opened 2026-08-27
[blue rolling fallback] Dust position

peyha:


src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
```diff
@@ -22,6 +22,8 @@ import {SafeApproveLib} from "../libraries/SafeApproveLib.sol";
 /// which case the destination position debt and collateral can be difficult to predict.
 /// @dev Contrary to Midnight, Blue positions can be liquidated because of interest accrual, which should be taken into
 /// account when deciding/approving the rolling configuration.
+/// @dev Nothing prevents rollers from leaving a small amount of debt on Midnight that would be unprofitable to roll
+/// though incentiveAtEnd can mitigate the size of the remaining debt.
```
chrmatt: How would incentiveAtEnd mitigate the size of the remaining debt? People also want to avoid any liquidation, even small ones. I think the better argument is that incentiveAtEnd can be chosen large enough to make even small rolls profitable and the large incentive for a small position is still a small cost overall.

MathisGD (approved):

src/periphery/blue-fallback-rolling/BlueFallbackRolling.sol
```suggestion
/// @dev Nothing prevents rollers from leaving a small amount of debt on Midnight that would be unprofitable to roll.
```

MathisGD: I removed this detail

MathisGD (approved):

adhusson (approved):

chrmatt (approved):

PR #1177, opened 2026-09-01
[Certora] MaxRepaidHealthy - support more than two collaterals

jhoenicke:
This uses a solidity helper to compute maxDebt covered by collateral, instead of doing it in CVL.  This simplifies the spec and allows to drop the restriction to two collaterals.  There is still the restriction because we need a finite CVL function to match the market with the ghost market, though.  The current spec allows three collaterals and uses an loop unrolling depth of 3.

### DIFF
--- a/certora/specs/MaxRepaidHealthy.spec
+++ b/certora/specs/MaxRepaidHealthy.spec
@@ -74,14 +74,7 @@ persistent ghost address globalMarketLoanToken;
 
 persistent ghost uint256 globalMarketChainId;
 
-// Exactly two collaterals is not a restriction on the result. Liquidating touches a single collateral, so the
-// whole contribution of every other collateral to maxDebt enters the reasoning as one arbitrary non-negative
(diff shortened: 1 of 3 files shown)


MathisGD (approved):

QGarchery (approved):

certora/specs/MaxRepaidHealthy.spec
```diff
@@ -73,14 +74,7 @@ persistent ghost address globalMarketLoanToken;
 
 persistent ghost uint256 globalMarketChainId;
 
-// Exactly two collaterals is not a restriction on the result. Liquidating touches a single collateral, so the
-// whole contribution of every other collateral to maxDebt enters the reasoning as one arbitrary non-negative
-// value, and one extra collateral with an arbitrary amount, price and LLTV already realizes every such value.
-// The second collateral therefore plays the role of the arbitrary otherCollatContribution of the Rocq proof,
-// and a market with more collaterals is covered by the same argument.
```
QGarchery, line 80: This reasoning is interesting, but we have it in the rocq proof still

PR #1186, opened 2026-09-09
no-op for `withSig`

QGarchery:
see https://github.com/morpho-org/midnight/pull/1179#discussion_r3959297844 for motivation.

Note that:
- the idea is essentially to allow no-op for the `withSig` function
- it works against signature malleability in the following sense: if a tweaked signature was used to accept the digest, it would still be a no-op to call `setIsWhitelistedWithSig` with the initial signature


QGarchery:
Closing in favor of #1189, which is better because of gas usage and because it keeps the property that `newIsWhitelisted` is the whitelist status after the function is successfully called

PR #1190, opened 2026-09-12
[rate ratifier] Add set root with signature in SetterRateRatifier and remove EcrecoverRateRatifier

albert-llimos:


src/ratifiers/interfaces/ISetterRateRatifier.sol
```diff
@@ -4,11 +4,27 @@ pragma solidity >=0.5.0;
 
 import {IRatifier} from "../../interfaces/IRatifier.sol";
 
+struct Ratification {
+    bool isRatified;
```
MathisGD:
```suggestion
    bool isRootRatified;
```

src/ratifiers/interfaces/ISetterRateRatifier.sol
```diff
@@ -4,11 +4,27 @@ pragma solidity >=0.5.0;
...
+    uint128 nonce;
```
MathisGD:
```suggestion
    uint128 rootNonce;
```
maybe, to clarify that nonces are per root (not so intuitive)

src/ratifiers/SetterRateRatifier.sol
MathisGD, line 1: we could delete ECRecoverRatifier as part of this. and rename setterRateRatifier simply RateRatifier?

src/ratifiers/interfaces/ISetterRateRatifier.sol
albert-llimos: I named it `nonce`, similar to the whitelistEnterGate where nothing more is specified in the name but  I agree that it is better to name it `rootNonce`. Will do.

src/ratifiers/SetterRateRatifier.sol
albert-llimos: I was waiting first on getting confirmation to settle on this ratifier only but I can optimistically remove the `EcrecoverRatifier`, it seems likely that we will do that. Otherwise I will remove the commit.
I think I would prefer keeping the `SetterRateRatifier` naming though to be consistent with other ratifiers, since it has the same setter root mechanism as the SetterRatifier.

albert-llimos:
As part of removing the EcrecoverRateRatifier I have ported some tests from EcrecoverRateRatifierTest to SetterRateRatifier that were only present in the former, mainly testing the auction logic.
Also, without the EcrecoverRateRatifier, it means that the frontend test is back to only testing the EcrecoveRatifier so I have restored those to the same as in main (`FrontentSignatureTest.sol`and `sign-root.ts`).

src/ratifiers/SetterRateRatifier.sol
```diff
@@ -33,10 +41,48 @@ contract SetterRateRatifier is ISetterRateRatifier {
     /// tree might not be ratified or unratified by a single call to this function.
     function setIsRootRatified(address maker, bytes32 root, bool newIsRootRatified) external {
         require(maker == msg.sender || IMidnight(MIDNIGHT).isAuthorized(maker, msg.sender), Unauthorized());
-        isRootRatified[maker][root] = newIsRootRatified;
+        ratification[maker][root].isRootRatified = newIsRootRatified;
```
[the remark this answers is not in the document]
albert-llimos: I considered that but it seems to me that makers would either be using `setIsRootRatified`or `setIsRootRatifiedWithSig`, not both, and then increasing the nonce on `setIsRootRatified` seems conceptually wrong.
It is true that it could be problematic for a maker using both functions (is that even a reasonable usecase?) but it's easy to circumvent by just setting the root to `false` via `setIsWhitelistedWithSig`. Fwiw, this also applies to, for example, the WhitelistEnterGate and it is not addressed there.
@MathisGD Double checking with you that you agree with this, as I don't have too deep insights on how exactly makers sign on/off-chain.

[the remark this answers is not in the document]
MathisGD: >or provide an equivalent signature-cancellation mechanism

we have this, through setIsRootRatifiedWithSig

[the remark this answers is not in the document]
MathisGD: This problem is here in all contract that have set and setWith sig (including all ERC20s that support permit), not sure why here it would be more of an issue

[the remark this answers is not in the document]
albert-llimos: Yes we agree then, resolving.

MathisGD (approved):
good to merge for me

PR #1180, opened 2026-09-04
[rate ratifier] Rate ratifier / Price Ratifier

albert-llimos:
Implementation of RateRatifier, inspired by the original rate ratifier [PoC](https://github.com/morpho-org/midnight/pull/841).

No Ecrecover version is implemented as the RateRatifier (Setter) will now be used with the `setIsRootRatifiedWithSig`.

A PriceRatifier is also implemented, which is the former `SetterRatifier`adding `setIsRootRatifiedWithSig`.

PR #1200, opened 2026-09-17
[whitelist-enter-gate] Log ambiguous callers

peyha:
Add a non-indexed `caller` as the first parameter `SetIsWhitelistedWithSig`.

For `SetIsWhitelistedWithSig` the caller is distinct from the recovered whitelister, so it tells who relayed the signature.


src/periphery/whitelist-enter-gate/WhitelistEnterGate.sol
```diff
@@ -40,13 +40,13 @@ contract WhitelistEnterGate is IWhitelistEnterGate {
     function setRoleSetter(bool creditSide, address newRoleSetter) external {
         require(msg.sender == roleSetter[creditSide], NotRoleSetter());
         roleSetter[creditSide] = newRoleSetter;
-        emit SetRoleSetter(creditSide, newRoleSetter);
+        emit SetRoleSetter(msg.sender, creditSide, newRoleSetter);
```
MathisGD: but we don't log when the caller is unambiguous right?

src/periphery/whitelist-enter-gate/WhitelistEnterGate.sol
```diff
@@ -81,7 +81,9 @@ contract WhitelistEnterGate is IWhitelistEnterGate {
             require(nonce < previousNonce, InvalidNonce());
             require(isWhitelisted[creditSide][account] == newIsWhitelisted, WhitelistedStatusChanged());
         }
-        emit SetIsWhitelistedWithSig(whitelister, creditSide, account, newIsWhitelisted, nonce, previousNonce);
+        emit SetIsWhitelistedWithSig(
+            msg.sender, whitelister, creditSide, account, newIsWhitelisted, nonce, previousNonce
+        );
```
MathisGD, line 86: here it's ambiguous, so we should add it indeed

src/periphery/whitelist-enter-gate/interfaces/IWhitelistEnterGate.sol
```diff
@@ -22,12 +22,13 @@ interface IWhitelistEnterGate is IEnterGate {
...
     event Constructor(address indexed creditRoleSetter, address indexed debtRoleSetter, bool creditOpen, bool debtOpen);
-    event SetRoleSetter(bool creditSide, address indexed newRoleSetter);
-    event SetIsWhitelister(bool creditSide, address indexed account, bool newIsWhitelister);
+    event SetRoleSetter(address caller, bool creditSide, address indexed newRoleSetter);
+    event SetIsWhitelister(address caller, bool creditSide, address indexed account, bool newIsWhitelister);
     event SetIsWhitelisted(
         address indexed whitelister, bool creditSide, address indexed account, bool newIsWhitelisted
     );
     event SetIsWhitelistedWithSig(
+        address caller,
```
albert-llimos, line 31: ~~We added the caller to the event in the Ratifiers [because it is a gated function](https://github.com/morpho-org/midnight/pull/1180#discussion_r4030409131). `setRoleSetter` and `setIsWhitelister` are permissioned but `setIsWhitelistedWithSig` is not. I'd say that we don't really need to log the caller for the latter. Is it for consistency you want to add it?~~
I see from Mathis comments that what you meant initially is ambiguous vs non-ambiguous, not permissioned vs not permissioned.

src/periphery/whitelist-enter-gate/WhitelistEnterGate.sol
peyha: yes this one is not mandatory, it can retrieved with the previous `SetRoleSetter` event on the same side
but it's cheap and simplify a bit so we could keep it, i'm fine with both

src/periphery/whitelist-enter-gate/interfaces/IWhitelistEnterGate.sol
peyha: I'd say there are two reasons
1. consistency with the ratifiers
2. because it's permissionless it could be useful to know the accounts calling `SetIsWhitelistedWithSig` (though it could still be done without it by inspecting traces but it's way less efficient)
though users will pay an extra 256 gas

src/periphery/whitelist-enter-gate/WhitelistEnterGate.sol
MathisGD: but elsewhere in midnight we don't do that right?

albert-llimos (approved):

src/periphery/whitelist-enter-gate/WhitelistEnterGate.sol
peyha: oh yes good point it's similar to logging the `feeSetter` in `setMarketSettlementFee`

peyha: removing it

src/periphery/whitelist-enter-gate/WhitelistEnterGate.sol
```diff
@@ -40,13 +40,13 @@ contract WhitelistEnterGate is IWhitelistEnterGate {
...
         roleSetter[creditSide] = newRoleSetter;
-        emit SetRoleSetter(creditSide, newRoleSetter);
+        emit SetRoleSetter(msg.sender, creditSide, newRoleSetter);
     }
 
     function setIsWhitelister(bool creditSide, address account, bool newIsWhitelister) external {
         require(msg.sender == roleSetter[creditSide], NotRoleSetter());
         isWhitelister[creditSide][account] = newIsWhitelister;
-        emit SetIsWhitelister(creditSide, account, newIsWhitelister);
+        emit SetIsWhitelister(msg.sender, creditSide, account, newIsWhitelister);
```
adhusson: same here, msg.sender is known

src/periphery/whitelist-enter-gate/WhitelistEnterGate.sol
adhusson: It's not whitelisted, so I would say it does not matter. But we do log msg.sender in [BlueBuyCallback](https://github.com/morpho-org/midnight/blob/96c3e2ce97c9371260afb569d83f9d98e55ca660/src/periphery/blue-buy-callback/BlueBuyCallback.sol#L65), not sure why. We don't log it in [WhitelistReceiveSharesGate](https://github.com/morpho-org/vault-v2/blob/a0ba9df0ea697a080c0de69c18b84738cfb3bef7/src/periphery/gates/WhitelistReceiveSharesGate.sol#L89) and [WhitelistSendAssetsGate](https://github.com/morpho-org/vault-v2/blob/a0ba9df0ea697a080c0de69c18b84738cfb3bef7/src/periphery/gates/WhitelistSendAssetsGate.sol#L102).

MathisGD (approved):

adhusson (approved):

